PR #70, opened 2015-01-15
Bw2 add dev install instructions for macosx

bw2:


brettpthomas:
This looks great. The only potential issue is changing `WSGI_APPLICATION` in `settings.py`. Your change is the correct value, since wsgi.py is in the root directory, but I'm not sure if that could break current deployments. However, I just tested on the production server and this looks fine.

PR #73, opened 2015-01-20
Upgrading to Django 1.7 and associated refactoring 

brettpthomas:
Updating app requirements to 1.7. This necessitated a few other changes. What you should note: 

- Using a new class `XBrowseBaseConfig(AppConfig)` for app instantiation. This is a huge win, as it gets rid of a weird dependency issue that was, among other issues, causing Postgres to break. I didn't realize until today that this hook was added in 1.7 :) 

- I got rid of all the South migrations and created new django migrations from scratch. This shouldn't affect any developer interface, though take care to use the `makemigrations` and `migrate` commands moving forward.


README.md
```diff
@@ -67,4 +67,9 @@ After you've loaded the 1000 Genomes test project, you can run integration tests
     python functional_tests.py
 
 These tests assume that the `1kg` project above has been loaded - they run
-a bunch of browser simulations to make sure that the correct search results are returned.
\ No newline at end of file
+a bunch of browser simulations to make sure that the correct search results are returned.
+
+
+- psycopg2 in pip
```
bw2: These last 2 lines seem out of place

xbrowse_server/base/apps.py
```diff
@@ -0,0 +1,21 @@
...
+        Some of the Stores in the Mall depend on data stored in the db.
+        For example, the Datastore (ie. VariantStore) needs to know which projects have access to
+        which custom reference populations.
+        We set that up here, rather than store the state in the Datastore itself, to reduce the complexity of the Datastore API.
+        """
+        Project = self.get_model('Project')
+        mall.x_custom_populations_map = {p.project_id: p.private_reference_population_slugs() for p in Project.objects.all()}
+
+        ReferencePopulation = self.get_model('ReferencePopulation')
+        mall.x_custom_populations = ReferencePopulation.get_annotator_spec()
```
bw2: Disregard if I just don't know the code well enough, but the get_annotator_spec() method has a confusing name and seems to only be used here. It contains one line (`[p.to_dict() for p in ReferencePopulation.objects.all()]`). Why not just put this line here directly and get rid of that method?

brettpthomas:
Great catches. Deleted the `get_annotator_spec()` method and removed weird notes on readme. Tests passing // ready to merge from this view.

PR #98, opened 2015-01-28
bw2 enable ssl

bw2:


brettpthomas:
The config looks great, but the code as is won't allow any non-SSL deployments (eg. EC2) 

Should we add an `ssl_enabled` option to puppet and only add these lines if so?

bw2:
@brettpthomas - great idea

PR #138, opened 2015-04-13
Use load_project_dir in OSX README

brettpthomas:
The current OSX project loading instructions reference old file paths. If we update to use `load_project_dir` instead of the old add_* commands, issues are fixed!


bw2:
Thanks @brettpthomas

PR #158, opened 2015-11-15
Select cannonical transcript

bw2:


xbrowse/annotation/vep_annotations.py
```diff
@@ -257,47 +257,64 @@ def get_csq_fields_from_vcf_desc(csq_desc):
...
-    num_annotations = len(vep_annotation)
-    if num_annotations == 0:
-        print 'Warning: no VEP annnotation'
-        return None
+    if not transcript_annotations:
+        raise ValueError("transcript_annotations is empty")
 
-    worst_value = 1000
-    worst_index = -1
-    for i in range(num_annotations):
```
harindra-a: xrange is a nice function to use here as well (vs range) with caveats. Good pros/cons listed here http://stackoverflow.com/questions/135041/should-you-always-favor-xrange-over-range

xbrowse/annotation/vep_annotations.py
```diff
@@ -257,47 +257,64 @@ def get_csq_fields_from_vcf_desc(csq_desc):
...
+    for i, transcript_annotation in worst_severity_annotations:
+        if transcript_annotation['canonical']:
+            worst_severity_annotation_index = i
+            break
+    else:
+        # otherwise, sort transcripts alphabetically by transcript id and return the 1st one
+        worst_severity_annotations.sort(key=lambda x: x[1]['feature'])
+        worst_severity_annotation_index, _ = worst_severity_annotations[0]
+
+    return worst_severity_annotation_index
```
harindra-a: As a small stylistic comment, I generally stick to small single task functions vs longer functions to help me partition tasks so it is easier to debug, but the trade-off of is that it causes having a lot of smaller functions floating around.

bw2: I agree.. in general smaller functions are better.

xbrowse/annotation/vep_annotations.py
bw2: I general stick with `range` due to the cons listed - unless it's 10,000+ numbers and optimization is warranted.

xbrowse/annotation/vep_annotations.py
harindra-a: sounds good to me, ready for merge in!

PR #161, opened 2015-12-17
Feature phenotips integration 0.001

harindra-a:
Ready for beta testing

### DIFF
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -40,9 +40,9 @@ class Command(BaseCommand):
                 if line.strip() == "" or line.startswith('#'):
                     continue
                 indiv_id_list.append(line.strip())
-
             sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
+        indiv_id_list=None
         if options.get('vcf'):
             vcf_path = options.get('vcf')
             if vcf_path.endswith('.gz'):
@@ -50,19 +50,29 @@ class Command(BaseCommand):
             else:
                 vcf = open(vcf_path)
             indiv_id_list = vcf_stuff.get_ids_from_vcf(vcf)
-            self.__add_individuals_to_phenotips(indiv_id_list,project_id)
-            sample_management.add_indiv_ids_to_project(project, indiv_id_list)
+            #sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
+        individual_details=None
         if options.get('ped'):
             fam_file = open(options.get('ped'))
-            sample_management.update_project_from_fam(project, fam_file)
+            individual_details = sample_management.update_project_from_fam(project, fam_file)
 
+        #Favor PED file rich information VCF file minimum information to create patients
+        if individual_details is not None:
+          self.__add_individuals_to_phenotips(indiv_id_list,project_id,individual_details)
+        else:
+          if indiv_id_list is not None:
+            self.__add_individuals_to_phenotips(indiv_id_list,project_id)
+          else:
+            print '\nno information in VCF to create patients with\n'
+            sys.exit()
+          
 
     #given a list of individuals add them to phenotips
-    def __add_individuals_to_phenotips(self,individuals,project_id):
+    def __add_individuals_to_phenotips(self,individuals,project_id,patient_details=None):
       '''given a list of individuals add them to phenotips '''
       for individual in  individuals:        
-        create_patient_record(individual,project_id)
+        create_patient_record(individual,project_id,patient_details)
 
         
       
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -51,7 +51,7 @@ class Command(BaseCommand):
             else:
                 vcf = open(vcf_path)
             indiv_id_list = vcf_stuff.get_ids_from_vcf(vcf)
-            self.__add_individuals_to_phenotips(indiv_id_list)
+            self.__add_individuals_to_phenotips(indiv_id_list,project_id)
             sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
         if options.get('ped'):
@@ -60,10 +60,10 @@ class Command(BaseCommand):
 
 
     #given a list of individuals add them to phenotips
-    def __add_individuals_to_phenotips(self,individuals):
+    def __add_individuals_to_phenotips(self,individuals,project_id):
       '''given a list of individuals add them to phenotips '''
       for individual in  individuals:        
-        create_patient_record(individual)
+        create_patient_record(individual,project_id)
 
         
       
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -82,9 +82,12 @@ class Command(BaseCommand):
       #for now only using gender information from the PED file.
       for individual in  individual_details:
         id=individual['indiv_id']
-        gender='F'
-        if individual['gender'] == 'male':
+        if individual['gender'] == 'female':
+          gender='F'
+        elif individual['gender'] == 'male':
           gender='M'
+        else:
+          raise ValueError
         extra_details={
                        'gender':gender}
         create_patient_record(id,project_id,extra_details)
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -14,25 +14,24 @@ class Command(BaseCommand):
         make_option('--sample-list'),
         make_option('--vcf'),
         make_option('--ped'),
-        make_option('--create_phenotips_patients'),
-        make_option('--phenotips_username'),
-        make_option('--phenotips_password'),
+        make_option('--create_phenotips_patients',
+                    action='store_true',
+                    dest='create_phenotips_patients',
+                    default=False,
+                    help='add this patient to phenotips'),
     )
 
     def handle(self, *args, **options):
       
-        if options['vcf'] is None or options['ped'] is None or len(args)==0:
-          print '\n\nPlease enter a VCF file (--vcf), a PED file (--ped), and a project ID (first positional argument)'
-          print 'for example: python manage.py add_individuals_to_project  myProjectId  --vcf myVcf.vcf  --ped myPed.ped\n'
-          print 'If you also wish to create patients in phenotips, please use the --create_phenotips_patients,--phenotips_username, --phenotips_password to do so. ALL of them are REQUIRED.\n'
+        if len(args)==0 or options['vcf'] is None and options['ped'] is None:
+          print '\n\nPlease enter a VCF file (--vcf), OR IDEALLY A PED file (--ped), and a project ID (first positional argument)'
+          print 'for example: python manage.py add_individuals_to_project  myProjectId --ped myPed.ped\n'
+          print 'If you also wish to create patients in phenotips, please use the --create_phenotips_patients\n'
           sys.exit()
           
-        if options['create_phenotips_patients'] is not None and (options['phenotips_username'] is None or options['phenotips_password'] is None):
-          print '\n\nplease note if you use the --create_phenotips_patients option, both --phenotips_username and --phenotips_password options are REQUIRED.\n\n'
-          sys.exit()
-
         project_id = args[0]
         project = Project.objects.get(project_id=project_id)
+        indiv_id_list=None
 
         if options.get('sample_list'):
             indiv_id_list = []
@@ -42,7 +41,6 @@ class Command(BaseCommand):
                 indiv_id_list.append(line.strip())
             sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
-        indiv_id_list=None
         if options.get('vcf'):
             vcf_path = options.get('vcf')
             if vcf_path.endswith('.gz'):
@@ -50,29 +48,46 @@ class Command(BaseCommand):
             else:
                 vcf = open(vcf_path)
             indiv_id_list = vcf_stuff.get_ids_from_vcf(vcf)
-            #sample_management.add_indiv_ids_to_project(project, indiv_id_list)
+            sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
         individual_details=None
         if options.get('ped'):
             fam_file = open(options.get('ped'))
             individual_details = sample_management.update_project_from_fam(project, fam_file)
 
-        #Favor PED file rich information VCF file minimum information to create patients
-        if individual_details is not None:
-          self.__add_individuals_to_phenotips(indiv_id_list,project_id,individual_details)
-        else:
-          if indiv_id_list is not None:
-            self.__add_individuals_to_phenotips(indiv_id_list,project_id)
+        #add to phenotips
+        if options.get('create_phenotips_patients'):
+          #Favor PED file rich information VCF file minimum information to create patients
+          if individual_details is not None:
+            self.__add_individuals_to_phenotips_from_ped(individual_details,project_id)
           else:
-            print '\nno information in VCF to create patients with\n'
-            sys.exit()
-          
+            if indiv_id_list is not None:
+              self.__add_individuals_to_phenotips_from_vcf(indiv_id_list,project_id)
+            else:
+              print '\nno information in VCF to create patients with\n'
+              sys.exit()
+        else:
+          print '\n----not adding these patients to local phenotips instance.----\n'
+            
 
     #given a list of individuals add them to phenotips
-    def __add_individuals_to_phenotips(self,individuals,project_id,patient_details=None):
+    def __add_individuals_to_phenotips_from_vcf(self,individuals,project_id):
       '''given a list of individuals add them to phenotips '''
       for individual in  individuals:        
-        create_patient_record(individual,project_id,patient_details)
+        create_patient_record(individual,project_id)
+        
+    #given a list of individuals add them to phenotips
+    def __add_individuals_to_phenotips_from_ped(self,individual_details,project_id,):
+      '''given a list of individuals add them to phenotips '''
+      #for now only using gender information from the PED file.
+      for individual in  individual_details:
+        id=individual['indiv_id']
+        gender='F'
+        if individual['gender'] == 'male':
+          gender='M'
+        extra_details={
+                       'gender':gender}
+        create_patient_record(id,project_id,extra_details)
 
         
       
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -24,8 +24,8 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
       
         if len(args)==0 or options['vcf'] is None and options['ped'] is None:
-          print '\n\nPlease enter a VCF file (--vcf), OR IDEALLY A PED file (--ped), and a project ID (first positional argument)'
-          print 'for example: python manage.py add_individuals_to_project  myProjectId --ped myPed.ped\n'
+          print '\n\nPlease enter a VCF file (--vcf), OR IDEALLY A PED file (--ped), and a project ID (first positional argument).'
+          print 'For example: python manage.py add_individuals_to_project  myProjectId --ped myPed.ped\n'
           print 'If you also wish to create patients in phenotips, please use the --create_phenotips_patients\n'
           sys.exit()
           
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -11,14 +11,23 @@ from xbrowse_server.phenotips.utilities import create_patient_record
 class Command(BaseCommand):
 
     option_list = BaseCommand.option_list + (
-        make_option('--sample-list'),
-        make_option('--vcf'),
-        make_option('--ped'),
-        make_option('--create_phenotips_patients',
+        make_option('--sample-list',
                     action='store_true',
-                    dest='create_phenotips_patients',
+                    dest='sample-list',
                     default=False,
-                    help='add this patient to phenotips'),
+                    help='A sample list to gather patient information from.'),
+        make_option('--vcf',
+                    action='store_true',
+                    dest='vcf',
+                    default=False,
+                    help='A VCF file to gather patient information from.'
+                    ),
+        make_option('--ped',
+                    action='store_true',
+                    dest='ped',
+                    default=False,
+                    help='A PED file to gather patient information from (PREFERRED due to richer information set).'
+                    ),
     )
 
     def handle(self, *args, **options):
@@ -31,7 +40,6 @@ class Command(BaseCommand):
           
         project_id = args[0]
         project = Project.objects.get(project_id=project_id)
-        indiv_id_list=None
 
         if options.get('sample_list'):
             indiv_id_list = []
@@ -50,47 +58,8 @@ class Command(BaseCommand):
             indiv_id_list = vcf_stuff.get_ids_from_vcf(vcf)
             sample_management.add_indiv_ids_to_project(project, indiv_id_list)
 
-        individual_details=None
         if options.get('ped'):
             fam_file = open(options.get('ped'))
             individual_details = sample_management.update_project_from_fam(project, fam_file)
 
-        #add to phenotips
-        if options.get('create_phenotips_patients'):
-          #Favor PED file rich information VCF file minimum information to create patients
-          if individual_details is not None:
-            self.__add_individuals_to_phenotips_from_ped(individual_details,project_id)
-          else:
-            if indiv_id_list is not None:
-              self.__add_individuals_to_phenotips_from_vcf(indiv_id_list,project_id)
-            else:
-              print '\nno information in VCF to create patients with\n'
-              sys.exit()
-        else:
-          print '\n----not adding these patients to local phenotips instance.----\n'
-            
-
-    #given a list of individuals add them to phenotips
-    def __add_individuals_to_phenotips_from_vcf(self,individuals,project_id):
-      '''given a list of individuals add them to phenotips '''
-      for individual in  individuals:        
-        create_patient_record(individual,project_id)
-        
-    #given a list of individuals add them to phenotips
-    def __add_individuals_to_phenotips_from_ped(self,individual_details,project_id,):
-      '''given a list of individuals add them to phenotips '''
-      #for now only using gender information from the PED file.
-      for individual in  individual_details:
-        id=individual['indiv_id']
-        if individual['gender'] == 'female':
-          gender='F'
-        elif individual['gender'] == 'male':
-          gender='M'
-        else:
-          raise ValueError
-        extra_details={
-                       'gender':gender}
-        create_patient_record(id,project_id,extra_details)
-
-        
       
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -2,11 +2,10 @@ from optparse import make_option
 import gzip
 import sys
 from django.core.management.base import BaseCommand
-
 from xbrowse_server.base.models import Project
 from xbrowse_server import sample_management
 from xbrowse.parsers import vcf_stuff
-from xbrowse_server.phenotips.utilities import create_patient_record
+
 
 class Command(BaseCommand):
 
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -11,20 +11,15 @@ class Command(BaseCommand):
 
     option_list = BaseCommand.option_list + (
         make_option('--sample-list',
-                    action='store_true',
                     dest='sample-list',
                     default=False,
                     help='A sample list to gather patient information from.'),
         make_option('--vcf',
-                    action='store_true',
                     dest='vcf',
-                    default=False,
                     help='A VCF file to gather patient information from.'
                     ),
         make_option('--ped',
-                    action='store_true',
                     dest='ped',
-                    default=False,
                     help='A PED file to gather patient information from (PREFERRED due to richer information set).'
                     ),
     )
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -6,7 +6,8 @@ from django.core.management.base import BaseCommand
 from xbrowse_server.base.models import Project
 from xbrowse_server import sample_management
 from xbrowse.parsers import vcf_stuff
-
+from django.conf import settings
+from xbrowse_server.phenotips.utilities import create_patient_record
 
 class Command(BaseCommand):
 
@@ -61,7 +62,8 @@ class Command(BaseCommand):
     #given a list of individuals add them to phenotips
     def __add_individuals_to_phenotips(self,individuals):
       '''given a list of individuals add them to phenotips '''
-      for individual in  individuals:
-        print individual
+      for individual in  individuals:        
+        create_patient_record(individual)
+
         
       
--- a/xbrowse_server/base/management/commands/add_individuals_to_project.py
+++ b/xbrowse_server/base/management/commands/add_individuals_to_project.py
@@ -6,7 +6,6 @@ from django.core.management.base import BaseCommand
 from xbrowse_server.base.models import Project
 from xbrowse_server import sample_management
 from xbrowse.parsers import vcf_stuff
-from django.conf import settings
 from xbrowse_server.phenotips.utilities import create_patient_record
 
 class Command(BaseCommand):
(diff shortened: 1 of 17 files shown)


xbrowse_server/base/management/commands/add_project.py
```diff
@@ -1,14 +1,43 @@
...
+from xbrowse_server.phenotips.utilities import add_new_user_to_phenotips
+from xbrowse_server.phenotips.utilities import get_uname_pwd_for_project
+from django.conf import settings
 
 class Command(BaseCommand):
 
     def handle(self, *args, **options):
+        if len(args)<2 or not args[0] or not args[1]:
+          print '\n'
+          print 'please provide a project ID and a project name as the first two arguments. If project name as spaces, please remember to enclose in double quotes'
```
bw2: would rather raise error if project name has spaces.

xbrowse_server/base/management/commands/add_individuals_to_project.py
```diff
@@ -50,18 +48,46 @@ def handle(self, *args, **options):
...
+        create_patient_record(individual,project_id)
+        
+    #given a list of individuals add them to phenotips
+    def __add_individuals_to_phenotips_from_ped(self,individual_details,project_id,):
+      '''given a list of individuals add them to phenotips '''
+      #for now only using gender information from the PED file.
+      for individual in  individual_details:
+        id=individual['indiv_id']
+        gender='F'
+        if individual['gender'] == 'male':
```
bw2: to double check assumptions, would be good to do
if individual['indiv_id'] == 'female': gender='F' 
if individual['indiv_id'] == 'male': gender='M' 
else: raise ValueError

xbrowse_server/base/management/commands/add_individuals_to_project.py
```diff
@@ -14,25 +14,24 @@ class Command(BaseCommand):
...
     )
 
     def handle(self, *args, **options):
       
-        if options['vcf'] is None or options['ped'] is None or len(args)==0:
-          print '\n\nPlease enter a VCF file (--vcf), a PED file (--ped), and a project ID (first positional argument)'
-          print 'for example: python manage.py add_individuals_to_project  myProjectId  --vcf myVcf.vcf  --ped myPed.ped\n'
-          print 'If you also wish to create patients in phenotips, please use the --create_phenotips_patients,--phenotips_username, --phenotips_password to do so. ALL of them are REQUIRED.\n'
+        if len(args)==0 or options['vcf'] is None and options['ped'] is None:
+          print '\n\nPlease enter a VCF file (--vcf), OR IDEALLY A PED file (--ped), and a project ID (first positional argument)'
```
bw2: Very minor point, but for consistency we should chose if error messages and method docs start with upper case or lower case. (eg. Please vs. please). I would vote for upper case.

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+from django.conf import settings
+import json
+import urllib2
+import base64
+import requests
+logger = logging.getLogger(__name__)
+import sys
+
+#create a patient record in phenotips
+#by convention username and password are project_id,project_idproject_id
```
bw2: would put this in the method doc string.

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
             </div>
             
+
+
+        <!-- Phenotips integration -->
+		<div id="phenotipsModal" class="modal" role="dialog">
+			<div class="modal-dialog">
+				<div class="modal-content">
+					<div class="modal-header">
+						<button type="button" class="close" data-dismiss="modal">&times;</button>
```
bw2: cool.. we should add this X to other xBrowse dialogs (like gene details in variant search results)

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
+					</div>
+				</div>
+			</div>
+		</div>
+		<script>
+		$('.phenotipsEditModalBtn').on('click',function(event) {
+			sessionStorage.setItem('currentInFocusIndividualId',this.id)
+			var patientDetails=this.id.split('_');
+			var patientId=patientDetails[0];
+			var project=patientDetails[1];
```
bw2: if the project id had an underscore, I think there'd be issues here.

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
                         <td>{% if indiv.maternal_id %}{{ indiv.maternal_id }}{% else %}-{% endif %}</td>
                         <td>{% if indiv.paternal_id %}{{ indiv.paternal_id }}{% else %}-{% endif %}</td>
                         <td>{% for ipheno in indiv.get_phenotypes %}{{ ipheno.phenotype.name }}{% empty %}-{% endfor %}</td>
+                        <td><button id={{ indiv }}_{{ project.project_id }} type="button" class="btn btn-info btn-sm phenotipsEditModalBtn">Edit in Phenotips</button></td>
```
bw2: can project_id be stored in some other custom html attribute?

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
+					height: 100%;
+					}
+
+			.modal-dialog {
+					display: inline-block;
+					text-align: left;
+					vertical-align: middle;
+					}
+
+			.modal-dialog {
```
bw2: combine into just one .modal-dialog block

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
+		/**
+		 * Given a patient ID (NAxxx,) show the phenotips edit page for it
+		 **/
+		function showEditPageForThisPatient(patientId,project) {
+			var uri = '/api/phenotips/proxy/edit/' + patientId + '?' + 'project=' + project;
+			$('#phenotipsEditPatientFrame').attr('src', uri);
+			$('#phenotipsModal').modal('show');
+		}
+		</script>
+		<style>
```
bw2: conventional wisdom is to put css styles in a separate .css file (though I never fully undrerstood why)

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
+		*/
+		<script type='text/javascript' src='{{ STATIC_URL }}phenotips_integration/phenotips_integration.js'></script>
+    	<link rel='stylesheet' type='text/css' href='{{ STATIC_URL }}phenotips_integration/phenotips_integration.css' />    
+		 -->
+		<script>
+		/**
+		 * Given a patient ID (NAxxx,) show the phenotips edit page for it
+		 **/
+		function showEditPageForThisPatient(patientId,project) {
+			var uri = '/api/phenotips/proxy/edit/' + patientId + '?' + 'project=' + project;
```
bw2: for the previous-page-showing issue, maybe something like $('#phenotipsEditPatientFrame').attr('src', "about:blank"); can be used to clear the iframe before new url is loaded.

xbrowse_server/templates/family/family_home.html
```diff
@@ -145,12 +146,83 @@ <h4 class="section-header">Individuals</h4>
...
+					</div>
+					<div class="modal-body"">
+						<iframe name="phenotipsEditPatientFrame" id="phenotipsEditPatientFrame" src="" width=1000 height=800 frameborder="0"></iframe>
+					</div>
+				</div>
+			</div>
+		</div>
+		<script>
+		$('.phenotipsEditModalBtn').on('click',function(event) {
+			sessionStorage.setItem('currentInFocusIndividualId',this.id)
```
bw2: sorry - what's sessionStorage and how/where is it used?

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+  '''authenticates to phenotips, fetches (GET) given results and returns that'''
+  try:
+    password_mgr = urllib2.HTTPPasswordMgrWithDefaultRealm()
+    request = urllib2.Request(url)
+    base64string = base64.encodestring('%s:%s' % (uname, pwd)).replace('\n', '')
+    request.add_header("Authorization", "Basic %s" % base64string)   
+    result = urllib2.urlopen(request)   
+    return result
+  except Exception as e:
+    return e
```
bw2: why not let exception propagate?  
Also, it might cause issues here: `if result is not None and result.getcode()==200`

xbrowse_server/base/management/commands/add_project.py
```diff
@@ -1,14 +1,43 @@
...
+          print 'example: python manage.py 1kg "1000 Genomes Project"\n'
+          sys.exit()
         project_id = args[0]
+        project_name = args[1]
```
bw2: where does project name appear in phenotips?

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+  except Exception as e:
+    print 'error in do_authenticated_PUT:',e,
+    return e
+  
+#do a PUT call to phenotips
+def do_authenticated_POST(uname,pwd,url,data,headers):
+  try:
+    request=requests.post(url,data=data,auth=(uname,pwd),headers=headers)
+  except Exception as e:
+    print 'error in do_authenticated_POST:',e,
```
bw2: for error handling inside moduels, I feel we should stick with python logging through-out (for consistency and so we can adjust format in one location) and also - for errors inside utilities functions, etc. we should raise exceptions instead of returning values to reduce the chances of silently ignoring errors and to clarify error handling code

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+#create a patient record in phenotips
+#by convention username and password are project_id,project_idproject_id
+#authentication is protected by access to machine/localhost
+def create_patient_record(individual_id,project_id,patient_details=None):
+  '''make a patient record'''
+  uri = settings.PHENOPTIPS_HOST_NAME + '/bin/PhenoTips/OpenPatientRecord?create=true&eid=' + individual_id
+  if patient_details is not None:
+    uri += '&gender='+patient_details['gender']
+  uname=project_id
+  pwd=project_id+project_id
```
bw2: get_uname_pwd_for_project ?

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+  
+#return the username and password for this project
+def get_uname_pwd_for_project(project_name):
+  uname=project_name
+  pwd=project_name+project_name
+  return uname,pwd
+
+
+#return the generic collaborator username and password for this project
+def get_generic_collaborator_uname_pwd_for_project(project_name):
```
bw2: matter of style, but since this function is essentially identical to 
get_uname_pwd_for_project(..), why not combine into 1 function with an additional can_edit=False arg?

xbrowse_server/base/management/commands/add_project.py
```diff
@@ -1,14 +1,43 @@
...
+      add_new_user_to_phenotips(project_name,
+                                project_name, 
+                                uname,
+                                settings.PHENOPTIPS_ALERT_CONTACT ,
+                                pwd)
+      #next create a user with ONLY VIEW privileges (the rights are determined when pateints are added in,
+      #this step merely creates the user
+      add_new_user_to_phenotips(project_name + '(view only)',
+                                project_name + '(view only)', 
+                                uname+'_view',
```
bw2: would call get_generic_collaborator_uname_pwd_for_project(project_id) to get this value to isolate the uname+"_view" logic.

xbrowse_server/templates/family/family_home.html
bw2: Where is currentInFocusIndividualId used?

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+  do_authenticated_POST(admin_uname,admin_pwd,url,data,headers)
+  
+  
+#do a PUT call to phenotips
+def do_authenticated_PUT(uname,pwd,url,data,headers):
+  try:
+    request=requests.put(url,data=data,auth=(uname,pwd),headers=headers)
+  except Exception as e:
+    print 'error in do_authenticated_PUT:',e,
+    return e
```
bw2: I'm fine with catching and logging the request errors here, but again I think better to re-raise the same error or throw a new one, than return a value.

xbrowse_server/phenotips/utilities.py
```diff
@@ -0,0 +1,121 @@
...
+        'property#first_name':new_user_first_name,
+        'property#last_name':new_user_last_name,
+        'property#email':email_address,
+        'property#password':new_user_pwd
+        }
+  url=settings.PHENOPTIPS_HOST_NAME + '/rest/wikis/xwiki/spaces/XWiki/pages/' + new_user_name + '/objects'
+  do_authenticated_POST(admin_uname,admin_pwd,url,data,headers)
+  
+  
+#adds a collaborator to an existing patient. Requires an existing collaborator username, patient_eid (PXXXX..)
```
bw2: Just curious - why are all collaborators added as 'read_only' ? 
Also, I really like these method docs, I would just put them in a method doc string.

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
+def fetch_phenotips_edit_page(request,eid):
+  '''test function to see if we can proxy phenotips'''  
+  current_user = request.user
+  
+  if request.GET.has_key('project'):
+    project_name=request.GET['project']  
+    #add project name to session for use in proxying
+    request.session['current_project_name']=project_name
```
bw2: this is cool

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    elif project.is_public:
+        auth_level = 'public'
+    elif project.can_view(request.user):
+        auth_level = 'viewer'
+    else:
+        raise Exception("Authx - cannot authenticate:fetch_phenotips_edit_page ")
+    request.session['current_auth_level']=auth_level
+    
+  else: 
+    #getting admin account to translate eid to id
```
bw2: Since this handler is triggered by the button, shouldn't request.GET.has_key('project') always be true?in which case shouldn't the "else" code raise an error to avoid stale values from potentially redirecting to another project?

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    project_name = request.session['current_project_name']
+    ext_id=request.session['current_ext_id']
+    auth_level = request.session['current_auth_level']
+    
+  #depending on auth level, pick either the full edit username or the view-only username for this
+  #project to fetch page
+  phenotips_uname,phenotips_pwd ='_','_'
```
bw2: would delete line 69 for clarity

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+  if auth_level=='admin':
+    phenotips_uname,phenotips_pwd = get_uname_pwd_for_project(project_name)
+  else:
+    phenotips_uname,phenotips_pwd = get_generic_collaborator_uname_pwd_for_project(project_name)
+  url= settings.PHENOPTIPS_HOST_NAME+'/bin/'+ ext_id
```
bw2: would check assumptions about auth_level value here.. 
(eg. if auth_level == 'view': url= settings.PHENOPTIPS_HOST_NAME+'/bin/'+ ext_id )
...

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    ext_id=convert_internal_id_to_external_id(eid,
+                                            phenotips_proj_specific_admin__uname,
+                                            phenotips_proj_specific_admin_pwd)
+    request.session['current_ext_id']=ext_id
+    
+    #no check current auth level and add that to session too
+    project = get_object_or_404(Project, project_id=project_name)
+    if project.can_admin(request.user):
+        auth_level = 'admin'
+    elif project.can_edit(request.user):
```
bw2: I think only need to check can_admin or can_view here since can_edit (due to unfortunate naming) is the same as can_view in this case.

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+        auth_level = 'editor'
+    elif project.is_public:
+        auth_level = 'public'
+    elif project.can_view(request.user):
+        auth_level = 'viewer'
+    else:
+        raise Exception("Authx - cannot authenticate:fetch_phenotips_edit_page ")
```
bw2: other xbrowse api calls do 
return HttpResponse('unauthorized')
in this case. Better error pages are definitely needed, but for now I better to do that for consistency.

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+  phenotips_uname,phenotips_pwd ='_','_'
+  if auth_level=='admin':
+    phenotips_uname,phenotips_pwd = get_uname_pwd_for_project(project_name)
+  else:
+    phenotips_uname,phenotips_pwd = get_generic_collaborator_uname_pwd_for_project(project_name)
+  url= settings.PHENOPTIPS_HOST_NAME+'/bin/'+ ext_id
+  if auth_level=='admin':
+    url= settings.PHENOPTIPS_HOST_NAME+'/bin/edit/data/'+ ext_id
+  if not request.GET.has_key('project'):
+    url += '?'
```
bw2: I'm probably misunderstanding something.. in what case is request.GET.has_key('project') == False?

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    counter=0
+    for param,val in request.GET.iteritems():
+      url += param + '=' + val
+      if counter < len(request.GET)-1:
+        url += '&'
+      counter+=1
+  if type(ext_id) != dict:
```
bw2: again would do this type of error handling with try/except

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    response = __add_back_phenotips_headers_response(result)
+    return response
+  else:
+    logger.error('phenotips.views:'+ext_id['error'])
+    raise Http404    
 
 
+#do a GET as a proxy for Phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
```
bw2: would add @login_required

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+  for param,val in request.GET.iteritems():
+    url += param + '=' + val
+    if counter < len(request.GET)-1:
+      url += '&'
+    counter+=1
+  return url
+
+#do a POST as a proxy for Phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
```
bw2: would add @login_required

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+  '''test function to see if we can proxy phenotips'''  
+  current_user = request.user
+  
+  if request.GET.has_key('project'):
+    project_name=request.GET['project']  
+    #add project name to session for use in proxying
+    request.session['current_project_name']=project_name
+    
+    #also put current ext_id into session object
+    phenotips_proj_specific_admin__uname,phenotips_proj_specific_admin_pwd = get_uname_pwd_for_project(project_name)
```
bw2: style thing, but would prefer shorter like project_uname, project_pwd

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
-    url= os.path.join(settings.PHENOPTIPS_HOST_NAME,'bin/get/PhenoTips/ExportPatient?eid='+eid)
-    password_mgr = urllib2.HTTPPasswordMgrWithDefaultRealm()
-    request = urllib2.Request(url)
-    base64string = base64.encodestring('%s:%s' % (uname, pwd)).replace('\n', '')
-    request.add_header("Authorization", "Basic %s" % base64string)   
-    result = urllib2.urlopen(request)   
-    print result.read()
+    url= os.path.join(settings.PHENOPTIPS_HOST_NAME,'bin/get/PhenoTips/ExportPatient?eid='+int_id)
+    result = do_authenticated_call_to_phenotips(url,uname,pwd)
+    updated_patient_record=eval(result.read())
```
bw2: if I understand this correctly, better to use json parsing instead of eval (http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html)

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+#do a POST as a proxy for Phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
+def proxy_post(request):
+  '''to act as a proxy  '''
+  try:    
+    if len(request.POST) != 0 and request.POST.has_key('PhenoTips.PatientClass_0_external_id'):
+      project_name = request.session['current_project_name']
+      uname,pwd = get_uname_pwd_for_project(project_name)
+      __process_sync_request_helper(request.POST['PhenoTips.PatientClass_0_external_id'],
```
bw2: I'm a bit confused here... the main purpose of __process_sync_request_helper is to log the patient info in mongo PHENOTIPS_EDIT_AUDIT, right?  If yes, I would argue for putting this after line 142. 
Also, __process_sync_request_helper does the ExportPatient API call to phenotips. Since hte user's changes to the phenotips record are in the POST request params, why not just get those and store them in the DB to avoid the extra api call? (this should still give us the audit trail of who changed what)

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
-  return JsonResponse(message)
+#test function to see if we can proxy phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
+def fetch_phenotips_edit_page(request,eid):
+  '''test function to see if we can proxy phenotips'''  
+  current_user = request.user
+  
+  if request.GET.has_key('project'):
+    project_name=request.GET['project']
```
bw2: woudl call it project_id rather than project_name

xbrowse_server/base/management/commands/add_individuals_to_project.py
harindra-a: good point, added-in and committed changes as suggested. I initially went the other way to reduce code by skipping a "if" statement by making female default and changing only if it was male. I agree the downside of that is it being susceptible to bad nomenclature in PED

xbrowse_server/base/management/commands/add_individuals_to_project.py
harindra-a: Great, I like consistency as well. Updated "f" in "for" to be uppercase and committed. I will make a small note on the side for us to start aggregating these styling preferences to share with any new contributors so our code base stays consistent.

xbrowse_server/phenotips/utilities.py
harindra-a: Updated and committed. This is a debate I have with myself all the time. Doc tags vs method comments. In the past I have done both on the same method, thinking method comments are for active developers and doc strings being for developers simply importing in the code as a library. That turned a bit ugly with mostly the same comment appearing twice, so I sort of tried out method comments only. I think if you prefer doc strings, we can stick to that exclusively since that would be the best of both worlds (library users and active developers).

xbrowse_server/templates/family/family_home.html
harindra-a: OK great. Yeah, I have used jquery and fancybox in the past for modals, but used bootstrap modal here for general consistency with our code base and liked it a lot. It was pretty easy to use and flexible.

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
 
 
+#do a GET as a proxy for Phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
+def proxy_get(request):
+  '''to act as a proxy for get requests '''
+  project_name = request.session['current_project_name']
+  project_phenotips_uname,project_phenotips_pwd = get_uname_pwd_for_project(project_name)
+  try:
```
bw2: since user could hypothetically modify the request to change the project, I think the 
```
project = get_object_or_404(Project, project_id=project_name)
if project.can_admin(request.user):
   ...
```
set of check needs to be repeated here also.

xbrowse_server/phenotips/views.py
```diff
@@ -1,90 +1,187 @@
...
+    logger.error('phenotips.views:'+ext_id['error'])
+    raise Http404    
 
 
+#do a GET as a proxy for Phenotips
+#exempting csrf here since phenotips doesn't have this support
+@csrf_exempt
+def proxy_get(request):
+  '''to act as a proxy for get requests '''
+  project_name = request.session['current_project_name']
```
bw2: I know phenotips account setup prevents user access to other people's projects, but as defensive programming, might be good to also repeat these checks level here in xbrowse:
```
project = get_object_or_404(Project, project_id=project_name)
if project.can_admin(request.user):
   ...
```

(And also in the other proxy_ method)

xbrowse_server/templates/family/family_home.html
harindra-a: I agree, I think I was looking at the 1kg dataset and thought the nomenclature for individual names were standardized and the risk of "_" being in a name was minimum. Though after yesterdays meeting with Hayley, I am thinking it might not be as minimum. For now changed it to a "|" separator. If you feel uncomfortable about using a separated string like this, I could generate a unique number per individual, and then a local cache that maps that number to id + project for functions downstream to use. That is a bit of extra code overhead and compute time for browser, so didn't go for it as the first solution and went with the simpler separated string. Let me know if you prefer the latter strategy. For now committed the changes using a "|" separator.

xbrowse_server/templates/family/family_home.html
harindra-a: Yes, good call,  and the way to go.. updated and committed. The project name is common to all individuals, so didn't need to be passed in with each click. Following was cleaner/simpler.

			sessionStorage.setItem('currentInFocusIndividualId',this.id)
			var patientId=this.id;
			var project_name="{{ project.project_id }}";
			showEditPageForThisPatient(patientId,project_name);

xbrowse_server/templates/family/family_home.html
harindra-a: yeah, I prefer it separate as well and that's how the code originally was as you found. But there is something different in the way xBrowse is serving static code that broke it. For testing I had re-inserted it.I will pull it out again soon. On the same note, at some point we need to stop xBrowse serving static content and outsource that piece to apache HTTPD www. For when we start scaling up.

xbrowse_server/templates/family/family_home.html
harindra-a: Done and committed in. Some animation to fill up this lag in loading time was on my wish-list for a while now and I was actually putting it off till some time cleared up, but this about:blank is a good alternative till we get time to add a "loading" animation/page"

xbrowse_server/templates/family/family_home.html
harindra-a: Taken out and committed. Good catch, it is no longer used, and left-over from an old function that used to use it and has been since deprecated.

xbrowse_server/phenotips/utilities.py
harindra-a: This ties into our strategy of having 2 phenotips accounts per project. 1 edit/view (manager) and the other view-only. In earlier discussions it was decided that there would be only 2 types of Phenotips uses. Mangers (owners of patient) who can edit/view and everybody else who can ONLY view. Now, the nomenclature might be misleading here I think. Any number of other users can be added per patient in varying roles. The decision was to add as read-only. 

I already made all the method docs into doc strings. Will stick to this across our code base for consistancy.

xbrowse_server/phenotips/views.py
harindra-a: thanks! Sessions came in pretty handy. In fact we might have to rely on this at some point to speed  up our system by minimizing Phenotip login connections to 1 per session.

xbrowse_server/phenotips/views.py
harindra-a: Done and committed. These are initialization habits left over from java and forgetting Python scope is simpler!

xbrowse_server/phenotips/views.py
harindra-a: Done, committed.

xbrowse_server/phenotips/views.py
harindra-a: Agreed, changed and committed.

xbrowse_server/phenotips/views.py
harindra-a: Done and committed. Didn't think it needed one here since it has no URL mappings. but good to be safe. Added logging too.

xbrowse_server/phenotips/views.py
harindra-a: Done and committed. Didn't think it needed one here since it has no URL mappings. but good to be safe. Added logging too.

xbrowse_server/phenotips/views.py
harindra-a: Done and committed. I think I might have got this snippet from elsewhere in xBrowse. but agree, about standardization. Will stick to  return HttpResponse('unauthorized') going forward.

xbrowse_server/base/management/commands/add_project.py
harindra-a: It's given as part of the "first name" of the username(s) created for that project in Phenotips. You should be able to see it in the top right hand corner

xbrowse_server/phenotips/utilities.py
harindra-a: -Switched to "get_uname_pwd_for_project", this was an old iteration of code that was missed.
-Regarding merging the two types of user creation methods: Maybe the naming was ambiguous here. User creation happens ONLY in method "add_new_user_to_phenotips". While method "add_read_only_collaborator_phenotips_patient" is for associating an existing phenotips username to a patient, with ONLY read-only capabilities. The 2 underlying API calls and purpose of methods are vastly different, so merging is probably not a good idea and not reduce any code. Does that clarify things? If you still feel strongly about this, if so I could rename these 2 functions around to help reduce any ambiguity.

xbrowse_server/phenotips/utilities.py
harindra-a: Yup, good point. Done and committed. Just old code from an older iteration.

xbrowse_server/base/management/commands/add_project.py
harindra-a: Actually I would rather not use the uname from that method here since this field is basically the first_name of the user generated and get's displayed on the top right corner of phenotypes. But I agree in taking out the business logic here. I made a new method called get_names_for_user that returns the correct credentials as a tuple. Committed in.

xbrowse_server/phenotips/views.py
harindra-a: I think this relates to a Phenotips proxy quirk since calls come on different auth levels here. While this is a good idea, I think it breaks the proxy strategy. Let's chat about his in person soon and maybe figure out a compromise.

xbrowse_server/templates/family/family_home.html
bw2: agree HTTPD is faster, but as is it's easier to run a dev server.

xbrowse_server/templates/family/family_home.html
bw2: cool yeah animation would be nice

xbrowse_server/phenotips/views.py
harindra-a: We actually talked about the __process_sync_request_helper using the POST vs the API call in a sprint meeting few weeks ago, and we decided on using the API since that reduced the dependency on the UI fields changing and not having to parse the POST too much. As of now, the JSON object the API sends back get's saved in Mongo quite easily along with other meta data (date, xBrowse username etc). Let me know if you prefer the former and I can put in the change. Though I think the API call might be cleaner here..

As for moving the sync request, you were correct, good catch, I moved the whole block further down.

xbrowse_server/phenotips/views.py
harindra-a: Yeah, I agree, I am not a fan of eval either. I always use JSON parsing, but unfortunately that did't work well here due JSON enoding on the phenotips side sending back backslashes (http://stackoverflow.com/questions/10314715/why-is-json-encode-adding-backslashes). I took out the JSON code and evaluated into a dict to make it work. Since this JSON is coming from the server side from Phenotips (and not from the external internet), I trusted it for during testing, but I agree this might be a hole where people can save something bad.

I put some time into this again and found a way around it via,
updated_patient_record=json.dumps(json.JSONDecoder().decode(result.read()))

This solved the issue. Took out eval, put above in, committed.

xbrowse_server/phenotips/utilities.py
harindra-a: Done, committed, dealing with exceptions as they happen

PR #189, opened 2016-05-04
Integration

bw2:


settings.py
```diff
@@ -224,7 +224,7 @@
 
 # when set to None, this *disables* the PhenoTips interface for all projects. If set to a list of project ids, it will
 # enable the PhenoTips interface for *all* projects except those in the list.
-PROJECTS_WITHOUT_PHENOTIPS = None
+PROJECTS_WITHOUT_PHENOTIPS = ('test')
```
bw2: we should probably leave this None so phenotips integration is disabled by default

PR #296, opened 2018-03-01
Gene level notes

hanars:


hanars:
I still need to fix a few small bugs with saving variant notes and how things look in modals, but most of this is ready to go

xbrowse_server/api/utils.py
```diff
@@ -253,6 +253,12 @@ def add_extra_info_to_variants_project(reference, project, variants):
     add_clinical_info_to_variants(variants)
 
 
+def add_notes_to_genes(genes, user):
+    for gene in genes:
+        notes = list(GeneNote.objects.filter(gene_id=gene['gene_id']).order_by('date_saved'))
```
bw2: minor point - no need to convert to list here

xbrowse_server/api/views.py
```diff
@@ -661,6 +662,86 @@ def add_or_edit_variant_tags(request):
...
 
 
+@login_required
+@log_request('delete_gene_note')
+def delete_gene_note(request, note_id):
+    ret = {
+        'is_error': False,
+    }
+
+    notes = GeneNote.objects.filter(id=note_id)
```
bw2: another minor point, but single object lookups, may be better to use this pattern instead of GeneNote.objects.filter:
```
try:
    note = GeneNote.objects.get(id=note_id)
    ...
except ObjectDoesNotExist:
    ret['is_error'] = True
```

xbrowse_server/api/views.py
```diff
@@ -661,6 +662,86 @@ def add_or_edit_variant_tags(request):
...
+    if not form.is_valid():
+        return JSONResponse({
+            'is_error': True,
+            'error': server_utils.form_error_string(form)
+        })
+
+    if form.cleaned_data.get('note_id'):
+        event_type = "edit_gene_note"
+
+        notes = GeneNote.objects.filter(
```
bw2: would again use `GeneNote.objects.get` instead of `GeneNote.objects.filter`, and query just by note_id since that should be unique. Maybe it can go in a `get_note_or_error(..)` func. that does both the lookup and permissions check.

xbrowse_server/api/forms.py
```diff
@@ -283,3 +283,9 @@ def clean(self):
                 continue
             cleaned_data['project_tags'].append(ProjectTag.objects.get(project=self.project, tag=tag_text))
         return cleaned_data
+
+
+class GeneNoteForm(forms.Form):
+    gene_id = forms.CharField(max_length=20)
+    note_text = forms.CharField(max_length=1000)
```
bw2: 1000 seems short. Would increase or change to a TextField?

xbrowse_server/staticfiles/js/base/gene_details.js
```diff
@@ -17,15 +18,57 @@ window.GeneDetailsView = Backbone.View.extend({
 
     template: _.template($('#tpl-gene-modal-content').html()),
 
+    bindEvents: function(){
```
bw2: can you use `events:` here?
Similar to `select_variants.js`
```
    events: {
        "change #variant-presets-select": "standardSelectChange",
        "click a.toggle-annotation-details": "toggleAnnotDetails",
        ...
    }

```

xbrowse_server/staticfiles/js/base/add_note.js
```diff
@@ -0,0 +1,122 @@
...
+	        submit_to_clinvar: this.$('#submit_to_clinvar').is(':checked'),
+            tags: '',
+        };
+    }
+});
+
+window.AddOrEditGeneNoteView = AddOrEditNoteView.extend({
+
+    note_type: 'Gene',
+    allow_clinvar_submission: false,
```
bw2: no need for clinvar submission field for GeneNotes

xbrowse_server/staticfiles/js/base/xbrowse.js
```diff
@@ -26,7 +27,10 @@ window.ModalQueueView = Backbone.View.extend({
     },
 
     hide: function() {
-    	this.$('.modal').modal('hide');
+        // Only hide if not already hidden
+        if (this.$('.modal').hasClass('in')) {
```
bw2: thanks for fixing modals!

xbrowse_server/api/forms.py
```diff
@@ -283,3 +283,9 @@ def clean(self):
...
+    note_id = forms.CharField(max_length=10, widget=forms.HiddenInput, required=False)
```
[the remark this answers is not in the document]
hanars: for edits note_id would be present

PR #300, opened 2018-03-06
Update constraint scores

hanars:
This refactors the way we do reference_settings so that when files change in the future you only need to change one constant instead fo three settings files.

This also changes the gene tags to use an updated constraint scores file that doesn't exist yet


anneodonnell:
Here are the updated gnomAD constraint scores from Konrad. I think this is the right ticket to put this in but I can make a new one if needed. https://storage.googleapis.com/gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_gene.txt.bgz

hanars:
Hi Anne, this is actually an old pull request, not a ticket. I don;t think there is a ticket for updating constraint scores, so please make a new one

PR #335, opened 2018-03-26
Functional data

hanars:


xbrowse_server/api/forms.py
```diff
@@ -285,6 +285,21 @@ def clean(self):
         return cleaned_data
 
 
+class VariantFunctionalDataForm(forms.Form):
+    class TagsField(forms.TypedMultipleChoiceField):
+        def to_python(self, value):
+            return value
+
+        def valid_value(self, value):
+            return type(value) is dict and {'tag', 'metadata'}.issuperset(value) and type(value.get('tag')) in (unicode, basestring)
```
bw2: cool use of issuperset

xbrowse_server/api/views.py
```diff
@@ -612,11 +613,9 @@ def add_or_edit_variant_tags(request):
             form.cleaned_data['xpos'],
             form.cleaned_data['ref'],
             form.cleaned_data['alt'],
+            force_create=True
```
bw2: I think it's better to leave the `force_create` logic outside of `get_single_variant(..)` since it only occurs in 2 places. 
A `get` method with 2 possible return values (the found variant or None) seems easier to understand.

PR #343, opened 2018-04-03
Single page app

hanars:


ui/redux/utils/reducerFactories.js
```diff
@@ -240,3 +240,46 @@ export const createObjectsByIdReducer = (updateActionType, initialState = {}, de
 
   return reducer
 }
+
+
+/**
+ * Factory function that creates a reducer for managing a state object that looks like:
+ *
+ * {
+ *    loading: false,
```
bw2: I'd vote for renaming `loading` => `isLoading` and `error` => `errorMessage` so it's easier to guess the var type.

ui/redux/rootReducer.js
```diff
@@ -0,0 +1,67 @@
...
+      (e) => { throw new SubmissionError({ _error: e.message }) },
+    ).post(values)
+  }
+}
+
+// root reducer
+const rootReducer = combineReducers(Object.assign({
+  projectCategoriesByGuid: createObjectsByIdReducer(UPDATE_PROJECT_CATEGORIES_BY_GUID),
+  projectsByGuid: createObjectsByIdReducer(RECEIVE_PROJECTS),
+  projectsLoading: loadingReducer(REQUEST_PROJECTS, RECEIVE_PROJECTS),
```
bw2: Similarly would rename this to getProjectsIsLoading

ui/redux/rootReducer.js
```diff
@@ -0,0 +1,67 @@
...
+  user: zeroActionsReducer,
+  form: formReducer,
+}, dashboardReducers))
+
+export default rootReducer
+
+// basic selectors
+export const projectsLoading = state => state.projectsLoading.loading
```
bw2: Similarly would rename this to getProjectsIsLoading

ui/redux/rootReducer.js
```diff
@@ -0,0 +1,67 @@
...
+      },
+      e => dispatch({ type: RECEIVE_PROJECTS, error: e.message, updatesById: {} }),
+    ).get()
+  }
+}
+
+export const updateProject = (values) => {
+  return (dispatch) => {
+    const urlPath = values.projectGuid ? `/api/project/${values.projectGuid}` : '/api/project'
+    const projectField = values.projectField ? `_${values.projectField}` : ''
```
bw2: Can this line be deleted and then just do
`${urlPath}/${action}_project_${values.projectField}` ?

Also would it be better to just return immediately if values.projectField is null?

It might be worth adding /** .. */ docs explaining what the `values` arg is.

hanars: well most of the time you are updating  projects with no specific fields so not having a project field makes sense and you end up with urls like `create_project`, `update_project`etc 

The reason projectField exists as an option at all is for `update_project_categories` which needs to be a separate url

PR #347, opened 2018-04-09
Project page spa

hanars:


seqr/views/apis/individual_api.py
```diff
@@ -118,42 +114,101 @@ def edit_individuals_handler(request, project_guid):
...
-    if modified_individuals_by_guid is None:
-        return create_json_response(
-            {}, status=400, reason="'modifiedIndividuals' not specified")
+    update_individuals = {
+        ind['individualGuid']: (Individual.objects.get(guid=ind['individualGuid']), ind) for ind in modified_individuals_list
+    }
 
-    # edit individuals
-    modified_individuals_list = list(modified_individuals_by_guid.values())
+    errors = []
```
bw2: what's the reason for switching away from `validate_fam_file_records` here?

seqr/views/apis/individual_api.py
```diff
@@ -118,42 +114,101 @@ def edit_individuals_handler(request, project_guid):
...
+
+        if created:
+            logger.info("Created family: %s", family)
+            if not family.display_name:
+                family.display_name = family.family_id
+                family.save()
+
+        families.append(family)
+
+        if family_id != model.family.family_id:
```
bw2: when would this happen?

seqr/views/apis/individual_api.py
```diff
@@ -118,42 +114,101 @@ def edit_individuals_handler(request, project_guid):
...
-    # TODO more validation
-    errors, warnings = validate_fam_file_records(modified_individuals_list)
     if errors:
-        return create_json_response({'errors': errors, 'warnings': warnings})
+        return create_json_response({'errors': errors}, status=400, reason='Invalid updates')
 
-    updated_families, updated_individuals = add_or_update_individuals_and_families(
-        project, modified_individuals_list)
+    individuals_by_guid = {}
+    families = []
```
bw2: would vote for just creating families_by_guid = {} here and populating it in the loop

seqr/views/utils/json_to_orm_utils.py
```diff
@@ -48,12 +49,12 @@ def update_family_from_json(family, json, verbose=False):
     _update_model_from_json(family, json, FAMILY_JSON_FIELD_MAP, verbose=verbose)
 
 
-def update_individual_from_json(individual, json, verbose=False, allow_unknown_keys=False):
+def update_individual_from_json(individual, json, verbose=False, allow_unknown_keys=False, save=True):
```
bw2: what would be a reason to set save=False? seems like it's currently always =True

seqr/views/utils/json_utils.py
```diff
@@ -36,13 +37,15 @@ def render_with_initial_json(html_page, initial_json):
...
     )
-
     html = loader.render_to_string(html_page)
 
     html = html.replace(
         "window.initialJSON=null",
         "window.initialJSON="+initial_json_str
     )
+
+    import pdb; pdb.set_trace()
```
bw2: haven't use this before. would be cool to see how you use it.

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -5,11 +5,25 @@
...
 import os
+from django.db.models.fields.files import ImageFieldFile
 
+from seqr.models import CAN_EDIT
+from seqr.views.utils.json_utils import _to_camel_case
 from family_info_utils import retrieve_family_analysed_by
 logger = logging.getLogger(__name__)
 
 
+def _get_json_for_record(record, fields, processed_fields={}, get_record_field=None, get_parent_guid=None):
```
bw2: Not sure I understand the reason for the processed_fields, get_record_field, get_parent_guid args or why the added complexity is justified for these getters. Would be nice to whiteboard it.

seqr/urls.py
```diff
@@ -65,15 +64,13 @@
 from seqr.views.apis.variant_search_api import query_variants_handler
 
 react_app_pages = [
-    'dashboard'
+    'dashboard',
+    'project/(?P<project_guid>[^/]+)/.*'
```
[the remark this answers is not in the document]
hanars: django gives priority to exact matches over `*` matches so it shouldn't be an issue

seqr/views/apis/individual_api.py
hanars: If you updated the `family_id` in the bulk update for individuals

seqr/views/utils/json_utils.py
hanars: it sets a breakpoint and drops you in a python interpreter at that point. Should definitly not be committed though

seqr/views/utils/orm_to_json_utils.py
hanars: I'm not convinced the added complexity is justified, tbh. The idea is that the same function would be used for getting all the record fields from an object and from a database cursor because there were a lot of bugs coming up because they weren't quite identical

seqr/views/apis/individual_api.py
hanars: It assumes that this is the first upload and that all individuals you care about are in the update data. So for example, if you modify an individual and not their child `validate_fam_file_records` would not check if the sex/famly for that individual were still valid because it would not know that the unmodified child exists.

seqr/views/apis/individual_api.py
hanars: I'll change this to a dict by guid here, but we need to do the actual `_get_json` call after all the updates are done so the individual list is accurate

seqr/views/utils/json_to_orm_utils.py
hanars: I think I was testing a different approach that I backed out of and forgot this. I'll remove it

seqr/views/utils/json_to_orm_utils.py
```diff
@@ -1,69 +1,34 @@
...
+def _update_model_from_json(model_obj, json, user=None, verbose=False, allow_unknown_keys=False):
     modified = False
     for json_key, value in json.items():
-        if json_key in unknown_keys:
+        orm_key = _to_snake_case(json_key)
+        if allow_unknown_keys and not hasattr(model_obj, orm_key):
             continue
-        orm_key = json_field_map[json_key]
         if getattr(model_obj, orm_key) != value:
+            if orm_key in model_obj._meta.internal_json_fields and not (user and user.is_staff):
```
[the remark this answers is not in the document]
hanars: yes

ui/pages/Project/utils/histogramSelectors.js
```diff
@@ -1,47 +0,0 @@
...
-/**
- * function that returns a mapping of familySize => number of families
- *
- * @param state {object} global Redux state
- */
-export const getFamilySizeHistogram = createSelector(
-  getFamilyGuidToIndividuals,
-  (familyGuidToIndividuals) => {
-    const familySizes = Object.values(familyGuidToIndividuals)
-      .map(individualsInFamily => Math.min(individualsInFamily.length, 5))
```
hanars: @bw the logic for familySizeHistogram came directly from here, which had the limit of 5

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -71,189 +29,70 @@ const ANALYSIS_TYPE_LABELS = {
...
-    </Grid>
-    <Grid stackable style={{ margin: '0px' }}>
-      <Grid.Column width={12} style={{ paddingLeft: '0' }}>
-        <SectionHeader>
-          Variant Tags
-        </SectionHeader>
-        <div style={{ display: 'block', padding: '0px 0px 10px 0px' }}>
+const ProjectOverview = (props) => {
+  const familySizeHistogram = props.families
+    .map(family => Math.min(family.individualGuids.length, 5))
```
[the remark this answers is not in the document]
hanars: I commented on the file this was copied from

ui/pages/Dashboard/components/table-body/EditProjectCategoriesModal.jsx
```diff
@@ -13,16 +13,17 @@ const EditProjectCategoriesModal = (props) => {
   const categories = Object.values(props.projectCategoriesByGuid).map((projectCategory) => {
     return { value: projectCategory.guid, text: projectCategory.name, key: projectCategory.guid }
   })
+  const NAME = `editProjectCategories-${props.project.projectGuid}`
```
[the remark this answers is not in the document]
hanars: I'm not sure what you mean

ui/pages/Project/components/FamilyTable/FamilyRow.jsx
```diff
@@ -0,0 +1,188 @@
...
+            <br />
+          </span>
+          <br />
+          <PedigreeImagePanel family={props.family} />
+        </Grid.Column>
+
+        <Grid.Column width={10} style={{ maxWidth: '950px' }}>
+          <TextFieldView
+            isVisible={props.showDetails}
+            isEditable={props.project.canEdit && !props.showInternalFields}
```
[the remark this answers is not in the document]
hanars: the `showInternalFields` boolean here is really used to differentiate between the case review page, where we want users to interact with those fields, and the main family page, where we do not. For all of the "core" fields, we want them to show up on the case review page but we don't want them being edited, the only fields that should be edited there are the internal ones. The configurability here is not about user preference, its about intended functionality of these pages

ui/pages/Project/utils/histogramSelectors.js
bw2: 🤦‍♂️

ui/pages/Project/components/FamilyTable/FamilyRow.jsx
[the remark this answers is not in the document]
bw2: I agree this works, but I think generally we want parameters defined & named in a way that makes it easy to guess how they affect component/function/or class behavior. If the code is understandable without knowing anything about the places that currently call it, I think this makes the set of components easier to work with, more robust, etc.

ui/pages/Dashboard/components/table-body/EditProjectCategoriesModal.jsx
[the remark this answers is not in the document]
bw2: sorry I meant either rename this const to camelCase (eg. "modalName"), or just use  the value (`editProjectCategories-${props.project.projectGuid}`) directly instead of creating a const first.

ui/pages/Project/components/FamilyTable/FamilyRow.jsx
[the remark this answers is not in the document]
hanars: I've refactored this-let me know what you think

ui/pages/Dashboard/components/table-body/EditProjectCategoriesModal.jsx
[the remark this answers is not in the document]
hanars: oh got it. I fixed that

ui/pages/Project/components/FamilyTable/FamilyRow.jsx
```diff
@@ -0,0 +1,170 @@
...
+export const INTERNAL_NOTES = 'internalCaseReviewNotes'
+export const INTERNAL_SUMMARY = 'internalCaseReviewSummary'
+
+const fieldRenderDetails = {
+  [DESCRIPTION]: { name: 'Family Description' },
+  [ANALYSIS_STATUS]: { component: AnalysisStatus },
+  [ANALYSED_BY]: { component: AnalysedBy },
+  [ANALYSIS_NOTES]: { name: 'Analysis Notes' },
+  [ANALYSIS_SUMMARY]: { name: 'Analysis Summary' },
+  [INTERNAL_NOTES]: { name: 'Internal Notes', internal: true },
```
bw2: Editing internal notes doesn't appear to work on Case Review page. 
Edit form appears, can be submitted, and generates a request, but the entered text doesn't appear on the page (even after refreshing)

ui/pages/Project/components/ProjectCollaborators.jsx
```diff
@@ -0,0 +1,48 @@
...
+                   <i><a href={`mailto:${c.email}`}>{c.email}</a></i> : null
+              }
+
+            </Table.Cell>
+            <Table.Cell style={{ padding: '2px 10px', textAlign: 'center', verticalAlign: 'top' }} className="noBorder">
+              <Popup
+                position="top center"
+                trigger={<b style={{ cursor: 'pointer' }}> {c.hasEditPermissions ? ' † ' : ' '}</b>}
+                content={"Has 'Manager' permissions"}
+                size="small"
```
bw2: extra line-breaks appearing in collaborator lists. 
Example showing only 1st letter:

image

ui/pages/Project/components/FamilyTable/FamilyRow.jsx
bw2: Same thing for a few other Family-level fields: Analysis Notes, Analysis summary.

bw2: Also, rich text formatting isn't getting saved:
image

bw2:
Looks great to me. Amazing work!

PR #351, opened 2018-04-11
seqr model field updates

bw2:
combine variant_annotation and variant_genotypes into a single field: saved_variant_json, and also add disease_area and is_functional_data_enabled


hanars:
Looks good to me!

PR #379, opened 2018-05-07
added HGMD and other annotations

bw2:


xbrowse/core/constants.py
```diff
@@ -394,12 +420,16 @@
     'name': 'intergenic_variant',
     'slug': 'intergenic_variant',
     'so': 'SO:0001628'},
-
 ]
 
+ANNOTATION_DEFINITIONS = CLINVAR_ANNOTATION_DEFINITIONS + VEP_ANNOTATION_DEFINITIONS
+ANNOTATION_DEFINITIONS_INTERNAL = ANNOTATION_DEFINITIONS + HGMD_ANNOTATION_DEFINITIONS + VEP_ANNOTATION_DEFINITIONS
+
```
hanars: `ANNOTATION_DEFINITIONS` already  includes `VEP_ANNOTATION_DEFINITIONS` so you are actually adding it twice to `ANNOTATION_DEFINITIONS_INTERNAL`

xbrowse/core/constants.py
```diff
@@ -507,23 +552,35 @@
 
         ]
     }
-
 ]
 
+
+ANNOTATION_GROUPS = CLINVAR_ANNOTATION_GROUPS + VEP_ANNOTATION_GROUPS
+ANNOTATION_GROUPS_INTERNAL = CLINVAR_ANNOTATION_GROUPS + HGMD_ANNOTATION_GROUPS + VEP_ANNOTATION_GROUPS
+
```
hanars: same issue here

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -162,13 +171,18 @@ def get_elasticsearch_variants(self, project_id, family_id=None, variant_filter=
             # TODO add caching
             indiv_id = _encode_name(family_individual_ids[0])
 
-            mapping = self._es_client.indices.get_mapping(str(elasticsearch_index)+"*")
+
+            matching_indices = []
+            mapping = self._es_client.indices.get_mapping(str(elasticsearch_index)+"*,rare_genomes_project_wgs_rgp_*")
```
hanars: why are we hardcoding RGP here?

xbrowse/core/constants.py
bw2: oops.. good call

xbrowse/core/constants.py
bw2: This one is alright I think

xbrowse/datastore/elasticsearch_datastore.py
bw2: that was a test for adding SVs, wasn't meant to be committed. Will take it out.

bw2:
Fixed those things.

PR #381, opened 2018-05-09
Update phenotips to v1.4

bw2:


harindra-a:
awesome about v1.4. Let me do a live test of dev with the v1.4 on my machine

PR #382, opened 2018-05-10
Performance fixes

hanars:
This fixes a few big issues we were having with re-rendering unnecessarily:

1. I changed how `createObjectsByIdReducer` works so we can dispatch one event when we receive all the data instead of 4. This is crucial, because when we dispatch sequential events the DOM re-renders after each dispatch which was causing huge slow-downs

2. I added reselect to a bunch of selectors that were missing it

3. I updated `ReduxFormWrapper` so it only updates when relevant props change

With these optimizations I can up the number fo families per page to 25. However, I can't get rid of pagination on the edit families/ individuals form because even rendering everything exactly once it takes too long

I put this up against the project-page-spa branch to make reviewing easier, but we probably want to deploy these together


ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
```diff
@@ -26,59 +25,50 @@ const CaseReviewStatusDropdown = props =>
...
-        component="select"
-        tabIndex="1"
-        style={{ margin: '3px !important', maxWidth: '170px', display: 'inline-block', padding: '0px !important', marginRight: '10px' }}
-      >
-        {
-          CASE_REVIEW_STATUS_OPTIONS.map(option => <option key={option.value} value={option.value}>{option.name}</option>)
-        }
-      </Field>
-    </ReduxFormWrapper>
+      fields={[{
```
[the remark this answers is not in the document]
hanars: you can't generally do this, this is something I built into the redux form wrapper component. The advantage over jsx is it doesn't just vanilla render the field, it actually does some parsing so errors are passed down to fields nicely. 

The reason I got around to changing this now is that they were rerendering unneccessarily whenever unrelated props on the form changed and this fixes that issue

ui/redux/rootReducer.js
```diff
@@ -12,27 +12,21 @@ import modalReducers from './utils/modalReducer'
  */
 
 // actions
+const RECEIVE_DATA = 'RECEIVE_DATA'
```
bw2: Calling this `RECIEVE_DATA` implies that the only way to update these datasetsByGuid or any xByGuid state trees is to receive data from the server. This seems to be true in general since we always want to save the changes on the server first before updating client state, but just to double check, do you agree that's the case?

ui/redux/utils/reducerFactories.js
```diff
@@ -215,9 +215,15 @@ export const createObjectsByIdReducer = (updateActionType, initialState = {}, de
           console.error(`Invalid ${updateActionType} action. action.updatesById is undefined: `, action)
           return state
         }
-
+        let updates = action.updatesById
```
bw2: what's the purpose of the key arg?

ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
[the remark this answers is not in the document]
bw2: got it. I need to go through this in detail. One additional question - in what cases would you want to use
fields vs. renderChildren ?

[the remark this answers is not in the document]
hanars: you would want to use fields whenever possible, but you would use render Children when the input is something that doesn't really look like a form- for example, the bulk upload "input" which is really a prompt to upload a file

ui/redux/rootReducer.js
hanars: I agree

ui/redux/utils/reducerFactories.js
hanars: It allows use to dispatch one `RECEIVE_DATA` action and have it update multiple different reducers by key, like `projectsByGuid`, `familiesByGuid`, etc

PR #388, opened 2018-05-14
Performance linting

hanars:
Adds a linting plugin that complains if we define objects/arrays/functions in props directly, and then fixes all the issues it complained about. Many of the changes had to do with passing in `style` objects, which I fixed by switching to styled components wherever possible


ui/pages/Dashboard/components/table-header/ProjectTableHeader.jsx
```diff
@@ -20,10 +20,8 @@ import {
 
 const TableHeaderCell = styled(Table.HeaderCell)`
   padding: 12px 10px 12px 3px !important; 
-  background-color: #F3F3F3 !important;
-  color: #333333 !important;
   font-weight: 500 !important;
-  height: 10px;
+  text-align: ${props => props.textAlign}
```
bw2: nice

ui/shared/components/page/PageHeader.jsx
```diff
@@ -22,6 +22,17 @@ const ProjectTitleContainer = styled.div`
...
   margin: 50px 0px 35px 0px;
   line-height: 1.2em;
+  
+  a.active {
+    color: #111;
+    font-weight: 750;
+    cursor: auto;
+  }
+`
+const NavLinkNoActive = styled(NavLink)`
```
bw2: What's the purpose of setting `display: none` when active?

PR #403, opened 2018-05-23
Sync seqr models

bw2:


seqr/model_utils.py
```diff
@@ -0,0 +1,203 @@
...
+    for key, value in field_mapping.items():
+        if _is_seqr_model(value):
+            value = find_matching_xbrowse_model(value)
+            if value is not None:
+                field_mapping[key] = value
+            else:
+                logging.info("ERROR: unable to find equivalent seqr model for %s: %s" % (key, value))
+                del field_mapping[key]
+
+    return xbrowse_kwargs
```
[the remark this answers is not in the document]
bw2: what do you mean by this?

seqr/model_utils.py
```diff
@@ -0,0 +1,203 @@
...
+def _create_xbrowse_model(seqr_model, **kwargs):
+    try:
+        seqr_model_class = seqr_model.__class__
+        seqr_model_class_name = seqr_model_class.__name__
+        xbrowse_kwargs = _convert_seqr_kwargs_to_xbrowse_kwargs(seqr_model, **kwargs)
+        xbrowse_model_class = SEQR_TO_XBROWSE_CLASS_MAPPING[seqr_model_class_name]
+        xbrowse_model_class_name = xbrowse_model_class.__name__
+        logging.info("_create_xbrowse_model(%s, %s)" % (xbrowse_model_class_name, xbrowse_kwargs))
+        xbrowse_model = seqr_model_class.objects.create(**xbrowse_kwargs)
+        seqr_model_foreign_key_name = "xbrowse_"+_to_snake_case(xbrowse_model_class_name)
```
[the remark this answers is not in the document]
bw2: oops, good point

seqr/model_utils.py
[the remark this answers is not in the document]
hanars: yeah, I think in general we should be trying to rely on explicit foreign key references because its much harder to make mistakes with those rather than having these complex per-model field mappings that are manually defined and any time you change either model you could accidentally break that lookup. Also, for the secondary query I think we should explicitly check that `seqr_locus_list` is null, because if it has a reference to a different model than we don't want it

seqr/views/apis/individual_api.py
```diff
@@ -399,9 +396,7 @@ def add_or_update_individuals_and_families(project, individual_records):
             set_patient_hpo_terms(project, individual.phenotips_eid, record.get('hpoTerms'), is_external_id=True)
 
         if not individual.display_name:
-            individual.display_name = individual.individual_id
-
-        individual.save()
+            update_seqr_model(individual, display_name=individual.individual_id)
 
         _deprecated_update_original_individual_data(project, individual)
```
hanars: if we are using `update_seqr_model` we shouldn't also need `_deprecated_update_original_individual_data`

seqr/views/apis/project_api.py
```diff
@@ -91,14 +92,10 @@ def update_project_handler(request, project_guid):
...
-        project.save()
-    if 'description' in request_json:
-        project.description = request_json.get('description')
-        project.save()
+        update_seqr_model(project, name=request_json.get('name'))
 
-    # keep new seqr.Project model in sync with existing xbrowse_server.base.models - TODO remove this code after transition to new schema is finished
-    _deprecated_update_original_project(project)
+    if 'description' in request_json:
+        update_seqr_model(project, name=request_json.get('description'))
```
hanars: typo: should be `description=` not `name=`

seqr/views/apis/project_api.py
```diff
@@ -225,20 +222,6 @@ def _deprecated_create_original_project(project):
...
-    if base_project:
-        base_project = base_project[0]
-        base_project.project_name = project.name
-        base_project.description = project.description
-        base_project.save()
-
-
 def _deprecated_delete_original_project(project):
     """DEPRECATED - delete project in original xbrowse schema to keep things in sync.
     Args:
```
hanars: probably should get rid of this too in favor of `delete_seqr_model`

seqr/model_utils.py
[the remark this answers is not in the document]
bw2: fixed

seqr/views/apis/individual_api.py
bw2: multiple models are updated here so will as is

seqr/model_utils.py
```diff
@@ -0,0 +1,226 @@
...
+    xbrowse_model = find_matching_xbrowse_model(seqr_model)
+    if not xbrowse_model:
+        return
+
+    xbrowse_kwargs = convert_seqr_kwargs_to_xbrowse_kwargs(seqr_model, **kwargs)
+
+    _update_model(xbrowse_model, **xbrowse_kwargs)
+
+
+def _to_snake_case(camel_case_str):
```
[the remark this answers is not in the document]
bw2: good point.

seqr/views/apis/individual_api.py
```diff
@@ -389,8 +388,8 @@ def add_or_update_individuals_and_families(project, individual_records, user=Non
         if created:
             # create new PhenoTips patient record
             patient_record = create_patient(project, individual.phenotips_eid)
-            individual.phenotips_patient_id = patient_record['id']
-            individual.case_review_status = 'I'
+            update_seqr_model(individual, phenotips_patient_id=patient_record['id'])
+            update_seqr_model(individual, case_review_status='I')
```
hanars: really nitpicky, but no reason not to do this in one line

seqr/views/utils/json_to_orm_utils.py
```diff
@@ -34,6 +42,16 @@ def _update_model_from_json(model_obj, json, user=None, verbose=False, allow_unk
...
+                    for key, value in xbrowse_kwargs.items():
+                        setattr(model_obj, key, value)
+            except Exception as e:
+                logger.error("Unable to update xbrowse model {0}: {1}".format(xbrowse_model, e))
 
     if modified:
         model_obj.save()
+
+        if xbrowse_model is not None:
+            xbrowse_model.save()
```
hanars: so this all works, but I have an idea that I think is more readable where instead of setting the fields as we go we keep track of all the fields that are actually getting updated and then just call `update_seqr_model` with those:

```
seqr_update_fields = {}

        orm_key = _to_snake_case(json_key)
        if allow_unknown_keys and not hasattr(model_obj, orm_key):
            continue
        if getattr(model_obj, orm_key) != value:
            if orm_key in model_obj._meta.internal_json_fields and not (user and user.is_staff):
                raise ValueError('User {0} is not authorized to edit the internal field {1}'.format(user, orm_key))
            if verbose:
                model_obj_name = getattr(model_obj, 'guid', model_obj.__name__)
                logger.info("Setting {0}.{1} to {2}".format(model_obj_name, orm_key, value))
            seqr_update_fields[orm_key] = value

    if seqr_update_fields:
        update_seqr_model(model_obj, **seqr_update_fields)
```

bw2: oh yeah, that's much nicer  👍

PR #423, opened 2018-06-15
move caching into get_elasticsearch_variants

bw2:


xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -694,14 +716,14 @@ def get_single_variant(self, project_id, family_id, xpos, ref, alt):
```
hanars: Is there a reason you got rid of caching for single variants?

PR #434, opened 2018-06-25
Fix "case review changed by" display

hanars:
I confirmed that there is no appreciable performance hit


seqr/views/pages/project_page.py
```diff
@@ -126,47 +126,22 @@ def _retrieve_families(cursor, project_guid, user):
...
-
-        FROM seqr_individual AS i
-          JOIN seqr_family AS f ON i.family_id=f.id
-          JOIN seqr_project AS p ON f.project_id=p.id
-        WHERE p.guid=%s
-    """.strip()
 
-    cursor.execute(individuals_query, [project_guid])
+    fields = Individual._meta.json_fields + Individual._meta.internal_json_fields + \
+             ['family__guid', 'case_review_status_last_modified_by__email']
```
bw2: does `case_review_status_last_modified_by__email` work when case_review_status_last_modified_by is None?

hanars: yes

PR #435, opened 2018-06-25
saved variant page tweaks

hanars:


ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -115,33 +108,55 @@ class SavedVariants extends React.Component {
     }
   }
 
+  navigateTag = (e, data) => {
+    const { familyGuid } = this.props.match.params
+    const isCategory = this.categoryOptions.includes(data.value)
+    const urlPath = `/project/${this.props.project.projectGuid}/saved_variants/${familyGuid ? `family/${familyGuid}/` : ''}`
+    const tag = data.value === ALL_FILTER ? '' : data.value
+    this.props.updateSavedVariantTable({ categoryFilter: isCategory ? data.value : null })
+    this.props.history.push(`${urlPath}${isCategory ? '' : tag}`)
```
bw2: maybe simpler to delete line 115 and do 

const tag = data.value

${isCategory || data.value === ALL_FILTER ? '' : tag }

ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -115,33 +108,55 @@ class SavedVariants extends React.Component {
     }
   }
 
+  navigateTag = (e, data) => {
```
bw2: maybe `navigateTag` => `navigateToTag` ?

ui/pages/Project/components/SavedVariants.jsx
hanars: `isCategory` is false and we need it explicitly to be the empty string in the url, so theres no way to do that without nesting ternary expressions

bw2:
can fix the tests after the merge to dev

PR #436, opened 2018-06-26
Remove dataset model

bw2:


seqr/migrations/0040_auto_20180603_1309.py
```diff
@@ -0,0 +1,105 @@
...
+    Dataset = apps.get_model('seqr', 'Dataset')
+    Sample = apps.get_model('seqr', 'Sample')
+
+    print("\nTransferring metadata from {0} Dataset records to {1} Sample records".format(
+        Dataset.objects.filter(analysis_type="VARIANTS").count(),
+        Sample.objects.count(),
+        ))
+
+    for sample in Sample.objects.all():
+        dataset = sample.dataset_set.filter(analysis_type="VARIANTS").order_by('-loaded_date', '-pk').first()
```
hanars: so this only transfers the most recent dataset right? but don't we also need to keep track of the oldest dataset so we know when it was first loaded?

seqr/models.py
```diff
@@ -327,7 +330,7 @@ def __unicode__(self):
         return self.individual_id.strip()
 
     def _compute_guid(self):
-        return 'I%06d_%s' % (self.id, _slugify(str(self)))
+        return 'I%07d_%s' % (self.id, _slugify(str(self)))
```
hanars: why change this?

seqr/models.py
```diff
@@ -397,115 +397,68 @@ class Sample(ModelWithGUID):
...
-        (ANALYSIS_TYPE_ALIGNMENT, 'Alignment'),
-        (ANALYSIS_TYPE_VARIANT_CALLS, 'Variant Calls'),
-        (ANALYSIS_TYPE_SV, 'SV Calls'),
-        (ANALYSIS_TYPE_BREAKPOINTS, 'Breakpoints'),
-        (ANALYSIS_TYPE_SPLICE_JUNCTIONS, 'Splice Junction Calls'),
-        (ANALYSIS_TYPE_ASE, 'Allele Specific Expression'),
+    DATASET_TYPE_READ_ALIGNMENTS = 'ALIGN'
+    DATASET_TYPE_VARIANT_CALLS = 'VARIANTS'
+    DATASET_TYPE_SV_CALLS = 'SV'
+    DATASET_TYPE_BREAKPOINTS = 'BREAK'
```
hanars: what is a breakpoint?

seqr/models.py
```diff
@@ -397,115 +397,68 @@ class Sample(ModelWithGUID):
...
-    # is added a second time, it would be assigned the same dataset id as before.
-    # This will allow datasets to be processed and loaded only once, but shared between projects if
-    # needed by using the same dataset_id in the Dataset records of different projects.
-    dataset_id = models.TextField(null=True, blank=True, db_index=True)   # elasticsearch index
+    sample_type = models.CharField(max_length=20, choices=SAMPLE_TYPE_CHOICES, null=True, blank=True)
+    dataset_type = models.CharField(max_length=20, choices=DATASET_TYPE_CHOICES, null=True, blank=True)
 
-    dataset_location = models.TextField(null=True, blank=True, db_index=True)
+    # The sample's id in the underlying dataset (eg. the VCF Id for variant callsets).
+    sample_id = models.TextField(db_index=True)
```
hanars: should these be unique?

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -213,28 +213,6 @@ def _get_json_for_sample(sample):
     return result
 
 
-def _get_json_for_dataset(dataset, add_sample_type_field=True):
```
hanars: I feel like the fact you didn't need to make any changes to `_get_json_for_samples` is a sign we built this pretty well

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -54,20 +54,23 @@ const ProjectOverview = (props) => {
           }
           {props.project.canEdit ? <span><br /><EditFamiliesAndIndividualsButton /></span> : null }<br />
         </div>
+        {/*
```
hanars: I'd rather this was deleted than commented out. These kinds of comments actually render to the DOM

ui/pages/Project/selectors.js
```diff
@@ -51,16 +49,20 @@ export const getProject = createSelector(
...
 
-export const getProjectDatasets = createSelector(getDatasetsByGuid, getProjectGuid, filterProjectEntities)
+export const getProjectFamiliesByGuid = createSelector(getFamiliesByGuid, getProjectGuid, filterProjectEntities)
+export const getProjectIndividualsByGuid = createSelector(getIndividualsByGuid, getProjectGuid, filterProjectEntities)
+export const getProjectSamplesByGuid = createSelector(getSamplesByGuid, getProjectGuid, filterProjectEntities)
 
-export const getProjectSamples = createSelector(getSamplesByGuid, getProjectGuid, filterProjectEntities)
+export const getProjectFamilies = createSelector(getProjectFamiliesByGuid, d => Object.values(d))
+export const getProjectIndividuals = createSelector(getProjectIndividualsByGuid, d => Object.values(d))
+export const getProjectSamples = createSelector(getProjectSamplesByGuid, d => Object.values(d))
```
hanars: I think it would be better if we got rid of these and just used the `getProject<Thing>ByGuid` selectors everywhere for consistancy

ui/shared/components/graph/HorizontalStackedBar.jsx
```diff
@@ -9,11 +9,15 @@ import { Link } from 'react-router-dom'
 import { ColoredIcon } from '../StyledComponents'
 
 
-const BarContainer = styled.div`
+const BarContainer = styled.div.attrs({
+  w: (props) => { return props.width ? `${props.width}px` : '100%' },
+  h: (props) => { return props.height ? `${props.height}px` : 'auto' },
+  lh: (props) => { return props.height ? `${props.height - 2}px` : 'inherit' },
+})`
```
hanars: I like it, but out of curiosity is this just a code style preference or is there actual performance/ bnug implications of this?

ui/shared/constants/sampleConstants.js
```diff
@@ -0,0 +1,2 @@
+
+export const DATASET_TYPE_VARIANT_CALLS = 'VARIANTS'
```
hanars: why not just put this in the main constants file?

seqr/models.py
```diff
@@ -98,6 +98,8 @@ class Project(ModelWithGUID):
     can_edit_group = models.ForeignKey(Group, related_name='+', on_delete=models.PROTECT)
     can_view_group = models.ForeignKey(Group, related_name='+', on_delete=models.PROTECT)
 
+    genome_version = models.CharField(max_length=5, choices=GENOME_VERSION_CHOICES, default=GENOME_VERSION_GRCh37)
```
hanars: I like this. Do we want to update the saved variant page to use this instead of the saved_variant_json version, or do we want to do that in a separate PR?

seqr/models.py
bw2: it's unique relative to all sample ids in the VCF or elasticsearch index, but not necessarily globally unique

seqr/models.py
bw2: instead of fully calling structural variants (which includes 2 or more break-points + deciding if it's a deletion, insertion, inversion, etc.), you can go halfway and just detect the DNA breakpoints

seqr/migrations/0040_auto_20180603_1309.py
bw2: yes. I didn't find a way to create new Sample records during the migration, so left that to update_projects_in_new_schema

ui/shared/components/graph/HorizontalStackedBar.jsx
bw2: not sure about actual performance gains, but in dev mode StyledComponents generates a warning suggesting this change.

ui/shared/constants/sampleConstants.js
bw2: oops, good point

seqr/models.py
bw2: whichever you prefer

seqr/models.py
bw2: I increased other ones, so figured it wouldn't hurt

ui/pages/Project/selectors.js
bw2: Sure, and probably also for getProjectSavedVariants?

bw2:
just committed PR fixes. Still needs some tests, but let me know what you think.

seqr/models.py
hanars: lets do it now so we don't forget

hanars:
can you fix the codacy fails, a couple of them look ligit

otherwise looks good

bw2:
yup, much of that dataset_utils and dataset_api code is broken, but it's also not currently used anywhere. I originally wrote it for the "Upload Dataset" form which was on the new Project page. I'm thinking once your new Project page PR is in, we can restore/fix the form + api code so that it replaces the add_vcf.py command to create all the metadata & records for a new dataset after it done loading into elasticsearch. (The form and handler code should be changed to take an elasticsearch index name and get all sample ids from there instead of from the original VCF file).

PR #438, opened 2018-06-26
Gene page ui tweaks

hanars:


ui/shared/components/panel/genes/GeneExpression.jsx
```diff
@@ -32,55 +67,85 @@ const xcoord = (d) => {
...
-  )
+  render() {
+    const { expression } = this.props
+    if (!expression) {
+      return <p>Expression data not available for this gene.</p>
+    }
+
+    const tissues = Object.keys(expression).sort()
+    const tissueExpression = tissues.reduce((acc, tissue) =>
+      ({ ...acc, [tissue]: expression[tissue].filter(d => d > 0) }), {},
```
bw2: note: we'll want to move this filter to when preprocessing/loading the expression matrix into elasticsearch

PR #440, opened 2018-06-27
Edit family from saved variant page

hanars:


ui/shared/components/PopupWithModal.jsx
```diff
@@ -0,0 +1,49 @@
...
+import { connect } from 'react-redux'
+import { Popup } from 'semantic-ui-react'
+
+import { getOpenModals } from 'redux/utils/modalReducer'
+
+const BehindModalPopup = styled(Popup)`
+  z-index: 500 !important;
+`
+
+class PopupWithModal extends React.PureComponent {
```
bw2: Why is a PopupWithModal component needed?

ui/shared/components/PopupWithModal.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+  static propTypes = {
+    openModal: PropTypes.bool,
```
bw2: why is the openModal prop necessary?

ui/shared/components/panel/variants/VariantFamily.jsx
```diff
@@ -159,7 +162,7 @@ const VariantFamily = ({ variant, project, family, individualsByGuid }) => {
       <IndividualCell>
         <Header size="small">
           Family<HorizontalSpacer width={5} />
-          <Popup
+          <PopupWithModal
```
bw2: didn't see this in the previous PR, but would vote for naming conventions that indicate the object type (eg. something like "VariantFamilyPanel" or "VariantFamilyInfoPanel" instead of "VariantFamily")

ui/shared/components/PopupWithModal.jsx
hanars: because when you click the edit button a modal popped up, but then when you would move your mouse to the modal the popup would close and the modal would unmount

ui/shared/components/PopupWithModal.jsx
hanars: 1) this is state, not a prop
2) so that the popup won't close while the modal is open

ui/shared/components/panel/variants/VariantFamily.jsx
hanars: okay, but that seems really unrelated to the changes in this PR so I'd rather do a separate one where we rename all the files with that convention

ui/shared/components/PopupWithModal.jsx
bw2: can the modal be mounted/rendered by some top-level component outside the popup, and just shown/hidden based on a boolean in redux state?

ui/shared/components/panel/variants/VariantFamily.jsx
bw2: sounds great

ui/shared/components/PopupWithModal.jsx
bw2: (I'll merge as is now, and if you agree, this can also go in a future PR)

hanars: technically yes, but not without entirely refactoring how `BaseFieldView` works and the redux state bit would be kind of messy. I'm also not sure why thats any cleaner than this. 

The modal isn't rendering in the DOM inside the popup, btw. Its just the way semantic UI controls it it is still unmounting even though it was mounted on the main document body

ui/redux/utils/modalReducer.js
```diff
@@ -24,9 +24,10 @@ export const setModalConfirm = (modalName, confirmMessage) => dispatch =>
...
 // root reducer
 export default {
-  modal: createObjectsByIdReducer(UPDATE_MODAL_STATE, null, { open: false, confirming: false, confirmOnClose: null }),
+  modal: createObjectsByIdReducer(UPDATE_MODAL_STATE),
 }
 
 // basic selectors
 export const getModalOpen = (state, modalName) => state.modal[modalName] && state.modal[modalName].open
 export const getModalConfim = (state, modalName) => state.modal[modalName] && (state.modal[modalName].confirming || null) && state.modal[modalName].confirmOnClose
+export const getOpenModals = state => Object.keys(state.modal).filter(modalName => state.modal[modalName].open)
```
bw2: just revisiting the modal dialogs code & state
 - would rename `open` => `isOpen` and `getModalOpen` => `getModalIsOpen`
 - `confirmOnClose` => `confirmOnClosePanel` (?) or  `confirmOnCloseContent` 
 - `getModalConfim` => `getModalConfirmOnClosePanel` etc.


Also, it would be good to have a standard way to document the purpose of components and props - especially ones where it isn't immediately obvious what they do - for example the "popup" prop in Modal.jsx
There's  https://react-styleguidist.js.org/docs/documenting.html , but we don't have to follow it religiously.

ui/shared/components/panel/variants/VariantFamily.jsx
bw2: Also, logically this would extend to renaming Family.jsx => FamilyPanel.jsx (unless there's a better word than "panel" for div/grid-type components), and most components in the 'variants' folder, so if you agree with this, we should rename a bunch of things

ui/shared/components/PopupWithModal.jsx
bw2: I'm confused about this.. need to play around with this in my local dev instance.

hanars: We can also talk in person if that helps

PR #441, opened 2018-06-27
Fix phenotips proxy

bw2:


bw2:
thanks yeah, double-checked on the dev server - bam/cram still work

PR #444, opened 2018-06-29
gene annotations on correct gene

hanars:
As part of the new saved variants page I updated how we annotate genes so in the old UI you can show gene lists and omim annotations on the relevant gene, instead of ambiguously at the bottom of the list of genes


bw2:
LGTM

PR #455, opened 2018-07-05
Edit datasets ui

hanars:


seqr/utils/es_utils.py
```diff
@@ -0,0 +1,9 @@
+import elasticsearch
+
+import settings
+
+VARIANT_DOC_TYPE = 'variant'
+
+
+def es_client():
```
bw2: would rename to `get_es_client()`

seqr/views/apis/dataset_api.py
```diff
@@ -41,48 +42,45 @@ def add_dataset_handler(request, project_guid):
...
-        return create_json_response(
-            {}, status=400, reason="Invalid request: 'form' not in request_json")
 
-    form_data = request_json['form']
+    logger.info("add_dataset_handler: received %s" % pformat(request_json))
 
-    logger.info("add_dataset_handler: received %s" % pformat(form_data))
-
-    if "sampleType" not in form_data or "datasetType" not in form_data or "genomeVersion" not in form_data or "datasetPath" not in form_data:
+    required_fields = ['sampleType', 'datasetType']
```
bw2: why isn't elasticsearchIndex required?

seqr/views/apis/samples_api.py
```diff
@@ -11,8 +12,12 @@ def match_sample_ids_to_sample_records(
         project,
         sample_ids,
         sample_type,
+        dataset_type,
+        elasticsearch_index=None,
         max_edit_distance=0,
-        create_sample_records=False):
+        create_sample_records=False,
+        sample_individual_mapping=None,
```
bw2: would add new params to doc string. Also, not sure how strongly you feel about these names, but would vote for renaming sample_individual_mapping to `sample_id_to_individual_id_mapping`

seqr/views/apis/samples_api.py
```diff
@@ -68,20 +90,23 @@ def match_sample_ids_to_sample_records(
...
+                    loaded_date=timezone.now(),
                 )
                 sample_id_to_sample_record[sample_id] = new_sample
 
-    return sample_id_to_sample_record
+    return sample_id_to_sample_record, created_sample_ids
 
 
-def find_matching_sample_records(project, sample_ids, sample_type):
+def find_matching_sample_records(project, sample_ids, sample_type, dataset_type, elasticsearch_index):
```
bw2: would add new args to doc string

seqr/views/apis/dataset_api.py
hanars: for adding bam/cram datasets theres no index

seqr/views/apis/dataset_api.py
```diff
@@ -41,48 +45,45 @@ def add_dataset_handler(request, project_guid):
...
+            dataset_name=dataset_name,
+            max_edit_distance=0,
+            ignore_extra_samples_in_callset=ignore_extra_samples_in_callset,
+            mapping_file_id=mapping_file_id,
+        )
+        updated_sample_json = get_json_for_samples(updated_samples, project_guid=project_guid)
+        response = {
+            'samplesByGuid': {s['sampleGuid']: s for s in updated_sample_json}
+        }
+        updated_individuals = {s['individualGuid'] for s in updated_sample_json if s['sampleId'] in created_sample_ids}
```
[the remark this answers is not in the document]
hanars: lines 83-85 only is one query though

also, we don't have all the samples here, anything that wasn't updated hasn't been fetched, but we want to update the individuals so they have the full list of sample guids including those that re not new

seqr/views/apis/individual_api.py
```diff
@@ -242,29 +228,17 @@ def receive_individuals_table_handler(request, project_guid):
...
-    filename = stream._name
-    #file_size = stream._size
-    #content_type = stream.content_type
-    #content_type_extra = stream.content_type_extra
-    #for chunk in value.chunks():
-    #   destination.write(chunk)
+    def parse_file(filename, stream):
+        pedigree_records, errors, warnings = parse_pedigree_table(filename, stream, user=request.user, project=project)
+        if errors:
+            raise Exception({'errors': errors, 'warnings': warnings})
```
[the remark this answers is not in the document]
hanars: it was so incredibly ugly and it meant doing checks so many levels deep if there are error messages, whereas this pops us out of the flow immediately regardless of where the error was triggered. I can create specific Exception class for this so we don't need is instance though

seqr/views/apis/dataset_api.py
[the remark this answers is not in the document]
bw2: oh I see.. my mistake

seqr/views/apis/individual_api.py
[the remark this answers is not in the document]
bw2: sounds great

seqr/views/apis/individual_api.py
```diff
@@ -35,6 +32,13 @@
 __AFFECTED_TO_EXPORTED_VALUE['U'] = ''
 
 
+class ErrorsWarningsException(Exception):
```
bw2: Would go with something like `ParseException` to indicate error context/cause similar to built-in exceptions

PR #456, opened 2018-07-05
Project page ui tweaks

hanars:


ui/shared/components/graph/HorizontalStackedBar.jsx
```diff
@@ -91,7 +96,7 @@ class HorizontalStackedBar extends React.Component {
           trigger={
             <NoWrap>
               {dataWithPercents.filter(d => d.percent >= minPercent).map(d =>
-                <BarSection key={d.name} to={linkPath && `${linkPath}/${d.name}`} color={d.color} percent={d.percent} />,
+                <BarSection key={d.name} to={(sectionLinks && linkPath) ? `${linkPath}/${d.name}` : linkPath} color={d.color} percent={d.percent} />,
```
bw2: would disable links entirely if not linkPath

hanars: it does, `BarSection` renders a plain div if the `to` property is falsey

PR #470, opened 2018-07-18
Locus list ui

hanars:
still needs tests but otherwise this is good to go


seqr/model_utils.py
```diff
@@ -210,6 +213,9 @@ def convert_seqr_kwargs_to_xbrowse_kwargs(seqr_model, **kwargs):
     if seqr_class_name == "Individual" and "family" in xbrowse_kwargs:
         xbrowse_kwargs["project"] = getattr(seqr_model, "family").project
 
+    if seqr_class_name == "LocusList" and 'name' in xbrowse_kwargs:
+        xbrowse_kwargs['slug'] = xbrowse_kwargs['name'].replace(' ', '-').lower()
```
bw2: might be better to call `slugify` method instead. (Other places in the code use it)

ui/shared/components/panel/genes/LocusListDetail.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+const mapStateToProps = (state, ownProps) => ({
+  locusList: getLocusListsByGuid(state)[ownProps.locusListGuid] || {},
+})
+
+const mapDispatchToProps = {
+  onSubmit: updateLocusList,
+}
+
+export { LocusListDetail }
```
bw2: Should this be export {  LoadedLocusListDetail }?

hanars: no, thats exported as the default on the next line, this is intentionally exporting the unconnected component for testing

seqr/model_utils.py
bw2: or django.utils.text.slugify

ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -116,35 +146,25 @@ class ReduxFormWrapper extends React.Component {
     } else if (this.props.submitFailed) {
       saveStatus = RequestStatus.ERROR
     }
+
+    // redux-form does not support throwing submission warnings so this is a work around
+    const submissionWarnings = this.props.warning || (this.props.error && this.props.error.map(error => error.warning).filter(warning => warning))
```
bw2: I was initially in favor of warnings but in retrospect there's no good way to fit them into a dialog like this, so would vote for getting rid of them - the error message when ignoreInvalidItems=false seems sufficient, and then the form can just close on submit when ignoreInvalidItems=true.

seqr/views/apis/locus_list_api.py
```diff
@@ -0,0 +1,233 @@
...
+                ))
+
+    gene_symbols_to_ids = get_gene_symbols_to_gene_ids(new_gene_symbols)
+    invalid_items += [symbol for symbol, gene_id in gene_symbols_to_ids.items() if not gene_id]
+    new_genes = get_genes([gene_id for gene_id in gene_symbols_to_ids.values() if gene_id] + list(new_gene_ids))
+    invalid_items += [gene_id for gene_id, gene in new_genes.items() if not gene]
+    new_genes = {gene_id: gene for gene_id, gene in new_genes.items() if gene}
+
+    if invalid_items and not request_json.get('ignoreInvalidItems'):
+        return None, None, invalid_items, 'This list contains invalid genes/ intervals. Update them, or select the "Ignore invalid genes and intervals" checkbox to ignore.'
```
bw2: could the invalid items be appended to this error message, so the error box shows one message instead of 2?

PR #475, opened 2018-07-18
Fix bugs

hanars:


ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -46,8 +46,14 @@ PathogenicityLink.propTypes = {
 }
 
 
+const clinvarUrl = (clinvar) => {
+  const baseUrl = 'http://www.ncbi.nlm.nih.gov/clinvar'
+  const variantPath = clinvar.variantId ? `/variation/${clinvar.variantId}` : `?term=${clinvar.alleleId}[alleleid]`
```
bw2: you want to default to alleleid (unless it's missing for some reason)

bw2: (the allele id is more specific than variant id)

seqr/views/utils/json_to_orm_utils.py
```diff
@@ -31,19 +31,13 @@ def update_individual_from_json(individual, json, verbose=False, user=None, allo
...
-
-
 def update_model_from_json(model_obj, json, user=None, verbose=False, allow_unknown_keys=False, immutable_keys=None):
     seqr_update_fields = {}
     internal_fields = model_obj._meta.internal_json_fields if hasattr(model_obj._meta, 'internal_json_fields') else []
 
     for json_key, value in json.items():
         orm_key = _to_snake_case(json_key)
-        if orm_key in (immutable_keys or []):
+        if orm_key in (immutable_keys or ['created_by']):
```
bw2: update_model_from_json seems like a low level method where it would be non-obvious to users why 'created_by' is treated as a special case?

bw2: Maybe better to remove `createdBy` before calling update_model_from_json?

ui/shared/components/panel/variants/Pathogenicity.jsx
hanars: got it

seqr/views/utils/json_to_orm_utils.py
hanars: because this function takes a valid json objects, and created_by is always a django user object, so this should never be allowed. since every single model has a created_by field, adding special casing to remove it everywhere seems much messier than just special casing it here

bw2: oh I see. you're right.

PR #477, opened 2018-07-19
Misc bug fixes

bw2:


seqr/views/utils/dataset_utils.py
```diff
@@ -192,8 +190,10 @@ def validate_variant_call_inputs(sample_type, dataset_path, project, elasticsear
     if dataset_path:
         if not dataset_path.endswith(".vcf.gz") and not dataset_path.endswith(".vds"):
             raise Exception("Variant call dataset path must end with .vcf.gz or .vds")
-        _validate_dataset_path(dataset_path)
-        _validate_vcf(dataset_path)
+
+        # TODO need to fix credentials
```
hanars: If you don't plan to fix this in the very near future, could you make a github issue so we don't forget about this?

settings.py
```diff
@@ -344,24 +344,28 @@
...
+PHENOTIPS_ADMIN_PWD = 'admin'
+PHENOTIPS_UPLOAD_EXTERNAL_PHENOTYPE_URL = "http://"+PHENOTIPS_SERVICE_HOSTNAME+":"+str(PHENOTIPS_PORT)+"/rest/patients/eid"
 
 # when set to None, this *disables* the PhenoTips interface for all projects. If set to a list of project ids, it will
 # enable the PhenoTips interface for *all* projects except those in the list.
 PROJECTS_WITHOUT_PHENOTIPS = []
 
 
-REDIS_SERVICE_HOSTNAME = os.environ.get('REDIS_SERVICE_HOST')
+REDIS_SERVICE_HOSTNAME = os.environ.get('REDIS_SERVICE_HOSTNAME')
```
hanars: are we sure this is also the env variable on the prod server? because if redis wasn't broken there this could break it, maybe?

ui/shared/components/form/EditRecordsForm.jsx
```diff
@@ -101,7 +101,7 @@ class EditRecordsForm extends React.Component
 
   formContent = () => {
     return (
-      <div>
+      <div style={{ marginBottom: this.props.records.length > ROWS_PER_PAGE ? '50px' : '0px' }}>
```
hanars: can we use styled componenets here?

settings.py
bw2: prod has both  REDIS_SERVICE_HOSTNAME="redis",  REDIS_SERVICE_HOST=<ip-address>. 

Many of the env vars are defined automatically by kubernetes, but this one is defined in 
REDIS_SERVICE_HOSTNAME deploy/kubernetes/shared-settings.yaml

hanars: cool. I figured you had already looked into it but just wanted to double check

seqr/views/utils/dataset_utils.py
bw2: good call.
https://github.com/macarthur-lab/seqr/issues/478

PR #494, opened 2018-07-25
Add seqr model for Analysed by

hanars:


seqr/views/utils/orm_to_json_utils.py
```diff
@@ -103,19 +102,22 @@ def _get_pedigree_image_url(pedigree_image):
 
     if add_individual_guids_field:
         prefetch_related_objects(families, 'individual_set')
+        prefetch_related_objects(families, 'familyanalysedby_set')
```
bw2: why only prefetch if add_individual_guids_field is True?

hanars: oh thats wrong, I'll fix it

PR #500, opened 2018-07-27
better warning for het males

hanars:


ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -29,6 +29,14 @@ const Allele = styled.span`
   font-style: ${(props) => { return props.isRef ? 'inherit' : 'italic' }};
 `
 
+const PAR_REGIONS = {
+  37: [[60001, 2699521], [154931044, 155260561]],
```
bw2: we get variant calls on both X and Y so would include the Y intervals also

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -29,6 +29,14 @@ const Allele = styled.span`
...
+  38: [[10001, 2781480], [155701383, 156030896]],
+}
+
+const isHemiVariant = variant =>
+  variant.chrom === 'X' && PAR_REGIONS[variant.genomeVersion].every(region => !(region[0] < variant.pos < region[1]))
```
bw2: cool, didn't know you could do a < x < b

ui/shared/components/panel/variants/VariantIndividuals.jsx
hanars: but there us a check for `variant.chrom === 'X'` so when would that happen?

bw2: It would have to be something like
```
const PAR_REGIONS = {
  37: {
    'X': [[60001, 2699521], [154931044, 155260561]],
    'Y': [..., ],
   },
  38: {
    'X': [..., ],
    'Y': [..., ],
  }

const isHemiVariant = variant =>
  (variant.chrom === 'X' ||   variant.chrom === 'Y')  && PAR_REGIONS[variant.genomeVersion][variant.chrom].every(region => !(region[0] < variant.pos < region[1]))
```

bw2: Also, it might be better to make the 'M' check part of isHemi:
```
const isHemiVariant = (variant, individual) => (individual.sex === 'M' && (variant.chrom === 'X' ||   variant.chrom === 'Y') ...)```

PR #501, opened 2018-07-27
fix dashboard sample counts

hanars:


seqr/views/pages/dashboard_page.py
```diff
@@ -225,7 +225,7 @@ def _add_sample_type_counts(cursor, projects_by_guid):
         SELECT
           p.guid AS project_guid,
           s.sample_type AS sample_type,
-          COUNT(*) AS num_samples
+          COUNT(distinct s.individual_id) AS num_samples
```
bw2: cool way to fix this

bw2: also should only count samples with dataset_type="VARIANTS"

bw2: and sample_status="loaded"

PR #507, opened 2018-07-30
adding gnomad popmax_or_global_AF, primate_ai, clinvar gold_stars. 

bw2:


seqr/views/apis/saved_variant_api.py
```diff
@@ -270,6 +271,7 @@ def _variant_details(variant_json, user):
             'clinsig': extras.get('clinvar_clinsig'),
             'variantId': extras.get('clinvar_variant_id'),
             'alleleId': extras.get('clinvar_allele_id'),
+            'goldStars': extras.get('clinvar_gold_stars'),
```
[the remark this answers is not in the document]
bw2: was thinking that could be added in a separate PR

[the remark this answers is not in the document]
hanars: I'd rather we didn't return it to the client side yet then. It will still be saved to the saved variant json so we can access it later when we want to show it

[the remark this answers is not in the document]
bw2: ok, I'll just add the ui

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -70,6 +70,9 @@ const Pathogenicity = ({ variant }) => {
               formatName={snakecaseToTitlecase}
             />,
           )}
+          {variant.clinvar.gold_stars != null &&
+            <Rating icon="star" defaultRating={variant.clinvar.gold_stars} maxRating={4} />
```
hanars: on the server side you use `goldStars` and here you use `gold_stars` so this won't work. Can you also post a quick screenshot of what this looks like?

seqr/views/apis/saved_variant_api.py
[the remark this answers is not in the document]
bw2: done

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -1,12 +1,23 @@
...
+const GrayStarIcon = styled(Icon).attrs({ name: 'star' })`
+  color: #D5D5D5;
+  margin: 0em 0.3em 0em 0em !important;
+`
+
+const FilledInStarIcon = styled(Icon).attrs({ name: 'star' })`
+  color: #FFB70A;
+  margin: 0em 0.3em 0em 0em !important;
+`
+
```
hanars: Icons support a color prop, so you could just use the builtin `<Icon name="star" color="yellow/grey" />` which I think would be clearer here. There is also a `fitted` prop if you need to decrease the margins

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -20,7 +20,12 @@ const PredictionValue = styled.span`
   text-transform: uppercase;
 `
 
-const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) => {
+const LinkButton = styled.span.attrs({ role: 'button', tabIndex: '0' })`
```
hanars: we have this as a utility already in `shared/components/buttons/ButtonLink`

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,39 +43,72 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
...
+            this.state.showMore ?
+              <div>
+                {
+                  MORE_PREDICTORS.map(field => <Prediction key={field} field={field} annotation={annotation} />)
+                }
+                <LinkButton onClick={() => this.setState({ showMore: false })}>hide</LinkButton>
+              </div>
+            :
+              MORE_PREDICTORS.some(field => annotation[field] != null) &&
+              <LinkButton onClick={() => this.setState({ showMore: true })}>show more..</LinkButton>
```
hanars: should have three dots in the ellipses

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,39 +43,72 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
     }
   } else {
     color = SEVERITY_MAP[value]
-    value = value.replace('_', ' ')
+    if (value) {
```
hanars: you return null if no value above, so you shouldn't need this check

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,39 +43,72 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
...
+      value = `${value}`.replace('_', ' ')
+    }
+  }
+
+  if (!name) {
+    name = field.replace(/_/g, ' ').toUpperCase()
```
hanars: given that we define what the fields are named, why not just continue to use field names that transform nicely?

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,38 +43,72 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
...
+
+    return annotation ?
+      <div>
+        <Prediction field="cadd_phred" annotation={annotation} dangerThreshold={20} warningThreshold={10} name="CADD" />
+        <Prediction field="dann_score" annotation={annotation} dangerThreshold={0.96} warningThreshold={0.93} name="DANN" />
+        <Prediction field="revel_score" annotation={annotation} dangerThreshold={0.75} warningThreshold={0.5} name="REVEL" />
+        <Prediction field="eigen_phred" annotation={annotation} dangerThreshold={1} warningThreshold={2}name="EIGEN" />
+        <Prediction field="mpc_score" annotation={annotation} dangerThreshold={2} warningThreshold={1} name="MPC" />
+        <Prediction field="primate_ai_score" annotation={annotation} dangerThreshold={0.7} warningThreshold={0.5} name="PMT AI" />
+        <Transition.Group animation="fade down" duration="500">
```
hanars: I think the behavior here needs a bit of tweaking, because if only one of the "main" predictors is present and say 3 of the `MORE_PREDICTORS` then the better behavior would be to just show those 4 predictors with no toggle. So rather than having a hardcoded set of predictors that are always show and some that are always toggled, I think it would make more sense to have a list of predictors and then filter them by the presence of a value and then slice it so the top 6(? 8? whatever) are rendered and any remainders are toggleable

seqr/management/commands/reload_saved_variant_json.py
```diff
@@ -31,20 +33,28 @@ def handle(self, *args, **options):
             logger.info("Project: " + project.name)
 
             saved_variants = SavedVariant.objects.filter(project=project, family__isnull=False).select_related('family')
+            if options['reset']:
+                for saved_variant in saved_variants:
+                    saved_variant.saved_variant_json = None
+                    saved_variant.save()
```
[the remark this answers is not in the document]
bw2: I also made this save variants immediately after retrieving them from elasticsearch - so that if a tunnel to ES times-out in the middle of a large project (eg. myoseq), most variants can still be updated.
Then, with this option if you first clear all saved_variant_json values, you can easily see which ones have been updated.

ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -175,15 +175,12 @@ class SavedVariants extends React.Component {
     ]
 
     const allShown = this.props.variantsToDisplay.length === this.props.totalVariantsCount
-    let shownSummary
-    if (allShown) {
-      shownSummary = this.props.variantsToDisplay.length > 0 ? 'all' : ''
```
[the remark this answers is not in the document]
bw2: The message with it shows up as `all n All Saved variants`, so it drops the extra 'all'

seqr/management/commands/reload_saved_variant_json.py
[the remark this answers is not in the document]
hanars: So a concern I have is that the way we currently load variants, we fetch a whole list of them and then add some additional annotations to everything in the list. While thats more relevant for mongo, it does still happen for a few fields for the es variants as well. Are you sure that the point at which its saving these is after those annotations have been added?

ui/shared/components/panel/variants/Predictions.jsx
bw2: can see some people preferring what you suggest, but not sure if everyone would since it's a little less predictable what gets shown where. Do you want to get feedback from people?

ui/shared/components/panel/variants/Predictions.jsx
bw2: nice, I'll use that.

ui/shared/components/panel/variants/Pathogenicity.jsx
bw2: oh yeah, thanks

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -126,10 +121,12 @@ const Frequencies = ({ variant }) => {
         header="Allele Counts"
         content={
           <div>
-            <div>This callset:<HorizontalSpacer width={10} /><b>{popCounts.AC}</b></div>
-            <div>Gnomad exomes:<HorizontalSpacer width={10} /><b>{popCounts.gnomadExomesAC}</b></div>
-            <div>Gnomad genomes:<HorizontalSpacer width={10} /><b>{popCounts.gnomadGenomesAC}</b></div>
-            <div>Topmed:<HorizontalSpacer width={10} /><b>{popCounts.topmedAC}</b></div>
+            <div>this callset:<HorizontalSpacer width={10} />{popCounts.AC} out of {popCounts.AN || '?'}</div>
```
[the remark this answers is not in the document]
bw2: yup I can just hide the line if AC or AN are null

ui/shared/components/panel/variants/Predictions.jsx
bw2: we could do both?
primate_ai_score is an ES index field so hard to change now

ui/shared/components/panel/variants/Predictions.jsx
hanars: If you feel strongly we can ask, but I feel pretty strongly about this one. If we were always showing all the predictors with some placeholder even if they were empty I would understand the desire to always have the same set shown vs hidden. However, given that we only show some of the predictors for each variant so there is no uniformity in whats shown anyways, I can't imagine a scenario where it is a better user experience to have to do an extra click

ui/shared/components/panel/variants/Predictions.jsx
hanars: we don't pass through field names though, we map them to different client side field names in the `_variant_details` method in `saved_variant_api`

bw2: that re-mapping in saved_variant_api will go away ideally. The current transformations are an artifact of having 2 different data sources (es. and mongo), but the ideal setup would be to pass elasticsearch results straight to client. I can get revert the regexp if you want though.

seqr/management/commands/reload_saved_variant_json.py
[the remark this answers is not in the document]
bw2: good point, but yes, variant_utils._retrieve_saved_variants_json calls add_extra_info_to_variants_project.

ui/shared/components/panel/variants/Pathogenicity.jsx
bw2: turns out fitted doesn't reduce margin enough and yellow/grey colors don't look right

bw2:
done

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -1,12 +1,23 @@
...
 
+const GrayStarIcon = styled(Icon).attrs({ name: 'star' })`
+  color: #D5D5D5;
+  margin: 0em 0.3em 0em 0em !important;
+`
+
+const YellowStarIcon = styled(Icon).attrs({ name: 'star' })`
+  color: #FFB70A;
+  margin: 0em 0.3em 0em 0em !important;
+`
```
hanars: I would rather we just used semantic-ui's defaults for colors unless there is a strong reason to do otherwise

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,38 +44,91 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
...
-      <Prediction field="fathmm" annotation={annotation} />
-      <Prediction field="cadd_phred" annotation={annotation} dangerThreshold={20} warningThreshold={10} />
-      <Prediction field="dann_score" annotation={annotation} dangerThreshold={0.96} warningThreshold={0.93} />
-      <Prediction field="revel_score" annotation={annotation} dangerThreshold={0.75} warningThreshold={0.5} />
-      <Prediction field="eigen_phred" annotation={annotation} dangerThreshold={1} warningThreshold={2} />
-      <Prediction field="mpc_score" annotation={annotation} dangerThreshold={2} warningThreshold={1} />
-    </div>
-    : null
-}
+export default class Predictions extends React.Component {
```
hanars: this should be `PureComponent`

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -38,38 +44,91 @@ const Prediction = ({ field, annotation, dangerThreshold, warningThreshold }) =>
...
+        {
+          morePredictorFields ?
+            <Transition.Group animation="fade down" duration="500">
+              {
+                this.state.showMore ?
+                  <div>
+                    {
+                      morePredictorFields.map(predictorField => <Prediction key={predictorField.field} field={predictorField.field} annotation={annotation} />)
+                    }
+                    <StyledButtonLink onClick={() => this.setState({ showMore: false })}>hide</StyledButtonLink>
```
hanars: rather than define functions within props which can cause unnecessary re-renders, could we instead have a `toggleShowMore` function defined on the class that toggles the state

ui/shared/components/panel/variants/Pathogenicity.jsx
bw2: Besides making it look better, idea was to match 
image
(https://www.ncbi.nlm.nih.gov/clinvar/variation/263800/)
semantic ui grey doesn't work for this (can see an example here https://semantic-ui.com/usage/theming.html)
Especially for specialized components like this, seems better to treat semantic ui as a starting point instead of sticking to unmodified semantic ui built-ins.

bw2:
done

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -24,17 +35,23 @@ const HGMD_CLASS_NAMES = {
...
 
 
-const PathogenicityLabel = ({ clinsig, formatName }) =>
-  <Label color={CLINSIG_COLOR[CLINSIG_SEVERITY[clinsig]] || 'grey'} size="medium" horizontal basic>
+const PathogenicityLabel = ({ clinsig, formatName, goldStars }) =>
+  <Label color={CLINSIG_COLOR[CLINSIG_SEVERITY[clinsig.toLowerCase()]] || 'grey'} size="medium" horizontal basic>
     {formatName ? formatName(clinsig) : clinsig}
+    {(goldStars != null) && [
+      <HorizontalSpacer key={-1} width={10} />,
+      [0, 1, 2, 3].map(i => (i < goldStars ?
```
hanars: This looks kind of messy. I know we started with the `Rating` component and then switched to this, but I think in the end it would be neater to use that. Since you are custom styling the color and padding anyways, why not style it on a component that does all this for you

bw2:
refactored clinvar stars

hanars:
My comment was that we should be using the semantic-ui `Rating` component, so these changes were pretty unrelated to that. It is nicer than before though, but iterating through a hard coded array of integers is never pretty

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -34,16 +33,20 @@ const HGMD_CLASS_NAMES = {
 }
 const hgmdName = hgmdClass => HGMD_CLASS_NAMES[hgmdClass]
 
+const ClinvarStars = ({ goldStars }) => goldStars != null &&
+  <StarsContainer>
+    {Array.from(Array(4).keys()).map(i => (i < goldStars ? <StarIcon key={i} goldstar="yes" /> : <StarIcon key={i} />))}
```
hanars: Why not just make `goldstar` a boolean prop?

bw2: It causes this warning
https://github.com/styled-components/styled-components/issues/1198

bw2:
I tried Rating, but it required more css to make it look right than the current option.

ui/shared/components/panel/variants/Pathogenicity.jsx
hanars: You need to do `const StarIcon = styled(({ goldstar, ...props }) => <Icon {...props} />).attrs({ name: 'star' })``

hanars: theres also a `ColoredIcon` component in `shared/components` that takes any valid CSS color for the color prop

PR #527, opened 2018-08-07
Add datasets fixes

hanars:


seqr/views/apis/samples_api.py
```diff
@@ -127,7 +127,10 @@ def find_matching_sample_records(project, sample_ids, sample_type, dataset_type,
         sample_id__in=sample_ids
     )
     if elasticsearch_index:
-        sample_query = sample_query.filter(Q(elasticsearch_index=elasticsearch_index) | Q(elasticsearch_index__isnull=True))
+        sample_query = sample_query.filter(
+            Q(elasticsearch_index=elasticsearch_index) |
```
bw2: wow yeah, that's subtle

seqr/views/utils/dataset_utils.py
```diff
@@ -123,16 +124,21 @@ def _update_samples_for_dataset(
...
-        if dataset_name:
-            sample.dataset_name = dataset_name
+            sample_update_json['dataset_file_path'] = sample_dataset_path_mapping[sample_id]
         if sample.sample_status != Sample.SAMPLE_STATUS_LOADED:
             not_loaded_samples.append(sample_id)
-            sample.sample_status = Sample.SAMPLE_STATUS_LOADED
-            sample.loaded_date = timezone.now()
-        sample.save()
+            sample_update_json['sample_status'] = Sample.SAMPLE_STATUS_LOADED
+            sample_update_json['loaded_date'] = timezone.now()
```
bw2: would run timezone.now() once before the loop, so that all the samples get the same timestamp

seqr/views/utils/variant_utils.py
```diff
@@ -0,0 +1,61 @@
...
+from xbrowse_server.base.models import Project as BaseProject
+from xbrowse_server.base.lookups import get_variants_from_variant_tuples
+
+
+def get_or_create_saved_variant(xpos=None, ref=None, alt=None, family=None, project=None, **kwargs):
+    if not project:
+        project = family.project
+    saved_variant, _ = SavedVariant.objects.get_or_create(
+        xpos_start=xpos,
+        xpos_end=xpos + len(ref) - 1,
```
bw2: in retrospect, since xpos_end is redundant in this case, how about we get rid of xpos_start/end and just have xpos

seqr/views/utils/variant_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        saved_variant = saved_variants_map[(var['xpos'], var['ref'], var['alt'], var['extras']['family_id'])]
+        _update_saved_variant_json(saved_variant, var)
+
+    return variants_json
+
+
+def _retrieve_saved_variants_json(project, variant_tuples):
+    project_id = project.deprecated_project_id
+    xbrowse_project = BaseProject.objects.get(project_id=project_id)
+    user = User.objects.filter(is_staff=True).first()  # HGMD annotations are only returned for staff users
```
bw2: These functions LGTM.. but will probably make sense to refactor once data is all in elasticsearch.
Particularly this line.. given the caching layers - it seems like the only place these permissions check(s)  should happen is in the code that retrieves the variant data from the cache or elasticsearch, or less ideally, the code  that generates the json response to the client.

seqr/views/utils/variant_utils.py
hanars: I like that idea, but thats out of the scope of this particular issue since it would require a bunch of updates unrelated files and a migration. Can you make a separate ticket?

seqr/views/utils/variant_utils.py
hanars: Agreed! Once we have all the data reloaded we will switch over all of this to use the new utilities for retrieving variants, which I assume we will write such that they don't do the HGMD filtering at the datastore level

PR #559, opened 2018-08-29
Use postgres reference data for new gene pages

hanars:


seqr/views/utils/gene_utils.py
```diff
@@ -1,45 +1,21 @@
...
-
-def get_gene(gene_id):
-    reference = get_reference()
-    gene = reference.get_gene(gene_id)
-    gene['expression'] = reference.get_tissue_expression_display_values(gene_id)
-    return _gene_json(gene)
+def get_gene(gene_id, user):
+    gene = GeneInfo.objects.get(gene_id=gene_id)
+    gene_json = get_json_for_gene(gene, user=user, add_notes=True)
+    # TODO load expression data into postgres reference_data
```
bw2: I think expression data should go into ES. 
It will only ever be queried by gene id, and would significantly affect postgres backup size.

reference_data/management/commands/update_all_reference_data.py
```diff
@@ -25,9 +25,11 @@ def add_arguments(self, parser):
 
     def handle(self, *args, **options):
         if not options["skip_gencode"]:
-            # download v19 and then v28 because there are 1000+ gene and transcript ids in v19 that
+            # download v19 and then v27 and then v28 because there are 1000+ gene and transcript ids in v19 that
             # gencode retired by the time of v28, but that are used in the gene constraint table and other datasets
+            # and also ~100 genes and transcripts taht were retired between v27 and v28
             update_gencode(19, reset=True)
+            update_gencode(27)
```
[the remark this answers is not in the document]
hanars: per the comment on the above line "~100 genes and transcripts that were retired between v27 and v28". Its breaking saved variants that are in those genes

seqr/views/utils/gene_utils.py
bw2: great point. pg_dump does have --exclude-schema=

bw2: One thing I forgot to add - where do you think is the right place to put add redis caching?

reference_data/management/commands/update_all_reference_data.py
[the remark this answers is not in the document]
bw2: didn't see that.. weird. Might be worth posting in seqr so people are aware that this happens

seqr/views/utils/gene_utils.py
hanars: well, does redis really improve performance over standard postgres queries? Also, probably better to have this conversation elsewhere since its not so relevant to these changes

reference_data/management/commands/update_all_reference_data.py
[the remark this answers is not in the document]
hanars: well currently we don't use reference data for anything and I'll just update it before deploying these changes, so it shouldn't have much of an effect

PR #573, opened 2018-09-06
Linux install improvements

bw2:


deploy/docker/pipeline-runner/shared/loftee/maxEntScan/splicemodels/me1s0acc7
```diff
@@ -0,0 +1,256 @@
+9.223072e-03
```
hanars: these don't really seem like files that belong in the seqr repo. could they live in a public bicket the way we do with reference data?

bw2: I didn't want to copy loftee code from a bucket because it's still changing and this way we have a version-controlled snapshot. The splicemodels are plain text and so small (4Mb) that it seems better to treat them as part of the code.

hanars: but  can't we just put a version-controlled copy in our own seqr bucket?

bw2: we could.. I've also modified it specifically for seqr to remove gerp-based flags since there's no gerp data for GRCh38

hanars: I still don't think our own version of loftee belongs in the seqr repo but if you feel strongly we can leave it

bw2: Konrad is also planning to put the code in a separate repo. How would you want to deploy it when the repo exists (and we potentially create a seqr-specific fork of it)?

hanars: well, we deploy hail which is in its own repo right? so maybe the same way as that?

bw2: ok, I'll move it to a bucket for now.

PR #577, opened 2018-09-13
Fix gene search genotypes

hanars:
If a variant exists in multiple indices in gene search we were essentially just using the genotypes from the final index instead of all of them, because for individuals absent in an index we assign them a genotype with num_alt -1

### DIFF
--- a/xbrowse_server/base/views/project_views.py
+++ b/xbrowse_server/base/views/project_views.py
@@ -881,7 +881,10 @@ def gene_quicklook(request, project_id, gene_id):
                 rare_variant_dict[variant_id] = variant
                 project_variants.append(variant)
             else:
-                rare_variant_dict[variant_id].genotypes.update(variant.genotypes)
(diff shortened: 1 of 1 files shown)


xbrowse_server/base/views/project_views.py
```diff
@@ -881,7 +881,10 @@ def gene_quicklook(request, project_id, gene_id):
                 rare_variant_dict[variant_id] = variant
                 project_variants.append(variant)
             else:
-                rare_variant_dict[variant_id].genotypes.update(variant.genotypes)
+                for indiv_id, genotype in variant.genotypes.items():
+                    existing_genotype = rare_variant_dict[variant_id].genotypes.get(indiv_id)
+                    if not existing_genotype or existing_genotype.num_alt < genotype.num_alt:
```
bw2: would use existing_genotype.num_alt == -1 here instead of existing_genotype.num_alt < genotype.num_alt

PR #578, opened 2018-09-17
add gtex expression to reference data tables

bw2:


reference_data/models.py
```diff
@@ -115,6 +115,18 @@ class GeneConstraint(models.Model):
     #pRec = models.FloatField()
 
 
+class GeneExpression(models.Model):
+    GTEX_TISSUE_TYPES = set([
```
hanars: instead of `set([...])` you can just do `{...}`

PR #590, opened 2018-09-20
Determine freq fields from index mapping

hanars:


xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -596,18 +617,7 @@ def get_elasticsearch_variants(
...
-                    'exac_v3_popmax_AF': float(hit["exac_AF_POPMAX"] or 0.0) if "exac_AF_POPMAX" in hit else None,
-                    'gnomad_exomes_AF': float(hit["gnomad_exomes_AF"] or 0.0) if "gnomad_exomes_AF" in hit else None,
-                    'gnomad_exomes_popmax_AF': float(hit["gnomad_exomes_AF_POPMAX_OR_GLOBAL"] or 0.0) if "gnomad_exomes_AF_POPMAX_OR_GLOBAL" in hit else (float(hit["gnomad_exomes_AF_POPMAX"] or 0.0) if "gnomad_exomes_AF_POPMAX" in hit else None),
-                    'gnomad_genomes_AF': float(hit["gnomad_genomes_AF"] or 0.0) if "gnomad_genomes_AF" in hit else None,
-                    'gnomad_genomes_popmax_AF': float(hit["gnomad_genomes_AF_POPMAX_OR_GLOBAL"] or 0.0) if "gnomad_genomes_AF_POPMAX_OR_GLOBAL" in hit else (float(hit["gnomad_genomes_AF_POPMAX"] or 0.0) if "gnomad_genomes_AF_POPMAX" in hit else None),
-                    'topmed_AF': float(hit["topmed_AF"] or 0.0) if "topmed_AF" in hit else None,
-                },
+                'db_freqs': {k: float(hit[k] or 0.0) if k in hit else (0.0 if k else None) for k, v in freq_fields.items()},
```
bw2: seems to me (0.0 if k else None) should be (0.0 if v is not None else None)

bw2: or maybe even 
 {k: float(hit[k] or 0.0) if k in hit else 0.0 for k, v in freq_fields.items() if v is not None}

hanars: no thats the whole point. If there is a key in the mapping but the key is not in the hit we still want to return 0

hanars: your proposal gets rid of keys entirely when there is nothing present, and the ui expects the key to be there with a null value

bw2: wait but here you're iterating over `k, v in freq_fields.items()` and saying `0.0 if k else None`, but freq_fields keys are just strings, right?

hanars: yeah but they are all defined in the form `"AF" if "AF" in index_fields else None` so if `k` is None it means there is no key in the mapping for that field at all, and otherwise there is a key in the mapping, which may or may not be present in the hit

bw2: even though freq_fields values are defined as `"AF" if "AF" in index_fields else None`, I don't see how k can be None. Like if you have
```
freq_fields = {'1kg_wgs_AF': None,
 '1kg_wgs_popmax_AF': None,
 'AF': None,
 'exac_v3_AF': None,
 'exac_v3_popmax_AF': None}
```
and then do 
```
print([k for k, v in freq_fields.items()])
```

hanars: oh duh, thanks for catching this

PR #635, opened 2018-10-31
Dev

hanars:


settings.py
```diff
@@ -324,7 +324,7 @@
 READ_VIZ_PASSWD=None
 
 
-KIBANA_HOSTNAME = os.environ.get('KIBANA_SERVICE_HOSTNAME', 'kibana')
```
hanars: @bw2 I changed this on purpose because for some reason the environment variable on the prod server was not staying set so every time I deployed kibana was going down. Can we please leave this alone and you can just set the environment variable you need locally?

bw2: all the other vars default to localhost (which is the correct value for local installs). It should be easy to fix prod - can we do that instead?

hanars: if you can get it to work then sure, but that variable was being wiped out every time I pulled and I'm not sure why and I don't have time to look into it. Before merging this to master I would like that resolved though

bw2: It looks like the settings (ConfigMap) are out of date on prod. I'll deploy the latest ones so next time the pod restarts it has that env var set to `kibana`

bw2: Until then, I've also added this to .bashrc on prod
```
export KIBANA_SERVICE_HOSTNAME='kibana'
```

PR #654, opened 2018-11-15
refactor deploy scripts to optionally attach to existing elasticsearch disks

bw2:


hanars:
I'd rather we not commit these as "test" settings to master. Can you change these to update the "real" settings before we merge this, or is there a reason why that doesn't make sense?

deploy/kubernetes/namespace.yaml
```diff
@@ -5,16 +5,16 @@ kind: Namespace
 metadata:
   name: {{ NAMESPACE }}
 ---
```
hanars: I'd rather we delete things than just comment them out

deploy/servctl_utils/other_command_utils.py
```diff
@@ -40,7 +43,7 @@
 COMPONENTS_TO_OPEN_IN_BROWSER = set([
     "cockpit",
     "elasticsearch",
-    "kibana",
+    "kibana", "es-kibana",
```
hanars: this should be on a new line

PR #661, opened 2018-11-21
Nested genotypes

hanars:
I'll want to update this with the correctly renamed parent/child -> variant/genotype structure before merging, but this is the bulk of it


seqr/views/utils/dataset_utils.py
```diff
@@ -147,14 +147,23 @@ def _update_samples_for_dataset(
...
     except TransportError as e:
         raise Exception(e.error)
 
+    #  Nested genotypes
+    if field_mapping.get(elasticsearch_index, {}).get('mappings', {}).get(VARIANT_DOC_TYPE, {}).get('join_field'):
+        s = elasticsearch_dsl.Search(using=es_client, index=elasticsearch_index)
+        s = s.params(size=0)
+        s.aggs.bucket('sample_ids', elasticsearch_dsl.A('terms', field='sample_id'))
+        response = s.execute()
+        return [agg['key'] for agg in response.aggregations.sample_ids.buckets]
```
bw2: wouldn't you need to call elasticsearch_datastore._decode_name on this (and also in the non-nested version)?

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -258,42 +264,99 @@ def get_elasticsearch_variants(
...
                 sample_id = family_individual_ids_to_sample_ids.get(indiv_id) or indiv_id
                 encoded_sample_id = _encode_name(sample_id)
-                q = Q('range', **{encoded_sample_id+"_num_alt": {'gte': 1}})
-                if atleast_one_nonref_genotype_filter is None:
-                    atleast_one_nonref_genotype_filter = q
-                else:
-                    atleast_one_nonref_genotype_filter |= q
+                genotype_filter = value
+                if type(genotype_filter) == int or type(genotype_filter) == basestring:
+                    genotype_filters[encoded_sample_id] = [('term', genotype_filter)]
```
bw2: I don't think you can guarantee in the current system that  encoded_sample_id is unique across families, etc.

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -258,42 +264,99 @@ def get_elasticsearch_variants(
...
-                    #logger.info("### ADDED FILTER: " + str(~Q('exists', field='filters')))
+        if is_nested:
+            #  Subquery for child docs with the requested sample IDs
+            sample_id_q = Q('terms', sample_id=encoded_sample_ids)
+
+            # Return inner hits with the requested samples
+            if indivs_to_consider:
+                s = s.filter(Q('has_child', type='genotype', query=sample_id_q, inner_hits={'size': len(indivs_to_consider)}))
+            else:
+                s = s.filter(Q('has_child', type='genotype', query=Q(), inner_hits={'size': 100}))
```
bw2: why is the limit 100 and not 3 or 10? either way would create a global constant or maybe there's a way to get rid of calls where indivs_to_consider is None?

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -258,42 +264,99 @@ def get_elasticsearch_variants(
...
+
+            if genotype_filters:
+                genotype_q = None
+                for encoded_sample_id, queries in genotype_filters.items():
+                    q = Q(queries[0][0], num_alt=queries[0][1])
```
bw2: q = Q(queries[0][0], num_alt=queries[0][1])  is quite difficult to read. Would just use the same pattern as 
genotype_q = None
if not genotype_q:
                        genotype_q = q
                    else:
                        genotype_q |= q

seqr/views/utils/dataset_utils.py
hanars: I forgot. But thinking about it we shouldn't need to encode/decode names for nested sample IDs anyways so I will take out the encoding here, rather than add the decoding below

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -511,11 +554,26 @@ def get_elasticsearch_variants(
...
+
             for individual_id, sample_id in family_individual_ids_to_sample_ids.items():
                 encoded_sample_id = _encode_name(sample_id)
-                num_alt = int(hit["%s_num_alt" % encoded_sample_id]) if ("%s_num_alt" % encoded_sample_id) in hit else -1
-                if num_alt is not None:
-                    all_num_alt.append(num_alt)
+
+                def _get_hit_field(field):
+                    if is_nested:
+                        gen_hit = genotypes_by_sample_id.get(encoded_sample_id, {})
```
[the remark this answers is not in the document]
hanars: Nope they aren't! I'll take it out

xbrowse/datastore/elasticsearch_datastore.py
hanars: This is copy-paste: https://github.com/macarthur-lab/seqr/pull/661/files/c45f3809551a30512065c2f56c109574a9b33a24#diff-bb4e5d5e6d2fce2f9ce02a49e3098e78L368

Under the current system a sample_id has to be unique within the callset or nothing works. For new search it might be wiorth revisiting that constraint but this just amintains existing behavior

xbrowse/datastore/elasticsearch_datastore.py
hanars: Thats the elasticsearch max number of inner hits that can be returned, but you are right it should be a constant. And this is used by gene search so it needs to stay. As a general rule, I'm not making any improvements to this search since I'm writing a clean new search, but in that new search I will clean up gene search so it does explicitly include the sample ids it wants

seqr/views/utils/dataset_utils.py
hanars: oh I misunderstood your comment, you meant we should be decoding the sample ids that aren't nested. i'll add that

xbrowse/datastore/elasticsearch_datastore.py
```diff
@@ -258,42 +266,103 @@ def get_elasticsearch_variants(
...
+                        else:
+                            q = Q(op, num_alt=val)
+                    q &= Q('term', sample_id=sample_id)
+                    if not genotype_q:
+                        genotype_q = q
+                    else:
+                        genotype_q |= q
+                s = s.filter(Q('has_child', type='genotype', query=genotype_q,  min_children=len(genotype_filters)))
+            elif indivs_to_consider:
+                s = s.filter(Q('has_child', type='genotype', query=(Q(Q('range', num_alt={'gte': 1}) & sample_id_q))))
```
bw2: I think there should be a single has_child query where the genotype_q also includes the min_ab and min_gq filters.. otherwise you could have one set of child-docs that match the num_alt filters, and another set that match the min_ab/min_gq ones.

PR #690, opened 2019-01-16
upgrade to latest igv.js

bw2:


bw2:
Having a hard time figuring out this error...
https://travis-ci.org/macarthur-lab/seqr/builds/480550576#L3926
@hanars do you know why these tests are breaking with the new IGV.js?

ui/shared/components/graph/IGV.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import igv from 'igv'
 import styled from 'styled-components'
-import style from 'igv/igv.css' //eslint-disable-line no-unused-vars
```
hanars: why is this no longer neccessary?

hanars:
@bw not sure why the changes are causing tests to fail... do they fail locally as well or just on travis? My guess is its a babel config thing: https://stackoverflow.com/questions/38296667/getting-unexpected-token-export

ui/shared/components/graph/IGV.jsx
bw2: IGV no longer has a css dependency (https://github.com/igvteam/igv.js?files=1#installation)

bw2:
it fails both locally and on Travis-CI - and seems to be this issue
https://github.com/facebook/jest/issues/2550
but the suggested fixed don't help.

hanars:
I think thats maybe why we were importanig it in the constructor instead of globally

PR #773, opened 2019-03-07
Search ui

hanars:


seqr/views/utils/orm_to_json_utils.py
```diff
@@ -555,3 +571,17 @@ def get_json_for_gene(gene, **kwargs):
     """
 
     return _get_json_for_model(gene, get_json_for_models=get_json_for_genes, **kwargs)
+
+
+def get_json_for_saved_searches(search, user):
+    def _process_result(result, search):
+        # Do not apply HGMD filters in shared searches for non-staff users
+        if not search.created_by and not user.is_staff and result['search'].get('pathogenicity', {}).get('hgmd'):
```
bw2: to avoid code that expects a very specific search json structure, maybe better to just have separate sets of saved searches for staff / non-staff?

seqr/fixtures/variant_searches.json
```diff
@@ -0,0 +1,269 @@
...
+    }
+},
+{
+    "model": "seqr.variantsearch",
+    "pk": 2,
+    "fields": {
+        "guid": "VS0000002_recessive_restrictiv",
+        "name": "Recessive Restrictive",
+        "search": {
+            "qualityFilter": {
```
bw2: given that the set of filters will evolve, should there be a "search_api_version" field inside "search"?

seqr/fixtures/variant_searches.json
```diff
@@ -0,0 +1,269 @@
...
+                "extended_splice_site": [
+                    "splice_region_variant"
+                ],
+                "essential_splice_site": [
+                    "splice_donor_variant",
+                    "splice_acceptor_variant"
+                ]
+            },
+            "inheritance": {
+                "filter": {
```
bw2: isn't the "filter" redundant with "mode"?

seqr/models.py
```diff
@@ -773,3 +782,38 @@ class Meta:
...
+    class Meta:
+        unique_together = ('created_by', 'name')
+
+        json_fields = ['guid', 'name', 'search']
+
+
+class VariantSearchResults(ModelWithGUID):
+    variant_search = models.ForeignKey('VariantSearch', on_delete=models.CASCADE)
+    families = models.ManyToManyField('Family')
+    search_hash = models.CharField(max_length=50)
```
bw2: would be nice to have docs about what info is included / not-included in search_hash and why it's needed instead of just using VariantSearchResults.guid

PR #887, opened 2019-06-03
Fix reference data install

bw2:


deploy/install_local.step6.install_seqr.sh
```diff
@@ -87,6 +87,19 @@ python -u manage.py collectstatic --no-input
...
+for table_name in \
+    reference_data_humanphenotypeontology reference_data_mgi reference_data_omim reference_data_primateai reference_data_transcriptinfo \
+    reference_data_dbnsfpgene reference_data_geneconstraint reference_data_geneexpression reference_data_geneinfo; do
+
+    psql -U postgres seqrdb -c "DROP TABLE ${table_name}"
+done
+
+REFERENCE_DATA_BACKUP_FILE=gene_reference_data_backup.gz
+wget -N https://storage.googleapis.com/seqr-reference-data/gene_reference_data_backup.gz -O ${REFERENCE_DATA_BACKUP_FILE}
+rm ${REFERENCE_DATA_BACKUP_FILE}
```
hanars: this looks like you just download and then delete the backup, where is the restore happening?

deploy/install_local.step6.install_seqr.sh
```diff
@@ -87,6 +87,19 @@ python -u manage.py collectstatic --no-input
 python -u manage.py loaddata variant_tag_types
 python -u manage.py loaddata variant_searches
 
+# download and restore gene reference data
+for table_name in \
+    reference_data_humanphenotypeontology reference_data_mgi reference_data_omim reference_data_primateai reference_data_transcriptinfo \
```
hanars: Would this work if you only dropped `reference_data_geneinfo`? Then you wouldn't need to maintain this list

bw2: oh yeah

PR #929, opened 2019-06-28
Multi expanded search categories

lizzij:
Allow multiple panels to be expanded at once, add "Expand All" and "Collapse All" button.


.gitignore
```diff
@@ -18,4 +18,8 @@ deploy/secrets/gcloud*/*
 deploy/secrets/gcloud*/*
 
 pedigree_images/*
-seqr_settings
\ No newline at end of file
+seqr_settings
+
+django_key
+gene_reference_data_backup.gz
```
hanars: you should actually delete this file and then remove this from gitignore

.gitignore
```diff
@@ -18,4 +18,8 @@ deploy/secrets/gcloud*/*
...
+semantic
```
hanars: you should actually delete this file and then remove this from gitignore

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -301,24 +305,50 @@ const PANELS = PANEL_DETAILS.map(panelDetails)
 const STAFF_PANELS = STAFF_PANEL_DETAILS.map(panelDetails)
 
 class VariantSearchFormContent extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { activeIndex: [] }
+  }
+
+  expandAll = () => {
+    this.setState({ activeIndex: [0, 1, 2, 3, 4, 5] })
```
hanars: rather than hardcoding this as a list of integers that will have to be maintained if we ever change the number of panels, I would compute this based on the length of `PANELS`. A neat way to do this in es6 is `[...PANELS.keys()]`

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -301,24 +305,50 @@ const PANELS = PANEL_DETAILS.map(panelDetails)
...
+  }
+
+  collapseAll = () => {
+    this.setState({ activeIndex: [] })
+  }
+
+  handleTitleClick = (e, itemProps) => {
+    const { index } = itemProps
```
hanars: A neat syntax you can use in es6 combines the above 2 lines as follows:
`handleTitleClick = (e, { index }) => {`

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -301,24 +305,50 @@ const PANELS = PANEL_DETAILS.map(panelDetails)
...
+          <Accordion fluid panels={this.props.user.isStaff ? STAFF_PANELS : PANELS} exclusive={false} activeIndex={this.state.activeIndex} onTitleClick={this.handleTitleClick} />
         </FormSection>
       </div>
     )
   }
 
   shouldComponentUpdate() {
     // Form content does not use passed props, so should never re-render on prop update
-    return false
+    return true
```
hanars: So what this function is doing is preventing react from unnecessarily re-rendering any of the internal content when props change. Changing this to true will cause it to ALWAYS re-render, even if none of the state or props have changed, which would be pretty bad for performance. Instead, we want this to return true if the state has changed, and return false otherwise, entirely ignoring the props. 

You can read more about this method here:
https://reactjs.org/docs/react-component.html#shouldcomponentupdate

I would recommend you try to figure out on your own how you can rewrite this function so it adds the behavior you need (actually updating when the state is changed) without getting rid of the previous behavior we want to keep (not updating when the props change). If you do find yourself struggling with this, feel free to ping me and I'd be happy to work through it with you

PR #937, opened 2019-07-02
Add gnomad sort frequency

lizzij:
Implemented gnomAD sort by genome frequency in search and for saved variants.


django_key
```diff
@@ -0,0 +1 @@
+9*?IVK:_v/L-+/X8GB8?%-hGd8J1^8p6o?bP?>B~\=bTeN
```
hanars: this file should not be committed

ui/shared/utils/constants.js
```diff
@@ -641,6 +642,7 @@ const VARIANT_SORT_OPTONS = [
     comparator: (a, b) =>
       VEP_CONSEQUENCE_ORDER_LOOKUP[a.mainTranscript.majorConsequence] - VEP_CONSEQUENCE_ORDER_LOOKUP[b.mainTranscript.majorConsequence],
   },
+  { value: SORT_BY_GNOMAD, text: 'gnomAD Genome Frequency', comparator: (a, b) => a.populations.gnomad_genomes.af - b.populations.gnomad_genomes.af },
```
hanars: This should say "gnomAD Genomes Frequency" plural instead of "Genome" singular

PR #952, opened 2019-07-16
Assign cases to analyst

lizzij:
- Assign cases to existing users with access to project or any staff user
- Show families that have been assigned to the current user


seqr/views/apis/users_api.py
```diff
@@ -36,6 +36,21 @@ def get_all_collaborators(request):
     return create_json_response(collaborators)
 
 
+@login_required(login_url=API_LOGIN_REQUIRED_URL)
+@csrf_exempt
+def get_all_analysts(request):
+    analysts = {}
+    for project in get_projects_user_can_view(request.user):
+        analysts.update(_get_project_collaborators(project, include_permissions=False))
```
hanars: So this is very close to what you want, but with one small caveat: per the original ticket " When editing, you should be able to select any existing user with access to **the project** or any staff user."  So rather than looking at all collaborators who share a project with the current user, you will want to include the project_guid in the URL and only get collaborators for that project. Let me know if you'd like to discuss this one more in person

seqr/views/apis/users_api.py
```diff
@@ -36,6 +36,21 @@ def get_all_collaborators(request):
...
+
+    for staff in User.objects.filter(is_staff=True):
+        staff_json = _get_json_for_user(staff)
+        staff_json = {staff_json['username']: staff_json}
+        analysts.update(staff_json)
```
hanars: So while this works fine, its rare that you want to create an object with one key-value pair, and then call update on another object. Usually, what you would do is directly assign the key-value pair to the final object, or you would create a single object with all the new data and call update once. So for example, you could 
```
for staff in User.objects.filter(is_staff=True):
        staff_json = _get_json_for_user(staff)
        analysts[staff_json['username']] = staff_json
```
or you could do
```
staff_analysts = {staff.username: _get_json_for_user(staff) for staff in User.objects.filter(is_staff=True)}
analysts.update(staff_analysts)
```

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -112,8 +112,10 @@ def _get_json_for_user(user):
         user = user._wrapped   # Django request.user actually stores the Django User objects in a ._wrapped attribute
 
     user_json = {_to_camel_case(field): getattr(user, field) for field in
-                ['username', 'email', 'first_name', 'last_name', 'last_login', 'is_staff', 'date_joined', 'id']}
+                ['username', 'email', 'first_name', 'last_name', 'last_login', 'is_staff', 'date_joined', 'id',
+                 'assigned_families']}
```
hanars: I'm not sure we need to return `assigned_families` at all. Because we have the `assigned_to` value on the family itself this actually becomes a lot of redundant data. We should discuss this more offline

ui/pages/Project/reducers.js
```diff
@@ -145,6 +145,34 @@ export const updateLocusLists = (values) => {
   }
 }
 
+export const loadUserOptions = () => {
```
hanars: this is actually a weird merge artifact, this function got moved to `redux/rootReducers` and then you ended up adding it back here by accident but its not longer needed here

ui/shared/components/panel/family.jsx
```diff
@@ -13,15 +13,20 @@ import TextFieldView from './view-fields/TextFieldView'
...
   FAMILY_ANALYSIS_STATUS_OPTIONS,
   FAMILY_FIELD_ANALYSIS_STATUS,
+  FAMILY_FIELD_ASSIGNED_ANALYST,
   FAMILY_FIELD_ANALYSED_BY,
   FAMILY_FIELD_FIRST_SAMPLE,
   FAMILY_FIELD_RENDER_LOOKUP,
   FAMILY_FIELD_OMIM_NUMBER,
   FAMILY_FIELD_PMIDS,
 } from '../../utils/constants'
+import { getUserOptions, getUserOptionsIsLoading, getUsersByUsername } from '../../../pages/Project/selectors'
```
hanars: One important thing to remember to do after you resolve merge conflicts is to run your code again and make sure you don't see any errors in the output for either the ui or the server. I'm pretty sure some of these functions got moved :)

ui/shared/components/panel/family.jsx
```diff
@@ -46,6 +51,56 @@ const mapSampleDispatchToProps = (state, ownProps) => ({
...
+  onChange: PropTypes.func,
+  value: PropTypes.any,
+}
+
+const EDIT_FIELDS = [
+  {
+    name: 'user',
+    label: 'Email',
+    component: connect(mapDropdownStateToProps, mapDropdownDispatchToProps)(AnalystEmailDropdown),
+    validate: value => (
```
hanars: I understand why you added this validation, and generally you would be right that its a good idea. However, when you are choosing values from a dropdown you generally don't need to validate them. For example, it would be really weird if there was an invalid email address in that dropdown. But as a user, it would be really confusing if you chose an un-editable option from a dropdown and then were told it was invalid

seqr/urls.py
```diff
@@ -138,13 +140,21 @@
 ]
 
 # NOTE: the actual url will be this with an '/api' prefix
+
+urlpatterns = []
+
+phenotips_urls = '^(?:%s)' % ('|'.join([
```
hanars: do you mind moving this back to where it was? Its not super important, it just made more sense there

seqr/views/apis/family_api.py
```diff
@@ -133,14 +135,39 @@ def update_family_fields_handler(request, family_guid):
...
+
 @login_required(login_url=API_LOGIN_REQUIRED_URL)
 @csrf_exempt
 def update_family_analysed_by(request, family_guid):
     """Updates the specified field in the Family model.
 
     Args:
         family_guid (string): GUID of the family.
-        field_name (string): Family model field name to update
+      u  field_name (string): Family model field name to update
```
hanars: I think this is a typo

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -181,6 +181,13 @@ def _process_result(result, family):
             result['individualGuids'] = [i.guid for i in family.individual_set.all()]
         if not result['displayName']:
             result['displayName'] = result['familyId']
+        if result['assignedAnalyst']:
+            result['assignedAnalyst'] = {
+                'fullName': result['assignedAnalyst'].get_full_name(),
+                'email': result['assignedAnalyst'].email,
+            }
+        else:
+            result['assignedAnalyst'] = ''
```
hanars: Its probably better to use `None` here, because thats the "falsey" version of objects and its better to always have the value of a field have the same type, and while the empty string is always falsey it is technically a different type

ui/pages/Project/selectors.js
```diff
@@ -379,3 +379,21 @@ export const getUserOptions = createSelector(
...
+  project => project.collaborators,
+)
+
+// analyst option selectors (add project collaborators to staff users)
+export const getAnalystOptions = createSelector(
+  getCollaborators,
+  getAllUsers,
+  (collaborators, users) => {
+    const staff = users.filter(user => user.isStaff)
+    const uniqueCollaborators = collaborators.filter(collaborator => !collaborator.isStaff)
```
hanars: nice, good catching of an edge case!

ui/redux/rootReducer.js
```diff
@@ -86,6 +86,19 @@ export const loadUserOptions = () => {
   }
 }
 
+export const loadStaffOptions = () => {
+  return (dispatch) => {
+    dispatch({ type: REQUEST_USERS })
+    new HttpRequestHelper('/api/users/get_all_staff',
```
hanars: I notice you aren't checking if these are already loaded before deciding whether or not to request these. Is the bug we were seeing related to that fixed?

ui/shared/components/panel/family.jsx
```diff
@@ -5,23 +5,27 @@ import { Grid, Popup, Icon } from 'semantic-ui-react'
 import { Link } from 'react-router-dom'
 import styled from 'styled-components'
 
-import { updateFamily } from 'redux/rootReducer'
-import { getProjectsByGuid, getFirstSampleByFamily } from 'redux/selectors'
+import { updateFamily, loadStaffOptions } from '../../../redux/rootReducer'
+import { getProjectsByGuid, getFirstSampleByFamily, getUserOptionsIsLoading, getUsersByUsername } from '../../../redux/selectors'
```
hanars: For imports from outside the same top level folder it is actually better style to use the absolute import (i.e. `'redux/selectors'` instead of the relative import (i.e. `'../../../redux/selectors`')

ui/shared/components/panel/family.jsx
```diff
@@ -46,6 +50,53 @@ const mapSampleDispatchToProps = (state, ownProps) => ({
 
 const FirstSample = connect(mapSampleDispatchToProps)(BaseFirstSample)
 
+const AnalystEmailDropdown = ({ load, loading, usersByUsername, onChange, value, ...props }) =>
+  <DataLoader load={load} loading={false} content>
+    <AddableSelect
```
hanars: I don't think this should be an `AddableSelect` I think it should be a regular select, because we only want them to be able to choose an existing user and not add a new one

ui/shared/components/panel/family.jsx
```diff
@@ -54,6 +105,12 @@ const familyFieldRenderProps = {
       <ColoredIcon name="stop" color={value.color} />
     ),
   },
+  [FAMILY_FIELD_ASSIGNED_ANALYST]: {
+    formFields: EDIT_FIELDS,
+    addConfirm: 'Are you sure you want to add the analyst to this family?',
+    fieldDisplay: value => (value ? <div>{(value.fullName) ? (value.fullName).concat(' - ') : ''}{value.email}</div> :
```
hanars: generally we only show the email if theres no name, not both at once

ui/redux/rootReducer.js
lizzij: Yes, it is fixed now.

ui/redux/rootReducer.js
```diff
@@ -86,6 +86,19 @@ export const loadUserOptions = () => {
   }
 }
 
+export const loadStaffOptions = () => {
```
hanars: So this function is identical to `loadUserOptions` except that the url is a tiny bit different. I would update this so `loadUserOptions` accepts a `staffOnly` parameter and conditionally creates the url. The you can define `loadStaffOptions` as 
```
export const loadStaffOptions = () => loadUserOptions(true)
```
The advantage to this is if for some reason in the future I need to change some of the behavior for fetching pr receiving users I only need to change it in one place instead of 2

ui/shared/components/panel/family.jsx
```diff
@@ -46,6 +50,53 @@ const mapSampleDispatchToProps = (state, ownProps) => ({
 
 const FirstSample = connect(mapSampleDispatchToProps)(BaseFirstSample)
 
+const AnalystEmailDropdown = ({ load, loading, usersByUsername, onChange, value, ...props }) =>
+  <DataLoader load={load} loading={false} content>
+    <Select
+      loading={loading}
+      additionLabel="Assigned Analyst: "
```
hanars: this prop shouldn't be used anymore, so regular selects don't have the addition option

seqr/views/apis/family_api.py
```diff
@@ -133,6 +135,35 @@ def update_family_fields_handler(request, family_guid):
...
+    assigned_analyst_username = request_json.get('assigned_analyst_username')
+
+    if not assigned_analyst_username:
+        return create_json_response(
+            {}, status=400, reason="'assigned analyst' is not specified")
+    try:
+        assigned_analyst = User.objects.get(username=assigned_analyst_username)
+    except Exception:
+        return create_json_response(
+            {}, status=400, reason="user does not have access")
```
hanars: this isn't really an accurate message, the actual error is "specified user des not exist"

seqr/models.py
```diff
@@ -6,7 +6,7 @@
 from django.contrib.auth.models import User, Group
 from django.contrib.postgres.fields import JSONField, ArrayField
 from django.db import models
-from django.db.models import options
+from django.db.models import options, ForeignKey
```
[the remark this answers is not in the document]
hanars: this is still true?

ui/shared/components/panel/family.jsx
hanars: this is still true

PR #961, opened 2019-07-22
Variant breakdown for family

lizzij:
show the breakdown across only the tags in the specific family on variant-specific pages.


ui/shared/components/graph/VariantTagTypeBar.jsx
```diff
@@ -5,22 +5,32 @@ import HorizontalStackedBar from '../graph/HorizontalStackedBar'
...
 
 
-export const getVariantTagTypeCount = (vtt, familyGuids) => (
-  familyGuids ? familyGuids.reduce((count, familyGuid) => count + (vtt.numTagsPerFamily[familyGuid] || 0), 0) : vtt.numTags
-)
+export const getVariantTagTypeCount = (vtt, familyGuids, variantGuid, variantsToDisplay) => {
+  if (familyGuids) {
+    return familyGuids.reduce((count, familyGuid) => count + (vtt.numTagsPerFamily[familyGuid] || 0), 0)
+  }
+  else if (variantGuid) {
```
hanars: From the ticket: 
```
variant-specific pages... should show the breakdown across only the tags in that specific family
```
What this code will do is show you the breakdown of tags for that single variant. However, this should be the breakdown of all the tags for the family. So what you want to do here is get the `familyGuids` from the variant and use those

ui/shared/components/graph/VariantTagTypeBar.jsx
```diff
@@ -5,22 +5,27 @@ import HorizontalStackedBar from '../graph/HorizontalStackedBar'
 import { EXCLUDED_TAG_NAME, REVIEW_TAG_NAME, NOTE_TAG_NAME } from '../../utils/constants'
 
 
-export const getVariantTagTypeCount = (vtt, familyGuids) => (
-  familyGuids ? familyGuids.reduce((count, familyGuid) => count + (vtt.numTagsPerFamily[familyGuid] || 0), 0) : vtt.numTags
-)
+export const getVariantTagTypeCount = (vtt, familyGuids, variantGuid, variantsToDisplay) => {
```
hanars: You have to be very careful when you change the function signature for functions that are exported. This could be used elsewhere and break. Rather than adding new arguments here, I think you should change how `familyGuid` is chosen in the first place. So instead of passing `variantGuid` and `variantsToDisplay` into the graph component, just change the props thats passed in from 
```
familyGuid={familyGuid}
variantGuid={variantGuid}
variantsToDisplay={this.props.variantsToDisplay}
```
to
```
familyGuid={variantGuid ?  (this.props.variantsToDisplay[0] || {}).familyGuids || [])[0] :  familyGuid}
```

ui/shared/components/graph/VariantTagTypeBar.jsx
```diff
@@ -5,9 +5,12 @@ import HorizontalStackedBar from '../graph/HorizontalStackedBar'
 import { EXCLUDED_TAG_NAME, REVIEW_TAG_NAME, NOTE_TAG_NAME } from '../../utils/constants'
 
 
-export const getVariantTagTypeCount = (vtt, familyGuids) => (
-  familyGuids ? familyGuids.reduce((count, familyGuid) => count + (vtt.numTagsPerFamily[familyGuid] || 0), 0) : vtt.numTags
-)
+export const getVariantTagTypeCount = (vtt, familyGuids) => {
```
hanars: you could probably just switch this back to the original ternary statment (`familyGuids ? familyGuids.reduce((count, familyGuid) => count + (vtt.numTagsPerFamily[familyGuid] || 0), 0) : vtt.numTags`)

ui/shared/components/graph/VariantTagTypeBar.jsx
```diff
@@ -33,15 +36,19 @@ const VariantTagTypeBar = ({ project, familyGuid, analysisGroup, sectionLinks =
...
+          familyGuid ? [familyGuid] : (analysisGroup || {}).familyGuids),
+        ...vtt }
     })}
   />
 )
 
 VariantTagTypeBar.propTypes = {
   project: PropTypes.object.isRequired,
   familyGuid: PropTypes.string,
+  variantGuid: PropTypes.string,
```
hanars: you should take these out since they aren't used

ui/shared/components/panel/variants/SavedVariants.jsx
```diff
@@ -232,7 +232,9 @@ class BaseSavedVariants extends React.Component {
                 height={30}
                 project={this.props.project}
                 analysisGroup={this.props.analysisGroup}
-                familyGuid={familyGuid}
+                familyGuid={variantGuid ? ((this.props.variantsToDisplay[0] || {}).familyGuids || [])[0] : familyGuid}
+                variantGuid={variantGuid}
+                variantsToDisplay={this.props.variantsToDisplay}
```
hanars: theres no need to pass these props in as they aren't used anymore

PR #963, opened 2019-07-23
Revert "Assign cases to analyst"

lizzij:
Reverts macarthur-lab/seqr#952


lizzij:
Fixed in the original branch "assign-cases-to-analyst" and added another pull request.

PR #967, opened 2019-07-23
Save as gene note

lizzij:
Seqr allows users to add notes about genes that are shared by all users and shown on the gene page/ detail modal (see https://seqr.broadinstitute.org/gene_info/ENSG00000121753 for an example). Seqr also has variant-level notes that analysts use while researching potential candidate variants. Sometimes, the variant level notes they have are pertinent to the whole gene, not just the variant, so there should be an option when adding a variant-level note to also make that note a public gene note. If that option is selected, the note will be created as both a variant-level note and a gene-level note.
If the variant is annotated in multiple genes, the gene note should only be created for the main transcript gene


seqr/views/apis/saved_variant_api.py
```diff
@@ -92,14 +93,39 @@ def create_saved_variant_handler(request):
 @login_required(login_url=API_LOGIN_REQUIRED_URL)
 @csrf_exempt
 def create_variant_note_handler(request, variant_guid):
+    save_as_gene_note = json.loads(request.body).get('saveAsGeneNote')
```
hanars: you should probably do `request_json = json.loads(request.body)` once at the begining fo this function, rather than re-parsing the json 3 times

seqr/views/apis/saved_variant_api.py
```diff
@@ -92,14 +93,39 @@ def create_saved_variant_handler(request):
...
-    return create_json_response({'savedVariantsByGuid': {variant_guid: {
-        'notes': [get_json_for_variant_note(tag) for tag in saved_variant.variantnote_set.all()]
-    }}})
+    return create_json_response({
+        'savedVariantsByGuid': variant_note,
+        'genesById': gene_note
+    })
+
+
+def _create_gene_note(gene_id, note_json, user):
```
hanars: I don't think making these into helper functions adds much. If there was a bunch of things being done in the function it would add readability, but these are both one liners. If there is shared functionality and the helper function is used in multiple places, thats good because it makes it easier to make updates in the future. But in general, a one line helper function thats only called once is less readable than just running the code in line

PR #971, opened 2019-07-25
Scroll pages at bottom

lizzij:
- add pagination bar at the bottom of the page
- add a scroll to top button at the bottom of the page
- pagination bar hidden when <= 1 page

### DIFF
--- a/deploy/install_local.step9.load_test_dataset.sh
+++ b/deploy/install_local.step9.load_test_dataset.sh
@@ -17,12 +17,14 @@ fi
 
 cd ${SEQR_DIR}/hail_elasticsearch_pipelines
 
-nohup python2.7 gcloud_dataproc/submit.py --cpu-limit 2 --num-executors 2  --run-locally hail_scripts/v01/load_dataset_to_es.py \
+PIPELINE_CPU_LIMIT=2
+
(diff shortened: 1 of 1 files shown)


ui/pages/Search/components/VariantSearchResults.jsx
```diff
@@ -37,16 +37,19 @@ const FIELDS = [
   VARIANT_SORT_FIELD_NO_FAMILY_SORT,
 ]
 
-
 const BaseVariantSearchResults = ({
   match, searchedVariants, variantSearchDisplay, searchedVariantExportConfig, onSubmit, load, unload, loading, errorMessage, totalVariantsCount,
 }) => {
+  const scrollToTop = () => window.scrollTo(0, 0)
```
hanars: you should actually define this as a constant outside of the  component, as it does not require any of the component props and theres therefore no reason to have it re-created on render.

ui/pages/Search/components/VariantSearchResults.jsx
```diff
@@ -91,6 +94,24 @@ const BaseVariantSearchResults = ({
           <Variants variants={searchedVariants} />
         </Grid.Column>
       </Grid.Row>
+      {searchHash &&
+        <LargeRow>
+          <Grid.Column width={11} floated="right" textAlign="right">
+            <ReduxFormWrapper
```
hanars: presumably, this is either the same or very similar to the pagination component at the top of the page. therefore, you should  make it its own component and then use that component here and at the top of the page

PR #973, opened 2019-07-26
Add default search categories

lizzij:
Add default empty values and deselect options: 
- Pathogenicity: "Any"
- Annotations: "All"
- Location: "None"
- Call Quality: "All Variants" 

For the Saved Search dropdown, add a way deselect it that resets the search to empty.


ui/pages/Search/components/SavedSearch.jsx
```diff
@@ -78,8 +78,8 @@ const SavedSearches = ({ options, savedSearchesByGuid, selectedSearch, load, loa
       loading={loading}
       placeholder="Select a Saved Search"
       value={(selectedSearch || {}).savedSearchGuid}
-      onChange={val => onChange(savedSearchesByGuid[val].search)}
-      options={options}
+      onChange={val => (val ? onChange(savedSearchesByGuid[val].search) : onChange({}))}
+      options={[{ text: 'None', value: null, category: 'Default Searches' }].concat(options)}
```
hanars: you should never construct new arrays in a prop definition, you should update the selector that creates `options` to include all the options you want it to have

ui/pages/Search/components/SavedSearch.jsx
```diff
@@ -78,8 +78,8 @@ const SavedSearches = ({ options, savedSearchesByGuid, selectedSearch, load, loa
       loading={loading}
       placeholder="Select a Saved Search"
       value={(selectedSearch || {}).savedSearchGuid}
-      onChange={val => onChange(savedSearchesByGuid[val].search)}
-      options={options}
+      onChange={val => (val ? onChange(savedSearchesByGuid[val].search) : onChange({}))}
```
hanars: nitpicky, but its a bit neater to do `val => onChange(savedSearchesByGuid[val].search || {})`

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -181,7 +206,22 @@ const PATHOGENICITY_PANEL = pathogenicityPanel(false)
 
 const ANNOTATION_PANEL = {
   name: 'annotations',
-  headerProps: { title: 'Annotations', inputProps: JsonSelectProps(ANNOTATION_FILTER_OPTIONS) },
+  headerProps: {
+    inputProps: {
+      component: Select,
+      options: ANNOTATION_FILTER_OPTIONS,
+      format: (val) => {
```
hanars: Given that you have essentially-copy-pasted a lot of shared functionality for all the panels, you should instead update the `JsonSelectProps` helper to do this logic for you

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -221,15 +261,15 @@ const LOCATION_PANEL = {
     title: 'Location',
     name: 'locus',
     inputSize: 5,
-    inputProps: { component: LocusListSelector, format: val => val || {} },
+    inputProps: { component: LocusListSelector, format: val => val || { locusListGuid: '' } },
```
hanars: why does having `locusListGuid` evaluate to an empty string instead of undefined change things?

ui/pages/Search/components/filters/LocusListSelector.jsx
```diff
@@ -12,14 +12,16 @@ class BaseLocusListDropdown extends React.Component
...
-        value={locusList.locusListGuid}
-        onChange={locusListGuid => onChange({ locusListGuid })}
-        options={projectLocusListOptions}
-      />
+      <div>
+        <Dropdown
+          inline
+          selection
+          label="Gene List"
+          value={locusList.locusListGuid === '' ? null : locusList.locusListGuid}
```
hanars: having an equivalence check for an empty string and then assigning it to null is VERY risky and confusing to read. Why not just assign it to null if you want it to be null? Also, whats wrong with it just being undefined?

ui/pages/Search/constants.js
```diff
@@ -187,30 +187,62 @@ export const STAFF_PATHOGENICITY_FIELDS = [
...
       [HGMD_NAME]: HGMD_OPTIONS.map(({ value }) => value),
     },
   },
-]
+}
+
+export const STAFF_PATHOGENICITY_FILTER_OPTIONS = [
+  ALL_PATHOGENICITY_FILTER, PATH_LIKELY_PATH_FILTER, NOT_BENIGN,
+].map(value => ({ value, ...STAFF_PATHOGENICITY_FILTER_LOOKUP[value] }))
+
```
hanars: In most places we define the `OPTIONS` as a list, and then we do a reduce to get the `LOOKUP`, rather than what you have done here which is the reverse, so this is kind of an anti-pattern. We only do the defined exported constants for specific filters if you actually need to be able to import them by themselves in other components, which you shouldn't need to do for most of these. What you should do here is leave the `STAFF_PATHOGENICITY_FILTER_OPTIONS` the same, and then do something like
```
export const STAFF_PATHOGENICITY_FILTER_LOOKUP = STAFF_PATHOGENICITY_FILTER_OPTIONS.reduce(
  (acc, option) => ({ ...acc, [option.value]: option }), {}
)
```

ui/redux/utils/configureStore.js
```diff
@@ -19,11 +19,38 @@ const persistStoreMiddleware = store => next => (action) => {
...
+//   initialState = {},
+// ) => {
+//
+//   PERSISTING_STATE.forEach((key) => { initialState[key] = loadState(key) })
+//
+//   console.log('Creating store with initial state:')
+//   console.log(initialState)
+//
+//   return createStore(rootReducer, initialState, enhancer)
+// }
```
hanars: this seems like an accidental commit?

PR #977, opened 2019-07-30
Success stories

lizzij:
In family page
- Select success story types from a dropdown menu
- Add success story


seqr/models.py
```diff
@@ -237,6 +244,15 @@ class Family(ModelWithGUID):
     assigned_analyst = models.ForeignKey(User, null=True, on_delete=models.SET_NULL,
                                     related_name='assigned_families')  # type: ForeignKey
 
+    # success_story_types = models.ManyToManyField('SuccessStoryType')
```
hanars: I think you forgot to remove this

seqr/models.py
```diff
@@ -237,6 +244,15 @@ class Family(ModelWithGUID):
...
+    success_story_types = ArrayField(models.CharField(
+        max_length=10,
```
hanars: max length can be 1, since your values are all one character

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -186,8 +186,6 @@ def _process_result(result, family):
                 'fullName': result['assignedAnalyst'].get_full_name(),
                 'email': result['assignedAnalyst'].email,
             }
-        else:
```
hanars: why get rid of this?

ui/shared/components/panel/view-fields/TagFieldView.jsx
```diff
@@ -96,18 +96,22 @@ const TagFieldView = ({ initialValues, field, tagOptions, popup, tagAnnotation,
...
       <span>
         {displayFieldValues.map((tag) => {
           const label = <ColoredLabel size="small" color={tag.color} horizontal content={tag.name || tag.text} />
           return (
-            <span key={tag.name}>
-              <HorizontalSpacer width={5} />
-              {popup ? popup(tag)(label) : label}
-              {tagAnnotation && tagAnnotation(tag)}
-            </span>
+            showIconOnly ?
```
hanars: Rather than adding all this  logic so you can display these as colored dots with text, just display them as colored labels the way this component defaults to. Its fine, even if its not whats shown in the ticket

ui/shared/utils/constants.js
```diff
@@ -117,6 +142,11 @@ export const FAMILY_FIELD_RENDER_LOOKUP = {
     component: BaseFieldView,
     submitArgs: { familyField: 'analysed_by' },
   },
+  [FAMILY_FIELD_SUCCESS_STORY_TYPE]: {
+    name: 'Success Story Type',
+    component: TagFieldView,
```
hanars: these should be `internal: true`

ui/shared/components/panel/view-fields/TagFieldView.jsx
lizzij: This ensures backward compatibility with FamilyVariantTag, cause all details such as color/categories etc were stored in the store for FamilyVariantTags, but for SuccessStoryType one Char is stored.

hanars: But `FAMILY_SUCCESS_STORY_TYPE_OPTIONS` can include the correct mapping between the stored value and displayed name, so you should be able to use that. You may need to update them so that `name` is the actual stored one-char value and then add a `label` field with the text you want to show

ui/shared/utils/constants.js
```diff
@@ -85,6 +86,28 @@ export const FAMILY_ANALYSIS_STATUS_OPTIONS = [
...
 ]
 
+// SUCCESS STORY
+
+export const FAMILY_SUCCESS_STORY_NOVEL_DISCOVERY = 'N'
+export const FAMILY_SUCCESS_STORY_ALTERED_CLINICAL_OUTCOME = 'A'
+export const FAMILY_SUCCESS_STORY_COLLABORATION = 'C'
+export const FAMILY_SUCCESS_STORY_TECHNICAL_WIN = 'T'
+export const FAMILY_SUCCESS_STORY_DATA_SHARING = 'D'
+export const FAMILY_SUCCESS_STORY_OTHER = 'O'
```
hanars: in general, you should only export constants if they are used outside of the file, which I don't think these are

PR #982, opened 2019-08-01
scroll pages at bottom

lizzij:
In variant search page:
- add a scroll to top button
- display pagination field when > 1 pages


ui/pages/Search/components/VariantSearchResults.jsx
```diff
@@ -91,6 +91,24 @@ const BaseVariantSearchResults = ({
           <Variants variants={searchedVariants} />
         </Grid.Column>
       </Grid.Row>
+      {searchHash &&
+        <LargeRow>
+          <Grid.Column width={11} floated="right" textAlign="right">
+            <ReduxFormWrapper
+              onSubmit={onSubmit}
+              form="editSearchedVariantsDisplay"
```
hanars: you should never have two forms with the same `form` prop, you should name this something different to differentiate it from the other one

PR #984, opened 2019-08-07
Edit variant notes in line

lizzij:
Clicking the edit button to toggle the note from a display-only view to an edit view in the page.


ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -268,7 +271,11 @@ class ReduxFormWrapper extends React.Component {
   }
 
   componentWillUpdate(nextProps) {
-    if (nextProps.submitSucceeded && nextProps.closeOnSuccess && !nextProps.noModal) {
+    if (nextProps.submitSucceeded && nextProps.closeOnSuccess && nextProps.noModal) {
+      this.props.closeInlineModal()
+      this.props.handleClose(true)
```
hanars: Before, if the `noModal` prop was passed `this.props.handleClose(true)` was NOT called, so this changes the behavior of every other time this component is rendered with that prop, almost definitely in a breaking way. Additionally, those forms do not have `closeInlineModal` passed in so calling it as a function will also break them. Please update this so it works for your components with the new prop without breaking backwards compatibility

ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -147,6 +147,9 @@ class ReduxFormWrapper extends React.Component {
     /* React child component class. Mutually exclusive with fields */
     children: PropTypes.node,
 
+    /* Call if submit succeeded */
+    closeInlineModal: PropTypes.func,
```
hanars: I would name this something more generic, like `onSubmitSucceeded`, so that you can pass any function you want to call after a successful submission

ui/shared/components/buttons/UpdateButton.jsx
```diff
@@ -8,35 +8,39 @@ import Modal from '../modal/Modal'
...
-        initialValues={initialValues}
-        fields={formFields}
-        showErrorPanel={showErrorPanel}
-        confirmDialog={confirmDialog}
-        submitButtonText={submitButtonText}
-        confirmCloseIfNotSaved
-      />
-    ) }) }
-  </Modal>
+) => {
```
hanars: It looks like nothing changed here functionally, so can you change this back to how it was?

ui/shared/components/panel/view-fields/BaseFieldView.jsx
```diff
@@ -18,76 +20,126 @@ const FieldValue = styled.div`
...
+    const hasValue = (fieldValue && (!Object.getOwnPropertyNames(fieldValue).includes('length') || fieldValue.length > 0)) || this.props.showEmptyValues
+    if (!this.props.isEditable && !hasValue) {
+      return null
+    }
+    const fieldId = this.props.initialValues[this.props.idField]
+    const modalId = this.props.isEditable ? `edit-${fieldId || 'new'}-${this.props.field}-${this.props.modalId}` : null
+
+    const updateButton = this.props.showInLine ?
+      <div>
+        {this.state.showInLineButton &&
```
hanars: rather than doing a conditional for `this.state.showInLineButton` followed immediately by another conditional for `!this.state.showInLineButton` you could just use a ternary:
```
{this.state.showInLineButton ?
  <ButtonLink  icon="write"  onClick={this.toggleButtonVisibility}  /> : 
  <Segment>
          <ReduxFormWrapper
            noModal
            onSubmit={this.props.onSubmit}
            closeInlineModal={this.toggleButtonVisibility}
            form={this.props.modalId}
            initialValues={this.props.initialValues}
            fields={this.props.formFields}
            showErrorPanel={this.props.showErrorPanel}
            confirmDialog={this.props.confirmDialog}
            confirmCloseIfNotSaved
          />
  </Segment>
}
```

ui/shared/components/panel/view-fields/BaseFieldView.jsx
```diff
@@ -18,76 +20,126 @@ const FieldValue = styled.div`
...
+          initialValues={this.props.initialValues}
+          formFields={this.props.formFields}
+          formContainer={<div style={this.props.modalStyle} />}
+          showErrorPanel={this.props.showErrorPanel}
+          confirmDialog={this.props.addConfirm}
+          size="tiny"
+        />
+      )
+
+    const editButton = this.props.isEditable && (this.props.formFields ?
```
hanars: So this is very nitpicky, but the way the control flow here is kind of confusing, because before if it wasn't editable `updateButton` was just never defined, and now you are always defining it and then sometimes just not using it. So for non-editable fields, or fields without `formFields` this is kind of confusing because there are all these unused components being made and its kind of hard to read through and follow the logic for what gets created.  I think rather than defining `updateButton` and then maybe assigning it to `editButton`, you should just initialize the `editButton` variable and then assign it using a series of if statememts:
```
let editButton
if (this.props.isEditable) {
   if (this.props.showInLine) {
     ...
   }
   else if (this.props.formFields) {
      editButton = ...
   } else {
      editButton = ...
   }
}
```

ui/shared/components/panel/view-fields/BaseFieldView.jsx
```diff
@@ -113,6 +165,7 @@ BaseFieldView.propTypes = {
   showErrorPanel: PropTypes.bool,
   modalId: PropTypes.string,
   modalSize: PropTypes.string,
+  confirmDialog: PropTypes.string,
```
hanars: why did you add this prop?

ui/shared/components/panel/view-fields/BaseFieldView.jsx
```diff
@@ -18,76 +20,126 @@ const FieldValue = styled.div`
...
-      ]}
-      {
-        hasValue && !props.hideValue &&
-        <FieldValue compact={props.compact} fieldName={props.fieldName}>
-          {props.fieldDisplay(fieldValue, props.compact, fieldId)}
-        </FieldValue>
-      }
-      {!props.fieldName && buttons}
-    </span>)
+  shouldComponentUpdate(nextProps, nextState) {
```
hanars: you should use `React.PureComponent` instead of `React.Component` and then you won't need this function at all

ui/shared/components/panel/view-fields/BaseFieldView.jsx
lizzij: ESLint: The no-nested-ternary rule disallows nested ternary expressions.

ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -268,7 +271,11 @@ class ReduxFormWrapper extends React.Component {
   }
 
   componentWillUpdate(nextProps) {
-    if (nextProps.submitSucceeded && nextProps.closeOnSuccess && !nextProps.noModal) {
+    if (this.props.onSubmitSucceeded && nextProps.submitSucceeded && nextProps.closeOnSuccess && nextProps.noModal) {
```
hanars: I think if you pass an `onSubmitSucceeded` callback and submit succeeded you should call the callback but not call `handleClose` at all. So I don't think you need to check `closeOnSuccess` or `noModal` here

ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -268,7 +271,11 @@ class ReduxFormWrapper extends React.Component {
   }
 
   componentWillUpdate(nextProps) {
-    if (nextProps.submitSucceeded && nextProps.closeOnSuccess && !nextProps.noModal) {
+    if (this.props.onSubmitSucceeded && nextProps.submitSucceeded) {
+      this.props.onSubmitSucceeded()
+      this.props.handleClose(true)
```
hanars: I don't think this should call `handleClose` here. I think you should make this its own `if` statement separate from the `if/else` block that already existed so that way it does 2 separate checks, one for if it should call `onSubmitSucceeded` and the other for if it should call `handleClose`

ui/shared/components/form/ReduxFormWrapper.jsx
```diff
@@ -319,14 +325,14 @@ const mapStateToProps = (state, ownProps) => ({
 })
 
 const mapDispatchToProps = (dispatch, ownProps) => {
-  return {
+  return (ownProps.onSubmitSucceeded ? null : {
```
hanars: this should be unnecessary, since you just won't call these functions in the case with `onSubmitSucceeded`. Also, in theory, you should be able to support both a success callback and a modal, so I would take this out

ui/shared/components/panel/variants/FamilyVariantTags.jsx
```diff
@@ -262,6 +262,7 @@ const FamilyVariantTags = (
           <NoteContainer>
             {savedVariant && savedVariant.notes.map(note =>
               <VariantNoteField
+                showInLine
```
hanars: I would actually move this prop directly into `VariantNoteField`, since you always want variant notes to have this prop

ui/shared/components/form/ReduxFormWrapper.jsx
lizzij: I might be wrong but I couldn't get the cancel button in `ButtonPanel` to work without this

ui/shared/components/panel/variants/FamilyVariantTags.jsx
hanars: Try passing the `noModal` prop as well

ui/shared/components/panel/variants/FamilyVariantTags.jsx
```diff
@@ -182,9 +182,11 @@ VariantTagField.propTypes = {
   family: PropTypes.object.isRequired,
 }
 
-const VariantNoteField = ({ action, note, variant, family, ...props }) => {
+const VariantNoteField = ({ action, note, variant, family, showInLine, noModal, ...props }) => {
```
hanars: you aren't using `showInLine` and `noModal` so theres no reason to declare them here

ui/shared/components/form/ReduxFormWrapper.jsx
hanars: with `noModal` you shouldn't need this

lizzij: on submit/cancel: `TypeError: Cannot read property 'confirmOnClose' of undefined`

hanars: That error is because `handleClose` is being called but it shouldn't be when `noModal` is passed. You need to figure out why that is happening and fix that issue, not get rid of the function entirely

ui/shared/components/form/RichTextEditor.jsx
```diff
@@ -92,7 +92,7 @@ class RichTextEditor extends React.Component {
             }}
           />
         </div>
-        <div style={{ border: '1px #DDD solid', padding: '10px' }}>
+        <div style={{ width: '590px', border: '1px #DDD solid', padding: '10px' }}>
```
hanars: this should be `min-width`

PR #988, opened 2019-08-12
Staff success story page

lizzij:
Choose "success story types", or choose to view all types. The table include:
- The family ID (including a link to the family page)
- The success story type
- The success story text

The "Filter" bar at the top filters on the success story text.


django_key
```diff
@@ -0,0 +1 @@
+IsD,%SQb+y~xhw1/`w+\q0"*'k8$'0Cb=.S[7j)10FltEsJ
```
hanars: This should not get committed

seqr/views/apis/staff_api.py
```diff
@@ -409,6 +409,25 @@ def discovery_sheet(request, project_guid):
...
 
+@staff_member_required(login_url=API_LOGIN_REQUIRED_URL)
+def success_story(request, success_story_types):
+    errors = []
+
+    if success_story_types == 'all':
+        all_families = Family.objects.filter(~Q(success_story=None))
+    else:
+        all_families = Family.objects.none()
+        for success_story_type in success_story_types:
```
hanars: What type is `success_story_types`? Its coming from a URL query parameter so it must be a string, but you are treating it as a list?

seqr/views/apis/staff_api.py
```diff
@@ -409,6 +409,25 @@ def discovery_sheet(request, project_guid):
...
+            families = Family.objects.filter(success_story_types__contains=[success_story_type])
+            all_families = all_families.union(families)
```
hanars: This can all be done in aa single query: https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/fields/#overlap

seqr/views/apis/staff_api.py
```diff
@@ -409,6 +409,25 @@ def discovery_sheet(request, project_guid):
...
+    rows = _generate_family_rows(all_families)
+
+    return create_json_response({
+        'rows': rows,
+        'errors': errors,
```
hanars: If nothing is ever put in errors theres no reason to have it

seqr/views/apis/staff_api.py
```diff
@@ -409,6 +409,25 @@ def discovery_sheet(request, project_guid):
...
+    errors = []
+
+    if success_story_types == 'all':
+        all_families = Family.objects.filter(~Q(success_story=None))
+    else:
+        all_families = Family.objects.none()
+        for success_story_type in success_story_types:
+            families = Family.objects.filter(success_story_types__contains=[success_story_type])
+            all_families = all_families.union(families)
+    rows = _generate_family_rows(all_families)
```
hanars: The helper function is really small and only used once, so I think its more readable just to do it in line here:
```
rows = [{
     "project_guid": family.project.guid,
     "family_guid": family.guid,
     "family_id": family.family_id,
     "success_story_types": family.success_story_types,
      "success_story": family.success_story,
      "row_id": family.guid,
} for family in all_families]
```

ui/pages/Staff/components/BaseReport.jsx
```diff
@@ -43,6 +43,7 @@ const BaseReport = ({ page, viewAllCategory, idField, defaultSortColumn, getDown
       data={data}
       columns={columns}
       loadingProps={LOADING_PROPS}
+      getRowFilterVal={getRowFilterVal}
```
hanars: I don't think you ended up using this, so you should take it back out

ui/pages/Staff/components/SuccessStory.jsx
```diff
@@ -0,0 +1,103 @@
...
+      editLabel="choose success story types"
+      field="successStoryTypes"
+      idField="row_id"
+      initialValues={formatInitialValue(match)}
+      tagOptions={FAMILY_SUCCESS_STORY_TYPE_OPTIONS}
+      onSubmit={value => history.push(`/staff/success_story/${value.successStoryTypes}`)}
+      showIconOnly
+      simplifiedValue
+      fieldDisplay={value => value.map(tag =>
+        <span>
```
hanars: Why not reuse this from the edit on the family page?

ui/pages/Staff/components/SuccessStory.jsx
```diff
@@ -0,0 +1,103 @@
...
+        </span>)}
+    />
+    or <NavLink to="/staff/success_story/all" activeStyle={ACTIVE_LINK_STYLE}>view all success stories</NavLink>
+    <HorizontalSpacer width={20} />
+    {filters}
+    <VerticalSpacer height={15} />
+    <SortableTable
+      striped
+      collapsing
+      horizontalScroll
```
hanars: Since there aren't so many columns, instead of doing `collapsing` with `horizontalScroll` and a `minWidth` on the columns, just leave these out and the table will automatically divide the width evenly among all columns

ui/pages/Staff/constants.js
```diff
@@ -90,6 +92,16 @@ const formatFamilySummary = row =>
...
+const formatIDLink = row =>
+  <div>
+    <Link to={`/project/${row.project_guid}/family_page/${row.family_guid}`} target="_blank">{row.family_id}</Link>
+  </div>
+
+const formatSuccessStoryTypes = row =>
+  row.success_story_types && <div>{row.success_story_types.map(tag => <div><small>{FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP[tag].name}</small></div>)}</div>
+
+const formatDiscoveryTags = row => <div>{row.extras_variant_tag_list}</div>
+// row.extras_variant_tag_list && <div>{row.extras_variant_tag_list.map(tag => <div><small>{tag}</small></div>)}</div>
```
hanars: I think this is an accidental comment?

ui/pages/Staff/constants.js
```diff
@@ -153,3 +165,10 @@ export const DISCOVERY_SHEET_COLUMNS = [
   { name: 'num_individuals_sequenced', content: '# of Individuals Sequenced' },
   { name: 'analysis_summary', content: 'Analysis Summary', style: { minWidth: '800px' } },
 ]
+
+export const SUCCESS_STORY_COLUMNS = [
+  { name: 'family_id', content: 'Family ID', format: formatIDLink, noFormatExport: true, style: { minWidth: '200px' } },
+  { name: 'success_story_types', content: 'Success Story Types', format: formatSuccessStoryTypes, style: { minWidth: '300px' } },
+  { name: 'success_story', content: 'Success Story', style: { minWidth: '564px' } },
+  { name: 'discovery_tags', content: 'Discovery Tags', format: formatDiscoveryTags, noFormatExport: true, style: { minWidth: '400px' } },
+]
```
hanars: Can you double check that when you download this it looks good?

ui/pages/Staff/constants.js
```diff
@@ -90,6 +92,16 @@ const formatFamilySummary = row =>
...
 
+const formatIDLink = row =>
+  <div>
+    <Link to={`/project/${row.project_guid}/family_page/${row.family_guid}`} target="_blank">{row.family_id}</Link>
+  </div>
+
+const formatSuccessStoryTypes = row =>
+  row.success_story_types && <div>{row.success_story_types.map(tag => <div><small>{FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP[tag].name}</small></div>)}</div>
+
+const formatDiscoveryTags = row => <div>{row.extras_variant_tag_list}</div>
```
hanars: If all you are doing is grabbing a different value from the object to display, why not just make `extras_variant_tag_list` the name and get rid of the format function?

ui/pages/Staff/constants.js
```diff
@@ -90,6 +92,16 @@ const formatFamilySummary = row =>
     }
   </div>
 
+const formatIDLink = row =>
+  <div>
+    <Link to={`/project/${row.project_guid}/family_page/${row.family_guid}`} target="_blank">{row.family_id}</Link>
+  </div>
+
+const formatSuccessStoryTypes = row =>
+  row.success_story_types && <div>{row.success_story_types.map(tag => <div><small>{FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP[tag].name}</small></div>)}</div>
```
hanars: These should be displayed the way they are on the family page

ui/pages/Staff/constants.js
lizzij: When downloaded, success story types are in short form. Otherwise, it looks fine to me.

seqr/views/apis/staff_api.py
```diff
@@ -409,6 +409,29 @@ def discovery_sheet(request, project_guid):
     })
 
 
+@staff_member_required(login_url=API_LOGIN_REQUIRED_URL)
+def success_story(request, success_story_types):
+
+    if success_story_types == 'all':
+        families = Family.objects.filter(~Q(success_story=None))
```
hanars: A better way to do this is using an `isnull` filter: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#std:fieldlookup-isnull

ui/pages/Staff/constants.js
```diff
@@ -13,7 +13,10 @@ import {
   INDIVIDUAL_FIELD_MATERNAL_ID,
   INDIVIDUAL_FIELD_SEX,
   INDIVIDUAL_FIELD_AFFECTED,
+  FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP,
 } from 'shared/utils/constants'
+import {ColoredIcon} from "../../shared/components/StyledComponents";
```
hanars: this is definitely failing the linter. Before pushing up code, please remember to always check and make sure you don't see any errors

ui/pages/Staff/constants.js
```diff
@@ -90,6 +93,24 @@ const formatFamilySummary = row =>
     }
   </div>
 
+const formatIDLink = row =>
+  <div>
+    <Link to={`/project/${row.project_guid}/family_page/${row.family_guid}`} target="_blank">{row.family_id}</Link>
+  </div>
```
hanars: why does this need to be in a div?

ui/pages/Staff/constants.js
```diff
@@ -90,6 +93,24 @@ const formatFamilySummary = row =>
...
+
+const formatSuccessStoryTypes = row =>
+  row.success_story_types &&
+  <div>
+    {row.success_story_types.map(tag =>
+      <div>
+        <ColoredIcon name="stop" color={FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP[tag].color} />
+        {FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP[tag].name}
```
hanars: You should have one shared component for displaying success story types, and that should be used everywhere they are displayed

ui/pages/Staff/constants.js
hanars: By short form do you mean the single letter code? Because if so, could you fix that?

ui/shared/utils/constants.js
```diff
@@ -108,6 +109,12 @@ export const FAMILY_SUCCESS_STORY_TYPE_OPTIONS_LOOKUP = FAMILY_SUCCESS_STORY_TYP
   return { [tag.value]: tag, ...acc }
 }, {})
 
+export const FAMILY_SUCCESS_STORY_TYPE_TAG = tag =>
```
hanars: this is a confusing variable name, I would expect something with this name to be the string value of the field key or something. For a function, you should use camel case and make it describe what it does, so maybe something like `successStoryTypeDisplay`?

hanars: Also, I think this should be the default display for the tag field view, which is colored labels right?

PR #1143, opened 2020-01-22
Docker compose for local installs

bw2:
`cd deploy; docker-compose up ` now brings up all containers, and seqr landing page shows up at http://localhost:8000 . Still need to 
- fix static images
- create first user account
- update and test pipeline-runner image  & data loading


deploy/docker-compose.yml
```diff
@@ -0,0 +1,91 @@
...
+      - 8080:8080
+    depends_on:
+      - init-postgres
+
+  seqr:
+    image: gcr.io/seqr-project/seqr
+    ports:
+      - 8000:8000
+    environment:
+      - SEQR_GIT_BRANCH=7ce6e4066fd7cf6d0bc2dda838c18a1818a9a73f
```
hanars: is this what we want here?

hanars:
codacy is making some recommendations about apt-get. Can you take a look and see if they make sense and if so implement them? They sound like good ideas but also its very possible they are wrong

deploy/docker-compose.yml
bw2: you're right, I'll change it to 'dev'

PR #1189, opened 2020-02-19
Fix postgres disk mount

bw2:


bw2:
I've deployed this to dev with
`./servctl deploy -d postgres gcloud-dev `
and restored from the latest backup.

PR #1229, opened 2020-03-25
Updated the testcase for edit_families_handler and added a testcase f…

ShifaSZ:
There are three updates:
1. Modify the edit_families_handler api test case by adding more input data to increase coverage.
2. Add a test case for delete_families_handler api.
3. Reformat the whole file and wrap some of the long lines

### DIFF
--- a/seqr/views/apis/family_api_tests.py
+++ b/seqr/views/apis/family_api_tests.py
@@ -164,7 +164,6 @@ class ProjectAPITest(TestCase):
         self.assertListEqual(response_json.keys(), ['info', 'errors', 'warnings', 'uploadedFileId'])
 
         url = reverse(edit_families_handler, args=[PROJECT_GUID])
-        # _check_login(self, url)
 
         response = self.client.post(url, content_type='application/json',
                 data=json.dumps({'uploadedFileId': response_json['uploadedFileId']}))
--- a/seqr/views/apis/family_api_tests.py
+++ b/seqr/views/apis/family_api_tests.py
@@ -10,6 +10,7 @@ from seqr.views.apis.family_api import update_family_pedigree_image, update_fami
 from seqr.views.utils.test_utils import _check_login
 
 FAMILY_GUID = 'F000001_1'
+FAMILY_GUID2 = 'F000002_2'
 
 PROJECT_GUID = 'R0001_1kg'
 EMPTY_PROJECT_GUID = 'R0002_empty'
@@ -30,9 +31,9 @@ class ProjectAPITest(TestCase):
         # send request with a "families" attribute
         req_values = {
             'families': [
-                {'familyGuid': FAMILY_GUID, 'description': 'Test'},
-                {PREVIOUS_FAMILY_ID_FIELD: '1', 'description': 'Test'},
-                {FAMILY_ID_FIELD: '13', 'description': 'Test'}
+                {'familyGuid': FAMILY_GUID, 'description': 'Test description 1'},
+                {PREVIOUS_FAMILY_ID_FIELD: '2', 'description': 'Test description 2'},
+                {FAMILY_ID_FIELD: '13', 'description': 'Test description 13'}
             ]
         }
         response = self.client.post(url, content_type='application/json',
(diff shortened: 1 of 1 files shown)


seqr/views/apis/family_api_tests.py
```diff
@@ -5,32 +5,62 @@
...
+                {FAMILY_ID_FIELD: '13', 'description': 'Test'}
+            ]
+        }
+        response = self.client.post(url, content_type='application/json',
+                                    data=json.dumps(req_values))
         self.assertEqual(response.status_code, 200)
         response_json = response.json()
 
         self.assertListEqual(response_json.keys(), ['familiesByGuid'])
         self.assertEqual(response_json['familiesByGuid'][FAMILY_GUID]['description'], 'Test')
```
hanars: If you are adding additional test cases to the request, please also test that the response contains the updated data. In general, if you modify the input to a test you should also be modifying the test on the response. You should add tests for the additional updated families here

seqr/views/apis/family_api_tests.py
```diff
@@ -5,32 +5,62 @@
...
+                {'familyGuid': FAMILY_GUID}
+            ]
+        }
+        response = self.client.post(url, content_type='application/json',
+                                    data=json.dumps(req_values))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['individualsByGuid', 'familiesByGuid'])
+        self.assertIsNone(response_json['familiesByGuid'][FAMILY_GUID])
```
hanars: whenever you test deleting, you should test that trying to delete the same item again fails

seqr/views/apis/family_api_tests.py
ShifaSZ: Okay. I will add response checking for the updated data.

seqr/views/apis/family_api_tests.py
ShifaSZ: The deletion looks successful for the item doesn't exist. So I will not try to delete it again.

ShifaSZ: But I will update the test to cover more lines in the code.

seqr/views/apis/family_api_tests.py
```diff
@@ -5,31 +5,69 @@
...
         _check_login(self, url)
 
-        # send request
-        req_values = {'families': [{'familyGuid': FAMILY_GUID, 'description': 'Test'}]}
-        response = self.client.post(url, content_type='application/json', data=json.dumps(req_values))
+        # send request with a "families" attribute
+        req_values = {
+            'families': [
+                {'familyGuid': FAMILY_GUID, 'description': 'Test description 1'},
+                {PREVIOUS_FAMILY_ID_FIELD: '2', 'description': 'Test description 2'},
```
hanars: So the previous ID field is designed to be used to change the ID, so the usage here is not actually testing a realistic use-case. You should have the previous ID as well as a new ID, and then you should check in the response that the ID for this individual has been successfully updated

ShifaSZ: Okay. Updated.

seqr/views/apis/family_api_tests.py
```diff
@@ -59,48 +98,77 @@ def test_update_family_pedigree_image(self):
...
+        self.assertListEqual(response_json.keys(), ['info', 'errors', 'warnings', 'uploadedFileId'])
+
+        url = reverse(edit_families_handler, args=[PROJECT_GUID])
+
+        response = self.client.post(url, content_type='application/json',
+                data=json.dumps({'uploadedFileId': response_json['uploadedFileId']}))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['familiesByGuid'])
```
hanars: Can you please add some tests that the family data here is correct

ShifaSZ: Response data checking is added.

PR #1233, opened 2020-03-26
Add ut auth apis test

ShifaSZ:
Added three new test cases for the Auth apis.

### DIFF
--- a/seqr/views/apis/auth_api_tests.py
+++ b/seqr/views/apis/auth_api_tests.py
@@ -9,6 +9,9 @@ from django.contrib.auth.models import User
 class AuthAPITest(TestCase):
     fixtures = ['users']
 
+    def setUp(self):
+        User.objects.create_user('test_new_user', 'test_new_user@test.com', 'password123')
+
     def test_login_view(self):
         url = reverse(login_view)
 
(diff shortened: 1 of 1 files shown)


seqr/views/apis/auth_api_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        req_values = {
+            'email': 'test_user@test.com',
+            'password': 'not_a_password'
+        }
+        response = self.client.post(url, content_type='application/json',
+                                    data=json.dumps(req_values))
+        self.assertEqual(response.status_code, 401)
+
+        # send login request with a correct password
+        User.objects.create_user('test_new_user', 'test_new_user@test.com', 'password123')
```
hanars: If you need to do some data set up for a test where you are not actually testing that the set up works, and especially if you need to do the same set up for all the tests in a TestCase the way you do here, you should actually use a `setUp` method on the class which django will automatically run before the test starts. So in this case you would have a setUp method that creates the test user with a known password. I know we aren't always good abotu this elsewhere in the code, but for the new tests I think we should try to follow best practices. Django has an example of how that looks here: https://docs.djangoproject.com/en/1.11/topics/testing/tools/#fixture-loading

ShifaSZ: Updated the test by using setUp().

PR #1234, opened 2020-03-26
Add ut project cat apis

ShifaSZ:
Added two project categories in the test case to increase coverage.

### DIFF
--- a/seqr/views/apis/project_categories_api_tests.py
+++ b/seqr/views/apis/project_categories_api_tests.py
@@ -13,6 +13,6 @@ class ProjectCategoriesAPITest(TestCase):
         _check_login(self, url)
 
         response = self.client.post(url, content_type='application/json', data=json.dumps({
-            'categories': []
+            'categories': ['PC000002_categry_with_unicde', 'PC000003_categry_with_unicde']
         }))
         self.assertEqual(response.status_code, 200)
--- a/seqr/views/apis/project_categories_api_tests.py
+++ b/seqr/views/apis/project_categories_api_tests.py
@@ -4,11 +4,11 @@ from django.urls.base import reverse
 from seqr.views.apis.project_categories_api import update_project_categories_handler
 from seqr.views.utils.test_utils import _check_login
 
+from seqr.models import Project, ProjectCategory
+
 PROJECT_GUID = 'R0001_1kg'
 PROJECT_CAT_GUID2 = 'PC000002_categry_with_unicde'
 PROJECT_CAT_GUID3 = 'PC000003_test_category_name'
-PROJECT_CAT_GUID4 = 'PC000004_new_project_category'
-PROJECT_CAT_GUID5 = 'PC000005_pc000002_categry_with'
 NEW_PROJECT_CAT_NAME = 'New project category'
 
 
@@ -19,6 +19,34 @@ class ProjectCategoriesAPITest(TestCase):
         url = reverse(update_project_categories_handler, args=[PROJECT_GUID])
         _check_login(self, url)
 
+        category_guids = [PROJECT_CAT_GUID2, NEW_PROJECT_CAT_NAME]
+        response = self.client.post(url, content_type='application/json', data=json.dumps({
+            'categories': category_guids
+        }))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['projectsByGuid', 'projectCategoriesByGuid'])
+        self.assertListEqual(response_json['projectsByGuid'].keys(), [PROJECT_GUID])
+        updated_guid_set = set(response_json['projectCategoriesByGuid'].keys())
+
+        project = Project.objects.get(guid=PROJECT_GUID)
+        project_category_guids_in_db = set()
+        # Exam all the project_category guids from the db against the updated guids.
+        for project_category in project.projectcategory_set.all():
+            if project_category.guid == PROJECT_CAT_GUID2:
+                # The old guid must not be updated
+                self.assertNotIn(project_category.guid, updated_guid_set)
+            else:
+                # The new project category must have as name as given one
+                self.assertEqual(project_category.name, NEW_PROJECT_CAT_NAME)
+                self.assertIn(project_category.guid, updated_guid_set)
+            project_category_guids_in_db.add(project_category.guid)
+        for updated_guid in updated_guid_set:
+            if updated_guid not in project_category_guids_in_db:
+                # The updated project category which doesn't exist in the db must have a None value
+                self.assertIsNone(response_json['projectCategoriesByGuid'][updated_guid])
+
         response = self.client.post(url, content_type='application/json', data=json.dumps({
             'categories': []
         }))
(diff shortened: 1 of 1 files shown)


seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,21 @@
...
 
         response = self.client.post(url, content_type='application/json', data=json.dumps({
-            'categories': []
+            'categories': ['PC000002_categry_with_unicde', 'PC000003_categry_with_unicde']
         }))
         self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['projectsByGuid', 'projectCategoriesByGuid'])
+        self.assertListEqual(response_json['projectsByGuid'].keys(), [PROJECT_GUID])
```
hanars: Can you test that you have the expected project categories in the response as well? And can you check that the project json now has the correct project category guids in the response

hanars: Can you add another request that removes one of these categories and check that that works?

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,21 @@
...
     fixtures = ['users', '1kg_project']
 
     def test_project_categories_api(self):
-        url = reverse(update_project_categories_handler, args=['R0001_1kg'])
+        url = reverse(update_project_categories_handler, args=[PROJECT_GUID])
         _check_login(self, url)
 
         response = self.client.post(url, content_type='application/json', data=json.dumps({
-            'categories': []
+            'categories': ['PC000002_categry_with_unicde', 'PC000003_categry_with_unicde']
```
hanars: There is a typo in the second guid, so thats not an existing category, it is creating a new category with that name. Thats actually an important code path to test, but it would probably be better if you used something like `new_category` as the name instead of something that looks like it is an existing guid

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,21 @@
...
 class ProjectCategoriesAPITest(TestCase):
     fixtures = ['users', '1kg_project']
 
     def test_project_categories_api(self):
-        url = reverse(update_project_categories_handler, args=['R0001_1kg'])
+        url = reverse(update_project_categories_handler, args=[PROJECT_GUID])
         _check_login(self, url)
 
         response = self.client.post(url, content_type='application/json', data=json.dumps({
-            'categories': []
```
hanars: This was actually testing that sending an empty list removes all categories. Please leave this test in, check the response to confirm that the project categories are all removed, and then add the test you have here to add them back

seqr/views/apis/project_categories_api_tests.py
ShifaSZ: Good idea. The intention is to create new category indeed. I will update the name.

seqr/views/apis/project_categories_api_tests.py
ShifaSZ: Updated.

seqr/views/apis/project_categories_api_tests.py
hanars: My comment here still stands - can you please test that the `projectCategoryGuids` in the project json are correct, both here and after you have added new categories below?

ShifaSZ: Added the content of "projectCategoryGuids" test for both api call responses.

hanars: Yes, but what I was saying is you need to test the value of the `projectCategoryGuids` field inside the PROJECT item. As in, inside the json in `projectsByGuid`

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,41 @@
...
+
+        response = self.client.post(url, content_type='application/json', data=json.dumps({
+            'categories': ['PC000002_categry_with_unicde', NEW_PROJECT_CAT_NAME]
+        }))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['projectsByGuid', 'projectCategoriesByGuid'])
+        self.assertListEqual(response_json['projectsByGuid'].keys(), [PROJECT_GUID])
+        self.assertListEqual(response_json['projectCategoriesByGuid'].keys(), [PROJECT_CAT_GUID4, PROJECT_CAT_GUID5])
```
hanars: You should never assume you deterministically know what the guid that is generated will be. In the past, tests like that have proven to be incredibly flappy/ unreliable. Instead, you should assert that the expected known guid is in the response, and that there are 2 guids returned. Then you should take the other, new guid and query the database for it and check that the object exist and has the name you expect. And then you can check the project json and make sure it has the expected 2 guids

seqr/views/apis/project_categories_api_tests.py
ShifaSZ: Do you mean I should query the database rather than just checking the responses?

seqr/views/apis/project_categories_api_tests.py
ShifaSZ: Okay. I thought the results are always determined for a test setup.

hanars: But the guid generation could change, we could add another test fixture and the ID would increment. In general, we want tests to be as stable as possible and since we don't care what the guid is we don't want to write fragile tests that rely on it always being the same

ShifaSZ: I updated the test result examination with guids from the model.

seqr/views/apis/project_categories_api_tests.py
ShifaSZ: Added examining the responses against the data in the database after the update_project_categories_handler API call.

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,62 @@
...
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['projectsByGuid', 'projectCategoriesByGuid'])
+        self.assertListEqual(response_json['projectsByGuid'].keys(), [PROJECT_GUID])
+        updated_guid_set = set(response_json['projectCategoriesByGuid'].keys())
+
+        project = Project.objects.get(guid=PROJECT_GUID)
+        project_category_guids_in_db = set()
+        # Exam all the project_category guids from the db against the updated guids.
+        for project_category in project.projectcategory_set.all():
```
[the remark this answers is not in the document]
ShifaSZ: Okay.

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,62 @@
...
+                self.assertNotIn(project_category.guid, updated_guid_set)
+            else:
+                # The new project category must have the same name as the given one
+                self.assertEqual(project_category.name, NEW_PROJECT_CAT_NAME)
+                self.assertIn(project_category.guid, updated_guid_set)
+            project_category_guids_in_db.add(project_category.guid)
+        for updated_guid in updated_guid_set:
+            if updated_guid not in project_category_guids_in_db:
+                # The updated project category which doesn't exist in the db must have a None value
+                self.assertIsNone(response_json['projectCategoriesByGuid'][updated_guid])
```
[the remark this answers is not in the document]
ShifaSZ: Updated

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,62 @@
...
         self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+
+        self.assertListEqual(response_json.keys(), ['projectsByGuid', 'projectCategoriesByGuid'])
+        self.assertListEqual(response_json['projectsByGuid'].keys(), [PROJECT_GUID])
+
+        project = Project.objects.get(guid=PROJECT_GUID)
+        project_category_guids_in_db = set()
+        for project_category in project.projectcategory_set.all():
+            project_category_guids_in_db.add(project_category.guid)
```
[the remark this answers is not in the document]
ShifaSZ: Updated.

seqr/views/apis/project_categories_api_tests.py
```diff
@@ -4,15 +4,62 @@
...
+        self.assertSetEqual(project_category_guids_in_db, set([]))
+        for project_category_guid in response_json['projectCategoriesByGuid'].keys():
```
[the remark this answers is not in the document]
ShifaSZ: All the response test iterations have been removed.

seqr/views/apis/project_categories_api_tests.py
hanars: While you are now doing a great job checking the content of `response_json['projectCategoriesByGuid']`, there are still no tests for `response_json['projectsByGuid'][PROJECT_GUID]['projectCategoryGuids']` as I requested

ShifaSZ: I got your point finally. Sorry for the delay. I kept taking response_json['projectsByGuid'][PROJECT_GUID]['projectCategoryGuids'] as response_json['projectCategoriesByGuid'] by mistake.

PR #1235, opened 2020-03-26
Add ut variant apis

ShifaSZ:
Added a test for the query_variants_handler api to increase the test coverage.

### DIFF
--- a/seqr/fixtures/1kg_project.json
+++ b/seqr/fixtures/1kg_project.json
@@ -83,6 +83,20 @@
         "last_modified_date": "2017-03-12T23:52:55.846Z",
         "name": "test category name",
         "projects": [
+            1, 2
+        ]
+    }
+},
+    {
+    "model": "seqr.projectcategory",
+    "pk": 4,
+    "fields": {
+        "guid": "PC000003_test_category_demo",
+        "created_date": "2017-03-12T23:52:55.846Z",
+        "created_by": null,
+        "last_modified_date": "2017-03-12T23:52:55.846Z",
+        "name": "Demo",
+        "projects": [
             1
         ]
     }
--- a/seqr/views/apis/variant_search_api_tests.py
+++ b/seqr/views/apis/variant_search_api_tests.py
@@ -135,17 +135,6 @@ class VariantSearchAPITest(TestCase):
         results_model = VariantSearchResults.objects.get(search_hash=SEARCH_HASH)
         mock_get_variants.assert_called_with(results_model, sort='xpos', page=1, num_results=100)
 
-        # Test new search with allProjectFamilies
-        url1 = reverse(query_variants_handler, args=[SEARCH_HASH1])
-        response = self.client.post(url1, content_type='application/json', data=json.dumps({
-            'allProjectFamilies': True, 'search': SEARCH
-        }))
-        self.assertEqual(response.status_code, 200)
-        response_json = response.json()
-        self.assertSetEqual(set(response_json.keys()), {
-            'searchedVariants', 'savedVariantsByGuid', 'genesById', 'search', 'variantTagsByGuid', 'variantNotesByGuid',
-            'variantFunctionalDataByGuid', 'familiesByGuid', 'locusListsByGuid'})
-
         # Test pagination
         response = self.client.get('{}?page=3'.format(url))
         self.assertEqual(response.status_code, 200)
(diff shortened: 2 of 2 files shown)


seqr/fixtures/1kg_project.json
```diff
@@ -82,6 +82,20 @@
...
+        "projects": [
+            1, 2
+        ]
+    }
+},
+    {
+    "model": "seqr.projectcategory",
+    "pk": 4,
+    "fields": {
+        "guid": "PC000003_test_category_demo",
```
hanars: I do not think you meant to commit this here

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -134,6 +135,17 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts):
         results_model = VariantSearchResults.objects.get(search_hash=SEARCH_HASH)
         mock_get_variants.assert_called_with(results_model, sort='xpos', page=1, num_results=100)
 
+        # Test new search with allProjectFamilies
+        url1 = reverse(query_variants_handler, args=[SEARCH_HASH1])
```
hanars: Can you make this a second test called something like `test_query_variants_all_projects`? This test is already getting very long and all the other tested  requests in it build off one another (as in, it is actually testing that seqr behaves correctly when you send multiple requests in a row for a single search which is a real user behavior as they do a search and then sort/page/download) while this is testing a completely different search activity

seqr/fixtures/1kg_project.json
ShifaSZ: Hmm..., it is for other test. I will remove it.

ShifaSZ: I found that this row of the data is trying to satisfy the search with "allProjectFamilies" option. Line  101 in variant_search_api.py is as below:
```
            omit_projects = ProjectCategory.objects.get(name='Demo').projects.all()
```
It uses the parameter "name='Demo' " to search the database. Therefore, I created a new row in the ProjectCategory table of 1kg_project with a name "Demo".
I'm not sure what the purpose of the line and why it uses a literal value "Demo".

seqr/views/apis/variant_search_api_tests.py
ShifaSZ: I moved the test for the search with "allProjectFamilies" to a separated test function.

seqr/fixtures/1kg_project.json
hanars: Ah got it

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -209,6 +210,25 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts):
...
+        response = self.client.post(url, content_type='application/json', data=json.dumps({
+            'allProjectFamilies': True, 'search': SEARCH
+        }))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertSetEqual(set(response_json.keys()), {
+            'searchedVariants', 'savedVariantsByGuid', 'genesById', 'search', 'variantTagsByGuid', 'variantNotesByGuid',
+            'variantFunctionalDataByGuid', 'familiesByGuid', 'locusListsByGuid'})
+
+        self.assertListEqual(response_json['searchedVariants'], VARIANTS_WITH_DISCOVERY_TAGS)
```
hanars: In general, if you use a mock its important to test that the mock was called as expected. You should have `mock_get_variants.assert_called_with` somewhere in this test, the way there is in the above test

ShifaSZ: After learning more about how mock work, I added the parameter check with mock_get_variants.assert_called_with.

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -209,6 +210,28 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts):
...
+    # Test new search with allProjectFamilies
+    @mock.patch('seqr.views.apis.variant_search_api.get_es_variants')
+    def test_query_variants_all_project(self, mock_get_variants):
+        url = reverse(query_variants_handler, args=[SEARCH_HASH])
+        _check_login(self, url)
+
+        mock_get_variants.side_effect = _get_es_variants
+
+        response = self.client.post(url, content_type='application/json', data=json.dumps({
+            'allProjectFamilies': True, 'search': SEARCH
```
hanars: I don't see anything in this test that actually tests that the `allProjectFamilies` part of this request is working as intended. If the function completely ignored that parameter, this test would still behave identically. What does that parameter do that is different than if it was absent? Make sure whatever it is, there is something in the test that checks it is doing that. While getting coverage up is important, having tests that cover every line of code but don't actually make any assertions about what those lines do isn't really helpful

ShifaSZ:
It seems that the test for "allProjectFamilies" doesn't make much sense. Therefore, I withdraw this PR.

hanars:
No, it is a crucial edge case that needs to be tested. Please keep the PR open and add a check that the families are being set as expected

hanars:
Talked offline, closing for now

PR #1237, opened 2020-04-01
Added a test case for the elasticsearch_status api.

ShifaSZ:
A test case for the elasticsearch_status api.

### DIFF
--- a/seqr/views/apis/staff_api_tests.py
+++ b/seqr/views/apis/staff_api_tests.py
@@ -7,101 +7,159 @@ from django.urls.base import reverse
 from seqr.views.apis.staff_api import elasticsearch_status
 from seqr.views.utils.test_utils import _check_login
 
-ES_CAT_ALLOCATION=[{
+ES_CAT_ALLOCATION = [{
     u'node': u'node-1',
     u'disk.used': u'67.2gb',
     u'disk.avail': u'188.6gb',
     u'disk.percent': u'26'
-  },
-  {u'node': u'UNASSIGNED',
-   u'disk.used': None,
-   u'disk.avail': None,
-   u'disk.percent': None
-  }]
+},
+    {u'node': u'UNASSIGNED',
+     u'disk.used': None,
+     u'disk.avail': None,
+     u'disk.percent': None
+     }]
 
-MISSED_INDEX="test_index_old"
+EXPECTED_DISK_ALLOCATION = [{
+    u'node': u'node-1',
+    u'diskUsed': u'67.2gb',
+    u'diskAvail': u'188.6gb',
+    u'diskPercent': u'26'
+},
+    {u'node': u'UNASSIGNED',
+     u'diskUsed': None,
+     u'diskAvail': None,
+     u'diskPercent': None
+     }]
 
-ES_CAT_INDICES=[{
+ES_CAT_INDICES = [{
     "index": "test_index",
     "docs.count": "122674997",
     "store.size": "14.9gb",
     "creation.date.string": "2019-11-04T19:33:47.522Z"
-  },
-  {
-    "index": "test_index_second",
-    "docs.count": "672312",
-    "store.size": "233.4mb",
-    "creation.date.string": "2019-10-03T19:53:53.846Z"
-  },
-  {
-    "index": "1kg_test_index",
-    "docs.count": "672312",
-    "store.size": "233.4mb",
-    "creation.date.string": "2019-10-03T19:53:53.846Z"
-  }]
+},
+    {
+        "index": "test_index_alias_1",
+        "docs.count": "672312",
+        "store.size": "233.4mb",
+        "creation.date.string": "2019-10-03T19:53:53.846Z"
+    },
+    {
+        "index": "test_index_alias_2",
+        "docs.count": "672312",
+        "store.size": "233.4mb",
+        "creation.date.string": "2019-10-03T19:53:53.846Z"
+    },
+    {
+        "index": "test_index_no_project",
+        "docs.count": "672312",
+        "store.size": "233.4mb",
+        "creation.date.string": "2019-10-03T19:53:53.846Z"
+    }
+]
 
-ES_CAT_ALIAS=[
-  {
-    "alias": "010203d4c1452eef3f8725bfdee23476",
-    "index": "test_index"
-  },
-  {
-    "alias": "1531a081753f4a135eef25b3496316dd",
-    "index": "1kg_test_index"
-  }]
+ES_CAT_ALIAS = [
+    {
+        "alias": "test_index_second",
+        "index": "test_index_alias_1"
+    },
+    {
+        "alias": "test_index_second",
+        "index": "test_index_alias_2"
+    }]
 
-TEST_INDEX_SRC_PATH="test_index_src_path"
-TEST_INDEX_SECOND_SRC_PATH="test_index_second_src_path"
-TEST_INDEX_1KG_SRC_PATH="test_index_1kg_src_path"
-ES_INDEX_MAPPING={
-  "test_index": {
-    "mappings": {
-      "variant": {
-        "_meta": {
-          "gencodeVersion": "25",
-          "genomeVersion": "38",
-          "sampleType": "WES",
-          "sourceFilePath": TEST_INDEX_SRC_PATH,
-        },
-        "_all": {
-          "enabled": False
+ES_INDEX_MAPPING = {
+    "test_index": {
+        "mappings": {
+            "variant": {
+                "_meta": {
+                    "gencodeVersion": "25",
+                    "genomeVersion": "38",
+                    "sampleType": "WES",
+                    "sourceFilePath": "test_index_file_path",
+                },
+                "_all": {
+                    "enabled": False
+                }
+            }
         }
-      }
+    },
+    "test_index_alias_1": {
+        "mappings": {
+            "variant": {
+                "_meta": {
+                    "gencodeVersion": "25",
+                    "hail_version": "0.2.24",
+                    "genomeVersion": "37",
+                    "sampleType": "WGS",
+                    "sourceFilePath": "test_index_alias_1_path",
+                },
+                "_all": {
+                    "enabled": False
+                },
+            }
+        }
+    },
+    "test_index_alias_2": {
+        "mappings": {
+            "variant": {
+                "_meta": {
+                    "gencodeVersion": "19",
+                    "genomeVersion": "37",
+                    "sampleType": "WES",
+                    "datasetType": "VARIANTS",
+                    "sourceFilePath": "test_index_alias_2_path"
+                },
+                "_all": {
+                    "enabled": False
+                },
+            }
+        }
+    },
+    "test_index_no_project": {
+        "mappings": {
+            "variant": {
+                "_meta": {
+                    "gencodeVersion": "19",
+                    "genomeVersion": "37",
+                    "sampleType": "WGS",
+                    "datasetType": "VARIANTS",
+                    "sourceFilePath": "test_index_no_project_path"
+                },
+                "_all": {
+                    "enabled": False
+                },
+            }
+        }
     }
-  },
-  "test_index_second": {
-    "mappings": {
-      "variant": {
-        "_meta": {
-          "gencodeVersion": "25",
-          "hail_version": "0.2.24",
-          "genomeVersion": "38",
-         "sampleType": "WGS",
-          "sourceFilePath": TEST_INDEX_SECOND_SRC_PATH,
-        },
-        "_all": {
-          "enabled": False
-        },
-      }
-    }
-  },
-  "1kg_test_index": {
-    "mappings": {
-      "variant": {
-        "_meta": {
-          "gencodeVersion": "19",
-          "genomeVersion": "37",
-          "sampleType": "WES",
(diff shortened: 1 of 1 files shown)


seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+    "creation.date.string": "2019-10-03T19:53:53.846Z"
+  },
+  {
+    "index": "test_index_old",
+    "docs.count": "672312",
+    "store.size": "233.4mb",
+    "creation.date.string": "2019-10-03T19:53:53.846Z"
+  },
+  {
+    "index": "1kg.vcf.gz",
```
hanars, line 36: this is a vcf name, never an index name. Please do not use it in the test data as that can be confusing

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+TEST_INDEX_OLD_SRC_PATH="test_index_old_src_path"
+TEST_INDEX_1KG_SRC_PATH="test_index_1kg_src_path"
+ES_INDEX_MAPPING={
+  "test_index": {
+    "mappings": {
+      "variant": {
+        "_meta": {
+          "gencodeVersion": "25",
+          "genomeVersion": "38",
+          "sampleType_1": "WES",
```
hanars, line 63: seqr doesn't care about the `_1` fields - please take them out of the test data

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+}
+from django.test import TestCase
+from django.urls.base import reverse
+
+from seqr.views.apis.staff_api import elasticsearch_status
+from seqr.views.utils.test_utils import _check_login
+
+class VariantSearchAPITest(TestCase):
```
hanars, line 141: Please use a descriptive test case name - maybe `StaffAPITest`

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+        },
+      }
+    }
+  }
+}
+from django.test import TestCase
+from django.urls.base import reverse
+
+from seqr.views.apis.staff_api import elasticsearch_status
+from seqr.views.utils.test_utils import _check_login
```
hanars, line 139: imports should always be at the top of the file

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
+# -*- coding: utf-8 -*-
+import json
```
hanars, line 2: this is unused

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+from seqr.views.utils.test_utils import _check_login
+
+class VariantSearchAPITest(TestCase):
+    fixtures = ['users', '1kg_project']
+    multi_db = True
+
+    @mock.patch('elasticsearch_dsl.index.Index.get_mapping')
+    @mock.patch('elasticsearch.client.cat.CatClient.aliases')
+    @mock.patch('elasticsearch.client.cat.CatClient.indices')
+    @mock.patch('elasticsearch.client.cat.CatClient.allocation')
```
hanars, line 148: In general, you should never patch 3 functions on the same class separately. Just patch the class once, and then set the return value for each functionon that same mock.

However, in this case you have gone too deeply into the class in your patching. You should really only patch libraries that are imported and used directly in the file being tested, never dive into the libraries themselves. seqr doesn't care how the internals of ibraries work and should not be testing them. I would either patch the helper function `get_es_client` which is what is imported here, or if you wanted you could patch the one library call within that function and patch `elasticsearch.Elasticsearch`. Either way, it would be a single patch for all the cat calls and then you would set the return values for the different cat calls by doing something like 
```
mock_client.cat.allocation.return_value = ES_CAT_ALLOCATION
mock_client.cat.indices.return_value = ES_CAT_INDICES
...
```

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['indices', 'errors', 'diskStats', 'elasticsearchHost'])
+        for index in response_json['indices']:
+            index_id = index['index']
+            mapping = ES_INDEX_MAPPING[index_id]
+            self.assertEqual(index['sourceFilePath'], mapping['mappings']['variant']['_meta']['sourceFilePath'])
+        node_list = []
+        for node in response_json['diskStats']:
+            node_list.append(node['node'])
```
hanars, line 167: This would be much more readable as a list comprehension: `node_list = [node['node'] for node in response_json['diskStats'] ]`

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+        mock_cat_aliases.return_value = ES_CAT_ALIAS
+        mock_get_mapping.return_value = ES_INDEX_MAPPING
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['indices', 'errors', 'diskStats', 'elasticsearchHost'])
+        for index in response_json['indices']:
+            index_id = index['index']
+            mapping = ES_INDEX_MAPPING[index_id]
+            self.assertEqual(index['sourceFilePath'], mapping['mappings']['variant']['_meta']['sourceFilePath'])
```
hanars, line 164: rather than testing that one field of every dictionary has the correct value, you should only have 2 dictionaries (maybe one that has an active project and one that doesn't) and check that all the dictionary values are correct using `assertDictEquals`

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,173 @@
...
+        _check_login(self, url)
+
+        mock_cat_allocation.return_value = ES_CAT_ALLOCATION
+        mock_cat_indices.return_value = ES_CAT_INDICES
+        mock_cat_aliases.return_value = ES_CAT_ALIAS
+        mock_get_mapping.return_value = ES_INDEX_MAPPING
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['indices', 'errors', 'diskStats', 'elasticsearchHost'])
```
hanars, line 160: I would recommend setting up the dummy data so there is an erorr message and test that it is correct

seqr/views/apis/staff_api_tests.py
ShifaSZ: The "1kg.vcf.gz" is from the fixture date 1kg_project.json. If it is improper, I will update the fixture data.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Okay, I will delete them. They are actually from the response of the productional system.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Thanks. Good catch. It was from a copy-paste.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Updated.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Updated

seqr/views/apis/staff_api_tests.py
ShifaSZ: Cool, I will try.

ShifaSZ: Since the function under test doesn't use helper functions, so I have to mock the library classes of methods.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Updated.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Good idea. I didn't know it can check the dict as a whole.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I have changed it to "1kg_test_index". It will be an indice without active project.

seqr/views/apis/staff_api_tests.py
ShifaSZ: It has been found out that the dicts are not the same type. So, it can't use "assertDictEquals". I have added all the meta fields and project item checking.

ShifaSZ: I removed index `test_index_old` from the elasticsearch response to generate an `errors` output. And I found that there is a bug in the api elasticsearch_status() in below lines: (The `p.name` must be `p`)
```
errors = ['{} does not exist and is used by project(s) {}'.format(
        index, ', '.join(['{} ({} samples)'.format(p.name, len(indivs)) for p, indivs in project_individuals.items()])
    ) for index, project_individuals in seqr_index_projects.items() if project_individuals]
```

seqr/views/apis/staff_api_tests.py
ShifaSZ: I removed "test_index_old" index from the mocked elasticsearch response to generate the error message.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I decided not to change 1kg_project.json by removing `1kg.vcf.gz`. It is not required by this test case.

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,142 @@
...
+        for index in response_json['indices']:
+            index_id = index['index']
+            mapping_var_meta = ES_INDEX_MAPPING[index_id]['mappings']['variant']['_meta']
+            for key in mapping_var_meta.keys():
+                self.assertEqual(mapping_var_meta[key], index[key])
+            if index_id == '1kg_test_index':
+                self.assertListEqual(index['projects'], [])
+
+        self.assertIn(MISSED_INDEX, response_json['errors'][0])
```
hanars, line 134: There is a deterministic error message, just use `assertListEqual` that the errors list looks exactly as expected

seqr/views/apis/staff_api_tests.py
hanars: yes but as I said during our convesation, I would like you to use a minimal subset of fields so you can clearly test everything. Having too much unneccessary test data makes test unreadable

seqr/views/apis/staff_api_tests.py
hanars: You should be able to deterministically know exactly what the content of the response dictionary is, and you should therefore be checking the entirety of the content. Iterating and using variables like this make it really hard to tell what the expected data looks like, which makes understanding future changes really difficult. What does it mean for the dicts to not have the same type? You are the one writing the dict it is being compared to, so you can make it the same

seqr/views/apis/staff_api_tests.py
```diff
@@ -0,0 +1,142 @@
...
+        mock_es_client = mock_elasticsearch.return_value
+        mock_es_client.cat.allocation.return_value = ES_CAT_ALLOCATION
+        mock_es_client.cat.indices.return_value = ES_CAT_INDICES
+        mock_es_client.cat.aliases.return_value = ES_CAT_ALIAS
+        mock_get_mapping.return_value = ES_INDEX_MAPPING
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['indices', 'errors', 'diskStats', 'elasticsearchHost'])
+        for index in response_json['indices']:
```
hanars, line 126: You should create objects that contain what you expect the index dictionary to have and test against that.  You don;t really need to test that every index is as expected, I would recommend testing 1 that has projects and 1 that doesn't. So for instance, the tests you would have here are 
```
self.assertEqual(len(response_json['indices']), 3)
self.assertDictEqual(response_json['indices'][0], EXPECTED_TEST_INDEX)
self.assertDictEqual(response_json['indices'][2], EXPECTED_1KG_INDEX)
```

seqr/views/apis/staff_api_tests.py
ShifaSZ: Okay, I will make expected result dicts and compare them again the test results.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I'm not sure whether the full error message checking is good or not for the stability of the test especially when there are minor changes to the API codes. But I will try to do as your suggestion.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Good. I will do as you've suggested.

ShifaSZ: I also reformatted the entire file. So there are manage changes. But you can just review the response data checking part and those expected data.

PR #1238, opened 2020-04-03
Add unit tests for other staff apis.

ShifaSZ:


seqr/fixtures/1kg_project.json
```diff
@@ -334,11 +334,36 @@
...
         "analysis_status": "Q",
+        "success_story": "Published with Gleeson and Reza (PMID 31668703)",
+        "success_story_types": ["C", "D"],
         "internal_analysis_status": null,
         "internal_case_review_notes": "<div><span style=\"text-decoration: underline;\"><span style=\"font-family: 'book antiqua', palatino, serif;\">case review notes for family 12</span></span></div>\n<ul>\n<li><span style=\"font-family: 'book antiqua', palatino, serif;\">note1</span></li>\n<li><span style=\"font-family: 'book antiqua', palatino, serif;\">note 2</span></li>\n<li><span style=\"font-family: 'book antiqua', palatino, serif;\">note 3</span></li>\n</ul>",
         "internal_case_review_summary": "<div><span style=\"font-family: 'courier new', courier, monospace;\"><strong>summary for family 12</strong></span></div>"
     }
 },
+{
+    "model": "seqr.family",
```
hanars, line 345: I general I try not to bloat our test feature with more data if we can repurpose existing data. Why do you need a new family, instead of tweaking one that already exists?

seqr/views/apis/staff_api_tests.py
```diff
@@ -195,3 +202,75 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
         mock_get_mapping.assert_called_with(doc_type='variant,structural_variant')
+
+    def test_mme_details(self):
+        url = reverse(mme_details)
+        _check_login(self, url)
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['metrics', 'genesById', 'submissions'])
```
hanars, line 213: please test some of the content of these. For example, test the metric keys are as expected, the length of the submissions are as expected, the keys in one of the submissions are as expected...

seqr/views/apis/staff_api_tests.py
```diff
@@ -195,3 +202,75 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
+
+    def test_seqr_stats(self):
+        url = reverse(seqr_stats)
+        _check_login(self, url)
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['individualCount', 'familyCount', 'sampleCountByType'])
```
hanars, line 222: As I said above, please test the content

seqr/views/apis/staff_api_tests.py
```diff
@@ -195,3 +202,75 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
+
+    def test_get_projects_for_category(self):
+        url = reverse(get_projects_for_category, args=[PROJECT_CATEGRORY_NAME])
+        _check_login(self, url)
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['projectGuids'])
```
hanars, line 231: As I said above, please test the content

seqr/views/apis/staff_api_tests.py
```diff
@@ -195,3 +202,75 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
+
+    def test_discovery_sheet(self):
+        non_project_url = reverse(discovery_sheet, args=[NON_PROJECT_GUID])
+        _check_login(self, non_project_url)
+
+        response = self.client.get(non_project_url)
+        self.assertEqual(response.status_code, 500)
```
hanars, line 238: 500s are unhandled errors and we don't generally test them. If there is a specific error handled in the code we want to test it would be a 400 error and we would also often want to rest the error message

seqr/views/apis/staff_api_tests.py
```diff
@@ -195,3 +202,75 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
+        self.assertListEqual(response_json.keys(), ['rows', 'errors'])
+        self.assertListEqual(response_json['rows'], [])
+        self.assertListEqual(response_json['errors'], ["No data loaded for project: Empty Project"])
+
+        url = reverse(discovery_sheet, args=[PROJECT_GUID])
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertListEqual(response_json.keys(), ['rows', 'errors'])
+        self.assertListEqual(response_json['errors'], [u'No data loaded for family: 13. Skipping...'])
```
hanars, line 254: Please test the content of the rows

seqr/fixtures/1kg_project.json
ShifaSZ: I was trying not to change the existing data because I don't know whether the other testcases are assuming the values of the data. But for this test, it must be safe to tweak the existing data because it just adds two extra fields. I will update the data.

ShifaSZ: I remember the reason for adding a new family now: I need to trigger an error which is caused by a family without any individual for the discovery_sheet API. The new family has no individual. Do you think it is ok to keep it?

seqr/views/apis/staff_api_tests.py
ShifaSZ: I didn't test the content of the response because I didn't know where they come from. After I did some investigations, I think I can test the content of the metric dictionary, the length and keys of genesById, and the length of submissions. Do you think that is okay?

seqr/views/apis/staff_api_tests.py
ShifaSZ: I could add tests against static values. But it could failed when the fixture data were changed. Or should I query the database to get the expected results? I do the test statically for now.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Add the content of the response_json['projectGuids'] test.

seqr/views/apis/staff_api_tests.py
ShifaSZ: The current code raises an exception without handling it when it receives a bad project guid. So it causes an 500 status response.  Should the code be changed to generate a 400s error at this case? I leave it unchanged for now.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Can I just test the length of the content list of the rows? Because I don't know what these contents from and will they change over time.

hanars: This is a crucial report and it is incredibly important that the content does not change. Please test at least one of the row's content.

ShifaSZ: Added testing one of the rows.

hanars:
The automated test runner says this PR is failing tests: https://travis-ci.org/github/macarthur-lab/seqr/builds/673045563

seqr/fixtures/1kg_project.json
hanars: Sure, but maybe make the family_id "no_individuals" or "empty_family" or something so we remember what its for in the future

seqr/views/apis/staff_api_tests.py
```diff
@@ -161,9 +170,48 @@
...
+     u'thoracic_cavity': u'N', u'non_patient_cell_model': u'N',
+     u't0_copy': u'2017-02-05T06:42:55.397Z', u'extras_pedigree_url': u'/media/ped_1.png',
+     u'family_id': u'1', u'genitourinary_system': u'N', u'coded_phenotype': u'',
+     u'animal_model': u'N', u'non_human_cell_culture_model': u'N', u'expression': u'N',
+     u'gene_name': u'RP11-206L10.5', u'breast': u'N'}
+
 
 class StaffAPITest(TestCase):
-    fixtures = ['users', '1kg_project']
+    fixtures = ['users', '1kg_project', 'reference_data', 'variant_searches']
```
hanars, line 214: The `variant_searches` fixture is actually production fixture data, not usually test data. Is there a reason we need it here?

seqr/views/apis/staff_api_tests.py
hanars: Yes please update it to be a 400, and update the test to test the error message

seqr/fixtures/1kg_project.json
ShifaSZ: Okay, I rename the display name of the family.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Updated

seqr/views/apis/staff_api_tests.py
ShifaSZ: Removed `variant_searches` but kept `reference_data`.

seqr/fixtures/1kg_project.json
hanars: Please rename the family_id field as I requested. There is no reason it needs to be the same as the pk

ShifaSZ: Changed the id to "no_individuals".

PR #1239, opened 2020-04-07
A Seqr management command testcase

ShifaSZ:
The first testcase for testing the normal functionality of the command add_project_tag.

### DIFF
--- a/seqr/management/commands/add_project_tag_tests.py
+++ b/seqr/management/commands/add_project_tag_tests.py
@@ -1,0 +1,27 @@
+from io import StringIO
+from django.core.management import call_command
+from django.test import TestCase
+
+TEST_ARGUMENTS = {
+    "project": "Test Project",
+    "name": "Test tag - Novel gene and phenotype",
+    "order": 22.0,
+    "category": "CMG Discovery Tags",
+    "description": "Gene not previously associated with a Mendelian condition",
+    "color": "#03441E"
+}
+
+
+class AddProjectTagTest(TestCase):
+    fixtures = ['users', '1kg_project']
+
+    def test_normal_command(self):
+        out = StringIO()
+        call_command('add_project_tag', '--project={}'.format(TEST_ARGUMENTS["project"]),
+            '--name={}'.format(TEST_ARGUMENTS["name"]), '--order={}'.format(TEST_ARGUMENTS["order"]),
+            '--category={}'.format(TEST_ARGUMENTS["category"]),
+            '--description={}'.format(TEST_ARGUMENTS["description"]),
+            '--color={}'.format(TEST_ARGUMENTS["color"]), stdout = out)
+
+        expectOutput = out.getvalue()
+        self.assertIn('', expectOutput)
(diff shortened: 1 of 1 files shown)


seqr/management/commands/add_project_tag_tests.py
```diff
@@ -17,11 +21,62 @@ class AddProjectTagTest(TestCase):
...
+        call_command('add_project_tag', '--project={}'.format(TAG_ARGUMENTS["project"]),
+            '--name={}'.format(TAG_ARGUMENTS["name"]), '--order={}'.format(TAG_ARGUMENTS["order"]),
+            '--category={}'.format(TAG_ARGUMENTS["category"]),
+            '--description={}'.format(TAG_ARGUMENTS["description"]),
+            '--color={}'.format(TAG_ARGUMENTS["color"]), stdout = out)
+
+        self.assertIn('', out.getvalue())
+
+        project = Project.objects.get(Q(name = TAG_ARGUMENTS["project"]))
+        variantTagType = VariantTagType.objects.get(name__iexact=TAG_ARGUMENTS["name"], project=project)
```
hanars, line 33: theres no reason for `iexact` here, you can just do `name=`

seqr/management/commands/add_project_tag_tests.py
```diff
@@ -17,11 +21,62 @@ class AddProjectTagTest(TestCase):
...
-
-        expectOutput = out.getvalue()
-        self.assertIn('', expectOutput)
+        call_command('add_project_tag', '--project={}'.format(TAG_ARGUMENTS["project"]),
+            '--name={}'.format(TAG_ARGUMENTS["name"]), '--order={}'.format(TAG_ARGUMENTS["order"]),
+            '--category={}'.format(TAG_ARGUMENTS["category"]),
+            '--description={}'.format(TAG_ARGUMENTS["description"]),
+            '--color={}'.format(TAG_ARGUMENTS["color"]), stdout = out)
+
+        self.assertIn('', out.getvalue())
```
hanars, line 30: All this test is that some string contains an empty string, which is always true for all strings in python. If theres no output you could just skip testing the output altogether, or if you want to check that there is no output you should use `asserEqual`

seqr/management/commands/add_project_tag_tests.py
```diff
@@ -17,11 +21,62 @@ class AddProjectTagTest(TestCase):
...
+
+    def test_missing_required_args(self):
+        out = StringIO()
+        error = ''
+        try:
+            call_command('add_project_tag',
+                stdout = out)
+        except CommandError as err:
+            error = err.message
+        self.assertEqual(error, 'Error: argument --project is required')
```
hanars, line 52: If you want to unit test exceptions you should use `assertRaises`: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

seqr/management/commands/add_project_tag_tests.py
ShifaSZ: Okay. It was from a copy-paste. Remove it.

seqr/management/commands/add_project_tag_tests.py
ShifaSZ: Oh, really?! Change it to assertEqual.

seqr/management/commands/add_project_tag_tests.py
ShifaSZ: Hmm... I remember it now. It is easy to forget if never use it. Changed to use assertRaises.

PR #1247, opened 2020-04-14
Added unit test for check_bam_cram_paths command.

ShifaSZ:
Two tests have been added. One is normal test while another is for exception test.
A change: BytesIO is used instead of StringIO because stdout is written with a str rather than a unicode (if a unicode is provided then it is casted to str). 
The check_bam_cram_paths implementation must replace print() with self.stdout.write() according to https://docs.djangoproject.com/en/3.0/howto/custom-management-commands/

### DIFF
--- a/seqr/management/commands/check_bam_cram_paths_tests.py
+++ b/seqr/management/commands/check_bam_cram_paths_tests.py
@@ -1,0 +1,27 @@
+import mock
+
+from io import BytesIO
+from django.core.management import call_command
+from django.test import TestCase
+EXPECTED_EXCEPTION_MSG = 'Error at /readviz/NA19675.cram (Individual: NA19675_1): Error accessing "/readviz/NA19675.cram" \n---- DONE ----\nChecked 1 samples\n1 failed samples: NA19675_1\n'
+EXPECTED_NORMAL_MSG = 'Error at /readviz/NA19675.cram (Individual: NA19675_1): Error accessing "/readviz/NA19675.cram" \n---- DONE ----\nChecked 1 samples\n1 failed samples: NA19675_1\n'
+
+
+class AddProjectTagTest(TestCase):
+    fixtures = ['users', '1kg_project']
+
+    @mock.patch('seqr.views.utils.dataset_utils.validate_alignment_dataset_path')
+    def test_normal_command(self, mock_validate_path):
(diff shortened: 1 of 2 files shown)


seqr/management/commands/check_bam_cram_paths_tests.py
```diff
@@ -0,0 +1,27 @@
...
+class CheckBamCramPathsTest(TestCase):
+    fixtures = ['users', '1kg_project']
+
+    @mock.patch('seqr.views.utils.dataset_utils.validate_alignment_dataset_path')
+    def test_normal_command(self, mock_validate_path):
+        mock_validate_path.return_value = ""
+        out = BytesIO()
+        call_command('check_bam_cram_paths', u'1kg project n\u00e5me with uni\u00e7\u00f8de', stdout=out)
+
+        self.assertEqual(EXPECTED_NORMAL_MSG, out.getvalue())
```
hanars, line 19: This would be more readable if you put the expected message directly down here. In general, defining global constants to check against is more readable when either a) you are checking against the same value multiple times or sometimes if b) its a really large object. But  in all other cases, its better to directly put the expected value inside the test

seqr/management/commands/check_bam_cram_paths_tests.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @mock.patch('seqr.views.utils.dataset_utils.validate_alignment_dataset_path')
+    def test_exception_command(self, mock_validate_path):
+        mock_validate_path.side_effect = Exception('Error accessing "/readviz/NA19675.cram"')
+        out = BytesIO()
+        call_command('check_bam_cram_paths', u'1kg project n\u00e5me with uni\u00e7\u00f8de', stdout=out)
+
+        self.assertEqual(EXPECTED_EXCEPTION_MSG, out.getvalue())
```
hanars, line 27: Same as above

seqr/management/commands/check_bam_cram_paths_tests.py
ShifaSZ: Updated

seqr/management/commands/check_bam_cram_paths_tests.py
ShifaSZ: Updated

PR #1251, opened 2020-04-17
Unit tests for Seqr management command reload_saved_variant_json

ShifaSZ:
There are two tests for two different command arguments and one for exception test.
This PR also fixed some of unicode issues for the loggings in the command.

### DIFF
--- a/seqr/management/commands/reload_saved_variant_json_tests.py
+++ b/seqr/management/commands/reload_saved_variant_json_tests.py
@@ -3,30 +3,81 @@ import mock
 
 from django.core.management import call_command
 from django.test import TestCase
+from django.db.models.query_utils import Q
+from seqr.models import Project
 
-PROJECT_NAME = '1kg project n\u00e5me with uni\u00e7\u00f8de'
+PROJECT_NAME = u'1kg project n\u00e5me with uni\u00e7\u00f8de'
 PROJECT_GUID = 'R0001_1kg'
-PROJECT_GUID2 = 'R0002_empty'
+PROJECT_GUID2 = 'R0003_test'
 SAVED_VARIANT_GUIDS = ['SV0000001_2103343353_r0390_100', 'SV0000002_1248367227_r0390_100']
-
+FAMILY_ID = '1'
 
 class ReloadSavedVariantJsonTest(TestCase):
     fixtures = ['users', '1kg_project']
 
     @mock.patch('logging.getLogger')
     @mock.patch('seqr.views.utils.variant_utils.update_project_saved_variant_json')
-    def test_normal_command(self, mock_update_json, mock_getLogger):
-        logger = mock_getLogger.return_value
+    def test_with_param_command(self, mock_update_json, mock_getLogger):
         mock_update_json.return_value = SAVED_VARIANT_GUIDS
+        logger = mock_getLogger.return_value
+
+        # Test with specific projects and a family id.
+        call_command('reload_saved_variant_json',
+                     PROJECT_NAME, PROJECT_GUID2,
+                     '--family-id=1'.format(FAMILY_ID))
+
+        projects_to_process = [PROJECT_NAME, PROJECT_GUID2]
+        projects = Project.objects.filter(Q(name__in=projects_to_process) | Q(guid__in = projects_to_process))
+        calls = [mock.call(project, family_id=FAMILY_ID) for project in projects]
+        mock_update_json.assert_has_calls(calls, any_order = True)
+
+        # Test for all projects and no specific family ids
+        call_command('reload_saved_variant_json')
+
+        projects = Project.objects.all()
+        calls = [mock.call(project, family_id=None) for project in projects]
+        mock_update_json.assert_has_calls(calls, any_order = True)
+
+        # Test with an exception.
+        mock_update_json.side_effect = Exception("Database error.")
         call_command('reload_saved_variant_json',
                      PROJECT_GUID,
-                     '--family-id=1')
+                     '--family-id=1'.format(FAMILY_ID))
 
-        calls = [
+        projects_to_process = [PROJECT_GUID]
+        projects = Project.objects.filter(Q(name__in=projects_to_process) | Q(guid__in = projects_to_process))
+        calls = [mock.call(project, family_id=FAMILY_ID) for project in projects]
+        mock_update_json.assert_has_calls(calls, any_order = True)
+
+        loggerInfoCalls = [
             mock.call(u'Project: 1kg project n\xe5me with uni\xe7\xf8de'),
             mock.call(u'Updated 2 variants for project 1kg project n\xe5me with uni\xe7\xf8de'),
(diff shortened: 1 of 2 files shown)


seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,83 @@
...
+PROJECT_GUID2 = 'R0003_test'
+SAVED_VARIANT_GUIDS = ['SV0000001_2103343353_r0390_100', 'SV0000002_1248367227_r0390_100']
+FAMILY_ID = '1'
+
+class ReloadSavedVariantJsonTest(TestCase):
+    fixtures = ['users', '1kg_project']
+
+    @mock.patch('logging.getLogger')
+    @mock.patch('seqr.views.utils.variant_utils.update_project_saved_variant_json')
+    def test_with_param_command(self, mock_update_json, mock_getLogger):
```
hanars, line 20: you should always use snake_case for variable name, not mix it with camelCase. So it should be `mock_get_logger`

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,83 @@
...
+        mock_update_json.return_value = SAVED_VARIANT_GUIDS
+        logger = mock_getLogger.return_value
```
hanars, line 22: I would name this `mock_logger` to be clearer

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # Test with specific projects and a family id.
+        call_command('reload_saved_variant_json',
+                     PROJECT_NAME, PROJECT_GUID2,
+                     '--family-id=1'.format(FAMILY_ID))
```
hanars, line 27: you are calling `format` but you hardcoded the family id. I think the hardcoded param is fine, so just take out the call to `format`

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,83 @@
...
+
+        projects_to_process = [PROJECT_NAME, PROJECT_GUID2]
+        projects = Project.objects.filter(Q(name__in=projects_to_process) | Q(guid__in = projects_to_process))
+        calls = [mock.call(project, family_id=FAMILY_ID) for project in projects]
```
hanars, line 31: You should never be updating multiple projects with a family id provided, even if that is technically possible it is not an intended use case. Please do this as 2 separate tests: one that multiple whole projects are updated, and one that one family in one project is updated

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,83 @@
...
+        call_command('reload_saved_variant_json',
+                     PROJECT_GUID,
+                     '--family-id=1'.format(FAMILY_ID))
+
+        projects_to_process = [PROJECT_GUID]
+        projects = Project.objects.filter(Q(name__in=projects_to_process) | Q(guid__in = projects_to_process))
+        calls = [mock.call(project, family_id=FAMILY_ID) for project in projects]
+        mock_update_json.assert_has_calls(calls, any_order = True)
+
+        loggerInfoCalls = [
```
hanars, line 52: please test the logger calls with the function call they are associated with, instead fo all at the end. You can reset mocks so between test calls of the command you should reset the logger mock

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Good catch.

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Updated

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Good catch. I will use `format` because I need to re-use the value of `FAMILY_ID`.

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Okay. Update the first test to be a single project and a family id test. The second test will be multiple projects with no family id provided.

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Thanks. mock_reset() is used now.

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,92 @@
...
+        mock_update_json.return_value = SAVED_VARIANT_GUIDS
+        mock_logger = mock_get_logger.return_value
+
+        # Test with a specific project and a family id.
+        call_command('reload_saved_variant_json',
+                     PROJECT_NAME,
+                     '--family-id={}'.format(FAMILY_ID))
+
+        projects_to_process = [PROJECT_NAME]
+        projects = Project.objects.filter(Q(name__in=projects_to_process) | Q(guid__in = projects_to_process))
```
hanars, line 30: Given that this is querying fixture data, please write a test query that directly gets what you expect (i.e. filtering for the single expected guid)

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,92 @@
...
+        mock_logger.reset_mock()
+
+        # Test for all projects and no specific family ids
+        call_command('reload_saved_variant_json')
+
+        projects = Project.objects.all()
+        calls = [mock.call(project, family_id=None) for project in projects]
+        mock_update_json.assert_has_calls(calls, any_order = True)
+
+        loggerInfoCalls = [
```
hanars, line 52: please use snake_case for variables

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Changed to query for a single project.

seqr/management/commands/reload_saved_variant_json_tests.py
ShifaSZ: Updated the naming.

seqr/management/commands/reload_saved_variant_json_tests.py
```diff
@@ -0,0 +1,85 @@
+#-*- coding: utf-8 -*-
+import mock
+
+from django.core.management import call_command
+from django.test import TestCase
+from django.db.models.query_utils import Q
```
hanars, line 6: This is not used

ShifaSZ: Sorry. Fixed, will check later.

PR #1252, opened 2020-04-20
Seqr management command lift_project_to_hg38 unit test

ShifaSZ:
Added a test for the basic features of lift_project_to_hg38. Exception tests are not included.
The lift_project_to_hg38.py was modified to fixed a unicode issue and a argument mismatch issue of the _update_variant_samples() method.

### DIFF
--- a/seqr/management/commands/lift_project_to_hg38.py
+++ b/seqr/management/commands/lift_project_to_hg38.py
@@ -39,7 +39,6 @@ class Command(BaseCommand):
         sample_ids, index_metadata = get_elasticsearch_index_samples(elasticsearch_index)
         validate_index_metadata(index_metadata, project, elasticsearch_index, genome_version=GENOME_VERSION_GRCh38)
         sample_type = index_metadata['sampleType']
-        dataset_path = index_metadata['sourceFilePath']
 
         matched_sample_id_to_sample_record = match_sample_ids_to_sample_records(
             project=project,
--- a/seqr/management/commands/lift_project_to_hg38.py
+++ b/seqr/management/commands/lift_project_to_hg38.py
@@ -32,7 +32,7 @@ class Command(BaseCommand):
         elasticsearch_index = options['es_index']
 
         project = Project.objects.get(Q(name=project_arg) | Q(guid=project_arg))
-        logger.info('Updating project genome version for {}'.format(project.name))
+        logger.info(u'Updating project genome version for {}'.format(project.name))
 
         # Validate the provided index
         logger.info('Validating es index {}'.format(elasticsearch_index))
@@ -82,7 +82,7 @@ class Command(BaseCommand):
                 ))
 
         # Lift-over saved variants
-        _update_variant_samples(matched_sample_id_to_sample_record, elasticsearch_index, dataset_path)
+        _update_variant_samples(matched_sample_id_to_sample_record, elasticsearch_index)
         saved_variants = get_json_for_saved_variants(saved_variant_models_by_guid.values(), add_details=True)
         saved_variants_to_lift = [v for v in saved_variants if v['genomeVersion'] != GENOME_VERSION_GRCh38]
 
--- a/seqr/management/commands/lift_project_to_hg38_tests.py
+++ b/seqr/management/commands/lift_project_to_hg38_tests.py
@@ -3,9 +3,9 @@ import mock
 import __builtin__
 from copy import deepcopy
 from django.core.management.base import CommandError
-from seqr.models import Family, Individual, SavedVariant
+from seqr.models import Family
 from pyliftover.liftover import LiftOver
-from seqr.views.utils.test_utils import VARIANTS
+from seqr.views.utils.test_utils import VARIANTS, SINGLE_VARIANT
 
 from django.core.management import call_command
 from django.test import TestCase
@@ -24,23 +24,33 @@ SAMPLE_IDS = ["NA19679", "NA19675_1", "NA19678", "HG00731", "HG00732", "HG00733"
 
 liftover_to_38 = LiftOver('hg19', 'hg38')
 
+LIFT_MAP = {
+    21003343353L: [('chr21', 3343400)],
+    1248367227L: [('chr1', 248203925)],
+    1001562437L: [('chr1',   1627057)],
+    1001560662L: [('chr1',  46394160)],
+}
 
-# To create a none return value for the variant at pos 1562437
+
 def mock_convert_coordinate(chrom, pos):
-    if pos == 1562437:
-        return
-    return liftover_to_38.convert_coordinate(chrom, pos)
+    pos = int(chrom.replace('chr', ''))*int(1e9) + pos
+    if pos in LIFT_MAP.keys():
+        return(LIFT_MAP[pos])
+
 
 @mock.patch('seqr.management.commands.lift_project_to_hg38.logger')
+@mock.patch('seqr.management.commands.lift_project_to_hg38.get_elasticsearch_index_samples')
 class LiftProjectToHg38Test(TestCase):
     fixtures = ['users', '1kg_project']
 
     @mock.patch.object(__builtin__, 'raw_input')
-    @mock.patch('seqr.management.commands.lift_project_to_hg38.get_elasticsearch_index_samples')
     @mock.patch('seqr.management.commands.lift_project_to_hg38.get_es_variants_for_variant_tuples')
-    def test_command(self, mock_get_es_variants, mock_get_es_samples, mock_input, mock_logger):
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.LiftOver')
+    def test_command(self, mock_liftover, mock_get_es_variants, mock_input, mock_get_es_samples, mock_logger):
         mock_get_es_samples.return_value = SAMPLE_IDS, INDEX_METADATA
         mock_get_es_variants.return_value = VARIANTS
+        mock_liftover_to_38 = mock_liftover.return_value
+        mock_liftover_to_38.convert_coordinate.side_effect = mock_convert_coordinate
         mock_input.return_value = 'y'
         call_command('lift_project_to_hg38', u'--project={}'.format(PROJECT_NAME),
                      '--es-index={}'.format(ELASTICSEARCH_INDEX))
@@ -52,16 +62,28 @@ class LiftProjectToHg38Test(TestCase):
             mock.call('Successfully lifted over 3 variants'),
             mock.call('Successfully updated 3 variants'),
             mock.call('---Done---'),
-            mock.call('Succesfully lifted over 3 variants. Skipped 4 failed variants. Family data not updated for 0 variants')
+            mock.call('Succesfully lifted over 3 variants. Skipped 3 failed variants. Family data not updated for 0 variants')
         ]
         mock_logger.info.assert_has_calls(calls)
 
         mock_get_es_samples.assert_called_with(ELASTICSEARCH_INDEX)
-        mock_get_es_variants.assert_called_with(mock.ANY, [(1001627057, u'G', u'C'), (1046394160, u'G', u'A'), (1248203925, u'TC', u'T'), (2003339582, u'GAGA', u'G')])
+
+        calls = [
+            mock.call('chr21', 3343353),
+            mock.call('chr1', 1562437),
+            mock.call('chr1', 248367227),
+            mock.call('chr1', 1560662),
+        ]
+        mock_liftover_to_38.convert_coordinate.assert_has_calls(calls, any_order = True)
+        
+        families = {family for family in Family.objects.filter(pk__in = [1, 2])}
+        self.assertSetEqual(families, mock_get_es_variants.call_args.args[0])
+        self.assertSetEqual(set([(1001627057, u'G', u'C'), (21003343400, u'GAGA', u'G'), (1248203925, u'TC', u'T'),
+                                 (1046394160, u'G', u'A')]), set(mock_get_es_variants.call_args.args[1]))
 
         # Test discontinue on lifted variants
         mock_logger.reset_mock()
-        mock_input.return_value = 'n'
+        mock_input.side_effect = 'n'
         with self.assertRaises(CommandError) as ce:
             call_command('lift_project_to_hg38', u'--project={}'.format(PROJECT_NAME),
                     '--es-index={}'.format(ELASTICSEARCH_INDEX))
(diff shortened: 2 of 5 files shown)


seqr/management/commands/lift_project_to_hg38.py
```diff
@@ -82,7 +82,7 @@ def handle(self, *args, **options):
                 ))
 
         # Lift-over saved variants
-        _update_variant_samples(matched_sample_id_to_sample_record, elasticsearch_index, dataset_path)
+        _update_variant_samples(matched_sample_id_to_sample_record, elasticsearch_index)
```
hanars, line 85: By removing this you actually make this variable  unused complete, which gets flagged by our style checker. Could you remove the variable declaration above on line 42?

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,357 @@
...
+  'hgvsp': 'ENSP00000413641.1:p.Leu48del',
+  'majorConsequence': 'frameshift_variant',
+  'consequenceTerms': [
+    'frameshift_variant',
+    'inframe_deletion',
+    'NMD_transcript_variant'
+  ]
+}
+
+ES_VARIANTS = [
```
hanars, line 110: Rather than copy/ pasting all these constants, there is a file called test_utils.py where a bunch of shared data is stored (things like expected family fields). Could you move these there instead?

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,357 @@
...
+        'variantId': '2-003339582-GAGA-G',
+        'xpos': 2003339582,
+        'end': None,
+        'svType': None,
+        'numExon': None,
+        '_sort': [2003339582],
+    },
+]
+
+SINGLE_ES_VARIANTS = [
```
hanars, line 220: Wouldn't "single" imply only one?

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,357 @@
...
+
+class LiftProjectToHg38Test(TestCase):
+    fixtures = ['users', '1kg_project']
+
+    @mock.patch.object(__builtin__, 'raw_input')
+    @mock.patch('logging.getLogger')
+    @mock.patch('seqr.views.utils.dataset_utils.get_elasticsearch_index_samples')
+    @mock.patch('seqr.utils.elasticsearch.utils.get_es_variants_for_variant_tuples')
+    @mock.patch('seqr.utils.elasticsearch.utils.get_single_es_variant')
+    def test_command(self, mock_single_es_variants, mock_get_es_variants, mock_get_es_samples, mock_getLogger, mock_input):
```
hanars, line 339: Can you please add testing for the error cases?

seqr/management/commands/lift_project_to_hg38.py
ShifaSZ: Okay

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: There are two calls to the mocked function `get_single_es_variant()` and this array contains the return_values these two calls.

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: Can I re-use the constants defined in seqr/utils/elasticsearch/es_utils_tests.py? If yes, then I don't have to do any copying.

hanars: If this was a direct copy-paste and they are the same there shouldn't be an issue of moving the data to test)utils and then importing the same dictionary in the different tests

seqr/management/commands/lift_project_to_hg38_tests.py
hanars: I see. The name here is a bit confusing - maybe use `SINGLE_ES_VARIANTS_RETURN_VALUES`

ShifaSZ: You are right. I have re-defined the constant by defining the elements of the array and allowing the test code to put them into arrays. There is less name confusing and data redundant.

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: Error cases have been added. I inserted new database rows or updated existing rows to generate errors. I'm not sure whether it is the right way to do test for the error cases.

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,269 @@
...
+        family = Family.objects.get(pk=3)
+        saved_variant = SavedVariant.objects.create(
+            xpos_start = 21005543353,
+            xpos_end = 21005543356,
+            ref = "GAGA",
+            alt = "G",
+            variant_id = "21-5543353-GAGA-G",
+            family = family
+        )
+        saved_variant.save()
```
hanars, line 137: Never create test data via the django API, except under extenuating circumstances. Instead, change the sample IDs that are returned from the mock

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,269 @@
...
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.LiftOver')
+    def test_command_other_exceptions(self, mock_liftover, mock_single_es_variants, mock_get_es_variants, mock_get_es_samples, mock_logger, mock_input):
+        mock_get_es_samples.return_value = SAMPLE_IDS, INDEX_METADATA
+
+        # Update a SavedVariant record to be a lifted variant
+        saved_variant = SavedVariant.objects.get(
+            variant_id = "21-3343353-GAGA-G",
+        )
+        saved_variant.saved_variant_json['genomeVersion'] = GENOME_VERSION_GRCh38
+        saved_variant.save()
```
hanars, line 167: Again, please create fixture data. If you find that adding fixture data causes unrelated tests to fail, you might want to try using project 3 instead of project 1, as it is used in far fewer other tests

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,269 @@
...
+
+        calls = [
+            mock.call(u'Updating project genome version for {}'.format(PROJECT_NAME)),
+            mock.call('Validating es index test_index'),
+            mock.call('Lifting over 3 variants (skipping 1 that are already lifted)')
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
+        project = Project.objects.get(name=PROJECT_NAME)
+        expected_families = {sv.family for sv in SavedVariant.objects.filter(family__project=project)}
```
hanars, line 228: Please do this as a family query filtering on a hardcoded set of guids. Test data is not dynamic, it should fail when things change. You should always hard code what you are expecting

seqr/views/utils/test_utils.py
```diff
@@ -117,3 +117,204 @@ def _proxy_request_stub(*args, **kwargs):
     'constraints', 'diseaseDesc', 'functionDesc', 'notes', 'omimPhenotypes', 'mimNumber', 'mgiMarkerId', 'geneNames',
 }
 GENE_DETAIL_FIELDS.update(GENE_FIELDS)
```
hanars, line 119: You also need to update the test you copied these from to import these

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: The data have been moved into the file seqr/views/utils/test_utils.py. Is the right file to put the data?

hanars: yes, but then you should remove it from the file it was originally in

ShifaSZ: Removed from the original file.

ShifaSZ: The original file seqr/utils/elasticsearch/es_utils_tests.py was updated.

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: Removed creating test data. Added a mocked SavedVariant.objects.filter() return_value.

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: It requires too much work to create new saved_variants for another project. Therefore, I keep using project 1. I changed the save_variant to have 'genomeVersion' 38 directly in the fixture data. It seems work.

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: `expected_families` is a set of SavedVariant model objects which normally are returned from queries. I think it is hard to provide statical hard coded data for these objects. Can I use `mock.ANY` to replace this expected called parameter?

seqr/views/utils/test_utils.py
ShifaSZ: I updated seqr/utils/elasticsearch/es_utils_tests.py.

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,249 @@
...
+    def test_command(self, mock_get_es_variants, mock_get_es_samples, mock_logger, mock_input):
+        mock_get_es_samples.return_value = SAMPLE_IDS, INDEX_METADATA
+        es_variants = deepcopy([ES_VARIANT1, ES_VARIANT2])
+        es_variants[0]['xpos'] = 1046394160
+        es_variants[0]['ref'] = 'G'
+        es_variants[0]['alt'] = 'A'
+        es_variants[0]['familyGuids'] = ['F000001_1']
+        es_variants[1]['xpos'] = 1248203925
+        es_variants[1]['ref'] = 'TC'
+        es_variants[1]['alt'] = 'T'
```
hanars, line 51: Why do you need to change the test fixture data? If its really necessary, do it at the top of the file and crate a constant, but ideally you should make the text work with the fixture or vice versa

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,249 @@
...
+            mock.call(u'Updating project genome version for {}'.format(PROJECT_NAME)),
+            mock.call('Validating es index test_index')
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
+        mock_get_es_samples.assert_called_with(ELASTICSEARCH_INDEX)
+
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.logger')
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.get_elasticsearch_index_samples')
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.Individual')
```
hanars, line 95: Never, ever, mock out a django model class. Change the test to work with the data in the test database and instead use a different return value for `mock_get_es_samples`

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,249 @@
...
+            mock.call(u'Updating project genome version for {}'.format(PROJECT_NAME)),
+            mock.call('Validating es index test_index'),
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
+        mock_get_es_samples.assert_called_with(ELASTICSEARCH_INDEX)
+
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.logger')
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.get_elasticsearch_index_samples')
+    @mock.patch('seqr.management.commands.lift_project_to_hg38.SavedVariant')
```
hanars, line 119: See above, you must never mock out the models

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -160,7 +160,7 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts):
             export_content[1],
             ['21', '3343400', 'GAGA', 'G', 'WASH7P', 'missense_variant', '', '', '', '', '', '', '', '', '', '', '', '',
              '', 'ENST00000623083.3:c.1075G>A', 'ENSP00000485442.1:p.Gly359Ser', '', '', '', '1',
-             'Tier 1 - Novel gene and phenotype (None)|Review (None)', '', '2', '', '', 'NA19675:1:46.0:0.702127659574', 'NA19679:0:99.0:0.0'])
```
hanars, line 163: This should not have changed, its an important thing to test. You need to change how you modified the fixture data so this is still correctly getting a tagged variant

seqr/management/commands/lift_project_to_hg38_tests.py
ShifaSZ: The data are the mock searching results from Elasticsearch. Therefore, they can't be in the fixture data.
I can put the updating to the top of the file as seqr/utils/elasticsearch/es_utils_tests.py does.

hanars: Sorry, fixture here just means "fixed test data" not the fixture.json file. I meant why doesn't `ES_VARIANT1` work as is?

seqr/views/apis/variant_search_api_tests.py
ShifaSZ: Rolled back.

seqr/management/commands/lift_project_to_hg38_tests.py
```diff
@@ -0,0 +1,235 @@
...
+    1248367227L: [('chr1', 248203925)],
+    1001562437L: [('chr1',   1627057)],
+    1001560662L: [('chr1',  46394160)],
+}
+
+
+def mock_convert_coordinate(chrom, pos):
+    pos = int(chrom.replace('chr', ''))*int(1e9) + pos
+    if pos in LIFT_MAP.keys():
+        return(LIFT_MAP[pos])
```
hanars, line 38: This works for the given test cases, but you want this function to always return something, since in general in python its bad to have a function that in some cases has a return value and in some cases doesn't. You could just do `return LIFT_MAP.get(pos)`

ShifaSZ: It's fair. I assumed it returns None when the key doesn't exist. I could be wrong. Updated.

hanars: Yeah, in python if you reach the end of a function without a specific return it just doesn't return anything, and you usually get a runtime error

PR #1253, opened 2020-04-20
fix failure to use custom configured ES port

ssadedin:
I found that even if I set ELASTICSEARCH_PORT in the django `settings.py`, SeqR would still try to query on port 9200. I tracked it down to the ElasticSearch constructor, which only uses the hostname and doesn't pass the port through. This change corrected the problem for me.


ssadedin:
will resubmit after squashing the commits that caused regressions

PR #1254, opened 2020-04-20
Fix custom elastic-search port not respected

ssadedin:
I found that even if I set ELASTICSEARCH_PORT in the django settings.py, SeqR would still try to query on port 9200. I tracked it down to the ElasticSearch constructor, which only uses the hostname and doesn't pass the port through. This change corrected the problem for me.


hanars:
Hey Simon, thanks for catching and fixing this! It looks like there are some unrelated commits somehow on this branch? Do you mind getting it cleaned up a bit before I merge it?

ssadedin:
Apologies, @hanars  - didn't realise those commits got included. Will resubmit - thanks!

ssadedin:
@hanars - see https://github.com/macarthur-lab/seqr/pull/1255 - this time I based it on the dev branch so you would not have switch over, and hopefully is clean this time.

Thanks!

PR #1256, opened 2020-04-21
add command to copy tags from one project to another

ssadedin:
There might be already a good way to do this but I found I often want to copy the tags from one project to another, so I made a simple SeqR management command for that.

Feel free to close if you already have a way to do this but I figured since it's useful for me I'd offer it for what it is worth.

### DIFF
--- a/seqr/management/commands/copy_project_tags.py
+++ b/seqr/management/commands/copy_project_tags.py
@@ -1,8 +1,8 @@
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import BaseCommand
 from django.db.models.query_utils import Q
 
 from seqr.models import Project, VariantTagType
-from seqr.model_utils import create_seqr_model
+
 
 class Command(BaseCommand):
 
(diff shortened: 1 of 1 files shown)


hanars:
Since most of the tags at this point are the non-project-specific tags and are added to all projects by default and we are trying to minimize the number of project specific tags we have, we've never needed this functionality. However, theres no reason not to have it!

seqr/management/commands/copy_project_tags.py
```diff
@@ -0,0 +1,30 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models.query_utils import Q
+
+from seqr.models import Project, VariantTagType
+from seqr.model_utils import create_seqr_model
```
hanars, line 5: This isn't used

seqr/management/commands/copy_project_tags.py
```diff
@@ -0,0 +1,30 @@
+from django.core.management.base import BaseCommand, CommandError
```
hanars, line 1: CommandError isn;t used

seqr/management/commands/copy_project_tags.py
ssadedin: have removed the unused imports in latest commits

seqr/management/commands/copy_project_tags.py
ssadedin: have removed the unused imports in latest commits

PR #1259, opened 2020-04-22
Add unit test for the staff anvil_export API

ShifaSZ:
The below data are created for the test:
1. A VariantTagType record is added to the fixture data 1kg_project.json. It is required because the API under test is accessing a VariantTagType record with a name "Known gene for phenotype". There is no existing record with the name.
2. A VariantTag record is added with the code in the test. It is not added to the 1kg_project.json because it breaks some other tests.
3. Some mock records of "Sample" responses from the Airtable
4. Some mock records of "Collaborator" responses from the Airtable
All Airtable response records are manually created according to the needs of the test.
5. Expected output file data is created.
Please review whether these test data are created properly.


seqr/views/apis/staff_api_tests.py
```diff
@@ -648,8 +648,8 @@ def test_anvil_export(self, mock_export_multiple_files):
         _check_login(self, url)
 
         # Create a SavedVariant row needed by this test
-        saved_variant = SavedVariant.objects.get(guid = "SV0000001_2103343353_r0390_100")
-        variant_tag_type = VariantTagType.objects.get(name = "Known gene for phenotype")
+        saved_variant = SavedVariant.objects.get(guid__exact = "SV0000001_2103343353_r0390_100")
```
hanars, line 651: You don't need to use `__exact`, by default if you just do `guid=...` it will do an exact lookup and its easier to read that way

seqr/views/apis/staff_api_tests.py
```diff
@@ -260,6 +265,261 @@
...
+         {'entity:discovery_id': u'NA20872', 'subject_id': u'NA20872', 'sample_id': u'NA20872'},
+         {'entity:discovery_id': u'NA20874', 'subject_id': u'NA20874', 'sample_id': u'NA20874'},
+         {'entity:discovery_id': u'NA20875', 'subject_id': u'NA20875', 'sample_id': u'NA20875'},
+         {'entity:discovery_id': u'NA20876', 'subject_id': u'NA20876', 'sample_id': u'NA20876'},
+         {'entity:discovery_id': u'NA20877', 'subject_id': u'NA20877', 'sample_id': u'NA20877'},
+         {'entity:discovery_id': u'NA20878', 'subject_id': u'NA20878', 'sample_id': u'NA20878'},
+         {'entity:discovery_id': u'NA20881', 'subject_id': u'NA20881', 'sample_id': u'NA20881'}]
+     ]
+]
+
```
hanars, line 522: having this much test data is very fragile. Can you test one row per file, the way you did for the discovery sheet?

seqr/views/apis/staff_api_tests.py
```diff
@@ -380,3 +640,31 @@ def test_success_story(self):
 
         self.assertEqual(len(response_json['rows']), 1)
         self.assertDictEqual(response_json['rows'][0], EXPECTED_SUCCESS_STORY)
+
+    @mock.patch('seqr.views.apis.staff_api.export_multiple_files')
+    @responses.activate
+    def test_anvil_export(self, mock_export_multiple_files):
+        url = reverse(anvil_export, args=[PROJECT_GUID])
```
hanars, line 647: Are you planning on doing a separate pr for the sample_metadata_export or do you want to do it as part of this PR?

seqr/views/apis/staff_api_tests.py
```diff
@@ -380,3 +640,31 @@ def test_success_story(self):
...
+        _check_login(self, url)
+
+        # Create a SavedVariant row needed by this test
+        saved_variant = SavedVariant.objects.get(guid__exact = "SV0000001_2103343353_r0390_100")
+        variant_tag_type = VariantTagType.objects.get(name__exact = "Known gene for phenotype")
+        variant_tag = VariantTag.objects.create(variant_tag_type = variant_tag_type)
+        variant_tag.saved_variants.add(saved_variant)
+        variant_tag.save()
```
hanars, line 655: You should put this in the fixture data, not have it dynamically created here

seqr/views/apis/staff_api_tests.py
```diff
@@ -380,3 +640,31 @@ def test_success_story(self):
...
+
+        # We will test the inputs of the export_multiple_files method.
+        # Outputs of the method are not important for this test.
+        mock_export_multiple_files.return_value = HttpResponse("Dummy text", content_type="text/plain")
```
hanars, line 659: If output doesn't matter then you shouldn't need to specify any return value at all

seqr/views/apis/staff_api_tests.py
```diff
@@ -380,3 +640,31 @@ def test_success_story(self):
...
+        responses.add(responses.GET, '{}/Samples'.format(AIRTABLE_URL),
+                      json=AIRTABLE_SAMPLE_RECORDS, status=200)
+        responses.add(responses.GET, '{}/Collaborator'.format(AIRTABLE_URL),
+                      json=AIRTABLE_COLLABORATOR_RECORDS, status=200)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+
+        mock_export_multiple_files.assert_called_with(EXPECTED_EXPORT_FILES,
+            u'1kg project n\u00e5me with uni\u00e7\u00f8de_AnVIL_Metadata',
+            add_header_prefix = True, file_format = 'tsv', blank_value = '-')
```
hanars, line 670: Instead of specifying all the `EXPECTED_EXPORT_FILES` here, I would change this to `mock_export_multiple_files.assert_called_with(mock.ANY, u'1kg project n\u00e5me with uni\u00e7\u00f8de_AnVIL_Metadata',
            add_header_prefix = True, file_format = 'tsv', blank_value = '-')` which will test all the other parameters to the call. Then you can access what files it was called with using something like `exported_files = mock_export_multiple_files.call_args.args[0]` and you can do a series of test around that, like that it has the proper length and then for each file check it has the expected filename, header, and contains one correct row

seqr/views/apis/staff_api_tests.py
ShifaSZ: Removed. I also removed the `__exact` for `name` because it makes no different in the test.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Yes. I included all the data because they are calling parameters which can only be tested as a whole. I'm going to put them into the HTTP response then I can test one row per file.

seqr/views/apis/staff_api_tests.py
ShifaSZ: My plan is to do it in the same PR after the anvil_export is confirmed to be good. But I can change it a separated PR if you see any issues.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I did put it in the fixture data. But it breaks some save_variant_api tests and some other tests. So I decided not to change the fixture data. If it must be in the fixture data, then I can find out the issue and make a new data which won't break other tests.

seqr/views/apis/staff_api_tests.py
ShifaSZ: The API still needs an HttpResponse even the test doesn't need any return value. I will update the comments to specify that.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Good! Thank you for telling me about `call_args`. It is very useful in this case.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I'm going to use `call_args` to test the data.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I probably do the sample_metadata_export next week with another PR.

seqr/views/apis/staff_api_tests.py
hanars: This should be fixture data. You may need to update other tests, thats okay

ShifaSZ: Fixed.

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -155,10 +154,8 @@ def test_saved_variant_data(self):
             'genesById', 'locusListsByGuid',
         })
         variants = response_json['savedVariantsByGuid']
-        self.assertSetEqual(
-            set(variants.keys()),
-            {'SV0000002_1248367227_r0390_100', 'SV0000001_2103343353_r0390_100', COMPOUND_HET_1_GUID, COMPOUND_HET_2_GUID}
-        )
+        guid_set = {'SV0000002_1248367227_r0390_100', 'SV0000001_2103343353_r0390_100', COMPOUND_HET_1_GUID, COMPOUND_HET_2_GUID}
+        self.assertSetEqual({guid for guid in variants.keys() if guid in guid_set}, guid_set)
```
hanars, line 158: This doesn't test anything. It says "assert that the subset of this set that is in `guid_set` equals `guid_set`, which is obviously true but meaningless. Please add the new saved variant object to a different family to prevent this issue

seqr/views/apis/staff_api_tests.py
```diff
@@ -1,11 +1,16 @@
 # -*- coding: utf-8 -*-
 import mock
 from datetime import datetime
+import responses
+from django.http import HttpResponse
+from seqr.models import VariantTag, VariantTagType, SavedVariant
```
hanars, line 6: These aren't used

seqr/views/apis/staff_api_tests.py
```diff
@@ -380,3 +490,35 @@ def test_success_story(self):
...
+        exported_files = mock_export_multiple_files.call_args.args[0]
+
+        self.assertListEqual([EXPECTED_PI_SUBJECT_FILE[0], EXPECTED_PI_SUBJECT_FILE[1]], [exported_files[0][0], exported_files[0][1]])
+        self.assertListEqual([EXPECTED_PI_SAMPLE_FILE[0], EXPECTED_PI_SAMPLE_FILE[1]], [exported_files[1][0], exported_files[1][1]])
+        self.assertListEqual([EXPECTED_PI_FAMILY_FILE[0], EXPECTED_PI_FAMILY_FILE[1]], [exported_files[2][0], exported_files[2][1]])
+        self.assertListEqual([EXPECTED_PI_DISCOVERY_FILE[0], EXPECTED_PI_DISCOVERY_FILE[1]], [exported_files[3][0], exported_files[3][1]])
+        self.assertIn(EXPECTED_PI_SUBJECT_FILE[2][0], exported_files[0][2])
+        self.assertIn(EXPECTED_PI_SAMPLE_FILE[2][0], exported_files[1][2])
+        self.assertIn(EXPECTED_PI_FAMILY_FILE[2][0], exported_files[2][2])
+        self.assertIn(EXPECTED_PI_DISCOVERY_FILE[2][0], exported_files[3][2])
```
hanars, line 524: This is a bit confusing to read. Instead of having `EXPECTED_PI_<X>_FILE` be a list of 3 elements, maybe make them lists of 2 elements so the assertions can look like `self.assertListEqual(EXPECTED_PI_<X>_FILE, exported_files[X][:1])` and then make separate variables for the content so you second set of assertions would look like `self.assertIn(EXPECTED_PI_<X>_ROW, exported_files[X][2])`

seqr/views/apis/saved_variant_api_tests.py
ShifaSZ: When the set(variants.key()) misses any elements in guid_set, then it will be an assertion error.  The test asserts that the guid_set is a subset of set(variants.keys()), or set(variants.keys()) contains guid_set. It is not always true.
The issue of using a different family is that I have to create a complete set of saved_variants, variant_tags, variant_tag_type (maybe), samples, individuals, and a family. I think they are too many. How do you think about that?

seqr/views/apis/staff_api_tests.py
ShifaSZ: Removed.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Good idea.

seqr/views/apis/saved_variant_api_tests.py
hanars: Oh sorry I misread this code. The much clearer way to write this is `set(variants.keys())`

seqr/views/apis/staff_api_tests.py
```diff
@@ -260,6 +271,105 @@
...
+    'maternal_id': u'NA19679', 'subject_id': u'NA19675_1', 'consanguinity': 'Present',
+    'family_id': u'1', 'entity:family_id': u'NA19675_1', 'paternal_id': u'NA19678'}
+
+EXPECTED_PI_DISCOVERY_FILE = [
+    u'1kg project n\xe5me with uni\xe7\xf8de_PI_Discovery',
+    ['entity:discovery_id', 'subject_id', 'sample_id', 'Gene-1', 'Gene_Class-1',
+     'inheritance_description-1', 'Zygosity-1', 'Chrom-1', 'Pos-1', 'Ref-1', 'Alt-1', 'hgvsc-1',
+     'hgvsp-1', 'Transcript-1', 'sv_name-1', 'sv_type-1', 'significance-1']
+]
+EXPECTED_PI_DISCOVERY_ROW = {'entity:discovery_id': u'NA19675_1', 'subject_id': u'NA19675_1', 'sample_id': u'NA19675'}
```
hanars, line 372: You need to test a discovery row that has gene data. In our call yesterday I specifically said this which is why I wanted to use saved variant 1. Its okay to instead use the other saved variant, but the here you should be using the row representing variant 2 to test against instead

seqr/views/apis/staff_api_tests.py
```diff
@@ -260,6 +271,105 @@
...
+    'hpo_present': u'HP:0001631|HP:0002011|HP:0001636', 'multiple_datasets': 'No', 'onset_category': u'Adult onset',
+    'subject_id': u'NA19675_1', 'family_guid': u'F000001_1', 'affected_status': 'Affected', 'pmid_id': '',
+    'project_id': u'1kg project n\xe5me with uni\xe7\xf8de'}
+
+EXPECTED_PI_SAMPLE_FILE = [
+    u'1kg project n\xe5me with uni\xe7\xf8de_PI_Sample',
+    ['entity:sample_id', 'subject_id', 'sample_id', 'dbgap_sample_id', 'sample_source',
+     'sample_provider', 'data_type', 'date_data_generation']
+]
+EXPECTED_PI_SAMPLE_ROW = {'entity:sample_id': u'NA19675_1', 'data_type': u'WES', 'subject_id': u'NA19675_1', 'sample_provider': '', 'sample_id': u'NA19675', 'date_data_generation': '2017-02-05'}
```
hanars, line 354: You need to test against a row that has the airtable data, so pick a row to use that has sample_provider set

seqr/views/apis/staff_api_tests.py
```diff
@@ -260,6 +271,105 @@
...
+}
+
+EXPECTED_PI_SUBJECT_FILE = [
+    u'1kg project n\xe5me with uni\xe7\xf8de_PI_Subject',
+    ['entity:subject_id', 'subject_id', 'prior_testing', 'project_id', 'pmid_id',
+     'dbgap_submission', 'dbgap_study_id', 'dbgap_subject_id', 'multiple_datasets', 'sex',
+     'ancestry', 'ancestry_detail', 'age_at_last_observation', 'phenotype_group', 'disease_id',
+     'disease_description', 'affected_status', 'onset_category', 'age_of_onset', 'hpo_present',
+     'hpo_absent', 'phenotype_description', 'solve_state']
+]
```
hanars, line 340: You need to test against a row that has the airtable data, so pick a row to use that has dbgap_subject_id set

seqr/views/apis/staff_api_tests.py
```diff
@@ -290,11 +400,11 @@ def test_elasticsearch_status(self, mock_elasticsearch, mock_get_mapping):
...
-        mock_es_client.cat.allocation.assert_called_with(format="json", h="node,disk.avail,disk.used,disk.percent")
-        mock_es_client.cat.indices.assert_called_with(format="json",
-                                                      h="index,docs.count,store.size,creation.date.string")
-        mock_es_client.cat.aliases.assert_called_with(format="json", h="alias,index")
-        mock_get_mapping.assert_called_with(doc_type='variant,structural_variant')
+        mock_es_client.cat.allocation.assert_called_with(format = "json", h = "node,disk.avail,disk.used,disk.percent")
+        mock_es_client.cat.indices.assert_called_with(format = "json",
+                                                      h = "index,docs.count,store.size,creation.date.string")
+        mock_es_client.cat.aliases.assert_called_with(format = "json", h = "alias,index")
+        mock_get_mapping.assert_called_with(doc_type = 'variant,structural_variant')
```
hanars, line 407: This does not conform with how any other code in the repo is styled. Please stick to consistent styling to keep the repo readable. Additionally, it creates work for me when I need to go through files and figure out what are real changes and what are just spacing changes. In the future, I will be asking you to do all the style cleanups before I will be reviewing PRs

ShifaSZ: I can't find the style checking tool on the github recently. How can I open it?
Sorry for creating work for you. Would it help if I make styling changes with a separated commit?

hanars: This isn't about the automated style checker, its about the fact that there should be nothing in the diff on github that isn't functionally relevant to the PR. If you just looked at the diff here you would see all these places where the code has different spacing that clearly don't impact the functionality

seqr/views/apis/staff_api_tests.py
ShifaSZ: To have gene data in the discovery rows, I updated the genotype data of the saved_variant 2. Now it seems work. Please review.

seqr/views/apis/staff_api_tests.py
ShifaSZ: I updated the mock responses from Airtable by matching the sample ID. It seems work now. Please review.

seqr/views/apis/staff_api_tests.py
ShifaSZ: As above.

seqr/views/apis/staff_api_tests.py
ShifaSZ: Understood.

PR #1269, opened 2020-04-27
Seqr management command unit tests for run_postgres_database_backup

ShifaSZ:
Added two test cases and adapted the command code to make the tests work.

### DIFF
--- a/seqr/management/commands/run_postgres_database_backup_tests.py
+++ b/seqr/management/commands/run_postgres_database_backup_tests.py
@@ -7,11 +7,11 @@ from django.core.management import call_command
 from django.test import TestCase
 
 
+@mock.patch('seqr.management.commands.run_postgres_database_backup.os')
+@mock.patch('seqr.management.commands.run_postgres_database_backup.datetime')
 class RunProtgresDatabaseBackupTest(TestCase):
 
     # Test the command without an argument
-    @mock.patch('seqr.management.commands.run_postgres_database_backup.os')
-    @mock.patch('seqr.management.commands.run_postgres_database_backup.datetime')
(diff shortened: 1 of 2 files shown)


seqr/management/commands/run_postgres_database_backup_tests.py
```diff
@@ -0,0 +1,69 @@
...
+            '/usr/local/bin/gsutil mv /postgres_backups/xwiki_unknown_backup_2020-04-27__20-16-01.txt.gz gs://db_back_bucket/postgres/xwiki_unknown_backup_2020-04-27__20-16-01.txt.gz\n',
+            out.getvalue())
+
+        mock_os.system.assert_called_with(
+            '/usr/local/bin/gsutil mv /postgres_backups/xwiki_unknown_backup_2020-04-27__20-16-01.txt.gz gs://db_back_bucket/postgres/xwiki_unknown_backup_2020-04-27__20-16-01.txt.gz')
+
+    # Test the command with different arguments
+    @mock.patch('seqr.management.commands.run_postgres_database_backup.os')
+    @mock.patch('seqr.management.commands.run_postgres_database_backup.datetime')
+    def test_command_with_arguments(self, mock_datetime, mock_os):
```
hanars, line 47: If all the tests in a module need to mock the same thing, you can put the mock decorators on the class itself instead of on each method. So in this case you would put them on top of `RunProtgresDatabaseBackupTest`, and then the test functions would still get the mocks as arguments the way they do now

ShifaSZ: Thank you for letting me know that.

PR #1278, opened 2020-05-05
Added a unit test for the reference_data management command update_dbnsfp_gene

ShifaSZ:
As the title.


reference_data/management/commands/utils/update_utils.py
```diff
@@ -36,7 +36,7 @@ def parse_record(record):
 
     @staticmethod
     def get_file_header(f):
-        return next(f).rstrip('\n\r').split('\t')
+        return f.readline().rstrip('\n\r').split('\t')
```
hanars, line 39: Why change this? Before it would skip any empty lines at the top of the file and now it doesn't

reference_data/management/commands/utils/update_utils.py
```diff
@@ -90,7 +90,8 @@ def update_records(reference_data_handler, file_path=None):
 
         for line in tqdm(f, unit=" records"):
             record = dict(zip(header_fields, line.rstrip('\r\n').split('\t')))
-            for record in reference_data_handler.parse_record(record):
+            records = reference_data_handler.parse_record(record)
+            for record in records:
```
hanars, line 94: Why change this? It makes no difference to the code execution, it doesn't make it any easier to read...

reference_data/management/tests/update_dbnsfp_gene_tests.py
```diff
@@ -0,0 +1,45 @@
...
+    @mock.patch('reference_data.management.commands.utils.update_utils.logger')
+    @mock.patch('reference_data.management.commands.utils.update_utils.tqdm')
+    @mock.patch('os.path.isfile')
+    @mock.patch.object(__builtin__, 'open')
+    @mock.patch('reference_data.management.commands.utils.download_utils.download_file')
+    def test_update_dbnsfp_gene_command(self, mock_download, mock_open, mock_isfile, mock_tqdm, mock_logger):
+        mock_isfile.return_value = False
+        mock_download.return_value = 'dbNSFP_gene'
+        f = mock_open.return_value.__enter__.return_value
+        f.readline.return_value = dbNSFP_HEADER
```
hanars, line 29: You should use `SimpleUploadedFile` for this, the way you do in the tests that test uploading and parsing files. Mock the return of mock_download to return a file and then you don't need to mock all the file functions

reference_data/management/tests/update_dbnsfp_gene_tests.py
```diff
@@ -0,0 +1,45 @@
...
+        mock_tqdm.return_value = dbNSFP_GENE_DATA
```
hanars, line 30: tqdm is a wrapper library that return the given iterable and prints some stuff about it. Don't mock it, as was discussed in previous PRs. If the print output is really messing with logging tests, then use a side_effect so it just returns whatever iterable it is given. But having a printing utility totally change the given value is really unintuitive and dpes not do a good job testing the given files

reference_data/management/commands/update_dbnsfp_gene.py
```diff
@@ -3,8 +3,6 @@
 from reference_data.management.commands.utils.update_utils import GeneCommand, ReferenceDataHandler
 from reference_data.models import dbNSFPGene
 
-logger = logging.getLogger(__name__)
```
hanars, line 6: By removing this the `logging ` import at the top of the file is no longer used and should be removed

reference_data/management/commands/utils/update_utils.py
ShifaSZ: I changed it because Python2 file object doesn't  have __next__() but has next() method and mock_open doesn't support next().
If it is required to keep the skipping empty lines feature, then I could do it in a Python2 only way and update it when migrating to Python3 according to [stack overflow](https://stackoverflow.com/questions/24779893/customizing-unittest-mock-mock-open-for-iteration) .

reference_data/management/commands/utils/update_utils.py
ShifaSZ: It is weird that it doesn't work without change it with mocking tqdm. I will change it back after removing tqdm mocking.

reference_data/management/tests/update_dbnsfp_gene_tests.py
ShifaSZ: Good point. I remember the SimpleUploadedFile I learned at the very beginning now. I will use it.

reference_data/management/tests/update_dbnsfp_gene_tests.py
ShifaSZ: Okay

reference_data/management/commands/update_dbnsfp_gene.py
ShifaSZ: Okay

reference_data/management/commands/utils/update_utils.py
ShifaSZ: Changed back to the original after removing mock_open.

reference_data/management/commands/utils/update_utils.py
ShifaSZ: Changed back after removing mocked tqdm.

reference_data/management/tests/update_dbnsfp_gene_tests.py
ShifaSZ: I found SimpleUploadedFile doesn't work for this case. It still needs to use `os` library to save the data to a temp file. Since we have to use a temp file, I can solved the issue by creating a temp file. Please review and confirm whether it is acceptable.

hanars: Okay so looking into this more it looks like best practice for unit tests that need temp files is to use the `tempfile` module and create the files you need in the test SetUp: https://simpleit.rocks/python/test-files-creating-a-temporal-directory-in-python-unittests/

ShifaSZ: Updated uses of the tempfile. 
Slightly different from the example in the link, a temp directory is created instead of just a temp file. The reason for this is that tempfile.TemporaryFile() opens (not just create) a temp file and tempfile.TemporayDirectory() is not supported in my environment.

reference_data/management/tests/update_dbnsfp_gene_tests.py
```diff
@@ -0,0 +1,67 @@
...
+        call_command('update_dbnsfp_gene')
+
+        mock_download.assert_called_with('http://storage.googleapis.com/seqr-reference-data/dbnsfp/dbNSFP4.0_gene')
+
+        calls = [
+            mock.call('Deleting 3 existing dbNSFPGene records'),
+            mock.call('Parsing file'),
+            mock.call('Creating 1 dbNSFPGene records'),
+            mock.call('Done'),
+            mock.call('Loaded 1 dbNSFPGene records from {}. Skipped 1 records with unrecognized genes.'.format(temp_file_path)),
```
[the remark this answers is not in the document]
ShifaSZ: To increase coverage, I removed the gene id of a record with a gene name "MIR1302-2HG". The current code will not count it as a record.
I updated the gene name to "MIR1302-2HG-W/O-GeneID". Hope it can be more clear.

reference_data/management/tests/update_dbnsfp_gene_tests.py
```diff
@@ -0,0 +1,67 @@
...
+            mock.call('Running ./manage.py update_gencode to update the gencode version might fix missing genes')
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
```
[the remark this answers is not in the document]
ShifaSZ: Added db data tests.

reference_data/management/tests/update_dbnsfp_gene_tests.py
```diff
@@ -0,0 +1,73 @@
...
+    def tearDown(self):
+        # Close the file, the directory will be removed after the test
+        shutil.rmtree(self.test_dir)
+
+    @mock.patch('reference_data.management.commands.utils.update_utils.logger')
+    @mock.patch('reference_data.management.commands.utils.download_utils.download_file')
+    def test_update_dbnsfp_gene_command(self, mock_download, mock_logger):
+        temp_file_path = os.path.join(self.test_dir, 'dbNSFP_gene')
+        with open(temp_file_path, 'w') as f:
+            f.write(u''.join(dbNSFP_GENE_DATA))
```
hanars, line 38: This should be in the test setUp, you can set `self. temp_file_path` so you can access it in the test

reference_data/management/tests/update_dbnsfp_gene_tests.py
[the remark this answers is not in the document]
hanars: I wasn't asking about the skipped gene, I was asking about the fact that there are 3 rows in the data and  the logger here only mentions 2 of them, I want to know what happened to the third. You have `OR4F5` get created and `MIR1302-2HG-W/O-GeneID` get skipped (which was getting skipped before you changed the name and was clearer the way it was). What happened to `A1BG-AS1`?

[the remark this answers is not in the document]
ShifaSZ: The gene `OR4F5` is loaded, `A1BG-AS1` is skipped, and the gene `MIR1302-2HG-W` is un-counted (neither loaded nor skipped) because it doesn't have a gene_id.

reference_data/management/tests/update_dbnsfp_gene_tests.py
ShifaSZ: Okay. I will try.

reference_data/management/tests/update_dbnsfp_gene_tests.py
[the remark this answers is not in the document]
hanars: Got it. It would be a bit more readable if you just took out the name for `MIR1302-2HG-W/O-GeneID` altogether and made it just a . and then changed A1BG-AS1 to something like `INVALID_GENE`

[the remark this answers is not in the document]
ShifaSZ: `MIR1302-2HG-W/O-GeneID` and `A1BG-AS1` are both the gene_**_name_**. I would like a test without a gene_**_id_**. The gene `MIR1302-2HG-W/O-GeneID` has a gene_**_id_** `.` already. To make less confusing, I will change the gene_name `MIR1302-2HG-W/O-GeneID` to `INVALID-GENE-NO-ID` and `A1BG-AS1` to `INVALID_GENE-UNKNOWN`.

PR #1281, opened 2020-05-06
Added unit test for the update_gencode management command for reference_data

ShifaSZ:
As the title.


ShifaSZ:
Need more updates before a PR.

PR #1283, opened 2020-05-11
postgres database backup script defaults and docs

bw2:


seqr/management/commands/run_postgres_database_backup.py
```diff
@@ -32,10 +40,12 @@ def run(cmd):
             backup_filename = "{db_name}_{deployment_type}_backup_{timestamp}.txt.gz".format(
                 db_name=db_name, deployment_type=options['deployment_type'], timestamp=timestamp)
 
-            run("/usr/bin/pg_dump -U postgres --host {postgres_host} {db_name} | gzip -c - > {backup_dir}/{backup_filename}".format(
+            run("pg_dump -U postgres --host {postgres_host} {db_name} | gzip -c - > {backup_dir}/{backup_filename}".format(
```
hanars, line 43: I'm pretty sure taking this out causes this command to fail in our cron job

seqr/management/commands/run_postgres_database_backup.py
```diff
@@ -32,10 +40,12 @@ def run(cmd):
...
                postgres_host=options['postgres_host'], db_name=db_name, backup_dir=backup_dir,
                 backup_filename=backup_filename
             ))
-            run("/usr/local/bin/gsutil mv {backup_dir}/{backup_filename} gs://{bucket}/postgres/{backup_filename}".format(
-                backup_dir=backup_dir, backup_filename=backup_filename, bucket=options['bucket']
-            ))
+            
+            if options.get('bucket'):
+                run("gsutil -m cp {backup_dir}/{backup_filename} gs://{bucket}/postgres/{backup_filename}".format(
```
hanars, line 49: ot of curiosity, why is `-m cp` better than `mv`?

bw2: cp seems less surprising than mv, and -m makes it faster. But I can change this back since mv was working ok.

seqr/management/commands/run_postgres_database_backup.py
bw2: good point. We should double check that /usr/bin and /usr/local/bin are on PATH in cron.

seqr/management/commands/run_postgres_database_backup.py
hanars: No I like the change, keep it. I was just wondering about the motivation

seqr/management/commands/run_postgres_database_backup.py
hanars: I'm pretty sure they weren't which is why backups were failing that time we lost all the data :(

hanars:
@bw2 what was the motivation behind this PR? Is this something we should still work toward merging?

PR #1287, opened 2020-05-14
Unit tests for the reference_data management command update_gencode

ShifaSZ:
As the title.

### DIFF
--- a/reference_data/management/tests/update_gencode_tests.py
+++ b/reference_data/management/tests/update_gencode_tests.py
@@ -134,22 +134,21 @@ class UpdateGencodeTest(TestCase):
     @mock.patch('reference_data.management.commands.update_gencode.logger')
     def test_update_gencode_command(self, mock_logger):
         # Test normal command function
-        call_command('update_gencode', '--reset', '--gencode-release=31', self.temp_file_path, '37')
+        call_command('update_gencode', '--gencode-release=31', self.temp_file_path, '37')
         calls = [
-            mock.call('Dropping the 0 existing TranscriptInfo entries'),
-            mock.call('Dropping the 49 existing GeneInfo entries'),
             mock.call(
                 'Loading {} (genome version: 37)'.format(self.temp_file_path)),
-            mock.call('Creating 2 GeneInfo records'),
+            mock.call('Creating 1 GeneInfo records'),
             mock.call('Creating 2 TranscriptInfo records'),
             mock.call('Done'),
             mock.call('Stats: '),
-            mock.call('  genes_created: 2'),
+            mock.call('  genes_skipped: 1'),
+            mock.call('  genes_created: 1'),
             mock.call('  transcripts_created: 2')
         ]
         mock_logger.info.assert_has_calls(calls)
 
-        gene_infos = [{
+        gene_infos = {gene.gene_id: {
             'start_grch37': gene.start_grch37,
             'chrom_grch37': gene.chrom_grch37,
             'coding_region_size_grch37': gene.coding_region_size_grch37,
@@ -159,10 +158,9 @@ class UpdateGencodeTest(TestCase):
             'gene_symbol': gene.gene_symbol,
             'end_grch37': gene.end_grch37,
             'strand_grch37': gene.strand_grch37
-        } for gene in GeneInfo.objects.all()]
-        self.assertEqual(len(gene_infos), 2)
-        self.assertDictEqual(gene_infos[0], {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 31, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
-        self.assertDictEqual(gene_infos[1], {'start_grch37': 621059, 'chrom_grch37': u'1', 'coding_region_size_grch37': 936, 'gencode_release': 31, 'gencode_gene_type': u'protein_coding', 'gene_id': u'ENSG00000284662', 'gene_symbol': u'OR4F16', 'end_grch37': 622053, 'strand_grch37': u'-'})
+        } for gene in GeneInfo.objects.filter(gene_id__in = ['ENSG00000223972', 'ENSG00000284662'])}
+        self.assertDictEqual(gene_infos['ENSG00000223972'], {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 27, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
+        self.assertDictEqual(gene_infos['ENSG00000284662'], {'start_grch37': 621059, 'chrom_grch37': u'1', 'coding_region_size_grch37': 936, 'gencode_release': 31, 'gencode_gene_type': u'protein_coding', 'gene_id': u'ENSG00000284662', 'gene_symbol': u'OR4F16', 'end_grch37': 622053, 'strand_grch37': u'-'})
 
         transcript_infos = {trans.transcript_id: {
             'start_grch37': trans.start_grch37,
@@ -177,17 +175,33 @@ class UpdateGencodeTest(TestCase):
 
         # Test normal command function without a --reset option
         mock_logger.reset_mock()
-        call_command('update_gencode', '--gencode-release=31', self.temp_file_path, '37')
+        call_command('update_gencode', '--reset', '--gencode-release=31', self.temp_file_path, '37')
         calls = [
+            mock.call('Dropping the 2 existing TranscriptInfo entries'),
+            mock.call('Dropping the 50 existing GeneInfo entries'),
             mock.call(
                 'Loading {} (genome version: 37)'.format(self.temp_file_path)),
-            mock.call('Creating 0 GeneInfo records'),
-            mock.call('Creating 0 TranscriptInfo records'),
+            mock.call('Creating 2 GeneInfo records'),
+            mock.call('Creating 2 TranscriptInfo records'),
             mock.call('Done'),
             mock.call('Stats: '),
-            mock.call('  genes_skipped: 2'),
-            mock.call('  transcripts_skipped: 2'),
-            mock.call('  genes_created: 0'),
-            mock.call('  transcripts_created: 0')
+            mock.call('  genes_created: 2'),
+            mock.call('  transcripts_created: 2')
         ]
         mock_logger.info.assert_has_calls(calls)
+
+        gene_infos = {gene.gene_id: {
+            'start_grch37': gene.start_grch37,
+            'chrom_grch37': gene.chrom_grch37,
+            'coding_region_size_grch37': gene.coding_region_size_grch37,
+            'gencode_release': gene.gencode_release,
+            'gencode_gene_type': gene.gencode_gene_type,
+            'gene_id': gene.gene_id,
+            'gene_symbol': gene.gene_symbol,
+            'end_grch37': gene.end_grch37,
+            'strand_grch37': gene.strand_grch37
+        } for gene in GeneInfo.objects.all()}
+        self.assertEqual(len(gene_infos), 2)
+        self.assertDictEqual(gene_infos['ENSG00000223972'], {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 31, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
+        self.assertDictEqual(gene_infos['ENSG00000284662'], {'start_grch37': 621059, 'chrom_grch37': u'1', 'coding_region_size_grch37': 936, 'gencode_release': 31, 'gencode_gene_type': u'protein_coding', 'gene_id': u'ENSG00000284662', 'gene_symbol': u'OR4F16', 'end_grch37': 622053, 'strand_grch37': u'-'})
+
--- a/reference_data/management/tests/update_gencode_tests.py
+++ b/reference_data/management/tests/update_gencode_tests.py
@@ -1,0 +1,175 @@
+import mock
+import __builtin__
+
+from django.core.management import call_command
+from django.test import TestCase
+from django.core.management.base import CommandError
+
+from reference_data.models import GENOME_VERSION_GRCh37, GENOME_VERSION_GRCh38
+
+BAD_FIELDS_GTF_DATA = [
+    'gene	11869	14412	.	+	.	gene_id "ENSG00000223972.4";\n',
+]
+
+GTF_DATA = [
+    # Comment
+    '#description: evidence-based annotation of the human genome, version 31 (Ensembl 97), mapped to GRCh37 with gencode-backmap\n',
+    # Existing gene_id
+    'chr1	HAVANA	gene	11869	14409	.	+	.	gene_id "ENSG00000223972.5_2"; gene_type "transcribed_unprocessed_pseudogene"; gene_name "DDX11L1"; level 2; hgnc_id "HGNC:37102"; havana_gene "OTTHUMG00000000961.2_2"; remap_status "full_contig"; remap_num_mappings 1; remap_target_status "overlap";\n',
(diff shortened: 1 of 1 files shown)


reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,159 @@
...
+        self.assertEqual(ce.exception.message, 'Error: argument --gencode-release is required')
+
+        # Test required argument out-of-range
+        with self.assertRaises(CommandError) as ce:
+            call_command('update_gencode', '--gencode-release=18')
+        self.assertEqual(ce.exception.message, 'Error: argument --gencode-release: invalid choice: 18 (choose from 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31)')
+
+        # Test genome_version out-of-range
+        with self.assertRaises(CommandError) as ce:
+            call_command('update_gencode', '--gencode-release=19', '/var/tmp', '39')
```
hanars, line 67: you should not use real files in real tmp directories, you should mock out all system calls

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,159 @@
...
+            call_command('update_gencode', '--reset', '--gencode-release=23', temp_bad_file_path, '37')
+        self.assertEqual(ve.exception.message, 'Unexpected number of fields on line #0: [\'gene\', \'11869\', \'14412\', \'.\', \'+\', \'.\', \'gene_id "ENSG00000223972.4";\']')
+
+    @mock.patch('reference_data.management.commands.update_gencode.logger')
+    @mock.patch('reference_data.management.commands.update_gencode.download_file')
+    def test_update_gencode_command_special_paths(self, mock_download, mock_logger):
+        # Test the code paths of generating urls, gencode_release == 19, and --reset option
+        mock_logger.reset_mock()
+        mock_download.return_value = self.temp_file_path
+        call_command('update_gencode', '--reset', '--gencode-release=19')
```
hanars, line 112: you need to test that all of these options work as expected - reset needs to actually reset the database and you need to test that the db models look as expected after this call. I would recommend moving the --reset testing into the main `test_update_gencode_command` test and keep this test isolated to testing url generation

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,159 @@
...
+            call_command('update_gencode', '--gencode-release=23', '/var/tmp2.gz', '37')
+        mock_isfile.assert_called_with('/var/tmp2.gz')
+        self.assertEqual(ce.exception.message, "Invalid genome_version for file: /var/tmp2.gz. gencode v23 and up must have 'lift' in the filename or genome_version arg must be GRCh38")
+
+        # Test --reset option and wrong number data feilds in a line
+        temp_bad_file_path = os.path.join(self.test_dir, 'bad.gencode.v23lift37.annotation.gtf.gz')
+        with gzip.open(temp_bad_file_path, 'w') as f:
+            f.write(u''.join(BAD_FIELDS_GTF_DATA))
+        with self.assertRaises(ValueError) as ve:
+            call_command('update_gencode', '--reset', '--gencode-release=23', temp_bad_file_path, '37')
```
hanars, line 103: why does reset have any impact on this test? The bad file should fail regardless of whether the db gets reset so adding it just makes the test confusing

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: In these test, the tmp file names in the argument will never be used for real file accesses (exception will happen before a file access). They are used for command argument checking only. Therefore, no needs to mock out file operations for these tests.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Okay. Moving `--reset` test to `test_update_gencode_command`.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Removed the reset option.

reference_data/management/tests/update_gencode_tests.py
hanars: Thats good then. Can you change it to something more obviously fake, maybe like `/local/path` or `/mock/dir` or something like that?

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,193 @@
...
+            f.write(u''.join(BAD_FIELDS_GTF_DATA))
+        with self.assertRaises(ValueError) as ve:
+            call_command('update_gencode', '--gencode-release=23', temp_bad_file_path, '37')
+        self.assertEqual(ve.exception.message, 'Unexpected number of fields on line #0: [\'gene\', \'11869\', \'14412\', \'.\', \'+\', \'.\', \'gene_id "ENSG00000223972.4";\']')
+        mock_logger.info.assert_called_with('Loading {} (genome version: 37)'.format(temp_bad_file_path))
+
+    @mock.patch('reference_data.management.commands.update_gencode.logger')
+    @mock.patch('reference_data.management.commands.update_gencode.download_file')
+    def test_update_gencode_command_url_generation(self, mock_download, mock_logger):
+        # Test the code paths of generating urls, gencode_release == 19, and --reset option
```
hanars, line 112: can you update this comment now that you moved --reset?

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,193 @@
...
+            mock.call('Creating 0 GeneInfo records'),
+            mock.call('Creating 0 TranscriptInfo records'),
+            mock.call('Done'),
+            mock.call('Stats: '),
+            mock.call('  genes_skipped: 2'),
+            mock.call('  transcripts_skipped: 2'),
+            mock.call('  genes_created: 0'),
+            mock.call('  transcripts_created: 0')
+        ]
+        mock_logger.info.assert_has_calls(calls)
```
hanars, line 193: You need to also test the db objects for this as well, reset should have deleted everything that was previously loaded

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,193 @@
...
+        self.assertDictEqual(transcript_infos['ENST00000332831'], {'start_grch37': 621059, 'end_grch37': 622053, 'strand_grch37': u'-', 'chrom_grch37': u'1', 'gene_id': u'ENSG00000284662'})
+
+        # Test normal command function without a --reset option
+        mock_logger.reset_mock()
+        call_command('update_gencode', '--gencode-release=31', self.temp_file_path, '37')
+        calls = [
+            mock.call(
+                'Loading {} (genome version: 37)'.format(self.temp_file_path)),
+            mock.call('Creating 0 GeneInfo records'),
+            mock.call('Creating 0 TranscriptInfo records'),
```
hanars, line 185: You need to test this such that db objects actually  get created, the way you did above

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Okay, change to /mock_path/...

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Okay

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: I use the reset option first and then without the option. Therefore, there are no new genes and transcripts created. Should I still test the db content? Or should I test it without the reset option first and then with a reset?

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: As above comment.

reference_data/management/tests/update_gencode_tests.py
hanars: reset deletes everything in the database and then reloads the models from the file, so there should be new records at the end of that run

hanars: Yeah, do it in the other order to make it clearer

reference_data/management/tests/update_gencode_tests.py
hanars: as above, you should do this in the order so there will be models to test. That said, in the future you can still make a db call and check that there are no models

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Updated

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Updated

reference_data/management/tests/update_gencode_tests.py
hanars: This still needs to be done

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,207 @@
...
+            'coding_region_size_grch37': gene.coding_region_size_grch37,
+            'gencode_release': gene.gencode_release,
+            'gencode_gene_type': gene.gencode_gene_type,
+            'gene_id': gene.gene_id,
+            'gene_symbol': gene.gene_symbol,
+            'end_grch37': gene.end_grch37,
+            'strand_grch37': gene.strand_grch37
+        } for gene in GeneInfo.objects.filter(gene_id__in = ['ENSG00000223972', 'ENSG00000284662'])}
+        self.assertDictEqual(gene_infos['ENSG00000223972'], {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 27, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
+        self.assertDictEqual(gene_infos['ENSG00000284662'], {'start_grch37': 621059, 'chrom_grch37': u'1', 'coding_region_size_grch37': 936, 'gencode_release': 31, 'gencode_gene_type': u'protein_coding', 'gene_id': u'ENSG00000284662', 'gene_symbol': u'OR4F16', 'end_grch37': 622053, 'strand_grch37': u'-'})
```
hanars, line 163: This is very hard to read, please do this line by line. Also, the script only created 1 gene - why are you testing 2?

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,207 @@
...
+        transcript_infos = {trans.transcript_id: {
+            'start_grch37': trans.start_grch37,
+            'end_grch37': trans.end_grch37,
+            'strand_grch37': trans.strand_grch37,
+            'chrom_grch37': trans.chrom_grch37,
+            'gene_id': trans.gene.gene_id
+        } for trans in TranscriptInfo.objects.all()}
+        self.assertEqual(len(transcript_infos), 2)
+        self.assertDictEqual(transcript_infos['ENST00000456328'], {'start_grch37': 11869, 'end_grch37': 14409, 'strand_grch37': u'+', 'chrom_grch37': u'1', 'gene_id': u'ENSG00000223972'})
+        self.assertDictEqual(transcript_infos['ENST00000332831'], {'start_grch37': 621059, 'end_grch37': 622053, 'strand_grch37': u'-', 'chrom_grch37': u'1', 'gene_id': u'ENSG00000284662'})
```
hanars, line 174: As above, please do this line by line. Also, you don't need to test all the fields for  both created transcripts - just check that there are 2 with the expected Ids and then check the fields for one of them

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Test normal command function without a --reset option
```
hanars, line 176: I think you mean "with" not "without"

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,207 @@
...
+            'gencode_release': gene.gencode_release,
+            'gencode_gene_type': gene.gencode_gene_type,
+            'gene_id': gene.gene_id,
+            'gene_symbol': gene.gene_symbol,
+            'end_grch37': gene.end_grch37,
+            'strand_grch37': gene.strand_grch37
+        } for gene in GeneInfo.objects.all()}
+        self.assertEqual(len(gene_infos), 2)
+        self.assertDictEqual(gene_infos['ENSG00000223972'], {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 31, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
+        self.assertDictEqual(gene_infos['ENSG00000284662'], {'start_grch37': 621059, 'chrom_grch37': u'1', 'coding_region_size_grch37': 936, 'gencode_release': 31, 'gencode_gene_type': u'protein_coding', 'gene_id': u'ENSG00000284662', 'gene_symbol': u'OR4F16', 'end_grch37': 622053, 'strand_grch37': u'-'})
```
hanars, line 206: As above, test line by line and only test all the fields for one entry

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Sorry. Updated this time.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: To ensure the skipped one hasn't been changed.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: I changed to define a method the prepare the dict for test. Let me know if it is okay.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Updated.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: As above.

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,220 @@
...
+            mock.call('Creating 2 GeneInfo records'),
+            mock.call('Creating 2 TranscriptInfo records'),
+            mock.call('Done'),
+            mock.call('Stats: '),
+            mock.call('  genes_created: 2'),
+            mock.call('  transcripts_created: 2')
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
+        gene_infos = {gene.gene_id: {
```
hanars, line 204: this isn't used

reference_data/management/tests/update_gencode_tests.py
```diff
@@ -0,0 +1,220 @@
...
+            'gencode_release': gene.gencode_release,
+            'gencode_gene_type': gene.gencode_gene_type,
+            'gene_id': gene.gene_id,
+            'gene_symbol': gene.gene_symbol,
+            'end_grch37': gene.end_grch37,
+            'strand_grch37': gene.strand_grch37
+        } for gene in GeneInfo.objects.all()}
+        self.assertEqual(GeneInfo.objects.all().count(), 2)
+        gene_info = _get_gene_infos(GeneInfo.objects.get(gene_id = 'ENSG00000223972'))
+        self.assertDictEqual(gene_info, {'start_grch37': 11869, 'chrom_grch37': u'1', 'coding_region_size_grch37': 0, 'gencode_release': 31, 'gencode_gene_type': u'transcribed_unprocessed_pseudogene', 'gene_id': u'ENSG00000223972', 'gene_symbol': u'DDX11L1', 'end_grch37': 14409, 'strand_grch37': u'+'})
```
hanars, line 217: Dictionaries are harder to read than line-by-line, and when something changes and they fail it is much harder to figure out what is different. If you already have a dictionary it makes sense to compare dictionaries, but it does not make sense to create 2 dictionaries just for the purpose of comparing them. Please do a line by line assertion as I asked. If you are really worried about copy-paste you could make a helper function that looks something like this, but its not neccessary:
```
def _assert_same_gene(gene, start_grch37=None, chrom_grch37=None, ...):
   self.assertEqual(gene. start_grch37, start_grch37)
   self.assertEqual(gene. chrom_grch37, chrom_grch37)
   ...
```

ShifaSZ: Okay, I can do line by line tests for some of their fields.

reference_data/management/tests/update_gencode_tests.py
ShifaSZ: Deleted.

PR #1292, opened 2020-05-15
Unit tests for reference_data management command update_omim

ShifaSZ:


reference_data/management/tests/update_omim_tests.py
```diff
@@ -0,0 +1,199 @@
...
+            mock.call('Creating 2 Omim records'),
+            mock.call('Done'),
+            mock.call('Loaded 2 Omim records from {}. Skipped 2 records with unrecognized genes.'.format(self.temp_file_path)),
+            mock.call('Running ./manage.py update_gencode to update the gencode version might fix missing genes')
+        ]
+        mock_logger.info.assert_has_calls(calls)
+
+        self.assertEqual(Omim.objects.all().count(), 2)
+        record = Omim.objects.get(gene__gene_symbol = 'OR4F5')
+        self.assertDictEqual({
```
hanars, line 181: This is pretty had to read, just do it as a line by line assertion
```
self.assertEqual(record.comments, 'linkage with rs1780324')
self.assertEqual(record. gene_description, Alkaline phosphatase, plasma level of, QTL 2')
...
```

reference_data/management/tests/update_omim_tests.py
```diff
@@ -0,0 +1,199 @@
...
+        mock_download.assert_not_called()
+        calls = [
+            mock.call('Deleting 2 existing Omim records'),
+            mock.call('Parsing file'),
+            mock.call('Creating 2 Omim records'),
+            mock.call('Done'),
+            mock.call('Loaded 2 Omim records from {}. Skipped 2 records with unrecognized genes.'.format(self.temp_file_path)),
+            mock.call('Running ./manage.py update_gencode to update the gencode version might fix missing genes')
+        ]
+        mock_logger.info.assert_has_calls(calls)
```
hanars, line 177: Why doesn't the mock logger have the logger calls about adding the phenotypic series?

reference_data/management/tests/update_omim_tests.py
ShifaSZ: Okay

reference_data/management/tests/update_omim_tests.py
ShifaSZ: Because I mock "reference_data.management.commands.utils.update_utils.logger" but not "reference_data.management.commands.update_omim.logger". It doesn't work if I mock the logging library. But I will try again to find a way make the logger works for both module at the same time.

ShifaSZ: They are different loggers with different module names. So I'm going to mock both of them.

PR #1299, opened 2020-05-20
Added unit tests for download_file method.

ShifaSZ:
Nodes for review:
1. The `tempfile` in the parameter `def download_file(url, to_dir=tempfile.gettempdir(), verbose=True):` can't be mocked because the default value of the parameter is generated before mocking being run. So I leave it un-mocked.
2. The exception in line #47 of `reference_data/management/commands/utils/download_utils.py` is un-reachable.
3. It seems that `responses` cannot mock the `urllib`. So the `urllib.urlopen()` is mocked separately.

### DIFF
--- a/reference_data/management/commands/utils/download_utils.py
+++ b/reference_data/management/commands/utils/download_utils.py
@@ -19,7 +19,7 @@ def download_file(url, to_dir=tempfile.gettempdir(), verbose=True):
     if not (url and url.startswith(("http://", "https://", "ftp://"))):
         raise ValueError("Invalid url: {}".format(url))
     local_file_path = os.path.join(to_dir, os.path.basename(url))
-    remote_file_size = get_remote_file_size(url)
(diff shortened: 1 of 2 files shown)


hanars:
So technically speaking, that "unreachable" line of code would be reachable if you called `get_remote_file_size` directly, although you are correct that it is impossible to hit that exception from within the call in `download_file`. Given that `get_remote_file_size` is never called elsewhere, what I would do is:
- Change the name of the function to start with a _ (i.e. `_get_remote_file_size`) which is the pythonic way of indicating that function is private to the file and shouldn't be imported elsewhere
- Change the function so there is just the first `if` clause to get the content length header, and then the function falls back to 0 with a comment about other methods not being implemented

hanars:
Friendly reminder that we should get this merged

hanars:
There were open comments and no approvals on the is PR. It was absolutely not okay to merge it

PR #1301, opened 2020-05-20
Set KIBANA_SERVICE_HOSTNAME in docker-compose.yml

rsolano1-uw:
Resolves https://github.com/macarthur-lab/seqr/issues/1300


bw2:
Thanks for troubleshooting this. 
I think the way to fix this is to add 
- KIBANA_SERVICE_HOSTNAME=kibana 
to
https://github.com/macarthur-lab/seqr/blob/master/docker-compose.yml#L80-L83

hanars:
I agree with @bw2 - I think the better solution here would be to modify the docker config, rather than modifying the settings defaults. Thanks so much for putting in the work to fix this issue!

rsolano1-uw:
Hey all,
I have updated the PR to take the environment variable approach.

PR #1302, opened 2020-05-22
Unit test permissions

hanars:
seqr has 3 primary types of user access: staff users with access to everything, manager users with full access to a subset of projects, and collaborator users who have read-only access to a subset of projects. However, due to limitation with how django creates test database groups and permissions it is impossible to create users with manager or collaborator access at the fixture level, and therefore we never tested that these access restrictions were enforced, we just used staff access or for non staff users we made them superusers. This means that there were some untestable code paths, and also that we have no test enforcement that these crucial access restrictions were actually being enforced.

This PR changes how test cases are logged in, by creating users with all 3 levels of access and requiring each test to specify what the minimum level of access for a url is and then testing that those without access actually don't have it

I also refactored many of the permission-checking utility functions to improve test coverage


seqr/views/utils/test_utils.py
```diff
@@ -1,25 +1,90 @@
...
 
-def login_non_staff_user(test_case):
-    test_user = User.objects.get(username='test_user_non_staff')
-    test_case.client.force_login(test_user)
+        edit_group = Group.objects.get(pk=2)
+        view_group = Group.objects.get(pk=3)
+        edit_group.user_set.add(cls.manager_user)
+        view_group.user_set.add(cls.manager_user, cls.collaborator_user)
+        assign_perm(user_or_group=edit_group, perm=CAN_EDIT, obj=Project.objects.all())
+        assign_perm(user_or_group=view_group, perm=CAN_VIEW, obj=Project.objects.all())
```
hanars, line 28: Due to limitations on how django loads built in permissions and groups, its impossible to create these permissions in the fixtures so they need to be created here at runtime

seqr/fixtures/users.json
```diff
@@ -29,7 +29,7 @@
     "fields": {
         "password": "pbkdf2_sha256$30000$y85kZgvhQ539$jrEC3L1IhCezUx3Itp+14w/T7U6u5XUxtpBZXKv8eh4=",
         "last_login": null,
-        "is_superuser": true,
+        "is_superuser": false,
```
ShifaSZ, line 32: It is not clear what's "is_superuser" for?

seqr/views/utils/test_utils.py
```diff
@@ -1,25 +1,90 @@
...
+            test_case (object): the django.TestCase or unittest.TestCase object
+            url (string): The url of the django view being tested.
+            permission_level (string): what level of permission this url requires
+         """
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 302)  # check that it redirects if you don't login
+
+        self.client.force_login(self.no_access_user)
+        if permission_level == self.ANY:
+            return
```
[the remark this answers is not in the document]
hanars: So "any" means "any authenticated user can access" which is different than just "anyone can access". There are several views in seqr that you can access as long as you are a seqr user but can't access if you aren;' logged in, which is what this test is for. Would changing this to `ANY_USER` make it more clear?

seqr/fixtures/users.json
hanars: `is_superuser` used to override all permissions and grant access to everything. We were using it as a workaround for tests because you can grant superuser access here in the fixture unlike real collaborator/manager permissions and then the test user can access everything being tested. But in real life seqr no one is a superuser so its not really testing our authentication well, so I took out all the code that grants superusers special access and then I also changed the fixture data here since it is no longer used

seqr/views/utils/test_utils.py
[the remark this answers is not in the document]
ShifaSZ: I prefer `LOGIN_USER` or `AUTHENTICATED_USER`.

seqr/fixtures/users.json
ShifaSZ: It makes sense.

PR #1303, opened 2020-05-22
Updated dataset apis to be Py2/3 compatible.

ShifaSZ:
Main changes:
1. Handle unicode, use `from __future__ import unicode_literals` to change all string literals to be unicode. Use explicit .encode() and .decode() for IOs.
2. Updated parse_file() to support both binary and text streams.
3. Use list() to convert .items(), .values(), .keys() into lists. But, keep views of .items()/.values()/.keys() unchanged. 
4. Use exception.args[0] to replace exception.message

To make the ut can run with Python3.7, below files are updated:
1. `seqr/migrations/0001_squashed_0067_remove_project_custom_reference_populations.py`: Removed the b from the argument of models.ManyToManyField(to=b'seqr.xxx')
2. `matchmaker/migrations/0001_initial.py`: Used list() to convert .keys() to a list.
3. `seqr/views/utils/export_utils.py`:  This file is actually used by dataset_api. Updated the IO encoding.
4. `requirement.txt`: Added future package. 
5. `constants.py` and `es_search.py` under `seqr/utils/elasticsearch`
6. All lambda functions with a tuple as argument also updated because they cause syntax error under Py3.7.


hanars:
Some high level comments:

- Most of the places you do an explicit cast for dict keys/values/items to a list I think it is unnecessary.  From https://portingguide.readthedocs.io/en/latest/dicts.html:
```
The following common operations work the same between views and lists,
as long as the underlying dict is not modified:

Iteration (e.g. for x in d.values())
Member testing (e.g. if x in d.values())
Length testing (e.g. len(d.values()))
```
- For exception messages, I think `str(e)` will work the same and be more readable that the `e.args[0]` syntax you were using

seqr/views/apis/dataset_api_tests.py
```diff
@@ -190,12 +192,11 @@ def test_add_variants_dataset(self, mock_es_search, mock_get_index_metadata, moc
...
         self.assertDictEqual(response_json['familiesByGuid'], {})
-        self.assertListEqual(response_json['samplesByGuid'].keys(), [sv_sample_guid])
+        self.assertListEqual(list(response_json['samplesByGuid'].keys()), [sv_sample_guid])
         self.assertEqual(response_json['samplesByGuid'][sv_sample_guid]['datasetType'], 'SV')
         self.assertTrue(response_json['samplesByGuid'][sv_sample_guid]['isActive'])
-        self.assertDictEqual(response_json['individualsByGuid'], {
-            'I000001_na19675': {'sampleGuids': [sv_sample_guid, existing_index_sample_guid]},
-        })
+        self.assertIn(sv_sample_guid, response_json['individualsByGuid']['I000001_na19675']['sampleGuids'])
+        self.assertIn(existing_index_sample_guid, response_json['individualsByGuid']['I000001_na19675']['sampleGuids'])
```
hanars, line 199: So this somewhat changes what is tested, because before it was testing a) that exactly one individual is returned, b) that the only key returned for the individual is `sampleGuids` and c) that there are exactly 2 sample guids. This updated test (which I assume is because the order of those guids isn't stable) would pass if something changed and for whatever reason unwanted additional samples or individuals were returned or additional individual fields were included, but we want it to fail if that changes. What I would recommend to continue to have the same level of testing without relying on a list order is:
```
self.assertListEqual(list(individualsByGuid['individualsByGuid'].keys()), ['I000001_na19675'])
self.assertListEqual(list(individualsByGuid['individualsByGuid']['I000001_na19675'].keys()), ['sampleGuids'])
self.assertSetEqual(set(response_json['individualsByGuid']['I000001_na19675']['sampleGuids']), {sv_sample_guid, existing_index_sample_guid})
```

seqr/views/apis/dataset_api_tests.py
```diff
@@ -280,11 +281,11 @@ def test_add_alignment_sample(self, mock_local_file_exists, mock_google_bucket_f
 
         self.assertSetEqual(set(response_json.keys()), {'igvSamplesByGuid', 'individualsByGuid'})
         self.assertEqual(len(response_json['igvSamplesByGuid']), 1)
-        sample_guid = response_json['igvSamplesByGuid'].keys()[0]
+        sample_guid = list(response_json['igvSamplesByGuid'].keys())[0]
```
hanars, line 284: while this works, I think the more pythonic way to do this is `sample_guid = next(guid for guid in response_json['igvSamplesByGuid'].keys())`

seqr/views/utils/export_utils.py
```diff
@@ -96,9 +96,9 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
             for filename, header, rows in files:
                 header_display = header
                 if add_header_prefix:
-                    header_display = map(
-                        lambda (i, k): '{}-{}'.format(str(i).zfill(2), k), enumerate(header)
-                    )
+                    header_display = list(map(
+                        lambda header_tuple: '{}-{}'.format(str(header_tuple[0]).zfill(2), header_tuple[1]), enumerate(header)
+                    ))
```
hanars, line 101: I think `list(map(...))` is less readable than a list comprehension (and honestly I'm not sure why we were using `map` in the first place) so I would change this to 
```
header_display = [{}-{}'.format(str(header_tuple[0]).zfill(2), header_tuple[1]) for header_tuple in enumerate(header)]
```

seqr/views/utils/file_utils.py
```diff
@@ -7,6 +9,7 @@
 import tempfile
 import openpyxl as xl
 
+from django.core.files.uploadedfile import InMemoryUploadedFile
```
hanars, line 12: This seems like something that is only added for testing purposes. I STRONGLY disagree with having conditional logic designed for that purpose, it makes the real code more clunky and it means we aren't actually testing what we want to be testing. If this is an issue with how the mock file isbeing passed, try using `BytesIO` as your test input instead of a utf-8 encoded string

seqr/views/utils/file_utils.py
```diff
@@ -86,25 +94,25 @@ def save_uploaded_file(request, process_records=None):
         raise ValueError("Received %s files instead of 1" % len(request.FILES))
 
     # parse file
-    stream = request.FILES.values()[0]
+    stream = list(request.FILES.values())[0]
```
hanars, line 97: similar to above, I think using `next` to get the first value is better than converting it to a list and getting the first element

seqr/views/utils/file_utils.py
```diff
@@ -86,25 +94,25 @@ def save_uploaded_file(request, process_records=None):
...
     filename = stream._name
 
     json_records = parse_file(filename, stream)
     if process_records:
         json_records = process_records(json_records, filename=filename)
 
     # save json to temporary file
-    uploaded_file_id = hashlib.md5(str(json_records)).hexdigest()
+    uploaded_file_id = hashlib.md5(str(json_records).encode('utf-8')).hexdigest()
```
hanars, line 105: I don't have any problem with this, but I'm curious why it is neccessary

seqr/views/utils/file_utils.py
```diff
@@ -86,25 +94,25 @@ def save_uploaded_file(request, process_records=None):
...
+    with gzip.open(serialized_file_path, "wt") as f:
         json.dump(json_records, f)
 
     return uploaded_file_id, filename, json_records
 
 
 def load_uploaded_file(upload_file_id):
     serialized_file_path = _compute_serialized_file_path(upload_file_id)
-    with gzip.open(serialized_file_path) as f:
+    with gzip.open(serialized_file_path, "rt") as f:
```
hanars, line 115: `t` is the default so why change it?

seqr/views/apis/dataset_api_tests.py
ShifaSZ: The change was because of the list order issue. Agree with your points. Update as your recommendation.

seqr/views/apis/dataset_api_tests.py
ShifaSZ: Hmm..., I try using next() + iter(). Is it okay?

seqr/views/utils/export_utils.py
ShifaSZ: Okay

seqr/views/apis/dataset_api_tests.py
hanars: Well theres no need to explicitly call `iter()`, the code I showed here implicitly calls it, but yes thats the idea

seqr/views/utils/file_utils.py
ShifaSZ: I agree that using isinstance(InMemoryUploadedFile) is not elegant. The issue is parse_file() can have an either binary (which is a InMemoryUploadedFile) or text stream inputs. I finally find a way to convert the binary input into a text input by using io.TextIOWrapper.

seqr/views/utils/file_utils.py
ShifaSZ: Okay

seqr/views/utils/file_utils.py
ShifaSZ: It is because `json_records` contains unicode and hash function requires unicode must be encoded. There is no issue before maybe because `json_records` contains ascii only.

seqr/views/utils/file_utils.py
ShifaSZ: The open modes of gzip are:
```
binary mode: 'r', 'rb', 'a', 'ab', 'w', 'wb', 'x', 'xb' 
text mode : 'rt', 'at', 'wt', or 'xt'
```
So, 'r' is a binary mode.

hanars: Where are you seeing that? In the documentation I see the `b` flag specifies binary, the `t` flag specifies text, and text is the default so if neither is provided it is text mode: https://docs.python.org/3/library/functions.html#open

seqr/views/utils/file_utils.py
hanars: But why are we supporting binary files at all? Is there any "real" use case with binary, or is it testing only? If testing only, you should change how we are testing, not the real code

seqr/views/utils/file_utils.py
ShifaSZ: The document you are looking at is for `open`. But in this case, it is about `gzip.open`.

seqr/views/utils/file_utils.py
ShifaSZ: It is not for testing only. I thought all the input data must be in text format. That was true if we don't support `.xls` files.
Since we need to support both text and binary data, the text decoding has to be done inside the parse_file() function. I will figure out how to do it and ask you review later.

seqr/views/apis/dataset_api_tests.py
ShifaSZ: `next(iter(response_json['igvSamplesByGuid']))` looks more concise. I'm going to use it.

hanars:
can you update this to resolve merge conflicts?

requirements.txt
```diff
@@ -20,3 +21,4 @@ slacker==0.13.0                  # library for sending slack messages
 slugify==0.0.1                   # used for encoding names for guids
 tqdm==4.40.2                     # convenient way to create progress bar for long-running command-line operations
 whitenoise==3.3.0                # simplified static file handling. Behind a major version due to missing Python 2 support
+csv23==0.3.2                     # provides the unicode-based API of the Python 3 csv module for Python 2 and 3.
```
ShifaSZ, line 24: `seqr/views/utils/file_utils.parse_file` uses `csv.read()` to process the csv stream. When there are unicodes in the stream, it is not possible to have a Python 2 and 3 compatible code. The csv23 package solves this issue.

seqr/utils/file_utils.py
```diff
@@ -35,7 +35,7 @@ def file_iter(file_path, byte_range=None):
         for line in _google_bucket_file_iter(file_path, byte_range=byte_range):
             yield line
     else:
-        with open(file_path) as f:
+        with open(file_path, 'rb') as f:
```
ShifaSZ, line 38: `'rb'` is necessary because when the file contents unicodes the Python 2 alway read out `str` data without decoding. It will cause errors when try to encode these `str` data back to bytes.
The issue can be solved by reading with binary mode and decoding them when the data are determined to be text.

seqr/views/apis/dataset_api_2_3_tests.py
```diff
@@ -126,7 +132,7 @@ def test_add_variants_dataset(self, mock_es_search, mock_get_index_metadata, moc
         mock_es_search.return_value.params.return_value.execute.return_value.aggregations.sample_ids.buckets = [
             {'key': 'NA19675'}, {'key': 'NA19679'}, {'key': 'NA19678_1'},
         ]
-        mock_file_iter.return_value = ['NA19678_1,NA19678']
+        mock_file_iter.return_value = BytesIO('NA19678_1,NA19678\n'.encode('utf-8'))
```
ShifaSZ, line 135: Use `BytesIO` here to make the `mock_file_iter` returns the same data type as the `file_iter()`.

seqr/utils/middleware.py
```diff
@@ -20,7 +20,7 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': str(exception.message)}
+            exception_json = {'message': str(exception)}
```
ShifaSZ, line 23: Removing `.message` is necessary for Python3.

seqr/views/apis/dataset_api.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import unicode_literals
+from builtins import str
```
ShifaSZ, line 2: `from builtins import str` is necessary when there are unicodes in the object standard Python 2 `str` will fail to convert the object to a string (`unicode` is required).

seqr/views/apis/dataset_api_2_3_tests.py
```diff
@@ -190,12 +197,12 @@ def test_add_variants_dataset(self, mock_es_search, mock_get_index_metadata, moc
...
-        self.assertListEqual(response_json['samplesByGuid'].keys(), [sv_sample_guid])
+        self.assertListEqual(list(response_json['samplesByGuid'].keys()), [sv_sample_guid])
         self.assertEqual(response_json['samplesByGuid'][sv_sample_guid]['datasetType'], 'SV')
         self.assertTrue(response_json['samplesByGuid'][sv_sample_guid]['isActive'])
-        self.assertDictEqual(response_json['individualsByGuid'], {
-            'I000001_na19675': {'sampleGuids': [sv_sample_guid, existing_index_sample_guid]},
-        })
+        self.assertSetEqual(set(response_json['individualsByGuid']['I000001_na19675']['sampleGuids']),
+                            set([sv_sample_guid, existing_index_sample_guid]))
+
```
ShifaSZ, line 205: Eliminated the occasional array element order issue.

seqr/views/apis/dataset_api_2_3_tests.py
```diff
@@ -226,9 +233,32 @@ def test_receive_alignment_table_handler(self):
...
+        file = tempfile.TemporaryFile()
+        wb.save(file)
+        file.seek(0)
+        data = file.read()
+        file.close()
+        f = SimpleUploadedFile('samples.xlsx', data)
+        response = self.client.post(url, data={'f': f})
+        self.assertEqual(response.status_code, 200)
+        self.assertListEqual(response.json()['info'], ['Parsed 2 rows from samples.xlsx'])
+
```
ShifaSZ, line 261: Added tests for uploading `.tsv` and `.xlsx` files.

seqr/views/utils/file_utils.py
```diff
@@ -86,25 +91,32 @@ def save_uploaded_file(request, process_records=None):
...
+    stream = next(iter(request.FILES.values()))
     filename = stream._name
 
+    if filename.endswith('.csv'):
+        try:
+            a = unicode
+        except:
+            stream = TextIOWrapper(stream.file, encoding = 'utf-8')
+    elif not filename.endswith('.xls') and not filename.endswith('.xlsx'):
+        stream = TextIOWrapper(stream.file, encoding = 'utf-8')
```
ShifaSZ, line 103: This a little ugly solution: if `unicode` is defined then do nothing for `.csv` file, otherwise wrap all non-excel files with `TextIOWrapper`.
We can't do the same thing as `file_iter()` data because uploaded data (`bytes`) can only be converted to `unicode` for Python 2. So we have to keep the data in `bytes` type otherwise they won't work with `csv.reader`.

seqr/views/utils/file_utils.py
```diff
@@ -86,25 +91,32 @@ def save_uploaded_file(request, process_records=None):
...
     json_records = parse_file(filename, stream)
     if process_records:
         json_records = process_records(json_records, filename=filename)
 
     # save json to temporary file
-    uploaded_file_id = hashlib.md5(str(json_records)).hexdigest()
+    uploaded_file_id = hashlib.md5(str(json_records).encode('utf-8')).hexdigest()
     serialized_file_path = _compute_serialized_file_path(uploaded_file_id)
-    with gzip.open(serialized_file_path, "w") as f:
+    with gzip.open(serialized_file_path, "wt") as f:
```
ShifaSZ, line 111: The change from `w` to `wt` is required by Python 3 otherwise it can only write binary data while json.dump() generates text data.

seqr/views/apis/dataset_api_2_3_tests.py
hanars: Can you also assert that the key in `'individualsByGuid'` is `'I000001_na19675'` and the only key in the individual json is `'sampleGuids'`

hanars: Also `{sv_sample_guid, existing_index_sample_guid}` is cleaner than `set([sv_sample_guid, existing_index_sample_guid])`

seqr/views/apis/dataset_api_2_3_tests.py
hanars: I don't think this is a good place for adding a test for xls file parsing. I think you should do that in the explicit tests for `parse_file` in `file_utils_tests` (which already exist). You can keep the tsv if you want because its relatively simple, although its not necessary

seqr/views/utils/file_utils.py
hanars: this should be done in `parse_file`, not here

seqr/views/utils/file_utils.py
```diff
@@ -37,7 +42,7 @@ def save_temp_file(request):
 
 def parse_file(filename, stream):
     if filename.endswith('.tsv') or filename.endswith('.fam') or filename.endswith('.ped'):
-        return [map(lambda s: s.strip().strip('"'), line.rstrip('\n').split('\t')) for line in stream]
+        return [[s.strip().strip('"') for s in line.rstrip('\n').split('\t')] for line in stream]
 
     elif filename.endswith('.csv'):
         return [row for row in csv.reader(stream)]
```
hanars, line 48: this is where you would do the `TextIOWrapper ` conversion

seqr/views/utils/file_utils.py
hanars: also, when you move it this is a really messy way to do it. what does `a = unicode` even do? `a` is never used and how would that assignment ever throw an exception?

ShifaSZ: `a = unicode` is for testing if `unicode` is defined or not. Maybe a better way is to use `unicode = unicode` or just `unicode`.

seqr/views/utils/file_utils.py
ShifaSZ: I'm not sure it can be done here or not because TextIOWrapper wraps BytesIO and it seems difficult to know if the input stream is a BytesIO or not. But, I will have a try to move the TextIOWrapper.

hanars: if it sis a stream and you wrap it in `TextIOWrapper` that should be fin. And if it is a string and you wrap it in `TextIOWrapper` that should also be fine. So i'm not sure what the issue is

seqr/views/apis/dataset_api_2_3_tests.py
ShifaSZ: Remove these tests. Adding them to the tests for file_utils.py.

seqr/views/apis/dataset_api_2_3_tests.py
hanars: this is still an issue

seqr/views/apis/dataset_api_2_3_tests.py
hanars: yes I agree

seqr/views/apis/dataset_api_2_3_tests.py
ShifaSZ: Fixed now.

seqr/views/apis/dataset_api_2_3_tests.py
```diff
@@ -1,10 +1,17 @@
...
 import mock
 import subprocess
 from datetime import datetime
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.urls.base import reverse
 
+from io import StringIO
+
+import openpyxl as xl
+import tempfile
```
hanars, line 13: these imports are unused

ShifaSZ: Delete `tempfile` and `openpyxl`

PR #1305, opened 2020-05-27
Update some necessary util files to enable running the unit test for dataset api with Python3

ShifaSZ:
Changes:
1. Wrapped keys/items/values as lists with list()
2. Removed the `b` from the argument of `models.ManyToManyField(to=b'seqr.xxx')` in `seqr/migrations/0001_squashed_0067_remove_project_custom_reference_populations.py`
3. Updated `seqr/views/utils/export_utils.py` and its unit tests for IO data encoding.
4. Added `future` package to `requirement-dev.txt`
5. `constants.py` and `es_search.py` under `seqr/utils/elasticsearch` are also updated in this PR because they are imported by `seqr/utils/elasticsearch/utils.py` which is imported by `seqr/views/utils/dataset_utils.py` and then dataset_api.

### DIFF
--- a/matchmaker/views/external_api.py
+++ b/matchmaker/views/external_api.py
@@ -161,7 +161,7 @@ matchbox on {insertion_date}, with seqr link
             if email.strip() != MME_DEFAULT_CONTACT_EMAIL]
         all_emails.update(emails)
         match_results.append((result_text, emails))
-    match_results = sorted(match_results, key=lambda result_text__: result_text__[0])
+    match_results = sorted(match_results, key=lambda result_tuple: result_tuple[0])
 
     base_message = """Dear collaborators,
 
--- a/matchmaker/views/external_api.py
+++ b/matchmaker/views/external_api.py
@@ -1,3 +1,4 @@
+from __future__ import unicode_literals
 import json
 import logging
 from django.core.mail.message import EmailMessage
@@ -149,7 +150,7 @@ def _generate_notification_for_incoming_match(results, incoming_query, incoming_
         individual = submission.individual
         project = individual.family.project
 
-        result_text = u"""seqr ID {individual_id} from project {project_name} in family {family_id} inserted into
+        result_text = """seqr ID {individual_id} from project {project_name} in family {family_id} inserted into
 matchbox on {insertion_date}, with seqr link
 {host}project/{project_guid}/family_page/{family_guid}/matchmaker_exchange""".replace('\n', ' ').format(
             individual_id=individual.individual_id, project_guid=project.guid, project_name=project.name,
@@ -160,9 +161,9 @@ matchbox on {insertion_date}, with seqr link
             if email.strip() != MME_DEFAULT_CONTACT_EMAIL]
         all_emails.update(emails)
         match_results.append((result_text, emails))
-    match_results = sorted(match_results, key=lambda (result_text, _): result_text)
+    match_results = sorted(match_results, key=lambda result_text__: result_text__[0])
 
-    base_message = u"""Dear collaborators,
+    base_message = """Dear collaborators,
 
     matchbox found a match between a patient from {query_institution} and the following {number_of_results} case(s) 
     in matchbox. The following information was included with the query,
@@ -183,7 +184,7 @@ matchbox on {insertion_date}, with seqr link
         incoming_query_contact_name=incoming_patient['patient']['contact'].get('name', '(sorry I was not able to read the information given for name'),
     )
 
-    message_template = u"""{base_message}{match_results}
+    message_template = """{base_message}{match_results}
 
     We sent this email alert to: {email_addresses_alert_sent_to}\n{footer}."""
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@ coverage==5.1
 django-compressor==2.1.1
 django-debug-toolbar==1.11                      # https://github.com/jazzband/django-debug-toolbar
 django-debug-toolbar-request-history==0.0.11    # for debugging database queries on async requests.
+future==0.17.1                                  # for Python2 and 3 compatibility
 mock                                            # mock objects for unit tests
 responses==0.10.6                               # mock HTTP responses for unit tests
 Jinja2==2.10.1                                  # jinja templates are used by servctl k8s deployment scripts
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,7 +2,6 @@ coverage==5.1
 django-compressor==2.1.1
 django-debug-toolbar==1.11                      # https://github.com/jazzband/django-debug-toolbar
 django-debug-toolbar-request-history==0.0.11    # for debugging database queries on async requests.
-future==0.17.1                                  # for Python2 and 3 compatibility
 mock                                            # mock objects for unit tests
 responses==0.10.6                               # mock HTTP responses for unit tests
 Jinja2==2.10.1                                  # jinja templates are used by servctl k8s deployment scripts
--- a/seqr/utils/elasticsearch/constants.py
+++ b/seqr/utils/elasticsearch/constants.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from reference_data.models import Omim, GeneConstraint
 from seqr.models import Individual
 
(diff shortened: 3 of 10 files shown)


seqr/views/utils/export_utils_tests.py
```diff
@@ -43,21 +43,21 @@ def test_export_table(self):
...
 
         # test csv format
-        response = export_multiple_files([['file1', header1, rows], ['file2', header2, rows]], 'zipfile')
+        response = export_multiple_files([[u'file1\xe3', header1, rows], ['file2', header2, rows]], 'zipfile')
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.get('content-disposition'), 'attachment; filename="zipfile.zip"')
         self.assertDictEqual(mock_zip_content, {
-            'file1.csv': 'col1,col2\nrow1_v1,row1_v2\nrow2_v1,',
-            'file2.csv': 'col1\nrow1_v1\nrow2_v1',
+            u'file1\xe3.csv': u'col1,col2\nrow1_v1\xe2,row1_v2\nrow2_v1,',
```
hanars, line 59: instead of prefixing all the test file strings with `u`, why not just use the future unicode import in this file?

matchmaker/views/external_api.py
```diff
@@ -160,9 +161,9 @@ def _generate_notification_for_incoming_match(results, incoming_query, incoming_
             if email.strip() != MME_DEFAULT_CONTACT_EMAIL]
         all_emails.update(emails)
         match_results.append((result_text, emails))
-    match_results = sorted(match_results, key=lambda (result_text, _): result_text)
+    match_results = sorted(match_results, key=lambda result_text__: result_text__[0])
```
hanars, line 164: Can you not destructure tuple args in python 3? That seems like a weird thing to get rid if. If we do need to change this, the variable name `result_text__` is a bit confusing. Maybe just use `result_tuple` or something like that?

requirements-dev.txt
```diff
@@ -2,6 +2,7 @@ coverage==5.1
 django-compressor==2.1.1
 django-debug-toolbar==1.11                      # https://github.com/jazzband/django-debug-toolbar
 django-debug-toolbar-request-history==0.0.11    # for debugging database queries on async requests.
+future==0.17.1                                  # for Python2 and 3 compatibility
```
hanars, line 5: you import `future` in `external_api.py` which is a real code file not a test file, so this needs to go in the main `requirements.txt` file

seqr/utils/elasticsearch/constants.py
```diff
@@ -246,16 +248,16 @@
 CORE_FIELDS_CONFIG = {
     'alt': {},
     'contig': {'response_key': 'chrom'},
-    'end': {'format_value': long},
+    'end': {'format_value': int},
```
hanars, line 251: this will absolutely break important seqr code. The values for variant start and end coordinates are larger than ints (at least in python 2). You need a different fix for this, although again I don't understand why this file is a blocker for the dataset tests?

hanars:
Many of these changes seem outside the scope of getting the dataset_api tests to pass. I understand why the migrations need to be python 3 compatible, but why matchmaker and elasticsearch code? Are you sure you are only running the dataset tests?

seqr/views/utils/export_utils.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import unicode_literals
+from builtins import str
```
hanars, line 2: you should never need to import builtins, they should already be in scope

seqr/views/utils/export_utils.py
```diff
@@ -52,21 +55,21 @@ def export_table(filename_prefix, header, rows, file_format, titlecase_header=Tr
         response = HttpResponse(content_type='text/tsv')
         response['Content-Disposition'] = 'attachment; filename="{}.tsv"'.format(filename_prefix)
         response.writelines(['\t'.join(header)+'\n'])
-        response.writelines(('\t'.join(map(unicode, row))+'\n' for row in rows))
+        response.writelines('\t'.join(list(map(str, row)))+'\n' for row in rows)
```
hanars, line 58: we explicitly had code to make sure you could export unicode and this takes out that functionality and restricts it to strings. Why? With unicode literals we could just take out the `map` entirely

seqr/views/utils/export_utils.py
```diff
@@ -52,21 +55,21 @@ def export_table(filename_prefix, header, rows, file_format, titlecase_header=Tr
...
         return response
     elif file_format == "json":
         response = HttpResponse(content_type='application/json')
         response['Content-Disposition'] = 'attachment; filename="{}.json"'.format(filename_prefix)
         for row in rows:
-            json_keys = map(lambda s: s.replace(" ", "_").lower(), header)
-            json_values = map(unicode, row)
-            response.write(json.dumps(OrderedDict(zip(json_keys, json_values)))+'\n')
+            json_keys = [s.replace(" ", "_").lower() for s in header]
+            json_values = row
```
hanars, line 65: theres no reason for this line of code, just change the line below it to use `rows` instead of `json_values`

seqr/views/utils/export_utils.py
```diff
@@ -96,17 +99,19 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
             for filename, header, rows in files:
                 header_display = header
                 if add_header_prefix:
-                    header_display = map(
-                        lambda (i, k): '{}-{}'.format(str(i).zfill(2), k), enumerate(header)
-                    )
+                    header_display = ['{}-{}'.format(str(i_k[0]).zfill(2), i_k[1]) for i_k in enumerate(header)]
```
hanars, line 102: `i_k` is a weird variable name, maybe do something descriptive like `indexed_column`

seqr/views/utils/export_utils.py
```diff
@@ -96,17 +99,19 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
...
                     header_display[0] = header[0]
                 content = DELIMITERS[file_format].join(header_display) + '\n'
                 content += '\n'.join([
                     DELIMITERS[file_format].join([row.get(key) or blank_value for key in header]) for row in rows
                 ])
-                if not isinstance(content, unicode):
-                    content = unicode(content, errors='ignore')
-                zip_file.writestr('{}.{}'.format(filename, file_format), content)
+                if not isinstance(content, str):
```
hanars, line 108: We absolutely need to support unicode here and added this code to make sure unicode works. Why would you take out unicode support and restrict it to strings?

seqr/views/utils/export_utils_tests.py
ShifaSZ: I didn't convert this file to Python2 and 3 compatible yet. But I AM able to use the unicode_literals to covert all the str literals to unicode before doing other converting.

ShifaSZ: I'm going to convert the whole file to be Py2/3 compatible after import unicode_literals. The only difference is to replace `self.assertRaisesRegexp()` with `self.assertRaisesRegex()`.

matchmaker/views/external_api.py
ShifaSZ: Python 3 doesn't support interpret the members of the tuple args. The only way is to use the index to access the tuple members. I can change the name to `result_tuple`.

requirements-dev.txt
ShifaSZ: Moved it to the requirements.txt

seqr/utils/elasticsearch/constants.py
ShifaSZ: I will add `from builtins import int` which will turn all `int` to be same as `long`.

ShifaSZ: The file is imported by the `seqr/utils/elasticsearch/utils.py` and `seqr/views/utils/data_utils.py`.

seqr/views/utils/export_utils.py
ShifaSZ: The `builtins` is part of the future package. It makes type `str` to become unicode.

seqr/views/utils/export_utils.py
ShifaSZ: The `row` is a list of some different types include `str`, `int` and maybe others. The `map()` converts all the members (especially non-str members) in the list into `str`.
The `str` here is imported from `builtins` which means it behaves the same as the `str` in Python 3 and it is unicode only.

seqr/views/utils/export_utils.py
ShifaSZ: Similar to the above comment, I have to add the map back to convert non-str members to str.

seqr/views/utils/export_utils.py
ShifaSZ: Okay. Change it to `index_value`.

hanars:
Closing this as the fixes for the unit tests are now directly in the main PR branch (https://github.com/macarthur-lab/seqr/pull/1303)

PR #1306, opened 2020-05-27
Unit test coverage

hanars:
Increases test coverage by:
- Adding unit tests
- Moving a complex function only used by deployment code (which we do not need to teest) out of the main seqr folder
- Updating mocks to mock lower level functions rather than utility functions to increase coverage for those functions


deploy/servctl_utils/shell_utils.py
```diff
@@ -31,3 +33,79 @@ def wait_for(procs):
 
     for proc in procs:
         proc.wait()
+
+
+def run(command,
```
hanars, line 38: this is directly copy/pasted from `seqr/utils/shell_utils.py` so I wouldn't worry about reviewing it

seqr/views/utils/file_utils_tests.py
```diff
@@ -0,0 +1,110 @@
...
+            'uploadedFileId': mock.ANY,
+        })
+
+    @mock.patch('seqr.views.utils.file_utils.xl.load_workbook')
+    def test_parse_file(self, mock_load_xl):
+        mock_load_xl.return_value.sheetnames = ['sheet1']
+        mock_load_xl.return_value.__getitem__.return_value = MOCK_EXCEL_SHEET
+
+        for ext, data in TEST_DATA_TYPES.items():
+            self.assertListEqual(parse_file('test.{}'.format(ext), BytesIO(data)), PARSED_DATA)
```
ShifaSZ, line 106: I prefer to use a `StringIO` as the argument of `parse_file()` because it will get unicode return values from `parse_file()`. As a reminder, the argument of `StringIO()` must be a unicode.

hanars: This causes many errors with unicode vs string encoding. I am going to leave it as is for now, and once we switch that file over to being python 3 compatible we can update this test

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1232,10 +1254,14 @@ def test_get_single_es_variant(self):
...
         self.assertDictEqual(variant, all_family_variant)
         self.assertExecutedSearch(
-            filters=[{'terms': {'variantId': ['2-103343353-GAGA-G']}}],
+            filters=[{'terms': {'variantId': ['1-248367227-TC-T']}}],
             size=2, index=','.join([INDEX_NAME, SV_INDEX_NAME]),
         )
 
+        with self.assertRaises(Exception) as cm:
+            get_single_es_variant(self.families, '10-10334333-A-G')
+        self.assertEqual(cm.exception.message, 'Variant 10-10334333-A-G not found')
```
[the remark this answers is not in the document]
hanars: Good point. I updated all the calls in this file

PR #1309, opened 2020-05-28
Travis python 3

hanars:


.travis.yml
```diff
@@ -1,16 +1,21 @@
 language: node_js
 
-python:
-- '2.7'
+jobs:
+  include:
+  - python: '2.7'
+    env: TEST_PATTERN='*_tests.py' COVERAGE_THRESHHOLD=90
+  - python: '3.7'
+    env: TEST_PATTERN='dataset_api_tests.py' COVERAGE_THRESHHOLD=20
```
ShifaSZ, line 8: It seems that it is not so easy to define a pattern to match all converted test files. How about we change the name of the test file to `*_2_3_tests.py` for all the converted tests? Then, we can use a simple pattern to cover all of them.

hanars: done!

PR #1310, opened 2020-06-01
Converted case_review_api to Py3.

ShifaSZ:
The file `seqr/utils/middleware.py` is updated because it is called by case_review_api. And its exception handling code is not compatible with Python3.


hanars:
You should change `case_review_api_tests.py` to `case_review_api_2_3_tests.py`

PR #1311, opened 2020-06-01
Converted seqr management command add_project_tag to Python3.

ShifaSZ:
Changes:
1. Add `from __future__ import unicode_literals` to convert all str literals to unicode
2. Change the exception result tests to accommodate the error message differences between the Django libraries of Python2 and 3.
  - Python 2 exception: 'Error: argument --project is required'
  - Python 3 exception: 'Error: the following arguments are required: --name, --order'
  Use regular expression assertion to avoid incompatibility between Python 2 and 3.

### DIFF
--- a/seqr/management/tests/add_project_tag_tests.py
+++ b/seqr/management/tests/add_project_tag_tests.py
@@ -45,20 +45,20 @@ class AddProjectTagTest(TestCase):
         out = StringIO()
         with self.assertRaises(CommandError) as err:
             call_command('add_project_tag', stdout = out)
-        self.assertEqual(err.exception.message, 'Error: argument --project is required')
+        self.assertRegex(str(err.exception), 'Error:.*argument.*--project')
 
         with self.assertRaises(CommandError) as err:
             call_command('add_project_tag',
                 '--project={}'.format(TAG_ARGUMENTS["project"]),
                 stdout = out)
-        self.assertEqual(err.exception.message, 'Error: argument --name is required')
+        self.assertRegex(str(err.exception), 'Error:.*argument.*--name')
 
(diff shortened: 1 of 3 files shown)


seqr/management/tests/add_project_tag_tests.py
```diff
@@ -45,20 +45,20 @@ def test_missing_required_args(self):
         out = StringIO()
         with self.assertRaises(CommandError) as err:
             call_command('add_project_tag', stdout = out)
-        self.assertEqual(err.exception.message, 'Error: argument --project is required')
+        self.assertRegex(str(err.exception), 'Error:.*argument.*--project')
```
hanars, line 48: In order to make sure the error message is as expected, could we change this to check that the error is one of the 2 exact messages we expect, rather than allowing anything with the argument name:
```
self.assertTrue(str(err.exception) in {'Error: argument --project is required', 'Error: the following arguments are required: --project'})
```

hanars:
Going forward, when you get a test file passing you should change the name from `<x>_tests.py` to `<x>_2_3_tests.py` so the test runner picks it up

seqr/management/tests/add_project_tag_tests.py
ShifaSZ: Okey, I will do that.

ShifaSZ: How about use `assertIn()` as below?
```
self.assertIn(str(err.exception), ['Error: argument --project is required',
       'Error: the following arguments are required: --project, --name, --order'])
```

hanars:
Now that the case_review_api PR is merged, if you change this back to being against dev does it work?

PR #1312, opened 2020-06-02
Add proband relationship

hanars:
This adds a new column to the Individual model to keep track of the relationship of the individual to the proband (primary affected family member). We need this data for reporting purposes. I also edited the sample metadata staff page to include this field as well as a couple other updates to that file that were requested


seqr/views/apis/staff_api_tests.py
```diff
@@ -576,37 +524,69 @@ def test_success_story(self):
         self.assertEqual(len(response_json['rows']), 1)
         self.assertDictEqual(response_json['rows'][0], EXPECTED_SUCCESS_STORY)
 
-    @mock.patch('seqr.views.apis.staff_api.export_multiple_files')
+    @mock.patch('seqr.views.utils.export_utils.zipfile.ZipFile')
```
hanars, line 527: This PR only changes the output of the sample metadata, and not the anvil download. However, the input to `export_multiple_files` for the anvil export changed, even though these change had no impact on the files in the end. Since what is really important with the anvil metadata is that the contents of the zipfile remained unchanged, I don't want the test to fail every time an intermediate step looks different, I only want it to fail if the final output changes. Therefore, I updated this test to mock out the lower level function and check the actual file content, rather than mocking out the higher level function

hanars:
> Very clean. Great job!
> I'm just curious about how the frontend and backend can sync for their proband relationship options.

Sadly, they don't really. You will note that there is an enumerated list of options for the backend in models.py and then a duplicate of that list for the frontend in constants.js 

This duplication exists for all enumerated fields in seqr, and isn't great. I have this ticket open to someday clean this up by passing all enumerations from the backend to the frontend at page load time but I haven't had the time to work on it yet: https://github.com/macarthur-lab/seqr-private/issues/535

PR #1318, opened 2020-06-04
Conver seqr igv_api to Python 2 and 3 compatible.

ShifaSZ:
Changes:
1. use `.decode('utf-8')` to convert `bytes` to unicode `str`
2. replace `__builtin__` with module path because it has been changed to `builtins` in Python 3. The module path is compatible for both Python 2 and 3.


seqr/views/apis/igv_api.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import unicode_literals
+from builtins import str, int
```
hanars, line 2: I'm confused about the interaction between `unicode_literals` and the builtin `str`. Why are we calling `str` in this module at all and why do we need the future builtin for it?

ShifaSZ: While running with Python 2, the `str` imported from `builtins` is different from the standard `str` (which is `__builtin__.str`). The result of `builtins.str()` is a `unicode` while that of `__builtin__.str()` is a `str` (Python 2 `str`).   
Both `__future__` and `builtins` are belonging to the `future` package.   
There is no interaction between `unicode_literals` and `builtins.str`.

hanars: Thanks! But why do we need the `str` builtin here at all though? It looks like `str` is only called once in the file and it should work just fine if in python 3 that returns unicode and in python 2 it doesn't

ShifaSZ: Yes. Using non-unicode `str` for Python 2 should be good in this case. I'm going to remove the importing of `str`.

ShifaSZ: Non-long `int` should be also good enough. I will also remove importing `builtins.int`.

ShifaSZ: Removed.

PR #1321, opened 2020-06-04
Converted locus_list_api to Python 2 and 3 compatible.

ShifaSZ:
One thing weird is that the Python 3 Django doesn't do longest url matching. It is solved by reordering the url entries. No change to the urls other locus_list_api yet.

### DIFF
--- a/seqr/views/apis/locus_list_api_2_3_tests.py
+++ b/seqr/views/apis/locus_list_api_2_3_tests.py
@@ -1,3 +1,5 @@
+# from __future__ import unicode_literals
+
 import json
 import mock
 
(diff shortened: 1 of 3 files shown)


seqr/views/apis/locus_list_api_2_3_tests.py
```diff
@@ -1,7 +1,8 @@
+from __future__ import unicode_literals
+
 import json
 import mock
 
-from django.test import TransactionTestCase
```
ShifaSZ, line 4: `TansactionTestCase` hasn't been used.

PR #1323, opened 2020-06-05
Convert check_bam_cram_paths command to Python 2 and 3 compatible.

ShifaSZ:
Changes:
1. Replace `BytesIO` with `StringIO` for the stdout of the command because Python 3 can't write `str` data to a binary IO.
2. Python 2 can't work with `StringIO` because the command wrapper force output into `str` type while `StringIO` requires `unicode`. Therefore, change `self.stdout.write()` to `self.stdout._out.write()` to bypass the wrapper.
Bypassing the wrapper and using the internal attribute is not good sometime. The other solution is to change `stdout` to `logging` for command outputs.


seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -21,8 +24,8 @@ def handle(self, *args, **options):
             except Exception as e:
                 individual_id = sample.individual.individual_id
                 failed.append(individual_id)
-                self.stdout.write('Error at {} (Individual: {}): {} '.format(sample.file_path, individual_id, e.message))
+                self.stdout._out.write('Error at {} (Individual: {}): {} \n'.format(sample.file_path, individual_id, str(e)))
 
-        self.stdout.write('---- DONE ----')
-        self.stdout.write('Checked {} samples'.format(len(samples)))
-        self.stdout.write('{} failed samples: {}'.format(len(failed), ', '.join(failed)))
+        self.stdout._out.write('---- DONE ----\n')
```
hanars, line 29: I understand why you made this change based on your explanation, but I think that using protected class members is really bad practice. I like your idea of switching to using logging instead

ShifaSZ: Updated.

PR #1324, opened 2020-06-05
Convert copy_project_tags command to Python 2 and 3 compatible.

ShifaSZ:
Some changes with the similar reasons as PR #1323


seqr/management/commands/copy_project_tags.py
```diff
@@ -26,5 +28,5 @@ def handle(self, *args, **options):
             tag.id = None
             tag.project = target_project
             tag.save()
-            self.stdout.write('Saved tag %s (new id = %d)' % (tag.name, tag.id))
+            self.stdout._out.write('Saved tag %s (new id = %d)\n' % (tag.name, tag.id))
```
hanars, line 31: same as my comment on the other PR, I think we should just switch this to logging

ShifaSZ: Replaced stdio with logging.

PR #1328, opened 2020-06-05
Converted reload_saved_variant_json command to be Python 2 and 3 compatible.

ShifaSZ:


hanars:
you should rename the test file to `_2_3_test.py`

seqr/management/tests/reload_saved_variant_json_tests.py
```diff
@@ -1,11 +1,13 @@
 #-*- coding: utf-8 -*-
```
hanars, line 1: you should rename this file to _2_3_test.py

ShifaSZ: Okay. Changed.

PR #1337, opened 2020-06-09
Converted seqr.view.utils.file_utils to Python 2 and 3 compatible.

ShifaSZ:
Besides the conventional changes, the special changes include:
1. Added TextIOWrapper for non excel files
2. Added csv and xlsx file tests to ensure save_uploaded_file() is working with a variety of file types
3. Changed the BytesIO in test_parse_file() to the StringIO. This is required because Python 3 tsv/csv parsing doesn't work with binary data.

Discussion: the tests for parse_file with xls data doesn't make sense because they are only provided with text data (via the StringIO) while xls/xlsx are binary data.


seqr/views/utils/file_utils_tests.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
hanars, line 2: you should rename this file to _2_3_tests.py so it will get tested in python 3 going froward

seqr/views/utils/file_utils_tests.py
```diff
@@ -9,6 +10,9 @@
 from seqr.views.utils.file_utils import save_temp_file, parse_file, load_uploaded_file
 from seqr.views.utils.test_utils import AuthenticationTestCase
 
+import openpyxl as xl
+from tempfile import NamedTemporaryFile
```
hanars, line 14: can you move these imports up so the app specific imports are all below the external imports

seqr/views/utils/file_utils_tests.py
```diff
@@ -96,15 +106,50 @@ def test_temp_file_upload(self):
             'parsedData': PARSED_DATA,
             'uploadedFileId': mock.ANY,
         })
+        # Remove temp file
+        load_uploaded_file(response_json['uploadedFileId'])
```
hanars, line 110: why is this necessary? everything gets written to temp directories so its not a big deal if the tests leave some files there

seqr/views/utils/file_utils_tests.py
```diff
@@ -96,15 +106,50 @@ def test_temp_file_upload(self):
...
+
+        # Test uploading with csv data
+        response = self.client.post('{}?parsedData=true'.format(url),
```
hanars, line 113: could you test the different data types in a loop, the way `test_parse_file` does

seqr/views/utils/file_utils_tests.py
```diff
@@ -96,15 +106,50 @@ def test_temp_file_upload(self):
...
+        # Remove temp file
+        load_uploaded_file(response_json['uploadedFileId'])
+
+        # Test uploading with xlsx data
+        wb = xl.Workbook()
+        ws = wb[wb.sheetnames[0]]
+        ws['A1'], ws['B1'], ws['C1'] = XLSX_PARSED_DATA[0]
+        ws['A2'], ws['B2'], ws['C2'] = XLSX_PARSED_DATA[1]
+        ws['A3'], ws['B3'], ws['C3'] = XLSX_PARSED_DATA[2]
+        with NamedTemporaryFile() as tmp:
```
hanars, line 130: setup code must be done in the test setup, not in the middle of the test

seqr/views/utils/file_utils_tests.py
ShifaSZ: ok

seqr/views/utils/file_utils_tests.py
ShifaSZ: ok

seqr/views/utils/file_utils_tests.py
ShifaSZ: I'm not sure about it. Delete these lines.

seqr/views/utils/file_utils_tests.py
ShifaSZ: Moved to setUp()

seqr/views/utils/file_utils_tests.py
ShifaSZ: Okay

seqr/views/utils/file_utils_2_3_tests.py
```diff
@@ -11,13 +15,13 @@
 
 TSV_DATA = b'Family ID	Individual ID	Notes\n\
 "1"	"NA19675"	"An affected individual, additional metadata"\n\
-"1"	"NA19678"	""'
+"1"	"NA19678"	"None"'
```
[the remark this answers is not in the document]
ShifaSZ: For `.xlsx` format `""` will get a parsed value of `None` rather than an empty string `''`. It need a special handling for the parsed data. I will try to update it.

seqr/views/utils/file_utils_2_3_tests.py
```diff
@@ -56,6 +60,23 @@ def _mock_cell(value):
...
+        ws['A2'], ws['B2'], ws['C2'] = PARSED_DATA[1]
+        ws['A3'], ws['B3'], ws['C3'] = PARSED_DATA[2]
+        with NamedTemporaryFile() as tmp:
+            wb.save(tmp)
+            tmp.seek(0)
+            xlsx_data = tmp.read()
+        self.upload_file_test_data_type = {
+            'tsv': TSV_DATA,
+            'csv': CSV_DATA,
+            'xlsx': xlsx_data
```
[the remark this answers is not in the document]
ShifaSZ: Okay. Test `xlsx` separately.

seqr/views/utils/file_utils_2_3_tests.py
```diff
@@ -37,6 +41,12 @@
     ['1', 'NA19678', ''],
 ]
 
+XLSX_PARSED_DATA = [
+    ['Family ID', 'Individual ID', 'Notes'],
+    ['1', 'NA19675', 'An affected individual, additional metadata'],
+    ['1', 'NA19678', None],
```
hanars, line 47: so actually excels parsed data should be the same as everything else and be an empty string, its a bug in `parse_file`. You should change ` padding = [None] * last_col_index` to ` padding = [''] * last_col_index`

ShifaSZ: Okay

seqr/views/utils/file_utils.py
```diff
@@ -51,8 +55,8 @@ def parse_file(filename, stream):
         rows = rows[:last_row_index+1]
         # all rows should have same column count
         last_col_index = max(max(i for i, val in enumerate(row) if val) for row in rows)
-        padding = [None] * last_col_index
-        rows = [(row + padding)[:last_col_index+1] for row in rows]
+        padding = [''] * last_col_index
+        rows = [(row[:max(i for i, val in enumerate(row) if val)+1] + padding)[:last_col_index+1] for row in rows]
```
ShifaSZ, line 59: `row + padding` will keep the trailing `None`. So, the `row` has to be replaced with `row[:max(i for i, val in enumerate(row) if val)+1]`

hanars: hmm I see the issue. I think a neater way to handle this would be to change the return from `_parse_excel_string_cell` from `cell_value` to `cell_value or ''`

ShifaSZ: Yes. Then, remove the padding? And replace all `None` value cells with `''`, not just the trailing ones?

ShifaSZ: I keep the trailing padding for just in case there are excel files with shorter rows.

seqr/views/utils/file_utils.py
```diff
@@ -64,7 +68,9 @@ def parse_file(filename, stream):
 
 def _parse_excel_string_cell(cell):
     cell_value = cell.value
-    if cell_value and cell.data_type == 'n' and int(cell_value) == cell_value:
+    if not cell_value:
+        cell_value = ''
```
hanars, line 72: This if/else needs to be in the other order, otherwise the value `0` will get removed when in reality we want it to be the string `"0"`

ShifaSZ: Oh... Should I check `0` and `None` separately?

hanars: No, like I said, change the order of the if/else:
```
if cell.data_type == 'n' and int(cell_value) == cell_value:
        cell_value = '{:.0f}'.format(cell_value)
elif not cell_value:
        cell_value = ''
```

ShifaSZ: So, the number `0` will become a string `'0'`?

seqr/views/utils/file_utils.py
```diff
@@ -64,9 +68,9 @@ def parse_file(filename, stream):
 
 def _parse_excel_string_cell(cell):
     cell_value = cell.value
-    if cell_value and cell.data_type == 'n' and int(cell_value) == cell_value:
+    if cell_value is not None and cell.data_type == 'n' and int(cell_value) == cell_value:
```
hanars, line 71: if `cell_value` is None than the part of this clause that is `cell.data_type == 'n'` will always fail, so adding the additional null check here is unnecessary. Just doing the single line change for `cell_value or ''` that I initially requested is really all you need to do

seqr/views/utils/file_utils_2_3_tests.py
```diff
@@ -52,10 +56,31 @@ def _mock_cell(value):
...
 class FileUtilsTest(AuthenticationTestCase):
     fixtures = ['users']
 
+    def setUp(self):
+        # Test uploading with xlsx data
+        wb = xl.Workbook()
+        ws = wb[wb.sheetnames[0]]
+        ws['A1'], ws['B1'], ws['C1'] = map(_convert_int, PARSED_DATA[0])
+        ws['A2'], ws['B2'], ws['C2'] = map(_convert_int, PARSED_DATA[1])
+        ws['A3'], ws['B3'] = map(_convert_int, PARSED_DATA[2][:2])
```
hanars, line 76: Given that this is a small file, this would be much more readable if you just wrote the data you want to write here, instead of doing a mapping on `PARSED_DATA`  (i.e. `ws[1] = [0, 'NA19678', '']...`)

ShifaSZ: Okay

seqr/views/utils/file_utils.py
hanars: bump

ShifaSZ: The problem is when `cell_value` is `None`, `int(cell_value)` will cause an exception.

ShifaSZ: If we do it as:
```
    if cell_value and cell.data_type == 'n' and int(cell_value) == cell_value:
```
Then `0` will be converted to `''`.

PR #1342, opened 2020-06-17
Converted some less util-functions depending apis to Python 2 and 3 compatible.

ShifaSZ:
Changes:
1. converted `dict.keys()/items()/values()` and `e.message` as usual
2. used `from future.standard_library import install_aliases` to solve the difference of `urllib` between Python 2 and 3.

### DIFF
--- a/seqr/views/apis/users_api.py
+++ b/seqr/views/apis/users_api.py
@@ -1,10 +1,5 @@
-from __future__ import unicode_literals
-from future.standard_library import install_aliases
-install_aliases()
-
-import urllib.parse as urlparse
-
 import json
+import urllib
 from anymail.exceptions import AnymailError
 from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth import login, authenticate
@@ -67,7 +62,7 @@ def forgot_password(request):
         """.format(
         full_name=user.get_full_name(),
         base_url=BASE_URL,
-        password_token=urlparse.quote_plus(user.password),
+        password_token=urllib.quote_plus(user.password),
     )
 
     try:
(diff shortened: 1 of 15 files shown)


seqr/views/apis/users_api.py
```diff
@@ -1,5 +1,10 @@
+from __future__ import unicode_literals
+from future.standard_library import install_aliases
+install_aliases()
```
hanars, line 3: what does this do? If it is something we need, can you add a comment here explaining it?

seqr/views/apis/users_api.py
```diff
@@ -1,5 +1,10 @@
...
+
+import urllib.parse as urlparse
```
hanars, line 5: I'm assuming this is a python 3 library support thing, but for my curiosity can you explain why this needs to update?

hanars:
This looks awesome! I just have some clarifying questions I'd like to understand before merging which is why I didn't approve yet, but it looks really good. The other thing I'd suggest is we should periodically bump the minimum python 3 test coverage in travis.yml. Right now it is at 23% but it looks like with this PR we are over 50

seqr/views/apis/users_api.py
ShifaSZ: It installs the aliases for `urllib.parse` for Python 2.
In Python 3, some of the methods of `urllib` have been moved under `urllib.parse`. For example, `urllib.quote_plus()` becomes `urllib.parse.quote_plus()`. The `future` package provides an alias of `urllib.parse` to support Python 2&3 compatible code.

seqr/views/apis/users_api.py
ShifaSZ: Python 3 has `urllib.parse` while Python 2 doesn't. The code:
```
from future.standard_library import install_aliases
install_aliases()
```
makes Python 2 support `urllib.parse` as well.

seqr/views/apis/users_api.py
ShifaSZ: I will add comments to the source file.

PR #1346, opened 2020-06-18
Converted seqr.utils to be compatible with Python 2 and 3.

ShifaSZ:
I don't add `from __future__ import unicode_literals` to the file `seqr/utils/middleware.py` in purpose. 
If `unicode_literals` is imported then the `'\n'` in `exception_json['traceback'] = traceback_message.split('\n')` at line 40 will become a `unicode` type. In this case, if `traceback_message` is a Python 2 `str` and has unicodes in it, it will be a casting error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jobs:
   - python: '2.7'
     env: TEST_PATTERN='*_tests.py' COVERAGE_THRESHHOLD=95
   - python: '3.7'
-    env: TEST_PATTERN='*_2_3_tests.py' COVERAGE_THRESHHOLD=23
+    env: TEST_PATTERN='*_2_3_tests.py' COVERAGE_THRESHHOLD=50
 
 node_js:
 - '11.8.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: node_js
+language: python
 
 jobs:
   include:
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 set -x
-pip install -r requirements.txt --user --upgrade
-pip install -r requirements-dev.txt --user --upgrade
+pip install -r requirements.txt --upgrade
+pip install -r requirements-dev.txt --upgrade
 cd ui/
 echo -ne "\n \n \3033[1B\nsemantic/\n" | npm install
 cd ..
--- a/seqr/utils/middleware.py
+++ b/seqr/utils/middleware.py
@@ -33,7 +33,7 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': str(exception)}
+            exception_json = {'message': exception.args[0]}
             traceback_message = traceback.format_exc()
             logger.error(traceback_message)
             if DEBUG:
--- a/seqr/utils/middleware.py
+++ b/seqr/utils/middleware.py
@@ -1,5 +1,5 @@
 from __future__ import unicode_literals
-from builtins import str as unicode_str
+from builtins import str
 
 from django.core.exceptions import PermissionDenied, ObjectDoesNotExist
 from django.utils.deprecation import MiddlewareMixin
@@ -36,10 +36,7 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            try:
-                exception_json = {'message': str(exception)}
-            except Exception:
-                exception_json = {'message': unicode_str(exception)}
+            exception_json = {'message': exception.message if hasattr(exception, 'message') else str(exception)}
             traceback_message = traceback.format_exc()
             logger.error(traceback_message)
             if DEBUG:
--- a/seqr/utils/middleware.py
+++ b/seqr/utils/middleware.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from django.core.exceptions import PermissionDenied, ObjectDoesNotExist
 from django.utils.deprecation import MiddlewareMixin
 from requests import HTTPError
@@ -33,10 +35,10 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': exception.args[0]}
+            exception_json = {'message': str(exception)}
             traceback_message = traceback.format_exc()
             logger.error(traceback_message)
             if DEBUG:
-                exception_json['traceback'] = traceback_message.split('\n')
+                    exception_json['traceback'] = traceback_message
             return create_json_response(exception_json, status=_get_exception_status_code(exception))
         return None
--- a/seqr/utils/middleware.py
+++ b/seqr/utils/middleware.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from builtins import str as unicode_str
 
 from django.core.exceptions import PermissionDenied, ObjectDoesNotExist
 from django.utils.deprecation import MiddlewareMixin
@@ -35,7 +36,10 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': str(exception)}
+            try:
+                exception_json = {'message': str(exception)}
+            except Exception:
+                exception_json = {'message': unicode_str(exception)}
             traceback_message = traceback.format_exc()
             logger.error(traceback_message)
             if DEBUG:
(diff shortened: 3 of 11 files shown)


seqr/utils/middleware.py
```diff
@@ -33,7 +33,7 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': str(exception)}
+            exception_json = {'message': exception.args[0]}
```
ShifaSZ, line 36: The reason for removing `str()` is: when the `exception` has unicode chars, if `exception` is a Python 2 `unicode` type, casting to Python 2 `str` will cause an error. Then, we have to import `str` from `future.builtins`, but it will cause another error if `exception` is a Python 2 `str`.

hanars, line 36: I'm not sure I like this change. I thought `str(exception)` was valid in python 3? Its entirely possible that there are subclasses of exceptions from thrid parties that implement there own string method and don't take args...

seqr/utils/redis_utils_2_3_tests.py
```diff
@@ -29,7 +29,10 @@ def test_safe_redis_get_json(self, mock_redis, mock_logger):
         mock_redis.return_value.get.side_effect = lambda key: key
         self.assertIsNone(safe_redis_get_json('test_key'))
         mock_logger.info.assert_called_with('Loaded test_key from redis')
-        mock_logger.warn.assert_called_with('Unable to fetch "test_key" from redis: No JSON object could be decoded')
+        warn_args = str(mock_logger.warn.call_args.args[0])
+        # Python 2 and 3 return different warning when encountering JSON decoding error
+        self.assertIn(warn_args, ['Unable to fetch "test_key" from redis: No JSON object could be decoded',
+                                  'Unable to fetch "test_key" from redis: Expecting value: line 1 column 1 (char 0)'])
```
hanars, line 35: The second error here does not in any way express to the user that the problem is with json, its a useless error, which means that error is being incorrectly handled in python 3. You need to fix the error handling, not the test. If the default message in python 3 does not in anay way include the word "json" , we need to add an explicit test for this

I'm also confused why this needs to change? This file was already running on travis in python 3 and it was passing with that JSON error. This makes me think that the problem is not the error message changing but rather that the update you made to the middleware actally returns a less useful message

hanars:
As I mentioned in my comment on the redis utils test, I think the change you made to the middleware is not good. I think the default should be to use the default string representation of the error, and if you can show me a case where that is failing then maybe we can wrap it in a try-except

Also, I'm not really worried about the issue with the `traceback` - we only return that when running locally to make debugging a bit easier but it is never called in production. I'd rather we import unicode literals and wrap the traceback in a try - except block and if theres a unicode error we just don't incude the traceback

seqr/utils/middleware.py
hanars: I would rather in the very rare case we have an uncaught error with unicode the error handling doesn't behave, rather than all the time retuning worse error messages. I think we should include the unicode literals and futre string, and wrap this in a try - catch to handle python 2

ShifaSZ: `str(exception)` does work in Python 3.

ShifaSZ: `try-except` maybe is better. I will try.

seqr/utils/redis_utils_2_3_tests.py
ShifaSZ: I did some experiments:
Python 2:
```
$ python
Python 2.7.16 |Anaconda, Inc.| (default, Sep 24 2019, 16:55:38) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.loads('')
  File "/Users/shifa/opt/anaconda2/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/Users/shifa/opt/anaconda2/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/shifa/opt/anaconda2/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
>>> 
```
Python 3:
```
$ python
Python 3.7.7 (default, May  6 2020, 04:59:01) 
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.loads('')
  File "/Users/shifa/opt/anaconda2/envs/seqr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/Users/shifa/opt/anaconda2/envs/seqr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/shifa/opt/anaconda2/envs/seqr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
>>> 
```
You can see the differences in error messages.

ShifaSZ: Why travis-ci didn't catch the error is weird.

seqr/utils/middleware.py
```diff
@@ -33,10 +36,13 @@ class JsonErrorMiddleware(MiddlewareMixin):
     @staticmethod
     def process_exception(request, exception):
         if request.path.startswith('/api'):
-            exception_json = {'message': str(exception)}
+            try:
+                exception_json = {'message': str(exception)}
+            except Exception:
+                exception_json = {'message': unicode_str(exception)}
```
hanars, line 42: I really dislike allowing python 2 strings here, especially since we will stop doing that when we switch to python 3. The idea behind this conversion is switch over to behaving in python 2 to the way it will in python 3, not to unnecessarily support python 2 behavior.  What I meant by using try-except here is we call `str(exception)` using the future `str` module, and if that throws an error we just return something generic like "An error occurred". Again, handling uncaught exceptions with unicode is an edge case I don't want us to worry about too much so I'd rather not write hard to maintain code to support it

seqr/utils/redis_utils_2_3_tests.py
hanars: Yeah its a problem, I see it passing here: https://travis-ci.org/github/macarthur-lab/seqr/jobs/699503352#L1113

The fact that the behavior on you machine differes from our test runner is a problem. The solution isn't to support both, its to figure out why this happened. One thing to start with is make sure your package versions match those used by travis - try running the command travis uses for this and see if it changes anything
```
pip install -r requirements.txt --user --upgrade
pip install -r requirements-dev.txt --user --upgrade
```

seqr/utils/middleware.py
hanars: thinking about this more, why not use `str` with `ignore` set so it will just strip out unicodes?

install_dependencies.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 set -x
-pip install -r requirements.txt --user --upgrade
-pip install -r requirements-dev.txt --user --upgrade
+pip install -r requirements.txt --upgrade
+pip install -r requirements-dev.txt --upgrade
```
ShifaSZ, line 5: Removing the `--user` option from the `pip` command because it causes errors while installing packages in `virtualenv`. I'm sure whether it could cause trouble or not when there is no `virtualenv`.

hanars: given that travis doesn't use virtualenv I'd prefer if you put this back

.travis.yml
```diff
@@ -1,11 +1,11 @@
-language: node_js
+language: python
```
hanars, line 1: thanks for catching this!

install_dependencies.sh
hanars: given that travis doesn't use virtualenv I'd prefer if you put this back

seqr/views/apis/staff_api_tests.py
```diff
@@ -486,7 +486,7 @@ def test_mme_details(self, mock_datetime):
         url = reverse(mme_details)
         self.check_staff_login(url)
 
-        mock_datetime.now.return_value = datetime(2020, 4, 27, 20, 16, 01)
+        mock_datetime.now.return_value = datetime(2020, 4, 27, 20, 16, 1)
```
ShifaSZ, line 489: `01` will cause the coverage tool to fail to collect coverage numbers for Python 3. Therefore, change it to `1`.

install_dependencies.sh
ShifaSZ: The errors were reported by travis-ci. So, it can't be changed back.

ShifaSZ: I believe it must be okay for travis-ci environment. I'm not sure whether there are any other environments using `install_dependencies.sh`. Removing `--user` will cause those environments to install packages to the system-wide.

seqr/utils/middleware.py
ShifaSZ: Updated

seqr/utils/redis_utils_2_3_tests.py
ShifaSZ: Changed to compare the first part of the warning message.

PR #1347, opened 2020-06-18
Seqr views utils

ShifaSZ:
The biggest impact while converting seqr.views.utils to Python 2 and 3 is still the unicode encoding. Especially, `seqr/views/utils/export_utils.py` and `seqr/views/utils/permissions_utils.py` make many of the `str` in the seqr apis become `unicode`.


seqr/views/apis/staff_api_tests.py
```diff
@@ -593,16 +593,16 @@ def test_anvil_export(self, mock_zip):
...
         self.assertEqual(mock_write_zip.call_count, 4)
         mock_write_zip.assert_has_calls([
-            mock.call('1kg project n\xc3\xa5me with uni\xc3\xa7\xc3\xb8de_PI_Subject.tsv', mock.ANY),
-            mock.call('1kg project n\xc3\xa5me with uni\xc3\xa7\xc3\xb8de_PI_Sample.tsv', mock.ANY),
-            mock.call('1kg project n\xc3\xa5me with uni\xc3\xa7\xc3\xb8de_PI_Family.tsv', mock.ANY),
-            mock.call('1kg project n\xc3\xa5me with uni\xc3\xa7\xc3\xb8de_PI_Discovery.tsv', mock.ANY),
+            mock.call(u'1kg project n\xe5me with uni\xe7\xf8de_PI_Subject.tsv', mock.ANY),
+            mock.call(u'1kg project n\xe5me with uni\xe7\xf8de_PI_Sample.tsv', mock.ANY),
+            mock.call(u'1kg project n\xe5me with uni\xe7\xf8de_PI_Family.tsv', mock.ANY),
+            mock.call(u'1kg project n\xe5me with uni\xe7\xf8de_PI_Discovery.tsv', mock.ANY),
```
ShifaSZ, line 605: Because of the string literals in `seqr/views/utils/export_utils.py` are converted into `unicode`, the `writestr()` calling arguments are changed.

seqr/views/utils/export_utils.py
```diff
@@ -25,6 +28,9 @@ def export_table(filename_prefix, header, rows, file_format='tsv', titlecase_hea
     Returns:
         Django HttpResponse object with the table data as an attachment.
     """
+    def _to_str(s):
+        return str(s, 'utf-8', errors = 'ignore') if not isinstance(s, str) else s
```
ShifaSZ, line 32: For Python 3, `str` objects can't be decoded. Therefore, it is necessary to check `isinstance(s, str)`.

seqr/views/utils/export_utils.py
```diff
@@ -34,21 +40,21 @@ def export_table(filename_prefix, header, rows, file_format='tsv', titlecase_hea
         response = HttpResponse(content_type='text/tsv')
         response['Content-Disposition'] = 'attachment; filename="{}.tsv"'.format(filename_prefix)
         response.writelines(['\t'.join(header)+'\n'])
-        response.writelines(('\t'.join(map(unicode, row))+'\n' for row in rows))
+        response.writelines(('\t'.join(map(_to_str, row))+'\n' for row in rows))
```
hanars, line 43: isn't `unicode` in python 2 basically the same as `str` in the future python 3? Because if so just swapping this out for future `str` should be fine. If we already weren't supporting unicode theres no reason to add support here

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -14,36 +16,37 @@ class ExportTableUtilsTest(TestCase):
...
-        self.assertRaisesRegexp(ValueError, '.*format.*',
+        self.assertRaisesRegex(ValueError, '.*format.*',
             lambda: export_table('test_file', header, rows, file_format='unknown_format')
         )
 
     @mock.patch('seqr.views.utils.export_utils.zipfile.ZipFile')
     def test_export_multiple_files(self, mock_zip):
         mock_zip_content = {}
         def mock_write_zip(file, content):
-            mock_zip_content[file] = unicode(content)
```
hanars, line 46: This is a crucial line of test code that is making sure this line is working - https://github.com/macarthur-lab/seqr/pull/1347/files#diff-f83ee929ae3748972d14c31ecc67bf2dR86

Since everything is encoded before write zip is called, this call should always run without error, and thats what this is testing. Taking this out means that encoding is not properly tested, which means a real life bug could go through

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -437,7 +438,7 @@ def _parse_datstat_export_format(rows):
 
 
 def _get_datstat_family_notes(row):
-    row = {k: unicode(v, errors='ignore') for k, v in row.items()}
+    row = {k: str(v, 'utf-8', errors='ignore') if not isinstance(v, str) else v for k, v in row.items()}
```
hanars, line 441: I don't like the conditional with the  type checking here - if it is an instance of `str` theres no harm wrapping it again and its much easier to read

seqr/views/utils/pedigree_info_utils_tests.py
```diff
@@ -26,16 +28,18 @@ def test_parse_pedigree_table(self):
             [['family_id', 'individual_id', 'sex', 'affected', 'father', 'mother'],
              ['', '', 'male', 'u', '.', 'ind2']], FILENAME)
         self.assertListEqual(records, [])
-        self.assertListEqual(
-            errors, ["Error while converting {} rows to json: Family Id not specified in row #1:\n{{'affected': 'u', 'maternalId': 'ind2', 'individualId': '', 'sex': 'male', 'familyId': '', 'paternalId': ''}}".format(FILENAME)])
+        self.assertIn(
+            errors[0], ["Error while converting {} rows to json: Family Id not specified in row #1:\n{{'familyId': '', 'individualId': '', 'sex': 'male', 'affected': 'u', 'paternalId': '', 'maternalId': 'ind2'}}".format(FILENAME),
```
hanars, line 32: change the way the error is generated to make the order consistently sorted in python 2. Or honestly, swap out the whole row json in the error for a line number. Supporting multiple error messages is a worst case scenario thing for library changes out of our control, not a first resort

seqr/views/utils/pedigree_info_utils_tests.py
```diff
@@ -205,21 +209,21 @@ def test_parse_datstat_pedigree_table(self):
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Current Age:__ Patient is deceased, age 33, due to heart attack, sample not available
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Age of Onset:__ 21
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Race/Ethnicity:__ White, Asian, Pacific; Not Hispanic
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Case Description:__ I have a really debilitating probably genetic condition. Ive seen many specialists.
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Clinical Diagnoses:__ Yes; SMAs
```
hanars, line 209: we were intentionally stripping unicode here - why did that change? This is a crucial import that does contain unicode and needs to remain stable

seqr/views/utils/pedigree_info_utils_tests.py
```diff
@@ -254,6 +258,7 @@ def test_parse_datstat_pedigree_table(self):
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Children:__ 
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Son, age 12, unaffected, unspecified availability
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; __Relatives:__ None"""
+        self.maxDiff = None
```
hanars, line 261: why is this added?

seqr/views/utils/pedigree_info_utils.py
hanars: Also, this changes the output of the report in a completely unacceptable way

seqr/views/utils/pedigree_info_utils_tests.py
ShifaSZ: It is for debugging. Remove it.

seqr/views/utils/export_utils.py
ShifaSZ: Hmm, maybe you are right. I made it too complicated.

seqr/views/utils/pedigree_info_utils_tests.py
ShifaSZ: Reduced the test scope to the first part of the message.

seqr/views/utils/pedigree_info_utils.py
ShifaSZ: `str(v, errors='ignore')` is a decoding. Python 3 can't decode `str` objects. That is the reason the conditional check is used.

seqr/views/utils/pedigree_info_utils_tests.py
ShifaSZ: The change is because of the change in:
```
def _get_datstat_family_notes(row):
-    row = {k: unicode(v, errors='ignore') for k, v in row.items()}
+    row = {k: str(v, 'utf-8', errors='ignore') if not isinstance(v, str) else v for k, v in row.items()}
```
`unicode(v, errors='ignore')` strips unicode chars in `v` when the type of `v` is `str`. But, after converting `pedigree_info_utils_tests.py` to the compatible mode, `v` becomes a `unicode` and `unicode(v, errors='ignore')` cause an exception because `unicode` can't be decoded.
After importing `builtins.str`, `str(unicode_v, errors='ignore')` isn't an exception but it won't strip unicode chars.

seqr/views/utils/export_utils_2_3_tests.py
ShifaSZ: According to the online [document](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.writestr)
```
ZipFile.writestr(zinfo_or_arcname, data, compress_type=None, compresslevel=None)

> Write a file into the archive. The contents is data, which may be either a str or a bytes instance; 
> if it is a str, it is encoded as UTF-8 first.
```
Explicitly encode the data is optional. It will be more concise if write `str` rather than encode them to `bytes`.

seqr/views/utils/test_utils.py
```diff
@@ -222,7 +224,7 @@ def _check_login(self, url, permission_level, request_data=None):
         'genotypes': {
             'NA19675': {
                 'sampleId': 'NA19675',
-                'ab': 0.7021276595744681,
+                'ab': 0.702127659574,
```
ShifaSZ, line 227: Removed 4 trailing digits in the decimal because Python 3 supports fewer decimals.

seqr/views/utils/export_utils_2_3_tests.py
ShifaSZ: As the discussion on today's meeting, `unicode(content)` will be changed to `str(content, 'ascii', errors='ignore') to strip out all unicode chars.

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/export_utils_2_3_tests.py  5
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils_2_3_tests.py  5
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
- seqr/views/utils/export_utils_2_3_tests.py  5
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
- seqr/views/utils/export_utils_2_3_tests.py  5
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
```diff
@@ -76,10 +79,11 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
                 content += '\n'.join([
                     DELIMITERS[file_format].join([row.get(key) or blank_value for key in header]) for row in rows
                 ])
-                if not isinstance(content, unicode):
-                    content = unicode(content, errors='ignore')
-                zip_file.writestr('{}.{}'.format(filename.encode('utf-8'), file_format), content)
+                if isinstance(content, str):
+                    content = content.encode('utf-8')
```
hanars, line 83: why is this necessary? `str` should be able to handle any of the data it would receive

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -437,7 +438,8 @@ def _parse_datstat_export_format(rows):
 
 
 def _get_datstat_family_notes(row):
-    row = {k: unicode(v, errors='ignore') for k, v in row.items()}
+    row = {k: v.encode('ascii', errors='ignore').decode() if isinstance(v, str) else str(v, 'ascii', errors='ignore')
```
hanars, line 441: if it is an instance of `str`, wouldn't `str(v, 'ascii', errors='ignore')` still work? so why add the complicated conditional?

seqr/views/utils/pedigree_info_utils_2_3_tests.py
```diff
@@ -26,16 +28,18 @@ def test_parse_pedigree_table(self):
...
 
         records, errors, warnings = parse_pedigree_table(
             [['family_id', 'individual_id', 'sex', 'affected', 'father', 'mother'],
              ['fam1', '', 'male', 'u', '.', 'ind2']], FILENAME)
         self.assertListEqual(records, [])
-        self.assertListEqual(
-            errors, ["Error while converting {} rows to json: Individual Id not specified in row #1:\n{{'affected': 'u', 'maternalId': 'ind2', 'individualId': '', 'sex': 'male', 'familyId': 'fam1', 'paternalId': ''}}".format(FILENAME)])
+        self.assertIn(
+            errors[0], ["Error while converting {} rows to json: Individual Id not specified in row #1:\n{{'familyId': 'fam1', 'individualId': '', 'sex': 'male', 'affected': 'u', 'paternalId': '', 'maternalId': 'ind2'}}".format(FILENAME),
+                     "Error while converting {} rows to json: Individual Id not specified in row #1:\n{{u'affected': u'u', u'maternalId': u'ind2', u'individualId': u'', u'sex': u'male', u'familyId': u'fam1', u'paternalId': u''}}".format(FILENAME)])
```
hanars, line 42: I like the way you tested the warning above by splitting on the newline better than the way you tested here with the 2 possible orders - can you do that here instead of this

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
ShifaSZ: The idea behind this is to strip out the unicode chars from the input data.  We could have a variety of ways to do the stripping depending on the input data types:
1. Python 2 `str`: do decoding to an 'ascii' text with either `str.decode()` or `future.builtins.str()`
2. Python 2 `unicode` or `future.builtins.newstr`: We need to encode the data to binary and then decode to `ascii` texts
3. Python 3 `str`: We need to encode the data to binary and then decode to `ascii` texts
So, it does encoding for #2 and #3 cases and does decoding for all cases.

Simple `str()` will cause errors in below cases:
. `str(data, errors='ignore')` if `data` is a Python 3 `str`
. `str(data)` or `str(data, errors='ignore')` if `data is a Python 2 `str` with unicode chars in it
We have to encode Python 2 `str` with unicodes and Python 3 `str` into binary before we can do `str()`.

seqr/views/utils/pedigree_info_utils.py
ShifaSZ: If it is an instance of Python 3 `str` then `str(v, 'ascii', errors='ignore')` will cause an error as below:
```
>>> v='abc'
>>> type(v)
<class 'str'>
>>> str(v, 'ascii', errors='ignore')
TypeError: decoding str is not supported
>>> 
```
The `str(v, 'ascii', errors='ignore')` works without errors for Python 2 but won't strip out unicodes for `unicode/newstr` instances:
```
>>> str_s='abc'
>>> str_s_with_unicode='\xe2abc'
>>> unicode_s=u'\xe2abc'
>>> newstr_s=str(str_s)
>>> newstr_s_with_unicode=str(unicode_s)
>>> type(str_s)
<type 'str'>
>>> type(str_s_with_unicode)
<type 'str'>
>>> type(unicode_s)
<type 'unicode'>
>>> type(newstr_s)
<class 'future.types.newstr.newstr'>
>>> type(newstr_s_with_unicode)
<class 'future.types.newstr.newstr'>
>>> str(str_s, 'ascii', errors='ignore')
'abc'
>>> str(str_s_with_unicode, 'ascii', errors='ignore')
'abc'
>>> str(unicode_s, 'ascii', errors='ignore')
'\xe2abc'
>>> str(newstr_s, 'ascii', errors='ignore')
'abc'
>>> str(newstr_s_with_unicode, 'ascii', errors='ignore')
'\xe2abc'
>>> 

```

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/pedigree_image_utils_2_3_tests.py  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

hanars:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_image_utils_tests.py  1
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/export_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/export_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/pedigree_info_utils_2_3_tests.py
ShifaSZ: Updated.

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -14,23 +16,24 @@ class ExportTableUtilsTest(TestCase):
...
-        rows = [['row1_v1', 'row1_v2'], ['row2_v1', 'row2_v2']]
+        rows = [['row1_v1\xe2', 'row1_v2'], ['row2_v1', 'row2_v2']]
 
         # test tsv format
         response = export_table('test_file', header, rows, file_format='tsv')
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.get('content-disposition'), 'attachment; filename="test_file.tsv"')
-        self.assertEqual(response.content, '\n'.join(['\t'.join(row) for row in [header]+rows]) + '\n')
+        self.assertEqual(response.content.decode('utf-8', errors = 'ignore'),
```
hanars, line 26: I don't want to test that the decoded content is decoded, I want to test that the raw content is decoded. This needs to not change

seqr/views/utils/pedigree_info_utils_2_3_tests.py
```diff
@@ -26,16 +27,17 @@ def test_parse_pedigree_table(self):
             [['family_id', 'individual_id', 'sex', 'affected', 'father', 'mother'],
              ['', '', 'male', 'u', '.', 'ind2']], FILENAME)
         self.assertListEqual(records, [])
-        self.assertListEqual(
-            errors, ["Error while converting {} rows to json: Family Id not specified in row #1:\n{{'affected': 'u', 'maternalId': 'ind2', 'individualId': '', 'sex': 'male', 'familyId': '', 'paternalId': ''}}".format(FILENAME)])
+        errors = errors[0].split('\n')
```
hanars, line 30: you need to keep testing that there is only 1 error

seqr/views/utils/pedigree_info_utils_2_3_tests.py
```diff
@@ -26,16 +27,17 @@ def test_parse_pedigree_table(self):
...
         self.assertListEqual(warnings, [])
 
         records, errors, warnings = parse_pedigree_table(
             [['family_id', 'individual_id', 'sex', 'affected', 'father', 'mother'],
              ['fam1', '', 'male', 'u', '.', 'ind2']], FILENAME)
         self.assertListEqual(records, [])
-        self.assertListEqual(
-            errors, ["Error while converting {} rows to json: Individual Id not specified in row #1:\n{{'affected': 'u', 'maternalId': 'ind2', 'individualId': '', 'sex': 'male', 'familyId': 'fam1', 'paternalId': ''}}".format(FILENAME)])
+        self.assertEqual(
+            errors[0].split('\n')[0], "Error while converting {} rows to json: Individual Id not specified in row #1:".format(FILENAME))
```
hanars, line 40: you need to keep testing that there is only 1 error

seqr/views/utils/export_utils_2_3_tests.py
ShifaSZ: Changed to encode the expected data.

seqr/views/utils/pedigree_info_utils_2_3_tests.py
ShifaSZ: Added

seqr/views/utils/pedigree_info_utils_2_3_tests.py
ShifaSZ: Added.

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -14,23 +16,23 @@ class ExportTableUtilsTest(TestCase):
...
-        rows = [['row1_v1', 'row1_v2'], ['row2_v1', 'row2_v2']]
+        rows = [['row1_v1\xe2', 'row1_v2'], ['row2_v1', 'row2_v2']]
 
         # test tsv format
         response = export_table('test_file', header, rows, file_format='tsv')
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.get('content-disposition'), 'attachment; filename="test_file.tsv"')
-        self.assertEqual(response.content, '\n'.join(['\t'.join(row) for row in [header]+rows]) + '\n')
+        self.assertEqual(response.content, 'column1\tcolumn2\nrow1_v1\xe2\trow1_v2\nrow2_v1\trow2_v2\n'.encode('utf-8'))
```
hanars, line 26: why does this need to change, it looks like it should evaluate to the same thing?

ShifaSZ: The main purpose of the change is to add `.encode('utf-8')` which is required by Python 3 since `response.content` is a binary.

hanars: then why not leave it as it is and just add the `encode`?

ShifaSZ: I thought the unicodes in the data should be removed. I would like to strip the unicodes in the expected data manually and explicitly rather than using `ascii` encoding. But, later I found the unicode is still there and simply added the char back without changing the whole statement back.
I can change it back.

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
```diff
@@ -77,10 +81,11 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
                 content += '\n'.join([
                     DELIMITERS[file_format].join([row.get(key) or blank_value for key in header]) for row in rows
                 ])
-                if not isinstance(content, unicode):
```
hanars, line 80: If the content is a python 2 string, wouldn't this behavior change?

seqr/views/utils/export_utils.py
```diff
@@ -77,10 +81,11 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
...
-                    content = unicode(content, errors='ignore')
-                zip_file.writestr('{}.{}'.format(filename.encode('utf-8'), file_format), content)
+                if isinstance(content, str):
+                    content = content.encode('utf-8')
+                content = str(content, 'ascii', errors='ignore') # Strip unicode chars in the content
+                zip_file.writestr('{}.{}'.format(filename, file_format), content)
         temp_file.seek(0)
         response = HttpResponse(temp_file, content_type='application/zip')
-        response['Content-Disposition'] = 'attachment; filename="{}.zip"'.format(zip_filename.encode('utf-8'))
```
hanars, line 85: why doesn't this need to be encoded anymore?

seqr/views/utils/export_utils.py
ShifaSZ: if `content` is a python 2 `str` then `isinstance(content, str)` will be `False` because we import `str` from `future.builtins`. So, there will be no encoding but will be followed by a decoding (`str()`). This behavior is the same as that of old codes: `not isinstance(content, unicode)` is True, and then it will be decoded with `unicode()`.

seqr/views/utils/export_utils.py
ShifaSZ: Because after `zip_filename` is encoded, it will be casted back to `str` and become part of the result string.
Please refer to below experiments:
Python 3:
```
>>> zip_filename='n\xe2me'
>>> zip_filename
'nâme'
>>> z=zip_filename.encode('utf-8')
>>> z
b'n\xc3\xa2me'
>>> type(z)
<class 'bytes'>
>>> a='attachment; filename="{}.zip"'.format(zip_filename.encode('utf-8'))
>>> b='attachment; filename="{}.zip"'.format(zip_filename)
>>> type(a)
<class 'str'>
>>> type(b)
<class 'str'>
>>> a
'attachment; filename="b\'n\\xc3\\xa2me\'.zip"'
>>> b
'attachment; filename="nâme.zip"'
>>> a==b
False
```
If `zip_filename.encode('utf-8')` is called then the result `a` is actually not the one we want. It embeds a string `b'n\xc3\xa2me'` which is the binary encoding of `zip_filename`.

Python 2:
```
>>> zip_filename=u'n\xe2me'
>>> z=zip_filename.encode('utf-8')
>>> z
'n\xc3\xa2me'
>>> type(z)
<type 'str'>
>>> a=u'attachment; filename="{}.zip"'.format(zip_filename.encode('utf-8'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
----- comment: error because casting `zip_filename.encode('utf-8')` to a `str` --------
>>> a='attachment; filename="{}.zip"'.format(zip_filename.encode('utf-8'))
>>> b='attachment; filename="{}.zip"'.format(zip_filename)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe2' in position 1: ordinal not in range(128)
----- comment: error because casting unicode `zip_filename` to a Python 2 `str` --------
>>> b=u'attachment; filename="{}.zip"'.format(zip_filename)
----- comment: this is our case because we import future 'unicode_literals' so that `'attachment; filename="{}.zip"'` is the same as `u'attachment; filename="{}.zip"'`. ---------
>>> type(a)
<type 'str'>
>>> type(b)
<type 'unicode'>
>>> print(a)
attachment; filename="nâme.zip"
>>> print(b)
attachment; filename="nâme.zip"
>>> a==b
__main__:1: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
False
```
Since `zip_filename` contains a non-ascii char, some of the castings will be failed because of no specification for 'errors' while encoding/decoding.

hanars: My concern here is I don't think the actual HTTP header can support unicode, I think it was relying on that binary encoding. Can you please update the unit tests to use a filename that has unicode in it and make sure it works in both versions of python

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/export_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
ShifaSZ: Sorry, I thought the `content-disposition` header could do encoding as the HTTP content. It actually does nothing. I found [an article](https://medium.com/@JeremyLaine/non-ascii-content-disposition-header-in-django-3a20acc05f0d) about this. It is really a little complicated. But I updated the code anyway. Please have a look to see if it is acceptable.

seqr/views/utils/export_utils.py
```diff
@@ -77,10 +86,16 @@ def export_multiple_files(files, zip_filename, file_format='csv', add_header_pre
...
+        try:
+            zip_filename.encode('ascii')
+            response['Content-Disposition'] = 'attachment; filename="{}.zip"'.format(zip_filename)
+        except UnicodeEncodeError:
+            response['Content-Disposition'] = "attachment; filename*=utf-8''{}".format(
+                quote('{}.zip'.format(zip_filename).encode('utf-8')))
```
ShifaSZ, line 100: Python 3 can work without `.encode('utf-8')` because the `quote()` will do encoding automatically. But, in Python 2, explicitly encoding is required.

seqr/views/utils/export_utils.py
ShifaSZ: There are some other functions that generate the `Content-Disposition` header. If the code for unicode work I will update all of them. And, it is better to validate the unicode filename from the front end.

seqr/views/utils/export_utils.py
hanars: This is all way more complicated. We should just do the same encoding for the filename that we do for the content itself and strip out all unicodes before we get here (so similar to
```
if isinstance(content, str):
       content = content.encode('utf-8')
content = str(content, 'ascii', errors='ignore')

File: seqr/views/utils/export_utils.py (hunk 18 above)
Comment (ShifaSZ) [reply seqr/views/utils/export_utils.py:85]: I tried it with `staff Pages/Anvil` and `Download AnVIL metadata for test_project_nâme`. According to the captured header below, the `content-disposition` header with Unicode looks working.
image

Comment (hanars) [reply seqr/views/utils/export_utils.py:85]: as I said, we should strip unicode out of the file name completely the same way we strip it out of the content

**Comment by hanars** [comment]


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
ShifaSZ: I would rather keep the existing behaviors unicodes in filenames than simply trip them out. So, I changed a little bit for Python 3 compatible by moving the `.encode('utf-8')` to the end of the string.

seqr/views/utils/export_utils.py
ShifaSZ: I keep the existing behaviors for unicodes. It is not ideal because some browsers may not get the right file name. But, it doesn't get worse. And, it maybe is important to continue supporting unicodes in the scenarios which it supports today. Is that okay?

hanars: I think it doesn't matter if the filename looks a bit weird so long as it can successfully download, so I think this is fine

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -85,7 +88,20 @@ def mock_write_zip(file, content):
         })
         mock_zip_content = {}
 
+        # test unicode chars in zip filename
```
hanars, line 91: you don't need like a separate test case for this, just modify the file name in the above test to have unicode

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -85,7 +88,20 @@ def mock_write_zip(file, content):
...
+        response = export_multiple_files(
+            [['file1', header1, rows], ['file2', header2, rows]], 'zipfilen\xe2me', file_format='tsv')
+        self.assertEqual(response.status_code, 200)
+        filename = response.get('content-disposition')
+        self.assertEqual(filename, 'attachment; filename="zipfilen\xc3\xa2me.zip"' if isinstance(filename, str)
+        else b'attachment; filename="zipfilen\xc3\xa2me.zip"')
```
hanars, line 97: so in general conditional logic in tests is not good, and python 3 doesn't even support `b'...'` notation so I really don't think this is a good way to test. Couldn't you cast it to bytes before checking equality?

seqr/views/utils/export_utils_2_3_tests.py
ShifaSZ: okay

seqr/views/utils/export_utils_2_3_tests.py
ShifaSZ: Python 3 doesn't support `u''` for unicode literals but it does support `b''` for bytes literals. I can change it to casting.

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/export_utils.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils_2_3_tests.py
```diff
@@ -65,10 +67,13 @@ def mock_write_zip(file, content):
...
 
+        # test tsv format with a filename in unicode
         response = export_multiple_files(
-            [['file1', header1, rows], ['file2', header2, rows]], 'zipfile', file_format='tsv')
+            [['file1', header1, rows], ['file2', header2, rows]], 'zipfilen\xe2me', file_format='tsv')
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.get('content-disposition'), 'attachment; filename="zipfile.zip"')
+        filename = response.get('content-disposition')
+        self.assertEqual(filename, 'attachment; filename="zipfilen\xc3\xa2me.zip"' if isinstance(filename, str)
+        else 'attachment; filename="zipfilen\xe2me.zip"'.encode('utf-8'))
```
hanars, line 76: This is a really fragile looking test and I'm not comfortable with it. Moreover, I'm now concerned that if the header type is bytes and not a string it might not actually work the way we want in real life. I really think we should just strip out unicode characters and use plain strings unless you can get this to work with a single predictable header

seqr/views/utils/export_utils.py
ShifaSZ: I'm quite confident that the latest code keeps the existing feature in Python 3 and won't break anything in Python 3. So I don't change it to trip out unicodes.

hanars:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- seqr/views/utils/pedigree_info_utils_2_3_tests.py  7
- seqr/views/utils/export_utils_2_3_tests.py  7
- seqr/views/utils/export_utils.py  1
- seqr/views/utils/orm_to_json_utils_2_3_tests.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5722441&bid=18721126)

seqr/views/utils/export_utils.py
ShifaSZ: Changed to stripping out all unicodes in the filename to remove the risk of unicodes in Http headers.

PR #1348, opened 2020-06-18
Unit test coverage

hanars:


seqr/views/apis/individual_api_tests.py
```diff
@@ -193,30 +216,62 @@ def test_individuals_table_handler(self):
         response_json = response.json()
         self.assertListEqual(response_json.keys(), ['individualsByGuid', 'familiesByGuid'])
```
ShifaSZ, line 217: It is better to use `assertSetEqual` to avoid the array order issue.

PR #1350, opened 2020-06-21
Converted Elasticsearch utils to Python 2 and 3 compatible.

ShifaSZ:
There are some changes that need to be reviewed and confirmed. Please see the inline comments.

### DIFF
--- a/seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
+++ b/seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
@@ -7,7 +7,7 @@ import mock
 from django.core.management import call_command
 from django.test import TestCase
 from django.core.management.base import CommandError
-from seqr.utils.elasticsearch.es_utils_tests import PARSED_VARIANTS
+from seqr.utils.elasticsearch.es_utils_2_3_tests import PARSED_VARIANTS
 from seqr.models import SavedVariant
 
 SAVED_VARIANT_GUID = 'SV0000001_2103343353_r0390_100'
--- a/seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
+++ b/seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
@@ -7,7 +7,7 @@ import mock
 from django.core.management import call_command
 from django.test import TestCase
 from django.core.management.base import CommandError
-from seqr.utils.elasticsearch.es_utils_2_3_tests import PARSED_VARIANTS
+from seqr.views.utils.test_utils import PARSED_VARIANTS
 from seqr.models import SavedVariant
 
 SAVED_VARIANT_GUID = 'SV0000001_2103343353_r0390_100'
--- a/seqr/utils/elasticsearch/es_gene_agg_search.py
+++ b/seqr/utils/elasticsearch/es_gene_agg_search.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from collections import defaultdict
 import logging
 
@@ -12,9 +14,11 @@ class EsGeneAggSearch(EsSearch):
     CACHED_COUNTS_KEY = None
 
     def aggregate_by_gene(self):
-        searches = {self._search}
+        searches = [self._search]
         for index_searches in self._index_searches.values():
-            searches.update(index_searches)
+            for index_search in index_searches:
+                if index_search not in searches:
+                    searches.append(index_search)
 
         for search in searches:
             agg = search.aggs.bucket(
@@ -88,8 +92,8 @@ class EsGeneAggSearch(EsSearch):
         loaded_compound_hets = self.previous_search_results.get('grouped_results', []) + \
                                self.previous_search_results.get('compound_het_results', [])
         for group in loaded_compound_hets:
-            variants = group.values()[0]
-            gene_id = group.keys()[0]
+            variants = next(iter(group.values()))
+            gene_id = next(iter(group))
             if gene_id and gene_id != 'null':
                 gene_counts[gene_id]['total'] += len(variants)
                 for family_guid in variants[0]['familyGuids']:
@@ -119,8 +123,8 @@ class EsGeneAggSearch(EsSearch):
                 loaded = sum(counts.get('loaded', 0) for counts in previous_search_results.get('loaded_variant_counts', {}).values())
                 if loaded == total_results:
                     for group in previous_search_results['grouped_results']:
-                        variants = group.values()[0]
-                        gene_id = group.keys()[0]
+                        variants = next(iter(group.values()))
+                        gene_id = next(iter(group))
                         if not gene_id or gene_id == 'null':
                             gene_id = next((
                                 gene_id for gene_id, transcripts in variants[0]['transcripts'].items()
--- a/seqr/utils/elasticsearch/es_gene_agg_search.py
+++ b/seqr/utils/elasticsearch/es_gene_agg_search.py
@@ -16,7 +16,7 @@ class EsGeneAggSearch(EsSearch):
     def aggregate_by_gene(self):
         searches = [self._search]
         for index_searches in self._index_searches.values():
-            searches = searches + [index_search for index_search in index_searches if index_search not in searches]
+            searches += [index_search for index_search in index_searches if index_search not in searches]
 
         for search in searches:
             agg = search.aggs.bucket(
--- a/seqr/utils/elasticsearch/es_gene_agg_search.py
+++ b/seqr/utils/elasticsearch/es_gene_agg_search.py
@@ -16,9 +16,7 @@ class EsGeneAggSearch(EsSearch):
     def aggregate_by_gene(self):
         searches = [self._search]
         for index_searches in self._index_searches.values():
-            for index_search in index_searches:
-                if index_search not in searches:
-                    searches.append(index_search)
+            searches = searches + [index_search for index_search in index_searches if index_search not in searches]
 
         for search in searches:
             agg = search.aggs.bucket(
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -129,7 +129,7 @@ class EsSearch(object):
 
     def filter_by_frequency(self, frequencies):
         q = Q()
-        for pop, freqs in frequencies.items():
+        for pop, freqs in sorted(frequencies.items()):
             if freqs.get('af') is not None:
                 filter_field = next(
                     (field_key for field_key in POPULATIONS[pop]['filter_AF']
@@ -227,6 +227,7 @@ class EsSearch(object):
         no_filter_indices = set()
         for index in self._indices:
             family_samples_by_id = self.samples_by_family_index[index]
+            family_samples_by_id = {key: family_samples_by_id[key] for key in sorted(family_samples_by_id.keys())}
             index_fields = self.index_metadata[index]['fields']
 
             genotypes_q = None
@@ -294,7 +295,7 @@ class EsSearch(object):
                 self._index_searches[index].append(self._search)
 
     def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary_search):
-        indices = sorted(set(self._indices), reverse = True)
+        indices = set(self._indices)
 
         paired_index_families = defaultdict(dict)
         if len(indices) > 1:
@@ -315,7 +316,7 @@ class EsSearch(object):
 
         seen_paired_indices = set()
         comp_het_q_by_index = {}
-        for index in indices:
+        for index in sorted(indices, reverse = True):
             family_samples_by_id = self.samples_by_family_index[index]
             index_fields = self.index_metadata[index]['fields']
             seen_paired_indices.add(index)
@@ -436,7 +437,7 @@ class EsSearch(object):
         return variant_results[:num_results]
 
     def _execute_multi_search(self, **kwargs):
-        indices = sorted(list(self._index_searches.keys()) or self._indices, reverse = True)
+        indices = sorted(self._index_searches.keys(), reverse = True) or self._indices
 
         if self.CACHED_COUNTS_KEY and not self.previous_search_results.get(self.CACHED_COUNTS_KEY):
             self.previous_search_results[self.CACHED_COUNTS_KEY] = {}
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -70,7 +70,7 @@ class EsSearch(object):
 
         if len(self.samples_by_family_index) != len(self.index_metadata):
             raise InvalidIndexException('Could not find expected indices: {}'.format(
-                ', '.join(set(self._indices) - set(self.index_metadata.keys()))
+                ', '.join(sorted(set(self._indices) - set(self.index_metadata.keys()), reverse = True))
             ))
 
         self.indices_by_dataset_type = defaultdict(list)
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -227,7 +227,6 @@ class EsSearch(object):
         no_filter_indices = set()
         for index in self._indices:
             family_samples_by_id = self.samples_by_family_index[index]
-            family_samples_by_id = {key: family_samples_by_id[key] for key in sorted(family_samples_by_id.keys())}
             index_fields = self.index_metadata[index]['fields']
 
             genotypes_q = None
@@ -251,7 +250,7 @@ class EsSearch(object):
                         continue
 
             if not genotypes_q:
-                for family_guid in family_samples_by_id.keys():
+                for family_guid in sorted(family_samples_by_id.keys()):
                     samples_by_id = family_samples_by_id[family_guid]
                     affected_status = self._family_individual_affected_status.get(family_guid)
 
@@ -325,7 +324,7 @@ class EsSearch(object):
             for pair_index, families in paired_index_families[index].items():
                 paired_families.update({family: pair_index for family in families})
 
-            for family_guid in family_samples_by_id.keys():
+            for family_guid in sorted(family_samples_by_id.keys()):
                 paired_index = paired_families.get(family_guid)
                 if paired_index and paired_index in seen_paired_indices:
                     continue
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from collections import defaultdict
 from copy import deepcopy
 import elasticsearch
@@ -43,7 +45,7 @@ class EsSearch(object):
 
         self._skipped_sample_count = defaultdict(int)
         if skip_unaffected_families:
-            for index, family_samples in self.samples_by_family_index.items():
+            for index, family_samples in list(self.samples_by_family_index.items()):
                 index_skipped_families = []
                 for family_guid, samples_by_id in family_samples.items():
                     affected_samples = [
@@ -63,12 +65,12 @@ class EsSearch(object):
             if len(self.samples_by_family_index) < 1:
                 raise Exception('Inheritance based search is disabled in families with no affected individuals')
 
-        self._indices = self.samples_by_family_index.keys()
+        self._indices = sorted(list(self.samples_by_family_index.keys()), reverse = True)
         self._set_index_metadata()
 
         if len(self.samples_by_family_index) != len(self.index_metadata):
             raise InvalidIndexException('Could not find expected indices: {}'.format(
-                ', '.join(set(self._indices) - set(self.index_metadata.keys()))
+                ', '.join(sorted(set(self._indices) - set(self.index_metadata.keys())))
             ))
 
         self.indices_by_dataset_type = defaultdict(list)
@@ -206,11 +208,15 @@ class EsSearch(object):
             for index in self._indices:
                 family_samples_by_id = self.samples_by_family_index[index]
                 affected_status = _get_family_affected_status(family_samples_by_id, inheritance_filter)
-                self._family_individual_affected_status.update(affected_status)
+                for family_guid in affected_status:
+                    if family_guid in self._family_individual_affected_status:
+                        self._family_individual_affected_status[family_guid].update(affected_status[family_guid])
+                    else:
+                        self._family_individual_affected_status[family_guid] = affected_status[family_guid]
 
         quality_filters_by_family = _quality_filters_by_family(quality_filter, self.samples_by_family_index, self._indices)
 
-        if inheritance_mode in {RECESSIVE, COMPOUND_HET} and not has_previous_compound_hets:
+        if inheritance_mode in [RECESSIVE, COMPOUND_HET] and not has_previous_compound_hets:
             if secondary_dataset_type:
                 self.update_dataset_type(secondary_dataset_type, keep_previous=True)
             self._filter_compound_hets(quality_filters_by_family, annotations_secondary_search)
@@ -262,7 +268,7 @@ class EsSearch(object):
                         if inheritance_mode:
                             inheritance_filter.update(INHERITANCE_FILTERS[inheritance_mode])
 
-                        if inheritance_filter.keys() == ['affected']:
+                        if list(inheritance_filter.keys()) == ['affected']:
                             raise Exception('Inheritance must be specified if custom affected status is set')
 
                         family_samples_q = _family_genotype_inheritance_filter(
@@ -277,7 +283,7 @@ class EsSearch(object):
                             family_samples_q |= x_linked_q
                     else:
                         # If no inheritance specified only return variants where at least one of the requested samples has an alt allele
-                        family_samples_q = _any_affected_sample_filter(samples_by_id.keys())
+                        family_samples_q = _any_affected_sample_filter(list(samples_by_id.keys()))
 
                     family_samples_q = _named_family_sample_q(family_samples_q, family_guid, quality_filters_by_family)
                     if not genotypes_q:
@@ -292,7 +298,7 @@ class EsSearch(object):
                 self._index_searches[index].append(self._search)
 
     def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary_search):
-        indices = set(self._indices)
+        indices = self._indices
 
         paired_index_families = defaultdict(dict)
         if len(indices) > 1:
@@ -434,7 +440,7 @@ class EsSearch(object):
         return variant_results[:num_results]
 
     def _execute_multi_search(self, **kwargs):
-        indices = self._index_searches.keys() or self._indices
+        indices = sorted(list(self._index_searches.keys()) or self._indices, reverse = True)
 
         if self.CACHED_COUNTS_KEY and not self.previous_search_results.get(self.CACHED_COUNTS_KEY):
             self.previous_search_results[self.CACHED_COUNTS_KEY] = {}
@@ -516,7 +522,7 @@ class EsSearch(object):
         if hasattr(raw_hit.meta, 'matched_queries'):
             family_guids = list(raw_hit.meta.matched_queries)
         elif self._return_all_queried_families:
-            family_guids = index_family_samples.keys()
+            family_guids = list(index_family_samples.keys())
         else:
             # Searches for all inheritance and all families do not filter on inheritance so there are no matched_queries
             alt_allele_samples = set()
@@ -835,7 +841,7 @@ class EsSearch(object):
         duplicates = 0
         results = {}
         for gene_compound_het_pair in compound_het_results:
-            gene = gene_compound_het_pair.keys()[0]
+            gene = next(iter(gene_compound_het_pair))
             compound_het_pair = gene_compound_het_pair[gene]
             if gene in results:
                 variant_ids = {variant['variantId'] for variant in compound_het_pair}
@@ -882,7 +888,7 @@ class EsSearch(object):
         num_compound_hets = 0
         num_single_variants = 0
         for variants_group in grouped_variants:
-            variants = variants_group.values()[0]
+            variants = next(iter(variants_group.values()))
 
             self.previous_search_results['grouped_results'].append(variants_group)
             if len(variants) > 1:
@@ -1122,7 +1128,7 @@ def _location_filter(genes, intervals, rs_ids, variant_ids, location_filter):
             q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop})
 
     filters = [
-        {'geneIds': (genes or {}).keys()},
+        {'geneIds': list((genes or {}).keys())},
         {'rsid': rs_ids},
         {'variantId': variant_ids},
     ]
@@ -1149,21 +1155,21 @@ def _pathogenicity_filter(pathogenicity):
         clinvar_clinical_significance_terms = set()
         for clinvar_filter in clinvar_filters:
             clinvar_clinical_significance_terms.update(CLINVAR_SIGNFICANCE_MAP.get(clinvar_filter, []))
-        pathogenicity_filter = Q('terms', clinvar_clinical_significance=list(clinvar_clinical_significance_terms))
+        pathogenicity_filter = Q('terms', clinvar_clinical_significance=sorted(list(clinvar_clinical_significance_terms)))
 
     if hgmd_filters:
         hgmd_class = set()
         for hgmd_filter in hgmd_filters:
             hgmd_class.update(HGMD_CLASS_MAP.get(hgmd_filter, []))
 
-        hgmd_q = Q('terms', hgmd_class=list(hgmd_class))
+        hgmd_q = Q('terms', hgmd_class=sorted(list(hgmd_class)))
         pathogenicity_filter = pathogenicity_filter | hgmd_q if pathogenicity_filter else hgmd_q
 
     return pathogenicity_filter
 
 
 def _annotations_filter(annotations):
-    vep_consequences = [ann for anns in annotations.values() for ann in anns]
+    vep_consequences = sorted([ann for anns in annotations.values() for ann in anns])
 
     consequences_filter = Q('terms', transcriptConsequenceTerms=vep_consequences)
 
@@ -1210,7 +1216,7 @@ def _get_sort(sort_key):
 
 
 def _sort_compound_hets(grouped_variants):
-    return sorted(grouped_variants, key=lambda variants: variants.values()[0][0]['_sort'])
+    return sorted(grouped_variants, key=lambda variants: next(iter(variants.values()))[0]['_sort'])
 
 
 def _get_compound_het_page(grouped_variants, start_index, end_index):
@@ -1218,7 +1224,7 @@ def _get_compound_het_page(grouped_variants, start_index, end_index):
     variant_results = []
     variant_count = 0
     for variants in grouped_variants:
-        curr_variant = variants.values()[0]
+        curr_variant = next(iter(variants.values()))
         if skipped < start_index:
             skipped += 1
         else:
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -250,7 +250,7 @@ class EsSearch(object):
                         continue
 
             if not genotypes_q:
-                for family_guid in sorted(family_samples_by_id.keys()):
+                for family_guid in family_samples_by_id.keys():
                     samples_by_id = family_samples_by_id[family_guid]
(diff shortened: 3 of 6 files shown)


seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
```diff
@@ -7,7 +7,7 @@
 from django.core.management import call_command
 from django.test import TestCase
 from django.core.management.base import CommandError
-from seqr.utils.elasticsearch.es_utils_tests import PARSED_VARIANTS
+from seqr.utils.elasticsearch.es_utils_2_3_tests import PARSED_VARIANTS
```
ShifaSZ, line 10: This test uses a constant defined by the elasticsearch utils test. I'm not sure whether we should move the constant to a common test util file.

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_gene_agg_search.py
```diff
@@ -12,9 +14,11 @@ class EsGeneAggSearch(EsSearch):
     CACHED_COUNTS_KEY = None
 
     def aggregate_by_gene(self):
-        searches = {self._search}
+        searches = [self._search]
         for index_searches in self._index_searches.values():
-            searches.update(index_searches)
+            for index_search in index_searches:
+                if index_search not in searches:
+                    searches.append(index_search)
```
ShifaSZ, line 21: Python 3 reports `TypeError: unhashable type: 'dict'` for `searches = {self._search}`. I changed to use a `list` rather than a `set`. The function should be equivalent. But I'm not sure whether it is a best way to do it.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -43,7 +45,7 @@ def __init__(self, families, previous_search_results=None, skip_unaffected_famil
 
         self._skipped_sample_count = defaultdict(int)
         if skip_unaffected_families:
-            for index, family_samples in self.samples_by_family_index.items():
+            for index, family_samples in list(self.samples_by_family_index.items()):
```
ShifaSZ, line 48: It should be okay without adding `list()`. But, Python 3 reports `dictionary changed size during iteration` error. After adding `list()`, the error disappeared.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -63,12 +65,12 @@ def __init__(self, families, previous_search_results=None, skip_unaffected_famil
             if len(self.samples_by_family_index) < 1:
                 raise Exception('Inheritance based search is disabled in families with no affected individuals')
 
-        self._indices = self.samples_by_family_index.keys()
+        self._indices = sorted(list(self.samples_by_family_index.keys()), reverse = True)
```
ShifaSZ, line 68: It is weird that the behavior of the code depends on the order of `self._indices` very much. Without sorting, the results of `elasticsearch` keep changing between each run.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -63,12 +65,12 @@ def __init__(self, families, previous_search_results=None, skip_unaffected_famil
...
         self._set_index_metadata()
 
         if len(self.samples_by_family_index) != len(self.index_metadata):
             raise InvalidIndexException('Could not find expected indices: {}'.format(
-                ', '.join(set(self._indices) - set(self.index_metadata.keys()))
+                ', '.join(sorted(set(self._indices) - set(self.index_metadata.keys())))
```
ShifaSZ, line 73: The purpose of adding `sorted()` here is to keep the expected test result to be stable.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -206,11 +208,15 @@ def filter_by_annotation_and_genotype(self, inheritance, quality_filter=None, an
             for index in self._indices:
                 family_samples_by_id = self.samples_by_family_index[index]
                 affected_status = _get_family_affected_status(family_samples_by_id, inheritance_filter)
-                self._family_individual_affected_status.update(affected_status)
+                for family_guid in affected_status:
+                    if family_guid in self._family_individual_affected_status:
+                        self._family_individual_affected_status[family_guid].update(affected_status[family_guid])
+                    else:
+                        self._family_individual_affected_status[family_guid] = affected_status[family_guid]
```
ShifaSZ, line 215: I'm not sure it is correct or not. The original code replaces the `affected_status` of a family with `family_guid` rather than updates them. I have changed it to update the `affected_status` of the family if the status exists.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -292,7 +298,7 @@ def _filter_by_genotype(self, inheritance_mode, inheritance_filter, quality_filt
                 self._index_searches[index].append(self._search)
 
     def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary_search):
-        indices = set(self._indices)
+        indices = self._indices
```
ShifaSZ, line 301: `set(self._indices)` makes the pre-sorted list out of order and causes un-predictable behavior in `elasticsearch` searching operations.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -434,7 +440,7 @@ def _process_single_search_response(self, parsed_response, page=1, num_results=1
         return variant_results[:num_results]
 
     def _execute_multi_search(self, **kwargs):
-        indices = self._index_searches.keys() or self._indices
+        indices = sorted(list(self._index_searches.keys()) or self._indices, reverse = True)
```
ShifaSZ, line 443: In Python 3, the order of `list(self._index_searches.keys())` is not fixed. Therefore, `sorted()` is added.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -1149,21 +1155,21 @@ def _pathogenicity_filter(pathogenicity):
         clinvar_clinical_significance_terms = set()
         for clinvar_filter in clinvar_filters:
             clinvar_clinical_significance_terms.update(CLINVAR_SIGNFICANCE_MAP.get(clinvar_filter, []))
-        pathogenicity_filter = Q('terms', clinvar_clinical_significance=list(clinvar_clinical_significance_terms))
+        pathogenicity_filter = Q('terms', clinvar_clinical_significance=sorted(list(clinvar_clinical_significance_terms)))
```
ShifaSZ, line 1158: The reason for adding `sorted()` is to make the test result has a predictable order.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -1149,21 +1155,21 @@ def _pathogenicity_filter(pathogenicity):
...
 
     if hgmd_filters:
         hgmd_class = set()
         for hgmd_filter in hgmd_filters:
             hgmd_class.update(HGMD_CLASS_MAP.get(hgmd_filter, []))
 
-        hgmd_q = Q('terms', hgmd_class=list(hgmd_class))
+        hgmd_q = Q('terms', hgmd_class=sorted(list(hgmd_class)))
```
ShifaSZ, line 1165: The reason for adding `sorted()` is to make the test result has a predictable order.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -1149,21 +1155,21 @@ def _pathogenicity_filter(pathogenicity):
...
         pathogenicity_filter = pathogenicity_filter | hgmd_q if pathogenicity_filter else hgmd_q
 
     return pathogenicity_filter
 
 
 def _annotations_filter(annotations):
-    vep_consequences = [ann for anns in annotations.values() for ann in anns]
+    vep_consequences = sorted([ann for anns in annotations.values() for ann in anns])
```
ShifaSZ, line 1172: The reason for adding `sorted()` is to make the test result has a predictable order.

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1333,7 +1338,7 @@ def test_invalid_get_es_variants(self):
         self.mock_es_client.indices.get_mapping.side_effect = lambda **kwargs: {}
         with self.assertRaises(InvalidIndexException) as cm:
             get_es_variants(results_model)
-        self.assertEqual(str(cm.exception), 'Could not find expected indices: test_index_sv, test_index')
+        self.assertEqual(str(cm.exception), 'Could not find expected indices: test_index, test_index_sv')
```
ShifaSZ, line 1341: Updated the order of the expected data after adding `sorted()`

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1527,14 +1532,14 @@ def test_filtered_get_es_variants(self):
...
-                                'inframe_insertion',
                                 'inframe_deletion',
+                                'inframe_insertion',
+                                'intergenic_variant',
                             ]
                         }},
                         {'terms': {
                             'clinvar_clinical_significance': [
-                                'Pathogenic', 'Likely_pathogenic', 'Pathogenic/Likely_pathogenic'
+                                'Likely_pathogenic', 'Pathogenic', 'Pathogenic/Likely_pathogenic'
```
ShifaSZ, line 1542: Updated the order of the expected data after adding `sorted()`

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1755,7 +1760,7 @@ def test_compound_het_get_es_variants_secondary_annotation(self):
 
         annotation_query = {'bool': {'should': [
             {'terms': {'transcriptConsequenceTerms': ['frameshift_variant']}},
-            {'terms': {'transcriptConsequenceTerms': ['intron', 'frameshift_variant']}}]}}
+            {'terms': {'transcriptConsequenceTerms': ['frameshift_variant', 'intron']}}]}}
```
ShifaSZ, line 1763: Updated the order of the expected data after adding `sorted()`

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1192,6 +1194,11 @@ def assertSameSearch(self, executed_search, expected_search_params):
         }
 
         if expected_search_params['filters']:
+            for i in range(len(expected_search_params['filters'])):
+                if 'bool' in expected_search_params['filters'][i] and 'must' in expected_search_params['filters'][i]['bool']:
+                    expected_search_params['filters'][i]['bool']['must'] = mock.ANY
+                if 'bool' in expected_search_params['filters'][i] and 'should' in expected_search_params['filters'][i]['bool']:
+                    expected_search_params['filters'][i]['bool']['should'] = mock.ANY
```
ShifaSZ, line 1201: There are many ordering issues in the `'filters'` parameters. Using `mock.ANY` will suppress the order errors but also hide possible real problems. Should I try to sort the items and do the comparison?

seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
hanars: yes we should move the constants to the main test_utils file where the other test constants live

seqr/utils/elasticsearch/es_gene_agg_search.py
hanars: this was working in python 2 and as far as I know this should not have changed between python 2 and 3. Can you find out why it changed?

seqr/utils/elasticsearch/es_search.py
hanars: since you added `sorted` to the assignment for `self._indices` above, this should already be sorted here

seqr/utils/elasticsearch/es_search.py
hanars: This probably does the same thing but possibly doesn't and is way longer and less readable so why change it?

seqr/utils/elasticsearch/es_search.py
hanars: it is crucial to deduplicate here which is what casting to a set does, so this is an unacceptable changed

seqr/utils/elasticsearch/es_utils_2_3_tests.py
hanars: it is unacceptable to make any changes here

seqr/utils/elasticsearch/es_search.py
ShifaSZ: Although `self._indices` is sorted, the new set generated by the subtraction is still possible out of order.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: If we agree on sorting `self._indices`, we can keep the existing code and it has the same behavior in both Python 2 and 3. If we don't sort `self._indices`, the result of the code will be different for Python 2 and 3. The new code doesn't depend on the order of `self._indices`.
I can change it back for now.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: The order is also important. Since deduplicate is crucial, I will try to sort the result of the casting.

seqr/management/tests/lift_variant_to_hg38_2_3_tests.py
ShifaSZ: Moved PARSED_VARIANTS and PARSED_SV_VARIANTS to seqr.views.utils.test_utils.

seqr/utils/elasticsearch/es_gene_agg_search.py
ShifaSZ: In Python 2:
```
>>> from elasticsearch_dsl import Search
>>> a=Search()
>>> b={a}
>>> b
set([<elasticsearch_dsl.search.Search object at 0x1031bb990>])
```
However, in Python 3:
```
>>> from elasticsearch_dsl import Search
>>> a=Search()
>>> b={a}
TypeError: unhashable type: 'Search'
```
According to the [document](https://stackoverflow.com/questions/17493307/creating-set-of-objects-of-user-defined-class-in-python), the class must provide `__eq__()` and `__hash__()` methods to allow its objects to be set elements. `elasticsearch_dsl.Search` does has `__eq__()` but doesn't have `__hash_()`. I did an experiment by adding below to the Search class:
```
    def __hash__(self):
        return hash(str(self))
```
Then, `searches = {self._search}` can run without an error.
These seem about Python 3. Python 2 needs neither `__eq__()` nor `__hash__()` for the set.

ShifaSZ: We don't like to change the elasticsearch library. So we have to avoid using a Search object as a set member.

hanars: but if it is lacking an `__eq__` method, how do we know `index_search not in searches` will work at all?

seqr/utils/elasticsearch/es_search.py
hanars: so this is an okay change to make, but its confusing to me because in python2 dicts are not ordered by default, but in python 3 there are. So I would expect the opposite

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_gene_agg_search.py
ShifaSZ: Hmm..., good question. But it has `__eq__` anyway.

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1050,6 +798,19 @@ def _set_cache(k, v):
 MOCK_LIFTOVERS['hg19'].convert_coordinate.side_effect = lambda chrom, pos: [[chrom, pos + 10]]
 
 
+def sort_array(a):
+    sd={str(d):d for d in a}
+    return [sd[i] for i in sorted(sd.keys())]
+
+def sort_search(filter):
```
hanars, line 805: Codacy found an issue: [Redefining built-in 'filter'](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912)

seqr/utils/elasticsearch/es_search.py
ShifaSZ: I saw the order of `set(self._indices) - set(self.index_metadata.keys())` changed from time to time in Python 3. But, since it is not ideal by adding `sorted()` for test purpose in production code, I updated the result assertion to accept different orders of the indices list.

hanars:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_utils_2_3_tests.py
hanars, line 805: Codacy found an issue: [Redefining built-in 'filter'](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912)

seqr/utils/elasticsearch/es_search.py
ShifaSZ: In Python 2, `dict.keys()` produces a sequence order (increasing) while starting from Python 3.7 it preserves the order of adding the keys. Below are the examples:
Python 2:
```
>>> d = {'a': 0, 'b': 1, 'c': 2}
>>> e = {'b':0, 'a':1, 'c': 2}
>>> d.keys()
['a', 'c', 'b']
>>> e.keys()
['a', 'c', 'b']
>>> 
```
Python 3.7 or higher:
```
>>> d = {'a': 0, 'b': 1, 'c': 2}
>>> e = {'b':0, 'a':1, 'c': 2}
>>> d.keys()
dict_keys(['a', 'b', 'c'])
>>> e.keys()
dict_keys(['b', 'a', 'c'])
>>> 
```

seqr/utils/elasticsearch/es_utils_2_3_tests.py
ShifaSZ: Added sorting to both executed and expected filters to avoid order issues while comparing the `list` objects.

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_gene_agg_search.py
```diff
@@ -12,9 +14,9 @@ class EsGeneAggSearch(EsSearch):
     CACHED_COUNTS_KEY = None
 
     def aggregate_by_gene(self):
-        searches = {self._search}
+        searches = [self._search]
         for index_searches in self._index_searches.values():
-            searches.update(index_searches)
+            searches = searches + [index_search for index_search in index_searches if index_search not in searches]
```
hanars, line 19: Given that sets don't work for these objects in python 3, why would we assume `in` will work here at all?

seqr/utils/elasticsearch/es_search.py
```diff
@@ -292,7 +294,7 @@ def _filter_by_genotype(self, inheritance_mode, inheritance_filter, quality_filt
                 self._index_searches[index].append(self._search)
 
     def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary_search):
-        indices = set(self._indices)
+        indices = sorted(set(self._indices), reverse = True)
```
hanars, line 297: This was turned into a set to improve the lookups for `index in indices`. If you are just going to immediately turn it back into a list theres no point in doing this

seqr/utils/elasticsearch/es_search.py
```diff
@@ -322,7 +324,7 @@ def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary
             for pair_index, families in paired_index_families[index].items():
                 paired_families.update({family: pair_index for family in families})
 
-            for family_guid in sorted(family_samples_by_id.keys()):
+            for family_guid in family_samples_by_id.keys():
```
hanars, line 327: why would you take away sorting?

seqr/utils/elasticsearch/es_search.py
```diff
@@ -434,7 +436,7 @@ def _process_single_search_response(self, parsed_response, page=1, num_results=1
         return variant_results[:num_results]
 
     def _execute_multi_search(self, **kwargs):
-        indices = self._index_searches.keys() or self._indices
+        indices = sorted(list(self._index_searches.keys()) or self._indices, reverse = True)
```
hanars, line 439: given `self._indices` is already sorted, I think this would be neater as `sorted(list(self._index_searches.keys()), reverse = True)  or self._indices`

seqr/utils/elasticsearch/es_gene_agg_search.py
ShifaSZ, line 19: I can't make `{self._search}` work for Python 3 because I can't modify the `Search` class of `Elasticsearch`. Using a `list` can be a workaround for this case.

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1194,9 +954,10 @@ def assertSameSearch(self, executed_search, expected_search_params):
         if expected_search_params['filters']:
             expected_search['query'] = {
                 'bool': {
-                    'filter': expected_search_params['filters']
+                    'filter': sort_filter(expected_search_params['filters'])
```
hanars, line 957: `expected_search_params` are created as synthetic data in the test. Just put them in in the correct order, don't post process them

seqr/utils/elasticsearch/es_utils_2_3_tests.py
```diff
@@ -1333,7 +1094,9 @@ def test_invalid_get_es_variants(self):
         self.mock_es_client.indices.get_mapping.side_effect = lambda **kwargs: {}
         with self.assertRaises(InvalidIndexException) as cm:
             get_es_variants(results_model)
-        self.assertEqual(str(cm.exception), 'Could not find expected indices: test_index_sv, test_index')
+        message = str(cm.exception).split(': ')
+        self.assertEqual(message[0], 'Could not find expected indices')
+        self.assertSetEqual(set(message[1].split(', ')), {'test_index_sv', 'test_index'})
```
hanars, line 1099: I thought you sorted the indices so this would be a predictable order

seqr/utils/elasticsearch/es_search.py
```diff
@@ -248,7 +250,7 @@ def _filter_by_genotype(self, inheritance_mode, inheritance_filter, quality_filt
                         continue
 
             if not genotypes_q:
-                for family_guid in sorted(family_samples_by_id.keys()):
+                for family_guid in family_samples_by_id.keys():
```
ShifaSZ, line 253: I found `sorted()` is unnecessary here because the order is not important for the codes in the loop.

seqr/utils/elasticsearch/es_gene_agg_search.py
hanars: I know, but given that sets don't work means that identity checking doesn't work, which means the line you added here that says `if index_search not in searches` shouldn't work either, so why is it here?

seqr/utils/elasticsearch/es_search.py
```diff
@@ -1020,10 +1022,10 @@ def _quality_filters_by_family(quality_filter, samples_by_family_index, indices)
     if any(quality_filter[field] for field in quality_field_configs.keys()):
         for index in indices:
             family_samples_by_id = samples_by_family_index[index]
-            for family_guid, samples_by_id in family_samples_by_id.items():
+            for family_guid, samples_by_id in sorted(family_samples_by_id.items()):
                 quality_q = Q()
-                for sample_id in samples_by_id.keys():
-                    for field, config in quality_field_configs.items():
+                for sample_id in sorted(samples_by_id.keys()):
+                    for field, config in sorted(quality_field_configs.items()):
```
ShifaSZ, line 1028: The `sorted()` added here is required because there are `&` operations on query filters such as `quality_q &= ~Q(q) | ~Q('term', samples_num_alt_1=sample_id)` and `quality_q &= ~Q(q)`. 
I did an experiment which shows that query filter `&` operation is order dependent:
```
>>>sample1_q = Bool(should=[Term(samples_num_alt_1=u'HG00732'), Term(samples_num_alt_2=u'HG00732')])
>>>sample2_q = Bool(should=[Term(samples_num_alt_1=u'HG00731'), Term(samples_num_alt_2=u'HG00731')])
>>>sample2_q & sample1_q
Bool(
  minimum_should_match=1,
  must=[Bool(minimum_should_match=1, should=[Term(samples_num_alt_1=u'HG00732'), Term(samples_num_alt_2=u'HG00732')])],
  should=[Term(samples_num_alt_1=u'HG00731'), Term(samples_num_alt_2=u'HG00731')])
>>>sample1_q & sample2_q
Bool(
  minimum_should_match=1,
  must=[Bool(minimum_should_match=1, should=[Term(samples_num_alt_1=u'HG00731'), Term(samples_num_alt_2=u'HG00731')])],
  should=[Term(samples_num_alt_1=u'HG00732'), Term(samples_num_alt_2=u'HG00732')])
```
From Elasticsearch point of view, the results of `sample2_q & sample1_q` and `sample1_q & sample2_q` are the same. But, when making comparison for testing, they are different. To make the results predictable, we have to keep the order when use the `&` operator on query filters.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -1058,7 +1060,7 @@ def _family_genotype_inheritance_filter(inheritance_mode, inheritance_filter, sa
                 individual_genotype_filter[individual.guid] = REF_REF
 
     is_sv_comp_het = inheritance_mode == COMPOUND_HET and 'samples' in index_fields
-    for sample_id, sample in samples_by_id.items():
+    for sample_id, sample in sorted(samples_by_id.items()):
```
ShifaSZ, line 1063: Same reason as above.

seqr/utils/elasticsearch/es_utils_2_3_tests.py
ShifaSZ: I removed the sorting in es_search.py which looks like below:
```
            raise InvalidIndexException('Could not find expected indices: {}'.format(
                ', '.join(set(self._indices) - set(self.index_metadata.keys()))
            ))
```
Although `self._indices` is sorted, the element order of `set(self._indices) - set(self.index_metadata.keys())` can change each run.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: I think `sorted()` is not necessary because the code in the loop is not depending on the order.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: True. Change it to `sorted(self._index_searches.keys(), reverse = True) or self._indices`.

seqr/utils/elasticsearch/es_gene_agg_search.py
ShifaSZ: The base class of `Search` which is `Request` has an `__eq__` method. The method is used when identity is being checked.

ShifaSZ: `set()` requires both `__eq__` and `__hash__` while `in` requires only `_eq_`.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: The order of the indices is real import according to the codes from es_search.py line 208~211:
```
            for index in self._indices:
                family_samples_by_id = self.samples_by_family_index[index]
                affected_status = _get_family_affected_status(family_samples_by_id, inheritance_filter)
                self._family_individual_affected_status.update(affected_status)
```
The index comes after will overwrite the previous index's `affected_status` with the same family.

If the lookup performance is a concern, I could move the `sorted()` to a few lines after. It looks like below:
```
    def _filter_compound_hets(self, quality_filters_by_family, annotations_secondary_search):
        indices = set(self._indices)

        paired_index_families = defaultdict(dict)
        if len(indices) > 1:
            sv_indices = [
                index for index in self.indices_by_dataset_type[Sample.DATASET_TYPE_SV_CALLS] if index in indices
            ]
            variant_indices = [
                index for index in self.indices_by_dataset_type[Sample.DATASET_TYPE_VARIANT_CALLS] if index in indices
            ]
       <<more codes skipped>>

        seen_paired_indices = set()
        comp_het_q_by_index = {}
        for index in sorted(indices, reverse = True):
```

seqr/utils/elasticsearch/es_utils_2_3_tests.py
ShifaSZ: Order changes to the expected data will be hard to review. Are you sure that you want that?

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_search.py
```diff
@@ -127,7 +129,7 @@ def filter(self, new_filter):
 
     def filter_by_frequency(self, frequencies):
         q = Q()
-        for pop, freqs in frequencies.items():
+        for pop, freqs in sorted(frequencies.items()):
```
ShifaSZ, line 132: Added sorting to keep Python 3 and 2 generate the same search filters.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: After removing the sorting on the filters, the `sorted()` here becomes necessary. Added it back.

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

seqr/utils/elasticsearch/es_search.py
ShifaSZ: After removing the sorting on the filters, the sorted() here becomes necessary. Added it back.
Ideally, we shouldn't add `sorted()` here because it doesn't change the search function but wastes some computation.

seqr/utils/elasticsearch/es_utils_2_3_tests.py
ShifaSZ: I have removed both sorting for executed and expected filters by adding a few more `sorted()` in the filter creation codes.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: I removed sorting here and changed back a `sorted()` in a line following.

seqr/utils/elasticsearch/es_gene_agg_search.py
hanars: in general, its better form to do `searches += ...` than `searches = searches + ...`

seqr/utils/elasticsearch/es_utils_2_3_tests.py
hanars: then you should sort `set(self._indices) - set(self.index_metadata.keys())`

seqr/utils/elasticsearch/es_gene_agg_search.py
ShifaSZ: Okay, I will make the change.

seqr/utils/elasticsearch/es_utils_2_3_tests.py
ShifaSZ: Okay, I will change them back.

hanars:


```diff

Complexity increasing per file
==============================
- seqr/utils/elasticsearch/es_utils_2_3_tests.py  19
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5731912&bid=18751204)

PR #1371, opened 2020-07-14
fix multi index search when indices have subset of individuals

hanars:
There is a bug in seqr (https://github.com/macarthur-lab/seqr/issues/1367) that occurs if a family has multiple indices containing different individuals. For instance, if the variant index has all 3 individuals, but the SV index only has 2 of them, seqr search gives an unhandled key error. This PR fixes it

### DIFF
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -65,7 +65,7 @@ class EsSearch(object):
             if len(self.samples_by_family_index) < 1:
                 raise Exception('Inheritance based search is disabled in families with no affected individuals')
 
-        self._indices = sorted(list(self.samples_by_family_index.keys()), reverse = True)
+        self._indices = sorted(list(self.samples_by_family_index.keys()))
         self._set_index_metadata()
 
         if len(self.samples_by_family_index) != len(self.index_metadata):
@@ -208,7 +208,10 @@ class EsSearch(object):
             for index in self._indices:
                 family_samples_by_id = self.samples_by_family_index[index]
                 affected_status = _get_family_affected_status(family_samples_by_id, inheritance_filter)
-                self._family_individual_affected_status.update(affected_status)
+                for family_guid, family_affected_status in affected_status.items():
+                    if family_guid not in self._family_individual_affected_status:
+                        self._family_individual_affected_status[family_guid] = {}
+                    self._family_individual_affected_status[family_guid].update(family_affected_status)
 
         quality_filters_by_family = _quality_filters_by_family(quality_filter, self.samples_by_family_index, self._indices)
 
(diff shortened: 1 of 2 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, families, previous_search_results=None, skip_unaffected_famil
             if len(self.samples_by_family_index) < 1:
                 raise Exception('Inheritance based search is disabled in families with no affected individuals')
 
-        self._indices = sorted(list(self.samples_by_family_index.keys()), reverse = True)
+        self._indices = sorted(list(self.samples_by_family_index.keys()))
```
hanars, line 68: The edge case we are testing is if different indices have different subsets of individuals loaded for a family, which is true in our test data for family 2: `test_index_sv` has individuals 4 and 5, and `test_index` has individuals 4, 5, and 6. However, because of the iteration order `test_index` was updated last and since it has the superset there were no errors. By reversing the iteration order here so the index with fewer samples is looked at last we duplicate the production error, and a couple of test start to fail with the same key error we see in the original bug report. Since we only added the sorting here for test stability in the first place, there are no negative consequences for swapping the order, and now going forward that edge case will have a unit test so we don;t reintroduce the bug

seqr/utils/elasticsearch/es_search.py
```diff
@@ -208,7 +208,10 @@ def filter_by_annotation_and_genotype(self, inheritance, quality_filter=None, an
             for index in self._indices:
                 family_samples_by_id = self.samples_by_family_index[index]
                 affected_status = _get_family_affected_status(family_samples_by_id, inheritance_filter)
-                self._family_individual_affected_status.update(affected_status)
+                for family_guid, family_affected_status in affected_status.items():
+                    if family_guid not in self._family_individual_affected_status:
+                        self._family_individual_affected_status[family_guid] = {}
+                    self._family_individual_affected_status[family_guid].update(family_affected_status)
```
hanars, line 214: this is the actual bug fix. If index_1 has `affected_status` of `{'F2': {'4': 'A', '5': 'U'}}` and index_2 has `affected_status` of `{'F2': {'5': 'U', '6': 'U'}}` than under the old code `self._family_individual_affected_status['F2']` would equal either `{'4': 'A', '5': 'U'}` or `{'5': 'U', '6': 'U'}` depending on the sorted index order. Under this new code, and what we actually want, it becomes `{'4': 'A', '5': 'U', '6': 'U'}` and we stop getting key errors

seqr/utils/elasticsearch/es_search.py
ShifaSZ: Can we combine the individuals of the family from different indices instead of overwriting those from the previous index?

seqr/utils/elasticsearch/es_search.py
ShifaSZ: It is exactly what I mean for "combining individuals" in my previous comment.

PR #1387, opened 2020-07-22
test igv response headers

hanars:


hanars:
turns out IGV was broken before the python 3 deployment and this is not in any way the issue

PR #1397, opened 2020-07-24
Postgres upgrade

hanars:


deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -8,7 +8,7 @@
 import time
 
 from deploy.servctl_utils.other_command_utils import check_kubernetes_context, set_environment
-from hail_elasticsearch_pipelines.kubernetes.kubectl_utils import is_pod_running, get_pod_name, get_node_name, run_in_pod, \
+from deploy.servctl_utils.kubectl_utils import is_pod_running, get_pod_name, get_node_name, run_in_pod, \
```
hanars, line 11: I've been meaning for a while to move the utility functions we need into our own repo instead of importing them from a submodule. I finally bit the bullet here because there is a unicode handling bug that is fixed in the seqr `run` utility but not in the imported one, so I copied over the file we need and changed it to use our working utility function

hanars:


```diff

Complexity increasing per file
==============================
- deploy/servctl_utils/kubectl_utils.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/hanars/seqr/pullRequest?prid=5908917&bid=19357643)

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -211,9 +211,6 @@ def deploy_external_connector(settings, connector_name):
 def deploy_elasticsearch(settings):
     print_separator("elasticsearch")
 
-    if settings["DELETE_BEFORE_DEPLOY"]:
-        delete_pod("elasticsearch", settings)
-
```
hanars, line 216: Currently all the deploy helper functions call this conditional delete, than deploy docker, then deploy the pod. We want to wait and delete the pod until after the docker deploy is done to minimize downtime, but rather than moving this if block around 6 time I just moved it into `deploy_pod` so theres less copy-pasting

moved this into `deploy_pod` since every time that function is called this needs to be too and this way theres less copy-paste. Plu

deploy/servctl_utils/kubectl_utils.py
```diff
@@ -0,0 +1,167 @@
+import logging
+import time
+
+from deploy.servctl_utils.shell_utils import run
```
hanars, line 4: copy/pasted this file directly from hail_elasticsearch_pipelines where we were previously importing it from, and then changed this import to use out implementation of run instead of the one in the other repo. They are almost identical, except ours actually handles unicode errors properly. I made no other change to the file

deploy/servctl_utils/deploy_command_utils.py
ShifaSZ: It is good to move the utility file into our repository.

PR #1401, opened 2020-07-28
bump vulnerable dependency version

hanars:


ShifaSZ:
It failed in Python 2 tests. Do we care about it?

PR #1403, opened 2020-07-28
Added Google Auth

ShifaSZ:
This PR is only for reviewing an experimental Google Auth for the seqr. It is not going to be merged. The PR makes it easier for comparison, reviewing, and commenting.
In this experiment, the package `django-allauth` is used. Some packages have to be upgraded (see the details in the file `requirements.txt`.

Use the command `python manage.py runserver --nostatic` to run your own seqr. and run command `python manage.py createsuperuser` to create a superuser.

Usages and testing:
1. open `http://localhost:8000/accounts/login` to sign in locally
2. Add `sites`, `Google client ID` and `secret key` in the page of `http://localhost:8000/admin` [See this article](https://medium.com/@whizzoe/in-5-mins-set-up-google-login-to-sign-up-users-on-django-e71d5c38f5d5) for the details of configuration. (Need minimum `staff` level access).
3. Login in with your existing seqr account from `http://localhost:8000/login`, then, open `http://localhost:8000/accounts/social/connections/` to add/remove your Google Auth to your local account
4. open `http://localhost:8000/accounts/logout` to logout.
5. use `http://localhost:8000/accounts/google/login` to sign in with Google Auth.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,9 @@
-Django==1.11.29                  # core server-side framework. Behind a major version due to missing Python 2 support
+Django==3.0.8                    # core server-side framework. Behind a major version due to missing Python 2 support
 django-anymail==7.0.0            # for sending emails using cloud-based mail service providers
-django-guardian==1.5.1           # object-level permissions for database records. Behind a major version due to missing Python 2 support
+django-guardian==2.3.0           # object-level permissions for database records. Behind a major version due to missing Python 2 support
+django-allauth==0.42.0           # for Google Auth
 django-hijack==2.1.10            # allows admins to login as other user
-django-cors-headers==3.0.2       # allows CORS requests for client-side development
+django-cors-headers==3.4.0       # allows CORS requests for client-side development
 
 elasticsearch==6.4.0             # elasticsearch client
 elasticsearch-dsl==6.4.0	     # elasticsearch query utilities
@@ -18,4 +19,4 @@ requests-toolbelt==0.9.1         # for troubleshooting requests
 slacker==0.13.0                  # library for sending slack messages
 slugify==0.0.1                   # used for encoding names for guids
 tqdm==4.40.2                     # convenient way to create progress bar for long-running command-line operations
-whitenoise==3.3.0                # simplified static file handling. Behind a major version due to missing Python 2 support
+whitenoise==5.1.0                # simplified static file handling. Behind a major version due to missing Python 2 support
--- a/seqr/views/apis/auth_api.py
+++ b/seqr/views/apis/auth_api.py
@@ -46,6 +46,6 @@ def login_required_error(request):
 
     This is used to redirect AJAX HTTP handlers to the login page.
     """
-    assert not request.user.is_authenticated()
+    assert not request.user.is_authenticated
 
     return create_json_response({}, status=401, reason="login required")
--- a/seqr/views/apis/saved_variant_api_tests.py
+++ b/seqr/views/apis/saved_variant_api_tests.py
@@ -75,6 +75,7 @@ COMPOUND_HET_5_JSON = {
 
 class SavedVariantAPITest(AuthenticationTestCase):
     fixtures = ['users', '1kg_project']
+    databases = {'reference_data', 'default'}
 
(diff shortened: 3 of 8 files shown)


hanars:
Can you make sure to document somewhere your decision to use `django-allauth` here - did you look into other packages and choose this one for a reason? Or is this sort of a proof of concept so you picked one to see if ti worked, but it might be worth looking into other options? Whatever the motivation, this is the kinfd of thing that would be good to keep recorded in the google doc you have going

requirements.txt
```diff
@@ -1,8 +1,9 @@
-Django==1.11.29                  # core server-side framework. Behind a major version due to missing Python 2 support
+Django==3.0.8                    # core server-side framework. Behind a major version due to missing Python 2 support
```
hanars, line 1: I was actually planning to bump a lot of these packages in the coming weeks anyways! Good to know that it doesn't break seqr to do so :)

settings.py
```diff
@@ -180,8 +187,23 @@
 AUTHENTICATION_BACKENDS = (
     'django.contrib.auth.backends.ModelBackend',
     'guardian.backends.ObjectPermissionBackend',
+    'allauth.account.auth_backends.AuthenticationBackend',
```
hanars, line 190: so this looks like we support multiple auth backends - is that the intention here?

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -75,6 +75,7 @@
 
 class SavedVariantAPITest(AuthenticationTestCase):
     fixtures = ['users', '1kg_project']
+    databases = {'reference_data', 'default'}
```
ShifaSZ, line 78: It is required by the new version of the Django to access the database in a different 'application'.

settings.py
ShifaSZ: The `django.contrib.auth.backends.ModelBackend` is always required event for Google Auth. I'm not sure about `guardian.backends.ObjectPermissionBackend`

seqr/views/apis/auth_api.py
```diff
@@ -46,6 +46,6 @@ def login_required_error(request):
 
     This is used to redirect AJAX HTTP handlers to the login page.
     """
-    assert not request.user.is_authenticated()
+    assert not request.user.is_authenticated
```
ShifaSZ, line 49: The `is_authenticated` attribute has changed from a function to a variable in the new version of the WSGI request.

seqr/views/react_app.py
```diff
@@ -22,7 +22,7 @@ def no_login_main_app(request, *args, **kwargs):
     user_token = kwargs.get('user_token')
     if user_token:
         initial_json['newUser'] = _get_json_for_user(User.objects.get(password=user_token))
-    elif not request.user.is_anonymous():
+    elif not request.user.is_anonymous:
```
ShifaSZ, line 25: The `is_anonymous` attribute has changed from a function to a variable in the new version of the WSGI request.

ShifaSZ:
> Can you make sure to document somewhere your decision to use `django-allauth` here - did you look into other packages and choose this one for a reason? Or is this sort of a proof of concept so you picked one to see if ti worked, but it might be worth looking into other options? Whatever the motivation, this is the kinfd of thing that would be good to keep recorded in the google doc you have going  

I have just picked one of the packages randomly after searching on the Internet. I can look into more options to see what's the situation.

requirements.txt
ShifaSZ: The old version won't work with the `django-allauth` package.

hanars:
> > Can you make sure to document somewhere your decision to use `django-allauth` here - did you look into other packages and choose this one for a reason? Or is this sort of a proof of concept so you picked one to see if ti worked, but it might be worth looking into other options? Whatever the motivation, this is the kinfd of thing that would be good to keep recorded in the google doc you have going
> 
> I have just picked one of the packages randomly after searching on the Internet. I can look into more options to see what's the situation.

I think looking into a few options would be good - maybe as a separate branch for each approach? Or if you do some research and decide this is the best one, just make sure to document that decision somewhere

PR #1414, opened 2020-08-05
add sv tag types

hanars:
Adds tag types as requested in https://github.com/macarthur-lab/seqr-private/issues/894

We add tag type fixture data in 2 places: `variant_tag_types` is the master list of all tag types which is loaded whenever new instances of seqr are set up, and should always include all out core tag types. `new_tag_types` is a transitional list that allows us to add the new types to the existing database by running `./manage.py loaddata new_tag_types`, and once that command has been run its okay to update it to a different set of tags for the next update as we do here

### DIFF
--- a/seqr/fixtures/new_variant_tag_types.json
+++ b/seqr/fixtures/new_variant_tag_types.json
@@ -3,51 +3,48 @@
     "model": "seqr.varianttagtype",
     "pk": null,
     "fields": {
-        "guid": "VTT_test_segregation",
-        "created_date": "2018-04-27T00:00:00.000Z",
+        "guid": "VTT_validate_sv",
+        "created_date": "2020-08-05T00:00:00.000Z",
         "created_by": null,
         "last_modified_date": null,
         "project": null,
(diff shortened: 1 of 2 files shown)


ShifaSZ:
> Adds tag types as requested in [macarthur-lab/seqr-private#894](https://github.com/macarthur-lab/seqr-private/issues/894)
> 
> We add tag type fixture data in 2 places: `variant_tag_types` is the master list of all tag types which is loaded whenever new instances of seqr are set up, and should always include all out core tag types. `new_tag_types` is a transitional list that allows us to add the new types to the existing database by running `./manage.py loaddata new_tag_types`, and once that command has been run its okay to update it to a different set of tags for the next update as we do here

I'm a little confused that the fixture data are not only used for testing but also for some real functions. But anyway, I believe it works. I have a question is why the new tag types are added to both `variant_tag_types` and `new_tag_types`? According to your comments, the `variant_tag_types` is for core tag types while `new_tag_types` is for transitional tag types.

PR #1427, opened 2020-08-18
Bump dependencies

hanars:


ui/package.json
```diff
@@ -36,6 +36,7 @@
     "fsevents": "^1.2.9",
     "html-webpack-plugin": "^3.2.0",
     "jest": "^24.9.0",
+    "mini-css-extract-plugin": "^0.9.0",
```
ShifaSZ, line 39: Is there any reason to move `min-css-extract-plugin` package in the file?

hanars: yes, it was incorrectly listed as a dependency when it is really a devDependency, and this moves it to the correct place

hanars:
Not sure why travis isn't reporting on this, but it ran and passed: https://travis-ci.org/github/macarthur-lab/seqr/builds/719064294

PR #1432, opened 2020-08-24
Fix update saved variants

hanars:
An endpoint was failing because the arguments being passed in to search were sets instead of lists, which caused search to fail


hanars:
Tests passing: https://travis-ci.org/github/macarthur-lab/seqr/builds/720791413

PR #1436, opened 2020-08-28
Fix predictor sort

hanars:
Fixes https://github.com/macarthur-lab/seqr-private/issues/912

`unmapped_type` is only used by elasticsearch if there is an index that is entirely missing a field thats is sorted on. When searching across variants and SVs this is used because SVs don't have any of the predictors. However, it turns out that not all regular variant indices use the same data type for field values, so while on the index used for testing this functionality when it was first implemented everything was a `float`, across all indices there are `float`, `half_float`, `double, and `keyword`. This updates the code to dynamically use the correct type for the given index and field

### DIFF
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -105,7 +105,7 @@ class EsSearch(object):
     def _set_index_metadata(self):
         self._set_index_name()
         from seqr.utils.elasticsearch.utils import get_index_metadata
-        self.index_metadata = get_index_metadata(self.index_name, self._client)
+        self.index_metadata = get_index_metadata(self.index_name, self._client, include_fields=True)
 
     def update_dataset_type(self, dataset_type, keep_previous=False):
         new_indices = self.indices_by_dataset_type[dataset_type]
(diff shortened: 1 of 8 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -1273,11 +1280,11 @@ def _get_compound_het_page(grouped_variants, start_index, end_index):
 
 
 def _parse_es_sort(sort, sort_config):
-    if sort in {'Infinity', '-Infinity'}:
+    if sort in {'Infinity', '-Infinity', None}:
         # ES returns these values for sort when a sort field is missing, using the correct value for the given direction
         sort = maxsize
     elif hasattr(sort_config, 'values') and any(cfg.get('order') == 'desc' for cfg in sort_config.values()):
-        sort = sort * -1
+        sort = float(sort) * -1
```
ShifaSZ, line 1287: It is confusing to me that `sort` can have a variety of types. If adding a brief description of the purpose of the function `_parse_es_sort()`, it will be helpful for understanding the changes of `sort`.

hanars: elasticsearch suports sorting on any type of data, its just whatever is in the field you specify. All of our sorts should be numeric, as we only support on numeric fields or using custom sorts that return numbers. However it turns out some of our predictor sort are loaded as strings and not numbers (i.e. the string `"0.01"`) so we need to explicitly cast them

hanars:
> Most of the PR is about `sort` of ES. Is there any document about that?
> The PR is approved since there is no real issue found but just a question.

There no seqr-specific documentation, but heres the ES documentation: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-sort.html

hanars:
passing tests: https://travis-ci.org/github/macarthur-lab/seqr/builds/722122839

PR #1443, opened 2020-09-02
Es test update

hanars:
Our tests for elasticsearch currently mock out the python elasticsearch client library. As part of upgrading ES we will be upgrading the client library as well. Since we mock it out everywhere it is used, if there are change to how the library behaves our test won;t'catch them. Therefore, I'm updating our unit tests to instead mock out the actual requests to elasticsearch, and I let the client library generate and parse those requests in the test just as it does in real life. 

Note that the `responses` library we use in most of our tests for mocking out http requests does not work for requests created in python by the `urllib3` library which is what is generated by elasticsearch, so I added the `urllib3-mock` library to test those requests


ShifaSZ (approved):
It is unfortunate that `responses` doesn't work with urllib3. But, you've done a great job with urllib3_mock. No issue found so far.

hanars:
passing tests: https://travis-ci.org/github/macarthur-lab/seqr/builds/723590653

PR #1462, opened 2020-10-01
Update locked package versions

hanars:
The last PR I put up bumped a lot the dependencies in our package-lock, but it turns out minor version bumps in babel break our build. This updates the lock file to include all the version bumps for vulnerable dependencies, but leaves our webpack dependencies unchanged


hanars:
sync is failing due to yargs-parser@5.0.0 however that was already a security issue on master so it is not actually new

hanars:
passing test: https://travis-ci.org/github/macarthur-lab/seqr/builds/732004128

ShifaSZ (approved):

PR #1475, opened 2020-10-19
Require csrf

hanars:
This adds in Django's support for requiring CSRF tokens to be passed on all POST/PUT/DELETE requests

### DIFF
--- a/seqr/views/apis/saved_variant_api_tests.py
+++ b/seqr/views/apis/saved_variant_api_tests.py
@@ -704,7 +704,7 @@ class SavedVariantAPITest(AuthenticationTestCase):
         url = reverse(update_saved_variant_json, args=['R0001_1kg'])
         self.check_manager_login(url)
 
-        response = self.client.get(url)
+        response = self.client.post(url)
         self.assertEqual(response.status_code, 200)
 
         self.assertDictEqual(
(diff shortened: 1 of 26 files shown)


settings.py
```diff
@@ -223,7 +225,7 @@
     "POSTMARK_SERVER_TOKEN": os.environ.get('POSTMARK_SERVER_TOKEN', 'postmark-server-token-placeholder'),
 }
 
-if os.environ.get('DEPLOYMENT_TYPE') == 'prod':
+if os.environ.get('DEPLOYMENT_TYPE') in {'prod', 'dev'}:
```
hanars, line 228: this condition was put in to differentiate between real deployed seqr and running locally. Since we are running all our security scans on dev, we want the dev deployment to behave like the prod one when it comes to security

ShifaSZ: Should we also add `staging` environment which is required the same as the `prod`? So that, the `dev` can be used for demo and testing.

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -704,7 +704,7 @@ def test_update_saved_variant_json(self, mock_get_variants):
         url = reverse(update_saved_variant_json, args=['R0001_1kg'])
         self.check_manager_login(url)
 
-        response = self.client.get(url)
+        response = self.client.post(url)
```
ShifaSZ, line 707: Is there any reason for changing `get` to `post`?

ShifaSZ (approved):
No issues found except a question.

settings.py
hanars: no, because these are actual values that exist in our deployment code and we don't have a staging or any near term plans to make one. If we add a staging environment, we should update the code then but adding impossible. values just makes the code harder to maintain

seqr/views/apis/saved_variant_api_tests.py
hanars: yeah I changed the client side code to do a POOST so it is CSRF verified so I changed the test to reflect that

PR #1478, opened 2020-10-22
Adding tests for social_auth.

ShifaSZ:
The tests need to be run twice with `TERRA_API_ROOT_URL` and without the setting.   
When running it without `TERRA_API_ROOT_URL`, it coverages all existing test cases with workspace attributes configured for the fixture projects. The new tests for the `terra_api_utils` and some of the tests for `dashboard_api` are not covered. These tests require `TERRA_API_ROOT_URL`.   
When the tests are run with `TERRA_API_ROOT_URL`, all test cases will be covered. The existing test cases run under a scenario of AnVIL signing-in enabled but no AnVIL data available. It emulates the situation of transitioning to AnVIL.  
Most of the seqr APIs haven't been tested against the permissions from AnVIL except the `dashboard_api`.


ssyms:
Hello! I'm with the AppSec team at Broad. Our scanner detected some Google access token in these commits. They appear to be for tests, so hopefully they are "fake". If they are real Google access tokens, could you follow the steps outlined here: https://dsp-security.broadinstitute.org/platform-security-categories/git/what-to-do-in-case-of-an-incident

hanars:
These are all test credentials so theres no real security risk However, @ShifaSZ should change all the credentials to be obviously fake (i.e. use something like `test_account_token` instead of something that looks real)

seqr/utils/social_auth_pipeline_tests.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class SocialAuthPipelineTest(TestCase):
+
+    @responses.activate
+    @mock.patch('seqr.utils.social_auth_pipeline.logger')
+    def test_validate_anvil_registration(self, mock_logger):
+        url = TERRA_API_ROOT_URL + 'register'
+        responses.add(responses.GET, url, status = 404)
+        r = validate_anvil_registration(GoogleOAuth2(), {'access_token': '', 'email': 'test@seqr.org'})
```
hanars, line 17: is this ACTUALLY calling google? or is the responses handling that?

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+from seqr.views.utils.terra_api_utils import get_anvil_billing_projects, get_anvil_profile, has_anvil_session,\
+    list_anvil_workspaces, get_anvil_workspace_acl
+from settings import TERRA_API_ROOT_URL
+
+
+class TerraApiUtilsCase(TestCase):
+    fixtures = ['users']
+
+    def setUp(self):
+        google_url = 'https://oauth2.googleapis.com/token'
```
hanars, line 16: constants should be all caps and defined at the top of the file

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+        content = b'{"access_token":"ya29.c.KpMB4QdMNGgxWQvj3PxOgbQqfWQznoHuuoY03vN_A-W_W3vkdXJ6q-a539jSZEWc602cBL_RfNUH4GZV-d8rosfzQwBHoCVxfQs-Su2jBgKEdTui53jfjyU9T9LxshDl8ov31yk8Pi8f9D9ER7EUp-W1mP9gSy9e4YWkLb_q4hdJRR3BJywlQIWvPEfawpjC5Jpt_qGJ","expires_in":3599,"token_type":"Bearer"}'
+        responses.add(responses.POST, google_url, status = 200, body = content)
+        self.user = User.objects.get(email__iexact = 'test_user@test.com')
+        has_anvil_session(self.user)
```
hanars, line 20: `has_anvil_session` sounds like something that just checks whether or not the user has a session and returns a boolean. What good does calling it here do? If its actually setting up a session you need to rename it

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class TerraApiUtilsCase(TestCase):
+    fixtures = ['users']
+
+    def setUp(self):
+        google_url = 'https://oauth2.googleapis.com/token'
+        content = b'{"access_token":"ya29.c.KpMB4QdMNGgxWQvj3PxOgbQqfWQznoHuuoY03vN_A-W_W3vkdXJ6q-a539jSZEWc602cBL_RfNUH4GZV-d8rosfzQwBHoCVxfQs-Su2jBgKEdTui53jfjyU9T9LxshDl8ov31yk8Pi8f9D9ER7EUp-W1mP9gSy9e4YWkLb_q4hdJRR3BJywlQIWvPEfawpjC5Jpt_qGJ","expires_in":3599,"token_type":"Bearer"}'
+        responses.add(responses.POST, google_url, status = 200, body = content)
+        self.user = User.objects.get(email__iexact = 'test_user@test.com')
```
hanars, line 19: you don't use `iexact` you just use `=`

seqr/views/utils/terra_api_utils_tests.py
hanars: you also need to test the cases where the user does not have permission to access terra

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+
+        profile = get_anvil_profile(self.user)
+        self.assertDictEqual(profile, {'enabled': {'ldap': True, 'allUsersGroup': True, 'google': True},
+            'userInfo': {'userEmail': 'sf-seqr@my-seqr.iam.gserviceaccount.com', 'userSubjectId': '108344681601016521986'}})
+
+        responses.replace(responses.GET, url, status = 404)
+        with self.assertRaises(Exception) as ec:
+            _ = get_anvil_profile(self.user)
+        self.assertEqual(str(ec.exception),
+            'Error: called Terra API "register" got status: {} with a reason: {}'.format(404, 'Not Found'))
```
hanars, line 50: please hardcode strings that can't change, do not use constants as formatting variables all it does is make the code less readable

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+                   '{"accessLevel": "PROJECT_OWNER","public": false,"workspace": {"attributes": {"description": "A workspace for seqr project"},"authorizationDomain": [],"bucketName": "fc-6a048145-c134-4004-a009-42824f826ee8","createdBy": "zhangshifa07504@gmail.com","createdDate": "2020-09-09T15:12:30.142Z","isLocked": false,"lastModified": "2020-09-09T15:12:30.145Z","name": "seqr-project 1000 Genomes Demo","namespace": "my-seqr-billing","workflowCollectionName": "6a048145-c134-4004-a009-42824f826ee8","workspaceId": "6a048145-c134-4004-a009-42824f826ee8"},"workspaceSubmissionStats": {"runningSubmissionsCount": 0}}]')
+        workspaces = list_anvil_workspaces(self.user)
+        self.assertEqual(len(workspaces), 3)
+        self.assertIsNotNone(workspaces[0]['public'])
```
hanars, line 61: you should check the workspace id or some other field thats more specific to the first entry here. The last entry in the list also has public=false

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,90 @@
...
+
+        responses.replace(responses.GET, url, status = 200,
+            body = '[{"accessLevel": "PROJECT_OWNER", "workspace": {"name": "1000 Genomes Demo", "namespace": "my-seqr-billing", "workspaceId": "237998e6-663d-40b9-bd13-57c3bb6ac593" }},'
+                   '{"accessLevel": "READER","workspace": {"name": "degenome","namespace": "degenome", "workspaceId": "2706d493-5fce-4fb2-9993-457c30364a06"}},'
+                   '{"accessLevel": "PROJECT_OWNER","workspace": {"name": "seqr-project 1000 Genomes Demo","namespace": "my-seqr-billing","workspaceId": "6a048145-c134-4004-a009-42824f826ee8"}}]')
```
hanars, line 66: why does the response data need to change here, your function doesn't actually do anything with the result so what value is there to having a different response other than making the code confusing. What you actually need to test here is that the responses were called with different url parameters than before when you pass in fields

hanars (changes requested):
You need to add tests to all the permission checking functions that check terra perissions as well as all the apis that use any of the terra helper functions

seqr/utils/social_auth_pipeline_tests.py
ShifaSZ: `responses` handle it.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: Okay

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: It must be `get_session()` which will create a session. I will update it.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: Okay

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: Okay

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: Change to check the whole `dict` must equal to the one predicted.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: The purpose of the test is to verify the filtering fields. The URL path should be something like `api/workspaces?accessLevel,workspace.name,workspace.namespace,workspace.workspaceId`. But the `responses` seems matching the path before the path is combined with parameters. So, the URL is the same as the previous one. That's why I have to use `responses.replace()` instead of `responses.add()`.

ShifaSZ: I probably change the function `list_anvil_workspaces()` to combine the parameters with the URL before calling the get method. It will be more friendly for the test.

hanars:
Some high-level notes are that we are going to need a few more tests than what you've added. In particular, we need to explicitly test behavior for both when `TERRA_API_ROOT_URL` is configured and when its not. In the new tests you add you will need to add both cases (i.e. test that it throws an error if you try to call any of the new functions when theres not root url).

Then in the existing tests for checking permissions and log in, we need to explicitly test how things work both when AnVIL is enabled and not. So for example, we need to add a test that an AnVIL user can access the project page if they have access to the corresponding workspace, and we need to test that existing users who had access to that project still have access even if they don't have workspace access. We also need to check that the function to get project collaborator lists correctly returns a list of both local and anvil users if it has a workspace name, but not if theres no workspace name configured.

 In short, these are really good tests for the new functions you have added, but we need to add tests everywhere they are called in the code (you can mock out these functions anywhere they are used if you would like, since you already test them in detail).

ShifaSZ:
> Some high-level notes are that we are going to need a few more tests than what you've added. In particular, we need to explicitly test behavior for both when `TERRA_API_ROOT_URL` is configured and when its not. In the new tests you add you will need to add both cases (i.e. test that it throws an error if you try to call any of the new functions when theres not root url).
> 
> Then in the existing tests for checking permissions and log in, we need to explicitly test how things work both when AnVIL is enabled and not. So for example, we need to add a test that an AnVIL user can access the project page if they have access to the corresponding workspace, and we need to test that existing users who had access to that project still have access even if they don't have workspace access. We also need to check that the function to get project collaborator lists correctly returns a list of both local and anvil users if it has a workspace name, but not if theres no workspace name configured.
> 
> In short, these are really good tests for the new functions you have added, but we need to add tests everywhere they are called in the code (you can mock out these functions anywhere they are used if you would like, since you already test them in detail).

I'm trying to re-use the existing tests (without changing) and added new tests for AnVIL. Not successful yet. I will let you know when it is okay.

ShifaSZ:
I updated the `seqr/views/apis/dashboard_api_tests.py` by adding a new test class `DashboardPageAnvilTest`. The new class inherits the existing test class `DashboardPageTest` and adds fixtures for AnVIL permissions. All the existing test cases will be re-run under new fixtures with AnVIL permissions configurations which is important to make sure new AnVIL permission won't break any existing features. Only the test cases which need special treatment for permissions from AnVIL are overridden. New test cases can also be added to cover more features with AnVIL access controls.

.travis.yml
```diff
@@ -13,7 +13,8 @@ install:
 - ./install_dependencies.sh
 
 script:
-- coverage run --source="./matchmaker","./seqr","./reference_data" --omit="*/migrations/*","*/apps.py" manage.py test -p '*_tests.py' -v 2 reference_data seqr matchmaker
+- TERRA_API_ROOT_URL=https://localhost/ coverage run --source="./matchmaker","./seqr","./reference_data" --omit="*/migrations/*","*/apps.py" manage.py test -p '*_tests.py' -v 2 reference_data seqr matchmaker
+- coverage run -a --source="./matchmaker","./seqr","./reference_data" --omit="*/migrations/*","*/apps.py" manage.py test -p '*_tests.py' -v 2 reference_data seqr matchmaker
```
ShifaSZ, line 17: I finally don't use `test detection` in the settings because it makes the tests always with the `TERRA_API_ROOT_URL` setting. We need to run the tests twice, one with `TERRA_API_ROOT_URL` and the other without it. The option `-a` of `coverage run` command can combine the results of the two runs.
I'm not sure whether it works with `Travis` or not by adding `TERRA_API_ROOT_URL=https://localhost/` in front of the test command. But, just try it first.

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +126,57 @@ def test_export_projects_table(self):
...
+    @mock.patch('seqr.views.utils.terra_api_utils.GOOGLE_SERVICE_ACCOUNT_INFO', GOOGLE_SERVICE_ACCOUNT_INFO)
+    class DashboardPageAnvilTest(DashboardPageTest):
+        fixtures = ['users', 'social_auth_data', '1kg_project']
+
+        @responses.activate
+        def test_dashboard_page_data(self):
+            responses.add(responses.POST, GOOGLE_ACCESS_TOKEN_URL, status=200, body=GOOGLE_TOKEN_RESULT)
+            responses.add(responses.POST, GOOGLE_API_TOKEN_URL, status=200, body=GOOGLE_TOKEN_RESULT)
+            responses.add(responses.GET, WORKSPACE_WITH_FIELDS_URL, status=200, body=WORKSPACE_RSP_NO_VALID_PROJECT)
+            responses.add(responses.GET, WORKSPACE2_ACL_URL, status=200, body=WORKSPACE2_ACL_RSP)
```
ShifaSZ, line 141: These are the wrappers that requested for the existing tests to run with `TERRA_API` enabled.

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +126,57 @@ def test_export_projects_table(self):
         response = self.client.get('{}?file_format=csv'.format(url))
         self.assertEqual(response.status_code, 400)
         self.assertDictEqual(response.json(), {'error': 'Invalid file_format: csv'})
+
+
+if anvil_enabled():
+    @mock.patch('seqr.views.utils.terra_api_utils.GOOGLE_SERVICE_ACCOUNT_INFO', GOOGLE_SERVICE_ACCOUNT_INFO)
+    class DashboardPageAnvilTest(DashboardPageTest):
```
hanars, line 133: So this code style is a bit hard to read, and I think it will be really difficult to extend this without tons of copy/paste to other places that also need to have full testing for the Anvil case  (i.e. the api calls to get all the collaborators for a project). What I'm proposing below is one way to update the tests to effectively run tests against both login modes, but I  am happy to meet to discuss if you don't like this approach or have any questions.

I think this could work a lot better/ be a lot more extensible to other test files if instead of linear inheritance of `DashboardPageAnvilTest` -> `DashboardPageTest` -> `AuthenticationTestCase` you instead used a mixin style. So what that would look like is you define an `AnvilAuthenticationTestCase` that inherits from `AuthenticationTestCase` but completely redefines defines the `_check_login` method and the `setUpTestData` to instead work with AnVIL login and not add them to user groups for permissions. This class will be where you mock out the `TERRA_API_ROOT_URL ` to make sure it is defined. You would ideally set up the users to have the same permissions in both (i.e. making sure that `collaborator_user` has the same project collaborator permissions in the mocked anvil as the mocked local groups).  Then instead of defining the main class as `class DashboardPageTest(AuthenticationTestCase)` you would do
```
class DashboardPageTest:
   <basically same code as now>
   ...

class LocalUserDashboardPageTest(AuthenticationTestCase, DashboardPageTest)

class AnvilDashboardPageTest(AnvilAuthenticationTestCase, DashboardPageTest):
   def test_dashboard_page_data(self):
        super(self, AnvilDashboardPageTest).test_dashboard_page_data()
       <anvil specific tests>
```

If you are struggling to get `responses` to work in the set up for the `AnvilAuthenticationTestCase`, you could instead patch the helper methods that make calls to AnVIL, and then just make sure that those functions are really well tested. So for example, in the `AnvilAuthenticationTestCase` define 
```
USER_WORKSPACE_MAP = {<a mapping between test users and the response bodies for what workspaces we want them to have access to>}
...
def setUp():
   patcher = mock.patch('seqr.views.utils.terra_api_utils.list_anvil_workspaces')
   patcher.start().side_effect = lambda user, *args: USER_WORKSPACE_MAP[user.email]
   self.addCleanup(patcher.stop)
```

hanars (changes requested):

.travis.yml
ShifaSZ: If there is an option for `manage.py test` which can be detected by the `settings.py` then it will be perfect.

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +126,57 @@ def test_export_projects_table(self):
         response = self.client.get('{}?file_format=csv'.format(url))
         self.assertEqual(response.status_code, 400)
         self.assertDictEqual(response.json(), {'error': 'Invalid file_format: csv'})
+
+
+if anvil_enabled():
```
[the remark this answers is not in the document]
ShifaSZ: The issue is related to whether the `social_django` application is loaded or not. If `social_django` is not loaded, the fixture data will crash the test (the data needs the UserSocialAuth model). So, the tests depending on `social_django` are run only when anvil_enabled is true.
Mocking settings won't load new apps.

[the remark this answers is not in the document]
hanars: but the whole test suite needs to run every time, and that is the ony acceptable way to have a test suite. You need to change the settings so `social_djang` is always loaded when tests are running, regardless of `TERRA_API_ROOT_URL`. As I mentioned yesterday, you can add conditional code to `settings.py` to detect if its a test run: https://github.com/broadinstitute/seqr/blob/19a1266e267ed2c5197c5f976d58114829865b08/settings.py#L423

.travis.yml
hanars: As I mentioned yesterday, we used to do this in seqr code. Here is an example: https://github.com/broadinstitute/seqr/blob/19a1266e267ed2c5197c5f976d58114829865b08/settings.py#L423

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Very heuristic idea. I have never think about doing it in this way. I still like to meet you next week for some questions.

.travis.yml
ShifaSZ: The problem with the detection in this way is that it always sets `TERRA_API_ROOT_URL` when it is testing. We need tests for both `TERRA_API_ROOT_URL` set and un-set.

seqr/views/apis/dashboard_api_tests.py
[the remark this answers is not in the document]
ShifaSZ: I see what you mean. We can load `social_django` app no matter the value of `TERRA_API_ROOT_URL`.

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Some thoughts for the meeting this afternoon:  
We need three kinds of tests with different permission settings: `pure local permissions`, `combined local and remote permissions`, and `pure remote permissions`. Some highlights of these tests are listed below.  
  
- Pure local permissions  
   - Test without `TERRA_API_ROOT_URL`  
   - All existing test cases should work  
  
- Combine local and remote permissions  
   - Test with `TERRA_API_ROOT_URL`  
   -  All existing test cases should work
   - Add a few new test cases for some corners to make sure:   
      - the users still have the existing permissions  
      - new permissions are added.  
  
- Pure remote permissions.  
   - Test with `TERRA_API_ROOT_URL`   
   - Provide a new `AnvilAuthenticationTestCase` class to:   
      - set all the local permissions to the mocked remote permissions  
   - Ideally, re-use all existing test cases

hanars: This is a really good summary of what we need. Looking forward to discussing

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
+
+# Tests for AnVIL access disabled
+class LocalDashboardPageTest(AuthenticationTestCase, DashboardPageTest):
+    fixtures = ['users', '1kg_project']
+
+
+# Test for permissions from AnVIL only
+@mock.patch('seqr.views.utils.terra_api_utils.TERRA_API_ROOT_URL', TEST_TERRA_API_ROOT_URL)
+class AnvilDashboardPageTest(AnvilAuthenticationTestCase, DashboardPageTest):
+    fixtures = ['users', 'social_auth_data', '1kg_project']
```
hanars, line 135: We don't usually include "data" in the fixture name, since fixtures are definitionally data, so I would change the filename from`social_auth_data` to `social_auth`

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
+
+    def setUp(self):
```
hanars, line 137: I think it would be better to move this into the `AnvilAuthenticationTestCase` base set up, since it seems likely that these mocks could be required  by most tests that were anvil authenticated

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
         self.assertDictEqual(response.json(), {'error': 'Invalid file_format: csv'})
+
+
+# Tests for AnVIL access disabled
+class LocalDashboardPageTest(AuthenticationTestCase, DashboardPageTest):
+    fixtures = ['users', '1kg_project']
+
+
+# Test for permissions from AnVIL only
+@mock.patch('seqr.views.utils.terra_api_utils.TERRA_API_ROOT_URL', TEST_TERRA_API_ROOT_URL)
```
hanars, line 133: you can add this mocking to the base `AnvilAuthenticationTestCase`, since you always want thins configured when testing AnVIL auth

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
+    @mock.patch('seqr.views.utils.permissions_utils.list_anvil_workspaces')
+    @mock.patch('seqr.views.utils.terra_api_utils._service_account_session')
+    @mock.patch('seqr.views.utils.terra_api_utils.time')
+    def test_dashboard_page_data(self, mock_time, mock_service_account_session, mock_list_anvil_workspaces):
+        # Users can see the projects that AnVIL allows
+        url = reverse(dashboard_page_data)
+        self.login_collaborator()
+        mock_time.time.return_value = 1603287741 + 10
+        mock_service_account_session.get.return_value = MIX_TEST_ALCS
+        mock_list_anvil_workspaces.return_value = MIX_TEST_WORKSPACES
```
hanars, line 190: Would there be a way to do the mixed tests in a way the reused the same terra mocking as the pure anvil test? When the data coming back from AnVIL is the same, it makes it really clear what is different in the Mix case vs the pure anvil case. See my comment on the `AnvilAuthenticationTestCase` about using helper methods to make it easier to test that there is different expected data

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
+
+    def setUp(self):
+        patcher = mock.patch('seqr.views.utils.permissions_utils.list_anvil_workspaces')
+        patcher.start().side_effect = get_workspaces_side_effect
+        self.addCleanup(patcher.stop)
+        patcher = mock.patch('seqr.views.utils.terra_api_utils._service_account_session')
+        patcher.start().get.side_effect = get_ws_acl_side_effect
+        self.addCleanup(patcher.stop)
+        patcher = mock.patch('seqr.views.utils.terra_api_utils.time')
+        patcher.start().return_value = 1603287741 + 10
```
hanars, line 145: can you add a comment to explain what this number is (or maybe make it a constant with a descriptive name)

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,116 @@ def test_export_projects_table(self):
...
+        self.addCleanup(patcher.stop)
+
```
hanars, line 147: you need to test that the dashboard page is calling all the mocked methods correctly. In the setup method, you can attach the patchers to the class so you can access them in the test:
```
patcher = mock.patch('seqr.views.utils.permissions_utils.list_anvil_workspaces')
self.mock_list_workspaces = patcher.start()
self.mock_list_workspaces.side_effect = get_workspaces_side_effect
```
Then you can use them the way you do mocks that were added by decorator:
```
def test_dashboard_page_data(self):
   super(AnvilDashboardPageTest , self).test_dashboard_page_data()
   self.mock_list_workspaces.assert_has_calls(...)
```

seqr/views/utils/test_utils.py
```diff
@@ -91,6 +91,93 @@ def _check_login(self, url, permission_level, request_data=None):
...
+        {
+            'public': ws['public'],
+            'accessLevel': ws['acl'][user.email]['accessLevel'],
+            'workspace':{
+                'namespace': ws['workspace_namespace'],
+                'name': ws['workspace_name']
+            }
+        } for ws in ANVIL_WORKSPACES if user.email in ws['acl'].keys()
+    ]
+
```
hanars, line 165: I really like these side effect methods. It makes it super clear how the permissions are set/ decided and it seems really simple to maintain and extend

seqr/views/utils/test_utils.py
```diff
@@ -91,6 +91,93 @@ def _check_login(self, url, permission_level, request_data=None):
...
+    def setUpTestData(cls):
+        cls.staff_user = User.objects.get(username='test_user')
+        cls.manager_user = User.objects.get(username='test_user_manager')
+        cls.collaborator_user = User.objects.get(username='test_user_non_staff')
+        cls.no_access_user = User.objects.get(username='test_user_no_access')
+
+        for project in Project.objects.all():
+            project.workspace_namespace = ANVIL_WORKSPACES[project.pk-1]['workspace_namespace']
+            project.workspace_name = ANVIL_WORKSPACES[project.pk - 1]['workspace_name']
+            project.save()
```
hanars, line 178: I STRONGLY dislike this. It's incredibly confusing to load a fixture and then modify it, especially to modify it in an inherited class so it's not even clear that that is happening. I think you need to set up the fixture data to have the correct data. I would rather have the anvil and local test cases differ from each other a bit than do a hack like this

An easy way to allow the tests to be reused with the different login mechanisms but have slightly different results is to add class level constants, so instead if `self.assertEqual(len(response_json['projectsByGuid']), 2)` do `self.assertEqual(len(response_json['projectsByGuid']), NUM_COLLABORATOR_PROJECTS)` and then each of you class definitions define a different value for `NUM_COLLABORATOR_PROJECTS`

hanars (changes requested):
I overall really like what you've done here!

seqr/views/utils/test_utils.py
ShifaSZ: Good idea to add login-mechanism-dependent results to the authentication class. I was hesitating about updating the fixture data. I could add the workspace information into the fixture data directly. But, I have the concern that it may not cover the cases of no workspace configurations for the local and mix tests. I really don't like to duplicate the fixture data for different tests. If I can use different predictions then it solves the problem.

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Yes. I would try to define a new mix-authentication class and define different mocking and expected results in it.

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Okay

seqr/views/utils/test_utils.py
hanars: I think add workspaces to project fixture 1 but not to project 2 or 3, or maybe to projects 1 and 2 but not 3 or 1 and 3 but not 2, whichever makes for the best tests. That will allow you to test all 3 different cases without needing to modify the fixtures in the tests.

seqr/views/apis/dashboard_api_tests.py
hanars: Hmmm I see what you are saying, but I was actually saying "what if the anvil permissions were the same in the anvil only and mix case". Then it would be really obvious to the reader what is different in the results between them, whereas if the permissions coming back from AnVIL are different its really hard to tell

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Done

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Done

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: The problem is that there will be no difference between local and mix if we use the same mocks as anvil only. It is hard to figure out if the anvil configurations have taken effect. Unless I make a slight change to the fixture data for the projects. For example, add `disable_staff_access` to one of the projects.
By the way, the new mix-auth class seems necessary because it enables both local and anvil permissions which won't happen in either local or anvil authentication classes.

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: The number is the timestamp for authentication. Updated

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Okay. But the pure anvil tests won't check the mocking because it re-uses the existing tests without changes. It must be safe because the local permissions are disabled so that the permissions can only come from the mocking.

seqr/views/utils/test_utils.py
ShifaSZ: I added workspaces to project 1 and 3, not 2.

seqr/views/utils/test_utils.py
```diff
@@ -91,6 +92,131 @@ def _check_login(self, url, permission_level, request_data=None):
...
+
+
+class MixAuthenticationTestCase(AnvilAuthenticationTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        AuthenticationTestCase.setUpTestData()
+
+    def setUp(self):
+        super(MixAuthenticationTestCase, self).setUp()
+        self.mock_list_workspaces.side_effect = lambda user, fields: MIX_TEST_WORKSPACES\
```
hanars, line 216: Is there a reason this needs to be different? What edge case does this test that you can't test by using a different user with the same AnVIL permissions? Would adding a test for the manager user allow you to cover that edge case while using the same permissions for everything? Or maybe actually inspecting some of the json itself (all the projects json includes whether or not the user has edit permissions, so if those differ between anvil and local and mix maybe that would be a way to test it)? Ideally, the mixed test should be different because the local and anvil permissions ARE different, and the mixed permissions are therefore different than both of them

If this is really necessary to have different mocked data in anvil, can you add a comment explaining why?

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,52 @@ def test_export_projects_table(self):
...
         self.assertDictEqual(response.json(), {'error': 'Invalid file_format: csv'})
+
+
+# Tests for AnVIL access disabled
+class LocalDashboardPageTest(AuthenticationTestCase, DashboardPageTest):
+    fixtures = ['users', '1kg_project']
+    NUM_COLLABORATOR_PROJECTS = 2
+    NUM_STAFF_PROJECTS = 3
+    NUM_STAFF_EXPORT_RECS = 3
+    NUM_STAFF_EXPORT_TSV_LINES = NUM_STAFF_EXPORT_RECS + 1
```
hanars, line 133: it is somewhat confusing to define class variables for things that are the same in all the inherited classes. Maybe move the variable definitions into the base class so its really clear whats different, and only leave the variables that are different between them in the subclasses

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -120,3 +122,52 @@ def test_export_projects_table(self):
...
+
+
+# Test for permissions from AnVIL only
+class AnvilDashboardPageTest(AnvilAuthenticationTestCase, DashboardPageTest):
+    fixtures = ['users', 'social_auth', '1kg_project']
+    NUM_COLLABORATOR_PROJECTS = 1
+    NUM_STAFF_PROJECTS = 3
+    NUM_STAFF_EXPORT_RECS = 3
+    NUM_STAFF_EXPORT_TSV_LINES = NUM_STAFF_EXPORT_RECS + 1
+
```
hanars, line 143: this class needs to also test that the terra calls were called correctly, similar to what you do on the mix test case

hanars (changes requested):

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: They must be different at least in two of the three classes. Otherwise, we don't have to define them and just use a constant in the test case.

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: Okay

seqr/views/utils/test_utils.py
ShifaSZ: The different mocked data is not always necessary especially when the local projects have left some permissions which can be filled by the anvil permissions. Base on current local fixture permissions, I probably can do as below table:
|                | local |         |              | Anvil |         |              | Mix |         |              |
|----------------|-------|---------|--------------|-------|---------|--------------|-------------|---------|--------------|
| project No.    | staff | manager | collaborator | staff | manager | collaborator | staff       | manager | collaborator |
| 1              | y     | y       | y            | na    | y      | na           | y           | y       | y            |
| 2              | y     | y       | y            | na    | na      | na           | y           | y       | y            |
| 3              | y     | y       | n            | na    | y      | y            | y           | y       | n            |
| total projects | 3     | 3       | `2`            | 3     | 2       | `1 `           | 3           | 3       | `3`            |  

The numbers of projects for the collaborator are different for local/anvil/mix test cases.

seqr/views/apis/dataset_api_tests.py
```diff
@@ -344,3 +343,50 @@ def test_add_alignment_sample(self, mock_local_file_exists, mock_subprocess):
             {sample_guid}
         )
         mock_subprocess.assert_called_with('gsutil ls gs://readviz/NA19679.bam', stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+
+"""
+# Tests for AnVIL access disabled
+class LocalDatasetAPITest(AuthenticationTestCase, DatasetAPITest):
+    fixtures = ['users', '1kg_project']
+"""
```
hanars, line 351: is this meant to be commented out?

seqr/views/apis/dataset_api_tests.py
```diff
@@ -344,3 +343,50 @@ def test_add_alignment_sample(self, mock_local_file_exists, mock_subprocess):
...
+
+
+def assertions(self, call_count):
```
hanars, line 354: can you rename this something a bit ore descriptive, like `assert_has_anvil_calls`

seqr/views/apis/project_api_tests.py
```diff
@@ -166,3 +166,54 @@ def test_empty_project_page_data(self):
...
+        super(AnvilProjectAPITest, self).test_create_update_and_delete_project()
+        self.mock_list_workspaces.assert_not_called()
+        self.mock_service_account.get.assert_not_called()
+
+    def test_project_page_data(self):
+        super(AnvilProjectAPITest, self).test_project_page_data()
+        self.mock_list_workspaces.assert_not_called()
+        self.mock_service_account.get.assert_called_with(
+            'api/workspaces/my-seqr-billing/anvil-1kg project n\u00e5me with uni\u00e7\u00f8de/acl')
+        self.assertEqual(self.mock_service_account.get.call_count, 5)
```
hanars, line 190: Can you update this test to test what the content of the project `collaborators` key in the response data? I would expect it to be different in the anvil and non-anvil cases so I would like that to be tested. So after line 94 above you could add something like 
```
collaborators = response_json['projectsByGuid'][PROJECT_GUID]['collaborators']
self.assertEqual(len(collaborators), NUM_COLLABORATORS)
self.assertEqual(len([c for c in collaborators if c['isAnvil']), NUM_ANVIL_COLLABORATORS)
```
And then for each of the local/anvil/mix cases you could define the appropriate variables

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -731,3 +731,127 @@ def test_update_variant_main_transcript(self):
         self.assertEqual(saved_variant.selected_main_transcript_id, transcript_id)
 
 
+# Tests for AnVIL access disabled
+class LocalSavedVariantAPITest(AuthenticationTestCase, SavedVariantAPITest):
+    fixtures = ['users', '1kg_project']
+
+
+def assertions(self, call_count):
```
hanars, line 739: same as above, can this be a more descriptive name?

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -473,3 +473,89 @@ def test_saved_search(self):
...
         response = self.client.get(delete_saved_search_url)
         self.assertEqual(response.status_code, 403)
+
+
+# Tests for AnVIL access disabled
+class LocalVariantSearchAPITest(AuthenticationTestCase, VariantSearchAPITest):
+    fixtures = ['users', '1kg_project', 'reference_data', 'variant_searches']
+
+
+def assertions(self, call_count, path=None):
```
hanars, line 483: same as above

seqr/views/utils/terra_api_utils_tests.py
hanars: Could you replace this with your fixture data, like `WORKSPACE_RSP_ONE_VALID_PROJECT`? If not this is fine

hanars (changes requested):

seqr/views/apis/dataset_api_tests.py
ShifaSZ: Forgot to remove the comment after debugging.

seqr/views/apis/project_api_tests.py
ShifaSZ: Good point. But I do it differently because all the collaborators have 'isAnvil' set when the Terra API is enabled. I compare the collaborator list against the expected list.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: It is a different one. This one is used only for one time. So I keep it as a literal value here.

seqr/views/apis/dashboard_api_tests.py
```diff
@@ -28,9 +31,8 @@
 ]
 
 
-class DashboardPageTest(AuthenticationTestCase):
-    fixtures = ['users', '1kg_project']
-
+class TestHelpers(object):
```
hanars, line 34: I strongly prefer the previous approach of having a top level `class DashboardPageTest(object):`. Why would you change this? I think its good for this test helper class to be unable to run without being mixed with an authentication case, so I don't like the idea of it being able to run with `BasicAuthTestCase`

seqr/views/apis/project_api_tests.py
```diff
@@ -166,3 +170,62 @@ def test_empty_project_page_data(self):
...
+
+
+# Tests for AnVIL access disabled
+class LocalProjectAPITest(AuthenticationTestCase, TestHelpers.ProjectAPITest):
+    fixtures = ['users', '1kg_project', 'reference_data']
+    PROJECT_COLLABORATORS = [{'dateJoined': '2017-03-12T23:09:54.180Z', 'displayName': 'Test Manager User', 'email': 'test_user_manager@test.com', 'firstName': 'Test Manager User', 'hasEditPermissions': True, 'hasViewPermissions': True, 'id': 11, 'isAnvil': False, 'isStaff': False, 'lastName': '', 'username': 'test_user_manager'},
+{'dateJoined': '2017-03-12T23:09:54.180Z', 'displayName': 'Test Non Staff User', 'email': 'test_user_no_staff@test.com', 'firstName': 'Test Non Staff User', 'hasEditPermissions': False, 'hasViewPermissions': True, 'id': 12, 'isAnvil': False, 'isStaff': False, 'lastName': '', 'username': 'test_user_non_staff'}]
+
+
+ANVIL_PROJECT_COLLABORATORS = [{'dateJoined': '2017-03-12T23:09:54.180Z', 'displayName': 'Test Manager User', 'email': 'test_user_manager@test.com', 'firstName': 'Test Manager User', 'hasEditPermissions': False, 'hasViewPermissions': True, 'id': 11, 'isAnvil': True, 'isStaff': False, 'lastName': '', 'username': 'test_user_manager'},
```
hanars, line 182: nitpicky, but its more readable if you define constants at the top of the file

seqr/views/apis/project_api_tests.py
```diff
@@ -166,3 +170,62 @@ def test_empty_project_page_data(self):
...
+        self.check_collaborator_login(url)
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 403)
+
+
+# Test for permissions from AnVIL and local
+class MixProjectAPITest(MixAuthenticationTestCase, TestHelpers.ProjectAPITest):
+    fixtures = ['users', 'social_auth', '1kg_project', 'reference_data']
+    PROJECT_COLLABORATORS = ANVIL_PROJECT_COLLABORATORS
```
hanars, line 214: For the mix case, is there any way to test that it combines permissions? Its okay not to if thats really hard

seqr/views/apis/project_api_tests.py
```diff
@@ -96,6 +97,9 @@ def test_project_page_data(self):
             response_json['projectsByGuid'][PROJECT_GUID]['lastAccessedDate'][:10],
             datetime.today().strftime('%Y-%m-%d')
         )
+        collaborators = [{k: c[k] for k in c if k != 'lastLogin'}  # remove key 'lastLogin' for the time keeps changing
```
hanars, line 100: if in the `PROJECT_COLLABORATORS` constant you define `'lastLogin': mock.ANY` it will work and be more readable here

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -731,3 +732,127 @@ def test_update_variant_main_transcript(self):
...
+        super(AnvilSavedVariantAPITest, self).test_create_update_and_delete_variant_note()
+        assert_no_list_ws_has_acl(self, 6)
+
+    def test_create_partially_saved_compound_het_variant_note(self):
+        super(AnvilSavedVariantAPITest, self).test_create_partially_saved_compound_het_variant_note()
+        assert_no_list_ws_has_acl(self, 2)
+
+    def test_create_update_and_delete_compound_hets_variant_note(self):
+        super(AnvilSavedVariantAPITest, self).test_create_update_and_delete_compound_hets_variant_note()
+        assert_no_list_ws_has_acl(self, 7)
```
hanars, line 777: why doe the pure AnVIL tests call the service account api many times, but the mix test case always calls it exactly once?

hanars (changes requested):

seqr/views/apis/dashboard_api_tests.py
ShifaSZ: If the class `DashboardPageTest` doesn't inherit any class like `BasicAuthTestCase` then many attributes used with the `DashboardPageTest` class are undefined. It is okay for running but it is hard to write or update the tests because the IDE can't provide any hints for the attributes.

seqr/views/apis/project_api_tests.py
ShifaSZ: Okay. Then, I will also define one constant for local collaborators so that two constants sit side-by-side.

seqr/views/apis/project_api_tests.py
ShifaSZ: It should be possible.

seqr/views/apis/project_api_tests.py
ShifaSZ: Since the three scenarios will have different collaborator lists, I delete the constant and use the literal data.

seqr/views/apis/project_api_tests.py
ShifaSZ: I add a pure local user to make the mix-case has a different collaborator list.

seqr/views/apis/project_api_tests.py
ShifaSZ: Good point.

seqr/views/apis/saved_variant_api_tests.py
ShifaSZ: The pure AnVIL tests fetch the permissions from AnVIL every time while the mix cases don't have to access AnVIL because they get the permissions local already.

seqr/views/apis/dashboard_api_tests.py
hanars: This nested class syntax is really hard to read. I disagree that the IDE hints make this easier to update, I think the weird nesting makes it harder, and makes it really unclear if the tests are supposed to be able to run on their own or not. I strongly prefer we change it back.

hanars: Here are some suggestions for how to trick the IDE into knowing what class something is meant to be, if you feel really strongly about it:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/205818219-How-to-tell-PyCharm-that-a-class-is-a-mixin-specifically-for-class-X

ShifaSZ: I will try `assert isinstance(self, BaseAuthTestCase)`.

hanars: okay I will wait to review until these changes are in

ShifaSZ: I will just change them back and use `assert isinstance(self, AuthenticationTestCase)` when needs.

seqr/views/apis/users_api_tests.py
```diff
@@ -9,29 +9,14 @@
...
     get_all_staff
-from seqr.views.utils.test_utils import AuthenticationTestCase
+from seqr.views.utils.test_utils import AuthenticationTestCase, AnvilAuthenticationTestCase,\
+    MixAuthenticationTestCase, WORKSPACE_FIELDS
 
 
 PROJECT_GUID = 'R0001_1kg'
 
 
-class UsersAPITest(AuthenticationTestCase):
```
hanars, line 18: Why split this test into a separate test case? I would expect this test to behave the same for all different modes of authentication, so why not leave it in the base `UsersAPITest` class and let it run with all 3 authentications to confirm it always works correctly?

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -473,3 +473,89 @@ def test_saved_search(self):
...
+        ]
+        self.mock_list_workspaces.assert_has_calls(calls)
+        self.mock_service_account.get.assert_called_with(
+            'api/workspaces/my-seqr-billing/anvil-1kg project n\u00e5me with uni\u00e7\u00f8de/acl')
+        self.assertEqual(self.mock_service_account.get.call_count, 1)
+
+    def test_query_all_project_families_variants(self):
+        super(AnvilVariantSearchAPITest, self).test_query_all_project_families_variants()
+        assert_no_list_ws_has_acl(self, 2, path='api/workspaces/my-seqr-billing/anvil-project 1000 Genomes Demo/acl')
+        # self.mock_list_workspaces.assert_not_called() self.mock_service_account.get.assert_not_called()
```
hanars, line 513: should this be commented out?

hanars (changes requested):

seqr/views/apis/users_api_tests.py
ShifaSZ: I thought the user creating/updating should be disabled when AnVIL is enabled. But, you are right, we need these features for the `mix` cases at least.

seqr/views/apis/variant_search_api_tests.py
ShifaSZ: Updated

hanars (approved):
Awesome work getting this done!

PR #1481, opened 2020-10-27
fix discovery sheet handling of updated hpo categories

hanars:
Fixes https://github.com/broadinstitute/seqr-private/issues/927

HPO updated their category mapping, which broke our parsing for the discovery sheet. This updates the discovery sheet to work with the new HPO categories per the discussion on the ticket, and also cleans up the category to discovery column mapping to make the code easier to understand/ more explicit


ShifaSZ (approved):
Looks good.

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/739323394

PR #1483, opened 2020-10-28
Travis matrix

hanars:
This updates our travis configuration so instead of running one travis job that sequentially runs 3 different test suites, it triggers 3 different jobs for the 3 different sets of tests. This will make it much clearer from the traivs output where we have failing tests, and should make the tests go faster


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/739728708

ShifaSZ (approved):
Looks great.

PR #1488, opened 2020-11-02
Fix log forwarding on redeploy

hanars:


deploy/docker/seqr/config/gunicorn_config.py
```diff
@@ -3,6 +3,4 @@
 workers = 1
 loglevel = 'info'
 timeout = 3600   # seconds (default is 30)
-errorlog = '-'  # '${INSTALL_DIR}/logs/gunicorn-error.log'
-accesslog = '-' # '${INSTALL_DIR}/logs/gunicorn-access.log'
```
hanars, line 7: this was causing us to get duplicate logs in stackdriver for every URL that was hit. Since we have the robust django access logs, we don't also want the less-complete duplicate gunicorn logs

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -301,6 +301,19 @@ def deploy_seqr(settings):
...
+    print_separator('re-deploying seqr')
+
+    seqr_pod_name = get_pod_name('seqr', deployment_target=deployment_target)
+    if not seqr_pod_name:
+        raise ValueError('No seqr pod found, unable to re-deploy')
+    sleep_until_pod_is_running('seqr', deployment_target=deployment_target)
+
+    run_in_pod(seqr_pod_name, 'git pull', verbose=True)
+    run_in_pod(seqr_pod_name, './manage.py migrate', verbose=True)
+    run_in_pod(seqr_pod_name, '/usr/local/bin/restart_server.sh')
```
hanars, line 314: This is the series of commands I was running when I ssh'd in and did the in-place deployment. It seems safer to have that as a script that runs, rather than me copy-pasting commands in

ShifaSZ (approved):
Can't find any issue.

PR #1490, opened 2020-11-03
Status endpoint

hanars:
Adds a status endpoint that we can set up for monitoring in stackdriver. This endpoint is publicly available, so it does not return any details about our tech stack/ what our dependent services are. Instead, it checks if dependent services are up and logs errors if they aren't and we will set up alerting on those errors


ShifaSZ (approved):
Looks good.

PR #1494, opened 2020-11-05
User display

hanars:
This PR does to things related to the  staff user page:
- It makes installing the "django hijak" package only happen locally, and then only shows the corresponding "Login" buttons on the user page locally. The hijak plugin is really useful for testing because it lets you log in as any user so you can test different configurations, but its really too dangerous for us to have in production
- adds a column to the users table to indicate if a user is active. In djngo, you don't delete users who no longer get access to the app, because that messes up all sorts of foreign key relationships, you mark them as inactive and then they can't log in but the user model still exists for tracking purposes


hanars:
Passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/741680096

hanars:
passig codacy: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6462236

ui/pages/Staff/components/Users.jsx
```diff
@@ -29,15 +40,15 @@ const COLUMNS = [
...
+const Users = React.memo(({ users, loading, load, hijak }) =>
   <DataLoader load={load} content loading={false}>
     <DataTable
       striped
       idField="username"
       defaultSortColumn="email"
       loading={loading}
       data={users}
-      columns={COLUMNS}
+      columns={hijak ? HIJAK_COLUMNS : COLUMNS}
```
ShifaSZ, line 51: Should it show HIJAK_COLUMNS only when the user is a hijak-authorized user?

settings.py
```diff
@@ -220,15 +215,17 @@
     DEBUG = False
 else:
     DEBUG = True
-    # Enable CORS for local development
-    INSTALLED_APPS.append('corsheaders')
+    # Enable CORS and hijak for local development
+    INSTALLED_APPS = ['corsheaders', 'hijack'] + INSTALLED_APPS
```
ShifaSZ, line 219: It reports `Forbidden (CSRF token missing or incorrect.)` in my environment. Are you sure it works in your environment?

ShifaSZ (changes requested):
There are two issues or questions.

ui/pages/Staff/components/Users.jsx
hanars: this is a staff-only page, and hijak only shows during local development which is just staff users, so this is fine

settings.py
hanars: good catch, I messed up a couple settings. I'll push up the change

hanars: this is fixed

ShifaSZ (approved):

PR #1502, opened 2020-11-12
Added AnVIL/Google authentication.

ShifaSZ:
All the updates for the AnVIL/Google authentication have been added to the latest `dev` branch in this PR.


hanars:
The security test is failing because you added a vulnerable dependency: https://app.snyk.io/org/hanars/project/15db81e1-2826-41b9-82da-b0e882e39d1e/pr-check/0f1ea7d1-c6ac-431b-a0f2-79b7d22e1492

Do we need this dependency in the end?

seqr/utils/social_auth_pipeline.py
```diff
@@ -0,0 +1,18 @@
+import logging
+import requests
+
+from django.shortcuts import redirect
+
+from seqr.views.utils.terra_api_utils import _get_call_args
```
hanars, line 6: you should never import protected methods in python (any method that starts with `_`). If this is a helper method that is needed outside the file where it is defined, rename it. Also, if this is intended to be used outside of the terra api file, you probably should name it something more descriptive, like `get_terra_call_args`

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+@mock.patch('seqr.views.utils.terra_api_utils.TERRA_API_ROOT_URL', TEST_TERRA_API_ROOT_URL)
+@mock.patch('seqr.views.utils.terra_api_utils.GOOGLE_SERVICE_ACCOUNT_INFO', GOOGLE_SERVICE_ACCOUNT_INFO)
+@mock.patch('seqr.views.utils.terra_api_utils.time')
+class TerraApiUtilsCase(TestCase):
+    fixtures = ['users', 'social_auth']
+
+    def setUp(self):
+        self.user = User.objects.get(email = 'test_user@test.com')
+        self.extra_data = {"expires": 3599, "auth_time": 1603287741, "token_type": "Bearer", "access_token": "ya29.EXAMPLE"}
```
hanars, line 20: there's no reason for this to be a dictionary attached to the test class, just define these as constants at the top of the file

hanars (changes requested):

hanars:
I'm seeing a lot of codacy errors that need to be fixed: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687

For all the ones where you have `Parameters differ from overridden '...' method` and then you have a bunch of `No value for argument '...' in method call`, you can fix that by changing 
```
def test_something(self):
        super(AnvilAPITest, self).test_something()
```
to 
```
def test_something(self, *args):
        super(AnvilAPITest, self).test_something(*args)
```

seqr/utils/social_auth_pipeline.py
ShifaSZ: Converted `_anvil_call()` to public function and named `anvil_call()` because `anvil_call()` has a higher level than `get_call_args()` has. Changed it call `anvil_call()` here.

ShifaSZ:
> The security test is failing because you added a vulnerable dependency: https://app.snyk.io/org/hanars/project/15db81e1-2826-41b9-82da-b0e882e39d1e/pr-check/0f1ea7d1-c6ac-431b-a0f2-79b7d22e1492
> 
> Do we need this dependency in the end?

I got an error when I tried to open the link: 
```
Unable to display this organization
The organization does not exist, or you do not have permissions to access it.
```
Have my account (Github: shifasz or zhangshifa07504@gmail.com) been added to the permission list?

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: Moved to a constant.

hanars:
> > The security test is failing because you added a vulnerable dependency: https://app.snyk.io/org/hanars/project/15db81e1-2826-41b9-82da-b0e882e39d1e/pr-check/0f1ea7d1-c6ac-431b-a0f2-79b7d22e1492
> > Do we need this dependency in the end?
> 
> I got an error when I tried to open the link:
> 
> ```
> Unable to display this organization
> The organization does not exist, or you do not have permissions to access it.
> ```
> 
> Have my account (Github: shifasz or [zhangshifa07504@gmail.com](mailto:zhangshifa07504@gmail.com)) been added to the permission list?

I have invited you at your Broad email address

seqr/views/utils/terra_api_utils.py
```diff
@@ -0,0 +1,185 @@
...
+        except Exception as ee:
+            logger.info('Refresh token failed. {}'.format(str(ee)))
+            raise TerraAPIException('Refresh token failed. {}'.format(str(ee)))
+    return social.extra_data['access_token']
+
+
+def anvil_call(method, path, user=None, headers=None, root_url=None, access_token=None, **kwargs):
+    if user:
+        access_token = _get_social_access_token(user)
+
```
hanars, line 157: so this is a function that can be called in 2 pretty fundamentally different ways - with a user where you need to get the access token, or with an access token already provided. But because everything is optional arguments and there isn't good argument validation, its really easy to mess up using this function, either by providing both user and token or buy providing neither, and in both cases you would get weird undefined behavior. I think it would be better to make `anvil_call` require `access_token` as a positional argument, and then write another function for anvil calls with a user that calls it
```
def user_anvil_call(user, method, path, **kwargs):
   access_token = _get_social_access_token(user)
   return anvil_call(method, path, access_token, **kwargs)
```

hanars (changes requested):

hanars:
I am still seeing many codacy issues, although fewer than before: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687

I am also very concerned about the insecure package. It doesn't seem like we really need this package - all it does is authorize the service account against google, which seems like its just 2 api calls that could honestly just be made with the requests library. Am I wrong about that?

seqr/views/utils/terra_api_utils.py
ShifaSZ: Good point.

requirements.txt
```diff
@@ -6,6 +6,7 @@ django-cors-headers==3.0.2       # allows CORS requests for client-side developm
 social-auth-app-django==4.0.0    # the package for Django to authenticate users with social medieas
 social-auth-core==3.3.3          # the Python social authentication package. Required by social-auth-app-django
 google-auth==1.22.1              # for Google service account client and secure sessions using access tokens
+rsa==4.6                         # required by google-auth
```
hanars, line 9: to be clear: the issue synk is reporting is that this is a downstream dependency of the google auth package you added and it is very vulnerable and unsafe to use. Adding the vulnerable downstream dependency explicitly does absolutely nothing to make this less of a security threat. We need to either a) remove the package that introduces vulnerability (which is google-auth) or b) document why it is absolutely essential that we use a vulnerable package. I d==o not think we need the google auth package. It is a security threat. We can use something else that isn't a security threat. Take it out

hanars (changes requested):

ShifaSZ:
> I am still seeing many codacy issues, although fewer than before: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687
> 
> I am also very concerned about the insecure package. It doesn't seem like we really need this package - all it does is authorize the service account against google, which seems like its just 2 api calls that could honestly just be made with the requests library. Am I wrong about that?

If we don't use the `google-auth` package then we have to use the service account secret to sign a JWT token and use this token to request an access token from Google.  We also need to handle token refreshing as well.

hanars:
> > I am still seeing many codacy issues, although fewer than before: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687
> > I am also very concerned about the insecure package. It doesn't seem like we really need this package - all it does is authorize the service account against google, which seems like its just 2 api calls that could honestly just be made with the requests library. Am I wrong about that?
> 
> If we don't use the `google-auth` package then we have to use the service account secret to sign a JWT token and use this token to request an access token from Google. We also need to handle token refreshing as well.

How much work is that? Are there other libraries that could handle this for us? Are there ways to only require the submodules we are actually using, to see if maybe they don't have the vulnerable downstream dependencies? How slow would it be to get a new token every time and not need to worry about refreshing?

Its not impossible for us to add a vulnerable dependency, but its a bad idea and we need to try hard not to do it. We need to be able to show that we tried to do this without the vulnerability, and that this was the only option

hanars:
You can always check the latest codacy results here: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequests

ShifaSZ:
> > > I am still seeing many codacy issues, although fewer than before: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687
> > > I am also very concerned about the insecure package. It doesn't seem like we really need this package - all it does is authorize the service account against google, which seems like its just 2 api calls that could honestly just be made with the requests library. Am I wrong about that?
> > 
> > 
> > If we don't use the `google-auth` package then we have to use the service account secret to sign a JWT token and use this token to request an access token from Google. We also need to handle token refreshing as well.
> 
> How much work is that? Are there other libraries that could handle this for us? Are there ways to only require the submodules we are actually using, to see if maybe they don't have the vulnerable downstream dependencies? How slow would it be to get a new token every time and not need to worry about refreshing?
> 
> Its not impossible for us to add a vulnerable dependency, but its a bad idea and we need to try hard not to do it. We need to be able to show that we tried to do this without the vulnerability, and that this was the only option

I tried uninstalling the `rsa` package and run seqr to find out whether we actually use it for the service account. I found that the package is not used. It is weird to me that we must use RSA algorithm to load the service account secret. So I looked into the `google-auth` library and found that it actually loads a cryptograph-based RSA algorithm instead of the pure python RSA from `rsa` package. The `rsa` package is a backup for the case of the cryptograph-based RSA does not present. The code is like below:
```
try:
    # Prefer cryptograph-based RSA implementation.
    from google.auth.crypt import _cryptography_rsa

    RSASigner = _cryptography_rsa.RSASigner
    RSAVerifier = _cryptography_rsa.RSAVerifier
except ImportError:  # pragma: NO COVER
    # Fallback to pure-python RSA implementation if cryptography is
    # unavailable.
    from google.auth.crypt import _python_rsa

    RSASigner = _python_rsa.RSASigner
    RSAVerifier = _python_rsa.RSAVerifier
```
Therefore, the `rsa` package will never be used by seqr even it is in the dependence list.

I could also write a service account client by myself. But I don't think Google recommends that.

Another alternative is to stop using a service account for now. There are some small changes that will be required to pass in the user information to the workspace ACL fetching function.

hanars:
I didn't realize stopping using the service account was an option. That is a a MUCH better approach, we should definitely do that if we can. What functionality would we lose without it?

ShifaSZ:
> I didn't realize stopping using the service account was an option. That is a a MUCH better approach, we should definitely do that if we can. What functionality would we lose without it?

Currently, if we stop using the service account, no functionality will be lost. I'm not sure if there would be any in the future.

hanars:
> > I didn't realize stopping using the service account was an option. That is a a MUCH better approach, we should definitely do that if we can. What functionality would we lose without it?
> 
> Currently, if we stop using the service account, no functionality will be lost. I'm not sure if there would be any in the future.

Great! So lets take it all out for now, and if we need it in the future we can revisit the issue then

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -693,11 +697,20 @@ def get_project_collaborators_by_username(project, include_permissions=True):
             collaborator, include_permissions, can_edit=True
         )
 
+    if project_has_anvil(project):
+        anvil_collaborators = get_workspace_collaborators(user, project.workspace_namespace, project.workspace_name)
+        collaborators.update({
+            collab.username: _get_collaborator_json(collab, include_permissions,
+                                                    can_edit = anvil_collaborators[collab.email] == CAN_EDIT,
+                                                    check_anvil = True)
```
hanars, line 705: so it looks like `check_anvil` here will set the `isAnvil ` attribute based on whether or not that user is currently logged in through google. But I think what we actually want here is for `isAnvil` to be True for all of these anvil collaborators. I think it would be clearer if you changed `_get_json_for_user` to take `is_anvil` as the optional argument, instead of `check_anvil`

seqr/views/utils/permissions_utils.py
```diff
@@ -19,14 +22,49 @@ def get_project_and_check_permissions(project_guid, user, **kwargs):
...
+        return None
+    if anvil_permission['accessLevel'] in ['WRITER', 'OWNER']:
+        return CAN_EDIT
+    return CAN_VIEW
+
+
+def anvil_has_perm(user, permission_level, project):
+    if not project_has_anvil(project):
+        return False
+    workspace_acl = user_get_workspace_acl(user, project.workspace_namespace, project.workspace_name)
```
hanars, line 40: do you really need to fetch the entire workspace access list to check if the currently authenticated user has access? It seems like since you are using the credentials of that user, there would be a faster way to check (for example, to check if the user is allowed to view the workspace, you could just do a head request against the workspace details and see if they get an access denier error, or something like that)

seqr/views/utils/test_utils.py
```diff
@@ -577,3 +715,11 @@ def call_request_json(self, index=-1):
     'numExon': 2,
     '_sort': [1049045387],
 }
+
+GOOGLE_API_TOKEN_URL = 'https://oauth2.googleapis.com/token'
+GOOGLE_OAUTH2_URL = 'https://accounts.google.com/o/oauth2/'
+GOOGLE_AUTHORIZATION_URL = '{}auth'.format(GOOGLE_OAUTH2_URL)
+GOOGLE_ACCESS_TOKEN_URL = '{}token'.format(GOOGLE_OAUTH2_URL)
+GOOGLE_REVOKE_TOKEN_URL = '{}revoke'.format(GOOGLE_OAUTH2_URL)
```
hanars, line 723: Do we need all of these now that we got rid of service accounts?

settings.py
```diff
@@ -112,6 +113,8 @@
             'context_processors': [
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',  # required for admin template
+                'social_django.context_processors.backends',
+                'social_django.context_processors.login_redirect',
```
hanars, line 117: can you add comments explaining what these are for

settings.py
```diff
@@ -286,3 +291,57 @@
 MME_SLACK_ALERT_NOTIFICATION_CHANNEL = 'matchmaker_alerts'
 MME_SLACK_MATCH_NOTIFICATION_CHANNEL = 'matchmaker_matches'
 MME_SLACK_SEQR_MATCH_NOTIFICATION_CHANNEL = 'matchmaker_seqr_match'
+
+#########################################################
+#  AnVIL Terra API specific settings
+#########################################################
+GOOGLE_AUTH_CONFIG_DIR = os.environ.get('GOOGLE_AUTH_CONFIG_DIR', '')
```
hanars, line 298: do we still need any of this now that we don't have service accounts?

hanars (changes requested):

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: Okay

seqr/views/utils/permissions_utils.py
ShifaSZ: Hmm... I re-visited the API list and found getting workspace details is the better one. The details of a workspace also have an `accessLevel` field which is what we need.

hanars: okay but what if a user has view access to a workspace but does not have permission to view the workspace ACL?

seqr/views/utils/test_utils.py
ShifaSZ: No. Delete them.

ShifaSZ: Actually, we still need these but we don't need the one (`GOOGLE_API_TOKEN_URL`) above

settings.py
ShifaSZ: Deleted the settings for the service account.

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -104,6 +107,7 @@ def _get_json_for_user(user):
         _to_camel_case(field): getattr(user, field) for field in [
         'username', 'email', 'first_name', 'last_name', 'last_login', 'is_staff', 'is_active', 'date_joined', 'id',
     ]}
+    user_json['isAnvil'] = is_google_authenticated(user) if is_anvil else False
```
hanars, line 110: This doesn't fix the issue I pointed out in my last review, it just changed the name of the parameter without fixing the behavior. When you are getting the json for the list of anvil-only collaborators, you want `isAnvil` to be true, regardless of whether or not that user ever has logged in with google. For the current user, you want to make `is_anvil` dependent on how they logged in, but for the collaborator json this should be unconditionally true. Here you should just do `user_json['isAnvil'] = is_anvil`, and then when you call the function for the current user you can do `_get_json_for_user(user, is_anvil=is_google_authenticated(user))`

seqr/views/utils/permissions_utils.py
```diff
@@ -19,14 +23,48 @@ def get_project_and_check_permissions(project_guid, user, **kwargs):
...
+    return CAN_VIEW if anvil_permission['accessLevel'] == 'READER' else None
+
+
+def anvil_has_perm(user, permission_level, project):
+    if not project_has_anvil(project):
+        return False
+    workspace_permission = user_get_workspace_access_level(user, project.workspace_namespace, project.workspace_name)
+    if not workspace_permission:
+        return False
+    workspace_permission['pending'] = False
```
hanars, line 44: why is this line needed? If the "pending" attribute isn't set here, you can always change the check in `_map_anvil_seqr_permission`  from `anvil_permission['pending']` to `anvil_permission.get('pending')`

seqr/views/utils/permissions_utils.py
```diff
@@ -19,14 +23,48 @@ def get_project_and_check_permissions(project_guid, user, **kwargs):
...
+    permission = _map_anvil_seqr_permission(workspace_permission)
+    return True if permission == CAN_EDIT else permission == permission_level
+
+
+def get_collaborator_permission_levels(user, workspace_namespace, workspace_name):
```
hanars, line 49: can you rename this something like `get_collaborator_anvil_permissions` or something to make it clear that this function is only checking the anvil permissions

seqr/views/utils/terra_api_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    logger.info('{} {} {} {} {}'.format(method.upper(), url, r.status_code, len(r.text), user))
+
+    return json.loads(r.text)
+
+
+def _user_anvil_call(method, path, user, **kwargs):
+    access_token = _get_social_access_token(user)
+    try:
+        return anvil_call(method, path, access_token, user=user, **kwargs)
+    except TerraNoAccessException:
```
hanars, line 94: I understand why we do this, but I'm a little worried about always ignoring all 404s and 403s. I think I would feel better if `anvil_call`  had an "allowed error" parameter that could also be passed through here, so for example in `user_get_workspace_access_level` you could do ` _user_anvil_call('get', path, user, allowed_error=403)` and then in `anvil_call` it would check if the status was an allowed error and if so return and empty dictionary and if not raise the error

hanars (changes requested):

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: I see the issue you are talking about now. But, in my opinion, not all the collaborators fetched from AnVIL have logged-in with AnVIL yet. So, it still needs to call `is_google_authenticated(collaborator)` to check the `is_anvil` for the collaborator in the list. Do you agree? 

Another reason why I used `check_anvil` as an optional parameter for `_get_json_for_user()` was that I didn't like to change the other places where call `_get_json_for_user()`.

I could change the function to `_get_json_for_user(user, is_anvil=None)` and it does `user_json['isAnvil] =  is_anvil if is_anvil != None else is_google_authenticated(user)`. Which means it will use `is_anvil` if provided. How do you think about it?

seqr/views/utils/permissions_utils.py
ShifaSZ: Good point.

seqr/views/utils/terra_api_utils.py
ShifaSZ: The issue is what response status codes can be silently ignored. We could have below status codes.
```
1. User has not registered (api: /register) : 404  -- not an error
2. Client ID is not whitelisted: 401 -- error
3. Access token is not valid or has expired: 401 -- error
4. Workspace does not exist or has no access permission (api: ws ACL or ws details): 404 -- not an error
5. User has access to a workspace but no ‘can share’ (api: ws ACL): 403 -- not an error
6. Other non-200 response status codes: error
```
We need to ignore both 403 and 404 for the workspace ACL requests. A single status code can't solve the issue. Maybe we should add an option to ignore both codes.

seqr/views/utils/orm_to_json_utils.py
hanars: so the `isAnvil` attribute for collaborators is for how they are displayed on the project page. So if a collaborator has access through Anvil but has never logged in to regular seqr at all, it would be really bad to show them as a local collaborator, given that their access comes purely from anvil. The big difference between the local and anvil sections is that local users can be edited and anvil users can't. So if a user has access through anvil, it doesn't matter how they are logging in at this point, theres no way to remove them from the project in seqr and therefore they should show up in the uneditable anvil section

If you want to leave `_get_json_for_user` (which I totally get) you could go back to using the `check_anvil` defaulted to True the way you did before and leave this line the same where it checks the google authentication if that is True. Then in the function to get  json for collaborators you set `check_anvil` to False, and then when you go through the list of anvil collaborator emails you update the `isAnvil` attribute directly

seqr/views/utils/terra_api_utils.py
hanars: 1) If a user has not registered with anvil they should be using seqr only in local mode and google log in should have failed. This is a absolutely an error
4) If seqr is trying to access a workspace that doesn't exist something has gone really wrong and this is an error

It looks to me like the 404 cases are actually errors, so the only error code that could ever be ignored is 403. I think it also depends on what api endpoint is being called- for get workspace ACL we expect there to be 403s for some users and the behavior we want is for it to return an empty dictionary and not report any issue. For checking if a user has access to a project, if a user does not have access we expect a 403 but the check permission functions throw errors when users shouldn't have access, so  I think throwing a `PermissionDenied` error would be the easiest way to deal with it. Given that its just this one error code that has an Exception calls thats naturally associated with it, I think instead of adding `allowed_error` we should just change the error handling in `anvil_call` to throw `PermissionDenied` for a 403 and the main Terra exception for everything else. Then the `get_workspace_acl` function would have error handling just for `PermissionDenied`, and everything else wouldn't and would allow seqr to use its default error handling for that

seqr/views/utils/orm_to_json_utils.py
hanars: Oh sorry I commented before seeing your latest code change. I think that approach is fine, so feel free to ignore that suggestion

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -693,11 +697,20 @@ def get_project_collaborators_by_username(project, include_permissions=True):
...
         )
 
+    if project_has_anvil(project):
+        permission_levels = get_collaborator_permission_levels(user, project.workspace_namespace, project.workspace_name)
+        collaborators.update({
+            collab.username: _get_collaborator_json(collab, include_permissions,
+                                                    can_edit = permission_levels[collab.email] == CAN_EDIT,
+                                                    is_anvil = True)
+            for collab in User.objects.filter(email__in = permission_levels.keys())
+        })
```
hanars, line 707: Sorry for not catching this earlier, but I just realized that if a user has access to a seqr project through AnVIL but has never logged in they won't show up, right? I think from the user's perspective, they want to know everyone who could access the project, regardless of whether or not they have yet (currently, for example, we show collaborators who have never logged in). You could add something here to force them to show up by just creating a json record for those email addresses:
```
for email, permission_level in in  permission_levels.items():
   if email not in collaborators:
      collaborators[email] = {
         'username': email, // just to ensure everything has a unique ID
         'email': email, 
         'isAnvil': True, '
         'hasViewPermissions': True, 
         'hasEditPermissions': permission_level == CAN_EDIT
      }
```

hanars (changes requested):

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: Yes. The original code only shows those AnVIL users who have logged in with Google at least once.
I agree that we'd better show all the users who have access to the project no matter they have logged in seqr locally or with AnVIL.
If we do as you've proposed, only those users who have never logged in seqr will be shown as `AnVIL` users. Is that what you want? I think we should show all users who can obtain permissions from AnVIL as `AnVIL` users.

hanars: We want all users who are in the workspace ACL to be shown as anvil users, regardless of if they've ever logged in. I realize that I got confused about how  the collaborators dictionary was keyed, so we should instead do something like 
```
if project_has_anvil(project):
        permission_levels = get_collaborator_permission_levels(user, project.workspace_namespace, project.workspace_name)
        users_by_email = {u.email: u for u in User.objects.filter(email__in = permission_levels.keys())}
        collaborators.update({
            collab.username: _get_collaborator_json(collab, include_permissions,
                                                    can_edit = permission_levels[collab.email] == CAN_EDIT,
                                                    is_anvil = True)
            for collab in users_by_email.values())
        })
        no_user_emails = set(permission_levels.keys()) - set(users_by_email.keys())
        for email in no_user_emails:
                collaborators[email] = {
                      'username': email, // just to ensure everything has a unique ID
                       'email': email, 
                       'isAnvil': True, '
                       'hasViewPermissions': True, 
                       'hasEditPermissions': permission_levels[email] == CAN_EDIT
                }
```

seqr/views/utils/terra_api_utils.py
ShifaSZ: The `/register` API is for the detection of the user's registration on AnVIL while signing in. It is not an error because when the exception detected the code will redirect back to the login page and won't just stop running.

We can't avoid a user access a workspace that he has no permission during the transitioning phase. When a user has local access to a project has no access to the corresponding workspace (because it may take time to configure all workspaces for the seqr users), if this user requests for the workspace details or ACL then it will be a 404 status.

Maybe we can throw a TerraForbidentException for 403 and a TerraNotFoundException for 404 in `anvil_call` and let the caller function to decide how to handle them.

hanars: "Maybe we can throw a TerraForbidentException for 403 and a TerraNotFoundException for 404 in anvil_call and let the caller function to decide how to handle them."

I like this approach. I think I would prefer to use the `PermissionDenied` exception for 403s because thats the standard exception we use all over when permission is denied so we can just let the default error handling deal with it in many cases

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: I see what you mean. My concern is that when a project manager wants to change a Google signed-in collaborator's local permissions, the manager has to firstly revoke the collaborator's AnVIL workspace access and then edit the collaborator locally.

hanars: theres really no point in updating local permissions for users who have access through anvil, because the anvil permissions override the local ones anyways. I think its an edge case thats okay to ignore

seqr/views/utils/terra_api_utils.py
ShifaSZ: Status code 403 means the access is forbidden. If we use `PermissionDenied`, I have the concern that if there is a real `PermissionDenied` during the call, it could result in miss-handling.

hanars: permission denied means that access is forbidden, I don't understand the difference?

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: I found `lastName` and `displayName` fields are required. To avoid further unexpected missing fields, I added all the fields as below:
```
                collaborators[email] = {
                    'username': email,  # to ensure everything has a unique ID
                    'email': email,
                    'isAnvil': True,
                    'hasViewPermissions': True,
                    'hasEditPermissions': permission == CAN_EDIT,
                    'displayName': email,
                    'lastName': email,
                    'is_staff': False,
                    'is_active': True,
                    'first_name': '', 'last_login': '', 'date_joined': '', 'id': '',
                }
```

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -693,11 +697,32 @@ def get_project_collaborators_by_username(project, include_permissions=True):
             collaborator, include_permissions, can_edit=True
         )
 
+    if project_has_anvil(project):
+        permission_levels = get_workspace_collaborator_perms(user, project.workspace_namespace, project.workspace_name)
+        for email, permission in permission_levels.items():
+            collaborator = User.objects.filter(email=email)
```
hanars, line 703: this is really inefficient in terms of database queries, which can be slow. Querying 100 users takes roughly the same time as querying 1 so I feel really strongly that you need to do one query here. Something like 
```
users_by_email = {u.email: u for u in User.objects.filter(email__in = permission_levels.keys())}
for email, permission in permission_levels.items():
    collaborator = users_by_email.get(email)
    if collaborator:
    ...
```

seqr/views/utils/terra_api_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    request_func = getattr(requests, method)
+    headers.update({'Authorization': 'Bearer {}'.format(access_token)})
+    r = request_func(url, headers=headers, **kwargs)
+
+    if r.status_code == 404 or r.status_code == 403:
+        logger.warning('{} {} {} {} {}'.format(method.upper(), url, r.status_code, len(r.text), user))
+        message = 'Warning: called Terra API: /{} got status: {} with a reason: {}'.format(path, r.status_code, r.reason)
+        if r.status_code == 404:
+            raise TerraNotFoundException(message)
+        raise PermissionDenied(message)
```
hanars, line 77: this would be a lot cleaner if you did separate if statements for 403 and 404. Also all 40X responses will be logged as warnings, so I don't think you need to also explicitly log here
```
if r.status_code == 404:
        raise TerraNotFoundException('Warning: called Terra API: {} /{} got status 404 with reason: {}'.format(method.upper(), path, r.reason)
if r.status_code == 403:
        raise PermissionDenied('Access denied (403) from Terra API: {} /{} with reason: {}'.format(method.upper(), path, r.reason)
```

seqr/views/utils/terra_api_utils.py
```diff
@@ -0,0 +1,157 @@
...
+              "canCompute": true
+            }
+          }
+    """
+    path = "api/workspaces/{0}/{1}/acl".format(workspace_namespace, workspace_name)
+    try:
+        return _user_anvil_call('get', path, user).get('acl', {})
+    except TerraNotFoundException:
+        return {}
+    except PermissionDenied:
```
hanars, line 156: you can do it as one line `except (TerraNotFoundException, PermissionDenied)`

hanars (changes requested):

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: Okay.

seqr/views/utils/terra_api_utils.py
ShifaSZ: Okay

ShifaSZ: It seems that for the Terra API. There is no other place that will process the status code. So the log is required here or the function which handles the exception.

hanars: unhandled exceptions in seqr are handled by the middleware and logged as exceptions. If you don't want that to happen, then you need to add exception handling for these errors whenever `anvil_call` is called, or you need to update middleware.py to properly handle them. But at any rate, something will eventually receive a `TerraNotFoundException` and either it will choose to ignore that exception, in which case not logging is fine, or the exception will bubble up and be logged in the middleware. I thin the best bet is to update `EXCEPTION_ERROR_MAP` in middleware.py to map a `TerraNotFoundException` to a 404

hanars:
So all of the codacy "Parameters differ from overridden " and "Possible hardcoded password" errors can be ignored, but the other codacy errors can't be, especially the  reimport error in terra_api_utils_tests.py

hanars:
codacy (failures are okay): https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6500687

hanars (approved):

PR #1503, opened 2020-11-13
expire cached search results

hanars:
Seqr caches search results, but when data reloads it is crucial that the cache is reset, or users see weird/ inaccurate results (see https://github.com/broadinstitute/seqr-private/issues/935)

One reason why resetting the cache fails is that there are so many keys to delete that the reset times out. While caching is crucial in the shorter terms (especially for recessive searches where there is a bunch of post processing that we cache) search itself always completes in a reasonable amount of time, so having the results cache expire after a couple weeks won't cause any serious performance changes to the user, and will ensure that users can never see seriously outdated results

### DIFF
--- a/seqr/utils/elasticsearch/es_utils_tests.py
+++ b/seqr/utils/elasticsearch/es_utils_tests.py
@@ -4,6 +4,7 @@ import jmespath
 import json
 import re
 from collections import defaultdict
+from datetime import timedelta
 from django.test import TestCase
 from elasticsearch.exceptions import ConnectionTimeout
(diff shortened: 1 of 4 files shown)


ShifaSZ (approved):
It looks good.

hanars:
passing test: https://travis-ci.org/github/broadinstitute/seqr/builds/743443324

hanars:
passing codacy: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6506066

PR #1506, opened 2020-11-17
Enables modsecurity and owasp rulesets on ingress-nginx

sjahl:
This is the less invasive option for SQR-68, where we enable the modsecurity plugin for our existing ingress-nginx.

My understanding is that this includes the ruleset from https://github.com/coreruleset/coreruleset, and therefore we might need to upgrade our ingress-nginx controller periodically (and perhaps initially) to stay up to date. By default, it operates in detection-only mode, so we'd need some log alerts to let us know what's happening. I can also look into how to configure it to actually act on threats once we're confident that it won't break anything.

One open question I have on this:
- It looks like the ingress-nginx project provides a helm chart to deploy itself. Did we use that to deploy this ingress originally, or did we simply use the [quickstart](https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud/deploy.yaml) ?

The other option I'm exploring for this to switch to a native GKE Ingress -- this one will take me a bit longer, so I wanted to leave this option open first. We can always switch to a GKE Ingress later.

### DIFF
--- a/deploy/kubernetes/nginx/nginx.yaml
+++ b/deploy/kubernetes/nginx/nginx.yaml
@@ -48,7 +48,12 @@ metadata:
   namespace: ingress-nginx
 data:
   enable-modsecurity: "true"
-  enable-owasp-modsecurity-crs: "true"
+  modsecurity-snippet: |
+    SecAuditLogType Serial
+    SecAuditLog /dev/stdout
+    SecAuditLogFormat JSON
+    SecRuleRemoveById 920350
+    Include /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf
 ---
 # docs @ https://kubernetes.github.io/ingress-nginx/user-guide/basic-usage/
 kind: Ingress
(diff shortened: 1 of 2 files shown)


hanars:
@bw2 did the initial deployment, so hopefully he can answer your question

hanars:
I would definitely recommend we do an initial upgrade for the  controller when we deploy this. Is there any way to update the `deploy_nginx` function to optionally do that, or at least add instruction for how to do that somewhere?

sjahl:
It looks like `deploy_nginx` will use whatever `master` is on the upstream repo, when it's called.

As a note for future work, tracking from master like this could unintentionally introduce some problems if the ingress-nginx project decides something isn't broken enough 😄

hanars:
> It looks like `deploy_nginx` will use whatever `master` is on the upstream repo, when it's called.
> 
> As a note for future work, tracking from master like this could unintentionally introduce some problems if the ingress-nginx project decides something isn't broken enough 😄

Yeah, we should change that so it uses a pinned version instead of master, if thats at all possible.

hanars:
passing codacy: https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6522917

hanars:
passing tests (failure is due to an unrelated test failing on dev that has since been fixed): https://travis-ci.org/github/broadinstitute/seqr/jobs/745510336

hanars (approved):

PR #1517, opened 2020-11-24
generate postgres password

hanars:
Creates a new postgres password whenever we deploy secrets

### DIFF
--- a/deploy/kubernetes/seqr/seqr.gcloud.yaml
+++ b/deploy/kubernetes/seqr/seqr.gcloud.yaml
@@ -72,15 +72,12 @@ spec:
               name: all-settings
         env:
         - name: POSTGRES_USERNAME
-          valueFrom:
-            secretKeyRef:
-              name: postgres-secrets
-              key: postgres.username
+          value: "{{ POSTGRES_USERNAME }}"
         - name: POSTGRES_PASSWORD
           valueFrom:
             secretKeyRef:
               name: postgres-secrets
-              key: postgres.password
+              key: password
         - name: OMIM_KEY
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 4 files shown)


deploy/kubernetes/shared-settings.yaml
```diff
@@ -15,6 +15,8 @@ NUM_NODES_PER_NODE_POOL: 4    # used when creating nodes for gcloud deployments
 
 GUNICORN_WORKER_THREADS: 4    # number of webserver threads
 
+POSTGRES_USERNAME: postgres
```
hanars, line 18: this is actually hardcoded in some of our manage commands so making it a secret is unnecessary

deploy/kubernetes/seqr/seqr.gcloud.yaml
```diff
@@ -72,15 +72,12 @@ spec:
...
-            secretKeyRef:
-              name: postgres-secrets
-              key: postgres.username
+          value: "{{ POSTGRES_USERNAME }}"
         - name: POSTGRES_PASSWORD
           valueFrom:
             secretKeyRef:
               name: postgres-secrets
-              key: postgres.password
+              key: password
```
sjahl, line 80: If we update the postgres password on deployment via `deploy_secrets()`, does seqr require a restart (or pod  redeployment) to pick up the new `POSTGRES_PASSWORD` env var value? Won't be an issue if we always redeploy seqr when deploying postgres, but could be if postgres or secrets get deployed on their own.

hanars: so we will need to redeploy seqr every time we redeploy secrets, but I generally do that anyways so it doesn't really change our deployment much. Redeploying postgres without redeploying secrets explicitly won't cause a new password to be generated so that won't affect seqr's deployment

sjahl (approved):

hanars:
codacy (error is okay to ignore): https://app.codacy.com/gh/macarthur-lab/seqr/pullRequest?prid=6562618

passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/745682975

PR #1528, opened 2020-12-02
Case review updates

hanars:
We have one seqr project that we run out of our lab where we have special functionality for doing case review. Currently, we permission this view by making it available to staff users in all projects, and then  the users just don't use it in all the unnecessary projects. Many of the users who use case review are people who only need access to the one project they review, but because the only way to access case review is as a staff user, they are granted staff access to all projects, which is not great. This changes the permissioning such that case review is enabled on a per project basis, and managers of that project will be able to access case review


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/747244249

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -189,12 +198,13 @@ def _process_result(result, family):
...
+    kwargs = {'additional_model_fields': _get_case_review_fields(families[0], families[0].project, user)}
     if project_guid or not skip_nested:
-        kwargs = {'nested_fields': [{'fields': ('project', 'guid'), 'value': project_guid}]}
+        kwargs.update({'nested_fields': [{'fields': ('project', 'guid'), 'value': project_guid}]})
     else:
-        kwargs = {'additional_model_fields': ['project_id']}
+        kwargs['additional_model_fields'] = (kwargs['additional_model_fields'] or []) + ['project_id']
 
-    return _get_json_for_models(families, user=user, process_result=_process_result, **kwargs)
+    return _get_json_for_models(families, user=user, process_result=_process_result,  **kwargs)
```
ShifaSZ, line 207: An extra blank is added before `**kwargs`

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -189,12 +198,13 @@ def _process_result(result, family):
     if add_individual_guids_field:
         prefetch_related_objects(families, 'individual_set')
 
+    kwargs = {'additional_model_fields': _get_case_review_fields(families[0], families[0].project, user)}
     if project_guid or not skip_nested:
-        kwargs = {'nested_fields': [{'fields': ('project', 'guid'), 'value': project_guid}]}
+        kwargs.update({'nested_fields': [{'fields': ('project', 'guid'), 'value': project_guid}]})
     else:
-        kwargs = {'additional_model_fields': ['project_id']}
+        kwargs['additional_model_fields'] = (kwargs['additional_model_fields'] or []) + ['project_id']
```
ShifaSZ, line 205: Using `kwargs['additional_model_fields'].append('project_id')` as the code for the individual looks nicer.

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -145,6 +146,12 @@ def _get_json_for_project(project, user, **kwargs):
     return _get_json_for_model(project, get_json_for_models=get_json_for_projects, user=user, **kwargs)
 
 
+def _get_case_review_fields(model, project, user):
+    if not (user and has_case_review_permissions(project, user)):
+        return None
```
ShifaSZ, line 151: It seems that returning `[]` is better than returning `None`.

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -245,6 +258,9 @@ def _process_result(result, individual):
             result['sampleGuids'] = [s.guid for s in individual.sample_set.all()]
             result['igvSampleGuids'] = [s.guid for s in individual.igvsample_set.all()]
 
+    kwargs = {
+        'additional_model_fields': _get_case_review_fields(individuals[0], individuals[0].family.project, user) or []
```
ShifaSZ, line 262: `or []` can be removed if `_get_case_review_fields()` return `[]` when there is no permission.

ShifaSZ (changes requested):
There may be some minor coding style issues.

ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/747691119

PR #1533, opened 2020-12-03
Added Google auth for local users.

ShifaSZ:
This PR adds the ability for existing users to be able to log in through Google, even before seqr is connected to AnVIL.
If the environment variable `GOOGLE_AUTH_CONFIG_DIR` is defined then Google auth is enabled. An existing seqr user can log in through Google if the user's email matches the Google account's email. Only when the environment variable `TERRA_API_ROOT_URL` is defined, seqr will use AnVIL for Google-signing-in user authentication.

### DIFF
--- a/seqr/utils/social_auth_pipeline.py
+++ b/seqr/utils/social_auth_pipeline.py
@@ -16,10 +16,14 @@ def validate_anvil_registration(backend, response, *args, **kwargs):
             return redirect('/login?anvilLoginFailed=true')
 
 
-def log_signed_in(backend, response, user=None, is_new=False, *args, **kwargs):
+def validate_user_exist(backend, response, user=None, *args, **kwargs):
     if not user:
-        logger.warning('Google user {} is trying to login without an existing seqr account'.format(response['email']))
+        logger.warning('Google user {} is trying to login without an existing seqr account ({}).'
+                       .format(response['email'], backend.name))
         return redirect('/login?googleLoginFailed=true')
+
+
+def log_signed_in(backend, response, is_new=False, *args, **kwargs):
     logger.info('Logged in {} ({})'.format(response['email'], backend.name))
     if is_new:
         logger.info('Created user {} ({})'.format(response['email'], backend.name))
--- a/seqr/utils/social_auth_pipeline.py
+++ b/seqr/utils/social_auth_pipeline.py
@@ -13,10 +13,13 @@ def validate_anvil_registration(backend, response, *args, **kwargs):
             anvil_call('get', 'register', response['access_token'])
         except TerraNotFoundException as et:
             logger.warning('User {} is trying to login without registration on AnVIL. {}'.format(response['email'], str(et)))
-            return redirect('/login?googleLoginFailed=true')
+            return redirect('/login?anvilLoginFailed=true')
 
 
-def log_signed_in(backend, response, is_new=False, *args, **kwargs):
+def log_signed_in(backend, response, user=None, is_new=False, *args, **kwargs):
+    if not user:
+        logger.warning('Google user {} is trying to login without an existing seqr account'.format(response['email']))
+        return redirect('/login?googleLoginFailed=true')
     logger.info('Logged in {} ({})'.format(response['email'], backend.name))
     if is_new:
         logger.info('Created user {} ({})'.format(response['email'], backend.name))
--- a/settings.py
+++ b/settings.py
@@ -327,33 +327,25 @@ SOCIAL_AUTH_POSTGRES_JSONFIELD = True
 SOCIAL_AUTH_URL_NAMESPACE = 'social'
 SOCIAL_AUTH_LOGIN_REDIRECT_URL = '/'
 
+SOCIAL_AUTH_PIPELINE = (
+    'social_core.pipeline.social_auth.social_details',
+    'social_core.pipeline.social_auth.social_uid',
+    'social_core.pipeline.social_auth.social_user',
+    'social_core.pipeline.social_auth.associate_by_email',
+    'social_core.pipeline.social_auth.associate_user',
+    'seqr.utils.social_auth_pipeline.log_signed_in',
+)
+
 if TERRA_API_ROOT_URL:
     SOCIAL_AUTH_GOOGLE_OAUTH2_AUTH_EXTRA_ARGUMENTS = {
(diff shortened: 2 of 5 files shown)


settings.py
```diff
@@ -315,25 +317,21 @@
...
+SOCIAL_AUTH_POSTGRES_JSONFIELD = True
+SOCIAL_AUTH_URL_NAMESPACE = 'social'
+SOCIAL_AUTH_LOGIN_REDIRECT_URL = '/'
 
-    SOCIAL_AUTH_GOOGLE_PLUS_AUTH_EXTRA_ARGUMENTS = {
-          'access_type': 'offline'
+if TERRA_API_ROOT_URL:
+    SOCIAL_AUTH_GOOGLE_OAUTH2_AUTH_EXTRA_ARGUMENTS = {
+        'access_type': 'offline',  # to make the access_token can be refreshed after expired (expiration time is 1 hour)
     }
```
hanars, line 333: is there a reason we don't want this in the non-anvil case?

settings.py
```diff
@@ -347,4 +345,15 @@
...
+        'social_core.pipeline.social_auth.social_details',
+        'social_core.pipeline.social_auth.social_uid',
+        'social_core.pipeline.social_auth.social_user',
+        'social_core.pipeline.user.get_username',
+        'social_core.pipeline.social_auth.associate_by_email',
+        'social_core.pipeline.social_auth.associate_user',
+        'social_core.pipeline.social_auth.load_extra_data',
+        'social_core.pipeline.user.user_details',
+        'seqr.utils.social_auth_pipeline.log_signed_in',
+    )
```
hanars, line 359: theres a lot of copy-paste here, which makes it hard to see the clear differences between the pipeline in the two cases. it might be helpful if you defined it and then edit it to make it work in the terra case, so something like

```
SOCIAL_AUTH_PIPELINE = [
   ...
]
if TERRA_API_ROOT_URL:
   ...
   SOCIAL_AUTH_PIPELINE.insert(0, 'seqr.utils.social_auth_pipeline.validate_anvil_registration')
   SOCIAL_AUTH_PIPELINE.insert(6, 'social_core.pipeline.user.create_user')
```

hanars (changes requested):

settings.py
ShifaSZ: We don't need the access_token in the non-anvil case. It can be a little more secure (less likely to leak the access_token) because the access_token will be expired soon and can't be refreshed with the default 'online' access type. Maybe I should try updating the pipeline to get rid of access_token completely because it will be much safer.

settings.py
ShifaSZ: Sounds good.

seqr/utils/social_auth_pipeline.py
```diff
@@ -13,10 +13,13 @@ def validate_anvil_registration(backend, response, *args, **kwargs):
...
             logger.warning('User {} is trying to login without registration on AnVIL. {}'.format(response['email'], str(et)))
-            return redirect('/login?googleLoginFailed=true')
+            return redirect('/login?anvilLoginFailed=true')
 
 
-def log_signed_in(backend, response, is_new=False, *args, **kwargs):
+def log_signed_in(backend, response, user=None, is_new=False, *args, **kwargs):
+    if not user:
+        logger.warning('Google user {} is trying to login without an existing seqr account'.format(response['email']))
+        return redirect('/login?googleLoginFailed=true')
```
hanars, line 22: this is important logic, but it doesn't belong in the `log_signed_in` helper step at the end of the pipeline. You should add a custom step named something like `validate_existing_user` that is run as soon in the pipeline as possible and is only in the non-anvil pipeline

settings.py
hanars: I see what you are saying about only wanting this in the anvil case and I agree. I think it is up to you whether you want to update the pipeline not to have the access token entirely or not, I don't think its necessary but if you want to do it go ahead

settings.py
```diff
@@ -315,36 +317,35 @@
...
-    )
-    INSTALLED_APPS.append('social_django')
+    SOCIAL_AUTH_PIPELINE = ('seqr.utils.social_auth_pipeline.validate_anvil_registration',) + \
+                           SOCIAL_AUTH_PIPELINE[:4] + \
+                           ('social_core.pipeline.user.get_username',
+                           'social_core.pipeline.user.create_user') + \
+                           SOCIAL_AUTH_PIPELINE[4:5] + \
+                           ('social_core.pipeline.social_auth.load_extra_data',
+                           'social_core.pipeline.user.user_details') + \
+                           SOCIAL_AUTH_PIPELINE[5:]
```
hanars, line 351: I know this was my suggestion, but seeing it its still hard to read. Maybe do something like

```
SOCIAL_AUTH_INITIAL_PIPELINE = (
    'social_core.pipeline.social_auth.social_details',
    'social_core.pipeline.social_auth.social_uid',
    'social_core.pipeline.social_auth.social_user',
    'social_core.pipeline.social_auth.associate_by_email',
    'seqr.utils.social_auth_pipeline.log_signed_in',
)
SOCIAL_AUTH_ASSOCIATE_USER_PIPELINE = ( 'social_core.pipeline.social_auth.associate_user',)
SOCIAL_AUTH_LOG_PIPELINE = ( 'seqr.utils.social_auth_pipeline.log_signed_in',)
...
if TERRA_API_ROOT_URL:
   ...
   SOCIAL_AUTH_PIPELINE = ('seqr.utils.social_auth_pipeline.validate_anvil_registration',) + \
                           SOCIAL_AUTH_INITIAL_PIPELINE + \
                           ('social_core.pipeline.user.get_username',
                           'social_core.pipeline.user.create_user') + \
                           SOCIAL_AUTH_ASSOCIATE_USER_PIPELINE + \
                           ('social_core.pipeline.social_auth.load_extra_data',
                           'social_core.pipeline.user.user_details') + \
                           SOCIAL_AUTH_LOG_PIPELINE
else:
   SOCIAL_AUTH_PIPELINE = SOCIAL_AUTH_INITIAL_PIPELINE + SOCIAL_AUTH_ASSOCIATE_USER_PIPELINE + SOCIAL_AUTH_LOG_PIPELINE
```

hanars (changes requested):

settings.py
ShifaSZ: My original thought is to remove the social auth records by updating the pipeline. In the end, I found there is no simple way to do that.

hanars (approved):

PR #1534, opened 2020-12-04
Adds Linkerd control plane deployment command and pod injection annotations

sjahl:


deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
```diff
@@ -14,10 +14,6 @@ spec:
       - secretName: elasticsearch-secrets
     fileRealm:
       - secretName: elasticsearch-secrets
-  http:
-    tls:
-      selfSignedCertificate:
-        disabled: true
```
hanars, line 20: how is seqr gonna have access to ES with this change. I don't see any code changes that would cause seqr to use the generated certificate?

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -245,6 +245,21 @@ def _set_elasticsearch_kubernetes_resources():
     if not has_kube_resource:
         run('kubectl apply -f deploy/kubernetes/elasticsearch/kubernetes-elasticsearch-all-in-one.yaml')
 
+
+def deploy_linkerd(settings):
+    print_separator('linkerd')
```
hanars, line 250: you should probably add this to the `DEPLOYMENT_TARGETS` config at the top of the file

deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
sjahl: Still working on this -- you can ignore this PR for now, still in draft status :)

sjahl: the short version, is that we'd need to add the CA generated by ECK as a secret for kibana, and then update the kibana configuration to use it for verification, and switch to an https url.

hanars (approved):

PR #1537, opened 2020-12-07
fix reporting for manual variants

hanars:
For manual variants there is more sparse data that causes an error during reporting. This fixes it


ShifaSZ (approved):

hanars:
passsing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/748175794

PR #1542, opened 2020-12-08
Adds a couple of WAF config updates to handle some potential false positives

sjahl:
Two actual changes, and one comment added here:

1) disables a rule which caught things like `within` and `sort` contained in seqr's URLs. The rule is meant to prevent windows command injection, so I don't think we need to worry too much about them.

2) Modify the default list of allowed Content-Types. We had many POST requests to a `/api/matchmaker/v1/match` uri, which had the content-type set to `application/vnd.ga4gh.matchmaker.v1.0+json`, and these looked like legitimate requests to me.

3) The changed comment is above the removal of rule 920350 so that we remember what that does. This OWASP rule blocks requests from localhost/private networks. Many of the requests coming through kubernetes (both actual app traffic and health checking) appear to be internal traffic, and need to be allowed. So this rule is disabled to allow the app to work.

### DIFF
--- a/deploy/kubernetes/nginx/nginx.yaml
+++ b/deploy/kubernetes/nginx/nginx.yaml
@@ -52,8 +52,19 @@ data:
     SecAuditLogType Serial
     SecAuditLog /dev/stdout
     SecAuditLogFormat JSON
+    # local requests
     SecRuleRemoveById 920350
+    # regex match for windows command injection, contained many false positives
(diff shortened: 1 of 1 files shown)


hanars (approved):

hanars:
Good catch on the matchmaker request, those are legitimate and are part of an external API we support that requires that content type

PR #1543, opened 2020-12-08
Superuser access

hanars:
This restricts 2 current staff-only views to superusers: the django admin that allows database editing, and the all users summary page

This also makes a clearer separation between the endpoint designed to return all details about all users (now restricted to superusers) and the endpoints designed to get the options for what collaborators are available to be added to a project. Note that adding collaborators to a project is only relevant in the non-anvil case, and if anvil is enabled users do not need to be added via the seqr UI and therefore shouldn't be listed as options


seqr/views/apis/project_api_tests.py
```diff
@@ -168,40 +169,40 @@ def test_empty_project_page_data(self):
...
+    'isActive': True, 'isStaff': False, 'isSuperuser': False, 'lastName': '', 'lastLogin': '',
+    'username': 'test_user_pure_anvil@test.com'}] + deepcopy(BASE_COLLABORATORS)
+for collab in ANVIL_COLLABORATORS:
+    collab['isAnvil'] = True
+
+LOCAL_COLLAB = {
+    'dateJoined': '2017-03-12T23:09:54.180Z', 'displayName': 'Test seqr local User', 'email': 'test_local_user@test.com',
+    'firstName': 'Test seqr local User', 'hasEditPermissions': False, 'hasViewPermissions': True, 'id': 14,
+    'isActive': True, 'isAnvil': False, 'isSuperuser': False, 'isStaff': False, 'lastLogin': None, 'lastName': '',
+    'username': 'test_local_user'}
```
ShifaSZ, line 194: Cool!

seqr/views/apis/users_api.py
```diff
@@ -28,20 +28,25 @@ def __init__(self, error, status_code=400, existing_user=None):
...
     else:
-        collaborators = {}
+        collaborators = set()
         for project in get_projects_user_can_view(request.user):
-            collaborators.update(get_project_collaborators_by_username(request.user, project, include_permissions=False))
+            collaborators.update(project.get_collaborators())
 
-    return create_json_response(collaborators)
+    return create_json_response({
+        user.username: _get_json_for_user(user, fields=USER_OPTION_FIELDS) for user in collaborators
```
ShifaSZ, line 40: Since only local users will be gotten, it is better to add parameter `is_anvil=False` to the function call of `_get_json_for_user()`. It can avoid unnecessary accesses to the Terra API.

seqr/views/apis/users_api.py
```diff
@@ -28,20 +28,25 @@ def __init__(self, error, status_code=400, existing_user=None):
...
+    })
 
 
 @login_required(login_url=API_LOGIN_REQUIRED_URL)
-def get_all_staff(request):
-    staff_analysts = {staff.username: _get_json_for_user(staff) for staff in User.objects.filter(is_staff=True)}
+def get_all_staff_options(request):
+    staff_analysts = {
+        staff.username: _get_json_for_user(staff, fields=USER_OPTION_FIELDS)
```
ShifaSZ, line 47: Adding `is_anvil=False` can save Terra API accesses.

seqr/views/apis/users_api_tests.py
```diff
@@ -309,93 +313,88 @@ def assert_has_anvil_calls(self):
...
         assert_has_anvil_calls(self)
 
-    def test_get_all_staff(self):
-        super(AnvilUsersAPITest, self).test_get_all_staff()
-        self.mock_list_workspaces.asssert_not_called()
-        self.mock_get_ws_acl.asssert_not_called()
+    def test_get_all_staff_options(self):
+        super(AnvilUsersAPITest, self).test_get_all_staff_options()
+        self.mock_list_workspaces.assert_not_called()
+        self.mock_get_ws_acl.assert_not_called()
```
ShifaSZ, line 333: Very bad mistakes I made. It is weird that `mock` can't catch the errors.

ShifaSZ (changes requested):

seqr/views/apis/superuser_api.py
```diff
@@ -0,0 +1,17 @@
...
+from seqr.views.utils.json_utils import create_json_response
+from seqr.views.utils.orm_to_json_utils import _get_json_for_user
+from settings import API_LOGIN_REQUIRED_URL
+
+logger = logging.getLogger(__name__)
+
+
+@user_passes_test(lambda u: u.is_active and u.is_superuser, login_url=API_LOGIN_REQUIRED_URL)
+def get_all_users(request):
+    users = [_get_json_for_user(user, is_anvil=False) for user in User.objects.exclude(email='')]
```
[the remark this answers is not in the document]
hanars: thats good in this case actually! But good to know

seqr/views/apis/users_api.py
hanars: since `USER_OPTION_FIELDS` does not include `is_anvil`, anvil isn't checked

seqr/views/apis/users_api.py
hanars: since `USER_OPTION_FIELDS` does not include `is_anvil`, anvil isn't checked

seqr/views/apis/users_api_tests.py
hanars: yeah I mean the point of mocks is that you can call any method from them and they don't error so they can be mocking anything, but I wish they had some built in edge cases for common typos, since this is pretty problematic

seqr/views/apis/users_api.py
ShifaSZ: okay

ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/748591519

PR #1544, opened 2020-12-08
properly escape quotes in the modesc config

sjahl:


sjahl:
Discovered after deploying that the modsecurity plugin did not like newlines, and needed single quotes escaped. This config is now on dev.

hanars (approved):

PR #1546, opened 2020-12-08
Remove staff priveleges

hanars:
This removes some privileges from staff users that we aren't going to support going forward. It removes the ability to edit and delete notes created by other users, as this should never be done in the course of normal application usage. It also removes the ability to see private gene lists created by other people


ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/748423527

PR #1548, opened 2020-12-09
Refactor _init_cluster_gcloud for managing node pools

sjahl:
Still need to fill out/update values in the prod yaml settings files, but wanted to get some quick feedback on the overall approach.

I thought the previous logic of creating node pools with a specific number of nodes, until a cluster-wide number of nodes was reached was a little confusing. The change here refactors that while loop, into something that iterates over a dictionary of node pools that are declared in the yaml settings (empty by default, and overridden in the dev/prod settings files)


hanars:
I like the overall approach here, I think its much clearer what node pools are being created. The only thing I can think of is it might be nice if for the node pool setting they could fall back to the defaults when not specified (i.e. `pool_settings['machine_type']` becomes `pool_settings.get('machine_type', DEFAULT_POOL_MACHINE_TYPE)`, since I think we will actually be using the same machine type for everything in prod right now) but if you think that makes it less clear I don't feel strongly

sjahl:
That's a great idea, I'll implement that. machine_type (and arguably num_nodes, too) is something we should fall back on, rather than try to initialize a pool without it.

sjahl:
I think we're at a point where this will deploy the node pool we want for elasticsearch. For now, I've left the elasticsearch master nodes nodeType selector as-is, so they should deploy into the default-pool. We can add another pool for them if you think that's useful. I've also updated the deployment targets to deploy elasticsearch into the seqr cluster.

I've left the separate elasticsearch cluster deployments intact. I think the cleanest approach will be to do another PR later that removes the references to the external connector and GKE cluster once we're confident

deploy/kubernetes/gcloud-dev-settings.yaml
```diff
@@ -2,13 +2,27 @@
 ####### SETTINGS THAT MAY NEED TO BE MODIFIED #######
 
 CLUSTER_NAME: 'seqr-cluster-dev'
-CLUSTER_MACHINE_TYPE: 'n1-highmem-4'   # n1-standard-2, n1-standard-4 .. n1-standard-64,  n1-highmem-2, n1-highcpu-2 ..
-CLUSTER_NUM_NODES: 1
+DEFAULT_POOL_MACHINE_TYPE: 'n1-highmem-4'   # n1-standard-2, n1-standard-4 .. n1-standard-64,  n1-highmem-2, n1-highcpu-2 ..
+DEFAULT_POOL_NUM_NODES: 2
+
+# TODO decide if we need multiple pools, or any pools at all in dev
```
hanars, line 8: long term we probably don't need multiple pools on dev

hanars:
> I think we're at a point where this will deploy the node pool we want for elasticsearch. For now, I've left the elasticsearch master nodes nodeType selector as-is, so they should deploy into the default-pool. We can add another pool for them if you think that's useful. I've also updated the deployment targets to deploy elasticsearch into the seqr cluster.

I think this configuration looks good! 

> I've left the separate elasticsearch cluster deployments intact. I think the cleanest approach will be to do another PR later that removes the references to the external connector and GKE cluster once we're confident

Agreed

deploy/kubernetes/postgres/postgres.gcloud.yaml
```diff
@@ -68,3 +68,5 @@ spec:
           mountPath: /var/lib/postgresql/data
       restartPolicy: Always
       dnsPolicy: ClusterFirst
+      nodeSelector:
```
hanars, line 71: we should also add this to kibana.gcloud.yaml

hanars (changes requested):

hanars:
this all looks good to me. We will need to do one manual step because the create cluster command will fail with "already created" which is what we expect, but it won't get updated to have a second node pool. I think  adding a second node to the default node pool by manually running one command is totally fine, but just wanted to call out that we will need to do it

sjahl:
I think this is good to go now? I ran this against dev on Friday, and it spun up an elasticsearch data pool. I think next step is deploying elasticsearch on the seqr dev cluster

hanars (approved):

seqr/utils/elasticsearch/utils.py
```diff
@@ -29,6 +30,11 @@ def get_es_client(timeout=60, **kwargs):
     }
     if ELASTICSEARCH_CREDENTIALS:
         client_kwargs['http_auth'] = ELASTICSEARCH_CREDENTIALS
+    if ELASTICSEARCH_PROTOCOL:
+        client_kwargs['scheme'] = ELASTICSEARCH_PROTOCOL
+    if ELASTICSEARCH_CA_PATH:
+        ssl_context = create_default_context(cafile=ELASTICSEARCH_CA_PATH)
```
hanars, line 36: how expensive an operation is this? search is already on the slower side, so if this is even a bit slow it might be instead worth running this once in settings.py when the app starts up

sjahl: I didn't notice it taking any perceptible amount of time when I was figuring out how to use it. I don't know that we really have the instrumentation in our monitoring to make a _real_ determination of whether it's adding slowness.

Looking through the docs, the function is basically establishing a group of security settings (i.e. 'how do i verify the server's certificate and what is my minimum tls version?') to apply to a connection. Given that all of our connections serve the same purpose of connecting to elasticsearch, it seems reasonable to re-use that context a little more generally, especially if we're worried about performance of the search itself.

I'll try moving it and see if things still work as expected.

hanars (approved):

PR #1554, opened 2020-12-10
grab all containers in a pod when following logs

sjahl:
With linkerd, every pod gets a second container that proxies and encrypts all its network traffic. When you have multiple containers in a pod, the `kubectl logs` command needs either a container name (in addition to the pod name), or the `-all-containers` flag, which follows logs from everything in the pod.

This PR adds that flag to our print_log method, since it seemed easier than maintaining a pod -> specific container mapping for all our services.

Also included in this fix are a couple of things that I noticed while debugging kibana and elasticsearch connections.


deploy/kubernetes/kibana/kibana.gcloud.yaml
```diff
@@ -26,6 +26,8 @@ spec:
       labels:
         name: kibana
         deployment: {{ DEPLOY_TO }}
+    spec:
+      automountServiceAccountToken: true
```
hanars, line 30: what does this do?

sjahl: This was documented here: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-service-mesh-linkerd.html#k8s-service-mesh-linkerd-kibana-apm

linkerd refused to mesh the kibana service until I added that. I'm not entirely sure if that has any effect if ECK is off in another cluster, but worth taking another look at it as we move elasticsearch back into the seqr cluster.

hanars (approved):

PR #1556, opened 2020-12-11
Use diifferent secrets for different envs

hanars:
This allows us to use different gcloud client ids for different environments. This doesn't change the command you need to run to deploy secrets

### DIFF
--- a/deploy/servctl_utils/deploy_command_utils.py
+++ b/deploy/servctl_utils/deploy_command_utils.py
@@ -75,7 +75,7 @@ SECRETS = {
     'nginx': ['{deploy_to}/tls.key', '{deploy_to}/tls.crt'],
     'seqr': [
         'omim_key', 'postmark_server_token', 'slack_token', 'airtable_key', 'django_key', 'seqr_es_password',
-        'google_client_id',  'google_client_secret'
+        '{deploy_to}/google_client_id',  '{deploy_to}/google_client_secret'
     ],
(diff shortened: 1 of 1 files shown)


deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -75,7 +75,7 @@
     'nginx': ['{deploy_to}/tls.key', '{deploy_to}/tls.crt'],
     'seqr': [
         'omim_key', 'postmark_server_token', 'slack_token', 'airtable_key', 'django_key', 'seqr_es_password',
-        'google_client_id',  'google_client_secret'
+        '{deploy_to}/google_client_id',  '{deploy_to}/google_client_secret'
```
ShifaSZ, line 78: Does the key path in `deploy/kubernetes/seqr/seqr.gcloud.yaml` need to be updated accordinarily?

ShifaSZ (changes requested):

deploy/servctl_utils/deploy_command_utils.py
hanars: nope! This isn't changing where the yaml file is supposed to be, its only changing the secrets, which aren't in source control. If you restore the latest serets backup, yo'll see that I added nested folders for these, with different keysin the dev and prod folders

ShifaSZ (approved):

PR #1558, opened 2020-12-11
enable the WAF

sjahl:
This changes the WAF from detect-only to enforcing. Requests that match one of the OWASP filters will receive a 403.


hanars:
how confident are we that all real traffic is able to get through? Do we need to leave this open for a few more days, or do you think this is good to go now?

sjahl:
I'm fairly confident -- everything that I'm seeing getting flagged looks like things we want to block. I can deploy this to dev, and then we should run through and exercise the critical workflows in the app, before putting it on prod.

hanars (approved):

PR #1562, opened 2020-12-15
adds a whitelist rule for requests to /complete/google

sjahl:
Requests for completing google authentication were getting caught by the WAF because they contained the string `.profile` in the request uri (`.profile` is a common file that gets looked for when scanning for inadvertently exposed secrets).

I believe this whitelist will turn off that rule when the request uri begins with `/complete/google`.


sjahl:
Deployed on dev, and verified that logging in with google was not possible before this change (403 Forbidden), and that I was able to log in successfully after adding the whitelist.

hanars (approved):

PR #1564, opened 2020-12-16
fix update family case review fields

hanars:
The recent updates to the case review pages changed permissioning such that case review fields were supposed to be updated via field-specific URLs, but I did not actually update the family fields to use these URLs. This makes that change and fixes those endpoints to work with the correct post body format


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/750057927

ShifaSZ (approved):
No issue found.

PR #1566, opened 2020-12-17
better error handling for duplicate saved search

hanars:


seqr/views/apis/variant_search_api_tests.py
```diff
@@ -483,6 +484,14 @@ def test_saved_search(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.json()['savedSearchesByGuid']), 4)
 
+        # Test cannot save different searches with the same name
+        body['filters'] = {'test': 'filter'}
+        with transaction.atomic():
```
hanars, line 489: this is required because otherwise the rest of the test crashes after an IntegrityError (even if it is properly handled)

ShifaSZ: Hope it won't crash after an IntegrityError in run-time.

ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/750299468

PR #1567, opened 2020-12-18
do not regenerate postgres password

hanars:
It turns out this really brings down seqr when we deploy secrets, as the docker image uses the secret when it initializes the db but does not keep it synced. We will manage credential rotation by doing a manual process once a year, until we can move to a google managed postgres instance. I will update the secrets backups with the appropriate new file


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/750299468

sjahl (approved):

PR #1568, opened 2020-12-18
Added setting to force social auth to use HTTPS for redirecting URL.

ShifaSZ:
Changed the social auth setting to force using HTTPS for the redirecting URL after authentication. It is necessary when the server is behind an HTTPS proxy.


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/750434604

hanars (approved):

PR #1570, opened 2020-12-21
remove owner permissions

hanars:
We have an owner level user group for projects but we don't use it anywhere and we don't need/ want it so this removes it. It also removes the object level permissions from the LocusList model,, since those aren't used at all either


ShifaSZ (approved):
All look good.

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/750884359

PR #1576, opened 2020-12-23
Servctl env cleanup

hanars:
Removes references to the now-deprecated elasticsearch deployment targets and cleans up the code since we no longer need to support different types of clusters. Also cleans up settings to make it a bit cleaerer what differs between dev and prod


sjahl (approved):
This all looks good to me -- much cleaner!

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/751238341

PR #1578, opened 2020-12-23
remove unnecessary deploy_to_prefix variable

hanars:
When the servctl scripts were written the idea was they could be used to deploy to gcloud for the broad deployment, or with minkub for local deployment. We have since removed all the minikub deployment code, but left in the deploy_to_prefix variable that toggled between these modes. This cleans up the code by removing it


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/751265577

sjahl (approved):
This looks good to me!

deploy/kubernetes/seqr/seqr.gcloud.yaml
```diff
@@ -154,14 +154,8 @@ spec:
...
         - name: KIBANA_SERVICE_PORT
           value: "{{ KIBANA_SERVICE_PORT }}"
-        - name: DEPLOY_TO
-          value: "{{ DEPLOY_TO }}"
-        - name: DEPLOY_TO_PREFIX
-          value: "{{ DEPLOY_TO_PREFIX }}"
         - name: DEPLOYMENT_TYPE
           value: "{{ DEPLOYMENT_TYPE }}"
-        - name: CLUSTER_NAME
-          value: "{{ CLUSTER_NAME }}"
```
ShifaSZ, line 164: I'm not sure the reason for delete `DEPLOY_TO` and `CLUSTER_NAME`.

sjahl: my understanding here was that the seqr pod itself doesn't actually need to know those values in the environment at runtime. but I could be wrong...

hanars: they are never used

PR #1579, opened 2021-01-04
Slack outage handling

hanars:
We should fail gracefully and continue normal seqr operation when slack is down


ShifaSZ (approved):
All look good.

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/752835107

PR #1584, opened 2021-01-05
Redis cache for Anvil API

ShifaSZ:
Added creating a cache entry in Redis when access AnVIL/Terra API successfully. The cache entry will be expired in a configurable time.


seqr/views/utils/terra_api_utils.py
```diff
@@ -67,6 +69,12 @@ def anvil_call(method, path, access_token, user=None, headers=None, root_url=Non
     url, headers = _get_call_args(path, headers, root_url)
     request_func = getattr(requests, method)
     headers.update({'Authorization': 'Bearer {}'.format(access_token)})
+    cache_key = 'terra_req__' + hashlib.md5((url + str(headers) + str(kwargs)).encode('utf-8')).hexdigest()
+    r_text = safe_redis_get_json(cache_key)
+    if r_text:
+        logger.info('Debug: Cache {} hit. {}'.format(cache_key, r_text))
+        return json.loads(r_text)
```
hanars, line 76: `safe_redis_get_json` should return it as loaded json so this should be unnecessary/ shouldn't actually work

seqr/views/utils/terra_api_utils.py
```diff
@@ -67,6 +69,12 @@ def anvil_call(method, path, access_token, user=None, headers=None, root_url=Non
     url, headers = _get_call_args(path, headers, root_url)
     request_func = getattr(requests, method)
     headers.update({'Authorization': 'Bearer {}'.format(access_token)})
+    cache_key = 'terra_req__' + hashlib.md5((url + str(headers) + str(kwargs)).encode('utf-8')).hexdigest()
```
hanars, line 72: This is not a user specific cache, which means this will not do what we need it to. For example, if you list workspaces a user has access to, and then 3 seconds later a different user makes that call, they will get back the previous users access list which is really bad!

I also don't think we should be caching all anvil requests, just the ones that get done really frequently as part of authorization. I would move this caching up a level into the function we actually want cached, and make sure the username is in the cache key

seqr/views/utils/terra_api_utils.py
```diff
@@ -67,6 +69,12 @@ def anvil_call(method, path, access_token, user=None, headers=None, root_url=Non
...
+    r_text = safe_redis_get_json(cache_key)
+    if r_text:
+        logger.info('Debug: Cache {} hit. {}'.format(cache_key, r_text))
```
hanars, line 75: I don't think we should be using the info logger to log debug statements. Either this should be a logger.debug log that isn't actually logged in prod or we should take out the word "debug" from the message

settings.py
```diff
@@ -337,11 +337,15 @@
 SOCIAL_AUTH_PIPELINE_ASSOCIATE_USER = ('social_core.pipeline.social_auth.associate_user',)
 SOCIAL_AUTH_PIPELINE_LOG = ('seqr.utils.social_auth_pipeline.log_signed_in',)
 
+TERRA_API_CACHE_EXPIRE_SECONDS = os.environ.get('TERRA_API_CACHE_EXPIRE_SECONDS', 60)
```
hanars, line 340: can you update `deploy/kubernetes/shared-settings.yaml` with a reasonable value for this (its probably good if its also 60), and update the env section of `deploy/kubernetes/seqr/seqr.gcloud.yaml` to include this variable:
```
 - name: TERRA_API_CACHE_EXPIRE_SECONDS
    value: "{{ TERRA_API_CACHE_EXPIRE_SECONDS }}"
```

hanars (changes requested):
Once you've made the changes requested, you should update the unit tests to check that the cache is working as expected.

seqr/views/utils/terra_api_utils.py
hanars: Also worth noting that codacy is flagging this line for "Use of insecure MD2, MD4, MD5, or SHA1 hash function."
By moving it up a level so you know what is being called, you can make a simpler cache key without needing to encode the whole request. For example, in `user_get_workspace_access_level` you could make the cache key something like `'terra_req__{}__{}'.format(path, user.username)`

seqr/views/utils/terra_api_utils.py
ShifaSZ: Really?! I should add a test to cover it. In the real code scenarios that I tried, no caches were used.

seqr/views/utils/terra_api_utils.py
ShifaSZ: I thought it has no harm to allow other users to get permission information for a project. But, I know it is wrong at least the workspace list must be per-user based.

ShifaSZ: I did a test for different users to send the same request but they don't hit the same cache for they have different headers which include the bearer tokens. But your suggestion is better because it is neater.

seqr/views/utils/terra_api_utils.py
hanars: If no caches are used then why on earth would we add this code??? Adding untested code we expect to never be used is ridiculous! There are multiple tests that assert that in a given single API call we make multiple identical calls to Terra! In all of those use cases I would absolutely expect us to be using the cache, and in fact that is the entire reason to write this code

ShifaSZ: I found that there are some tests that cover it. And I found that `json.loads()` works with a `dict` parameter. That's why I didn't catch the mistake.
There are 4 API calls, one for validation of account registration on AnVIL while logging in, one for workspace list when displaying the dashboard, one for access permissions of a workspace, one for the collaborator list of a workspace. The one that could be called often is the one for access permissions.
The reason that I didn't see cache hits yesterday is that I use an account with local permissions. After I switch my account, I can see multiple cache hits when calling the API for workspace access permissions.

seqr/views/utils/terra_api_utils.py
```diff
@@ -65,17 +65,12 @@ def _get_social_access_token(user):
     return social.extra_data['access_token']
 
 
-def anvil_call(method, path, access_token, user=None, headers=None, root_url=None, **kwargs):
+def anvil_call(method, path, access_token, user=None, headers=None, root_url=None):
```
hanars, line 68: you might also want to take out the `**kwargs` argument from `_user_anvil_call`, since it looks like its not used and its a bit confusing that you can pass through arbitrary kwargs to this function that might not be allowed

seqr/views/utils/terra_api_utils.py
```diff
@@ -110,12 +113,23 @@ def list_anvil_workspaces(user, fields=None):
 
 def user_get_workspace_access_level(user, workspace_namespace, workspace_name):
     path = "api/workspaces/{0}/{1}?fields=accessLevel".format(workspace_namespace, workspace_name)
+
+    cache_key = 'terra_req__' + hashlib.md5("{}__{}".format(user, path).encode('utf-8')).hexdigest()
```
hanars, line 117: What value is there to hashing this?

seqr/views/utils/terra_api_utils.py
```diff
@@ -110,12 +113,23 @@ def list_anvil_workspaces(user, fields=None):
...
+    r_text = safe_redis_get_json(cache_key)
```
hanars, line 118: I'm a little confused b the variable name `r_text`, as that sounds like a raw text response, and this should be a parsed response dictionary right?

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -21,6 +21,15 @@
...
     fixtures = ['users', 'social_auth']
 
+    def test_is_google_authenticated(self, mock_logger):
+        user = User.objects.get(email = 'test_user@test.com')
+        r = is_google_authenticated(user)
+        self.assertTrue(r)
+        user = User.objects.get(email = 'test_local_user@test.com')
+        r = is_google_authenticated(user)
+        self.assertFalse(r)
+        self.assertEqual(len(mock_logger.method_calls), 0)
```
hanars, line 31: why test that nothing was logged? If there is a case where something should be logged, why not test it?

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -148,3 +159,10 @@ def test_user_get_workspace_access_level(self, mock_time, mock_logger):
         mock_logger.warning.assert_called_with(
             'test_user called Terra API: GET /api/workspaces/my-seqr-billing/my-seqr-workspace?fields=accessLevel got status 404 with reason: Not Found')
         self.assertEqual(len(mock_logger.method_calls), 1)
+
+        # Test cache hit
+        url = '{}api/workspaces/my-seqr-billing/my-seqr-workspace?fields=accessLevel'.format(TEST_TERRA_API_ROOT_URL)
+        mock_redis.return_value.get.return_value = '{"accessLevel": "READER"}'
+        permission = user_get_workspace_access_level(user, 'my-seqr-billing', 'my-seqr-workspace')
+        self.assertDictEqual(permission, {"accessLevel": "READER"})
+        mock_redis.return_value.get.assert_called_with('terra_req__ec4582d9e5e0247385d63538686095cc')
```
hanars, line 168: can you add checks here that no requests were made

settings.py
hanars: this still needs doing

hanars (changes requested):
Can you look into the code and see if there are any places where we expect `list_anvil_workspaces` or `user_get_workspace_acl` to be called close together? If so we should add caching

seqr/views/utils/terra_api_utils.py
ShifaSZ: okay

seqr/views/utils/terra_api_utils.py
ShifaSZ: It used an access token as part of the key. The hash can protect the token from being logged. Now, it doesn't include sensitive information in the key. So I can get rid of the hash.

seqr/views/utils/terra_api_utils.py
ShifaSZ: I will rename it to `r`.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: The `mock_logger` is not been used in this test. But since it is a class modifier, I have to add it to the parameter of this test. Adding an assertion can avoid a warning of the parameter not used.

seqr/views/utils/terra_api_utils_tests.py
ShifaSZ: okay

seqr/views/utils/terra_api_utils_tests.py
hanars: but why do you need to mock the logger at all then?

settings.py
ShifaSZ: Kubernetes files are updated now.

deploy/kubernetes/shared-settings.yaml
```diff
@@ -41,6 +41,8 @@ UPLOADED_PEDIGREE_FILE_RECIPIENTS: ''
 # this allows cluster administration only from known machines or from machines on google cloud.
 MASTER_AUTHORIZED_NETWORKS: '69.173.112.0/21,69.173.127.232/29,69.173.127.128/26,69.173.127.0/25,69.173.127.240/28,69.173.127.224/30,69.173.127.230/31,69.173.120.0/22,69.173.127.228/32,69.173.126.0/24,69.173.96.0/20,69.173.64.0/19,69.173.127.192/27,69.173.124.0/23'
 
+TERRA_API_CACHE_EXPIRE_SECONDS: 3600
```
hanars, line 44: I don't think this is an acceptable amount of time to cache things. If someone had their access revoked in Terra it seems so unacceptable to continue to access seqr for an hour. Unless you have received explicit sign off from the compliance team about using such a long cache, please make this 60

hanars (changes requested):

hanars:
> Can you look into the code and see if there are any places where we expect `list_anvil_workspaces` or `user_get_workspace_acl` to be called close together? If so we should add caching

can you look into this and comment with your decision

ShifaSZ:
> list_anvil_workspaces

ShifaSZ:
> Can you look into the code and see if there are any places where we expect `list_anvil_workspaces` or `user_get_workspace_acl` to be called close together? If so we should add caching

The `list_anvil_workspaces` can be called from `dashboard_api`, `awesomebar_api`, `user_api`, and `variant_search_api`. It will be called for a single time in each API call. These APIs are not likely to be called closely. But, it may still make sense to caching for `list_anvil_workspaces` because of the amount of data (57k bytes) return from AnVIL if we make the expiration time longer (such as up to 1 hour).

The `user_get_workspace_acl` will be called from `project_api` when opening the project page. It is a one-time operation. It shouldn't need a cache. `user_get_workspace_acl` is also called when `add/update/delete` project collaborators. Each API call will only call once. I don't see any reason to add caching for it.

hanars:
> The `list_anvil_workspaces` can be called from `dashboard_api`, `awesomebar_api`, `user_api`, and `variant_search_api`. It will be called for a single time in each API call. These APIs are not likely to be called closely. But, it may still make sense to caching for `list_anvil_workspaces` because of the amount of data (57k bytes) return from AnVIL if we make the expiration time longer (such as up to 1 hour).

The `awesomebar_api` is the typeahead search bar in seqr, which means it gets hit really frequently by back-to-back requests, as it is hit again every time a letter is typed in the search bar. I think it is actually important to add caching for this case. Using the same amount of time for the cache timeout as in the other case should be fine

> The `user_get_workspace_acl` will be called from `project_api` when opening the project page. It is a one-time operation. It shouldn't need a cache. `user_get_workspace_acl` is also called when `add/update/delete` project collaborators. Each API call will only call once. I don't see any reason to add caching for it.

Agreed, no reason to add caching

deploy/kubernetes/shared-settings.yaml
```diff
@@ -42,6 +42,8 @@ POSTGRES_USERNAME: postgres
 # this allows cluster administration only from known machines or from machines on google cloud.
 MASTER_AUTHORIZED_NETWORKS: '69.173.112.0/21,69.173.127.232/29,69.173.127.128/26,69.173.127.0/25,69.173.127.240/28,69.173.127.224/30,69.173.127.230/31,69.173.120.0/22,69.173.127.228/32,69.173.126.0/24,69.173.96.0/20,69.173.64.0/19,69.173.127.192/27,69.173.124.0/23'
 
+TERRA_PERMS_CACHE_EXPIRE_SECONDS: 60
+TERRA_WORKSPACE_CACHE_EXPIRE_SECONDS: 300
```
hanars, line 46: You need to change `seqr.gcloud.yaml` to include these variables

hanars (changes requested):

hanars (approved):

PR #1586, opened 2021-01-05
exempt postgres until we can try linkerd's opaque ports

sjahl:
This change is in response to pg_dumps failing in seqr. My understanding is that connections to the postgres database are succeeding and normal app usage is fine, but dumps are performing very slowly (or the connections are timing out and leaving the dump stuck part-way through). We'll have to test linkerd's new opaque ports setting when 2.10 is released. This may help with timeouts, but may not help with performance; we'll need further investigation later if that doesn't fix the problem.

Once the ports are exempted from the mesh, dumps succeed in a reasonable amount of time.

### DIFF
--- a/deploy/kubernetes/postgres/postgres.gcloud.yaml
+++ b/deploy/kubernetes/postgres/postgres.gcloud.yaml
@@ -26,6 +26,7 @@ spec:
     metadata:
       annotations:
         linkerd.io/inject: enabled
+        config.linkerd.io/skip-inbound-ports: "5432"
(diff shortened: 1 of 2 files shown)


hanars:
At this point, what is linkerd enabled for?

hanars (approved):

sjahl:
Until we can try out opaque ports to see if they help with any of redis, postgres, or elasticsearch, it's only encrypting traffic between seqr and kibana.

PR #1589, opened 2021-01-07
Static files

hanars:
Fixes https://github.com/broadinstitute/seqr-private/issues/930 by disabling cross origin access for static files. This is default allowed to support using CDNs for static assests, but since we don't do that theres no harm in disabling thsis behavior and its more secure this way

This also tweaks local static file configuration to allow running seqr locally with the built assets on localhost:8000


ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/753558845

PR #1599, opened 2021-01-12
All project-family search timeouts

hanars:
The all project and family search has not worked for a long time due to timeouts (https://github.com/broadinstitute/seqr-private/issues/790). This changes the behavior of the search to do a single search across all permitted indices and only save the families that are actually hits for the search (instead of all families in seqr). This type of search is used only for finding other instances of a single variant across multiple projects, so these optimizations will work fine


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/754172295

ShifaSZ (approved):
All look good.

PR #1600, opened 2021-01-12
grant create project permission to superusers

hanars:
For local install users, the default superuser needs to be able to create projects, so we are allowing superusers to create projects as well as PMs


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/754156683

hanars:
Actually we need to take a different approach for local installs and PM/ analyst permissions. Closingtthis out and will put up a different solution instead

PR #1601, opened 2021-01-12
Added login restrictions to privileged users.

ShifaSZ:
All privileged users (data managers and superusers) must log in with Google Auth if Google Auth is enabled.

### DIFF
--- a/seqr/views/apis/auth_api.py
+++ b/seqr/views/apis/auth_api.py
@@ -11,7 +11,7 @@ import logging
 
 from seqr.views.utils.json_utils import create_json_response
 from seqr.views.utils.permissions_utils import user_is_data_manager
-from settings import SOCIAL_AUTH_GOOGLE_OAUTH2_KEY
+from seqr.views.utils.terra_api_utils import google_auth_enabled
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


seqr/views/apis/auth_api.py
```diff
@@ -28,7 +29,12 @@ def login_view(request):
     if users.count() != 1:
         return create_json_response({}, status=401, reason='Invalid credentials')
 
-    u = authenticate(username=users.first().username, password=request_json['password'])
+    user = users.first()
+    if SOCIAL_AUTH_GOOGLE_OAUTH2_KEY and (user_is_data_manager(user) or user.is_superuser):
```
hanars, line 33: it might be good to make a helper method somewhere for `_is_google_auth_enabled` that you can use both here and in react_app.py

seqr/views/apis/auth_api.py
```diff
@@ -28,7 +29,12 @@ def login_view(request):
...
+        logger.warning("Privileged user {} is trying to login without Google authentication.".format(user))
+        return create_json_response({}, status=401, reason='Privileged user must login with Google authentication.')
```
hanars, line 35: so for some reason, the HTTP reason doesn't parse well in the JS in production, so its best to also include a json body with the key `error`. So here that would be 
```
error = 'Privileged user must login with Google authentication.'
return create_json_response({'error': error}, status=401, reason=error)
```

seqr/views/apis/auth_api_tests.py
```diff
@@ -68,6 +69,48 @@ def test_login_view(self):
...
+        response = self.client.post(url, content_type='application/json',
+                                    data=json.dumps(req_values))
+        self.assertEqual(response.status_code, 401)
+        self.assertEqual(response.reason_phrase, 'Invalid credentials')
+
+    @mock.patch('seqr.views.apis.auth_api.SOCIAL_AUTH_GOOGLE_OAUTH2_KEY', 'test_key')
+    def test_login_view_with_google(self):
+        url = reverse(login_view)
+
+        # send login request with a privileged account (data manage or superuser) while Google auth is enabled
```
hanars, line 95: can you also add a test here to ensure that non-priveleged users CAN successfully log in when google auth is enabled? Testing any non-proveleged user is fine, you only need one

hanars (changes requested):

seqr/views/apis/auth_api.py
ShifaSZ: Okay

hanars (approved):

PR #1604, opened 2021-01-13
Local pm and analyst permissions

hanars:
Changes the way PM and analyst permissions are handles in local installs. For local installations, where the user groups are not defined, PM permissions are made available to superusers and analyst permissions are not made available to anyone. This makes no changes tot our current production seqr, which will continue to use groups for PMs and analysts


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/754349285

ShifaSZ (approved):
Awesome work.

PR #1613, opened 2021-01-13
Add AppSec Github Trivy Action

dinvlad:
Hi Team,

As discussed with @hanars, the DSP AppSec team would like to add our Docker image security scanner (Trivy) as a GitHub Action that will run on every PR. It will detect _Critical_ OS-level vulnerabilities (if any) in the image, and report it to you. You can then either:
* fix the issue (usually just by updating the packages in the Dockerfile), or
* ignore it, and document the reason with AppSec. Our NIST 800-53 compliance process has a timeline of 14 days for patching Critical vulnerabilities, or documenting why they are ignored.

Generally, after we sort out the results from the initial scan (if any), it should not return many positives, so this is a low-noise addition to your workflow. Generally (perhaps once a quarter) it can return 1-2 results, and it should be easy to fix those by updating the underlying base image and/or packages. The scanner only detects OS packages like Debian apt packages, and not Python packages (which are detected by Snyk and another external tool that we use), and we only report those that have a known fix.

As an alternative, you can adopt one of our recommended base images (https://github.com/broadinstitute/dsp-appsec-blessed-images), which are regularly scanned and patched for vulnerabilities. In that case, this GitHub Action will skip the scan so you won't have to worry about any findings at all.

Thanks!


dinvlad:
I can see that the image has been built successfully and is "clean" (woo-hoo!), but it currently takes ~38 minutes to build. If this is a problem, we can make it run every 2 weeks on your `master` branch instead - please let me know.

hanars:
> I can see that the image has been built successfully and is "clean" (woo-hoo!), but it currently takes ~38 minutes to build. If this is a problem, we can make it run every 2 weeks on your `master` branch instead - please let me know.

Really glad our image is clean! But I do think that build time will be a problem, especially given our conversation about how infrequently our image is rebuilt. I think switching this to running on master every 2 weeks would be a better approach for us

dinvlad:
Sounds good, I'll tweak it!

hanars:
Hi @dinvlad has this been updated to run against master na schedule instead of on PR?

dinvlad:
@hanars no not yet, I'm still working on it. I'll mention you here once ready! 👍

dinvlad:
@hanars ready for review!

hanars (approved):
awesome, thanks!

PR #1618, opened 2021-01-14
Airtable auth hardening

hanars:
In order to access airtable data confirm users have logged in through google and have Broad email accounts

### DIFF
--- a/seqr/views/utils/terra_api_utils_tests.py
+++ b/seqr/views/utils/terra_api_utils_tests.py
@@ -6,7 +6,8 @@ from django.contrib.auth.models import User
 
 from seqr.views.utils.test_utils import TEST_TERRA_API_ROOT_URL
 from seqr.views.utils.terra_api_utils import list_anvil_workspaces, user_get_workspace_acl,\
-    anvil_call, user_get_workspace_access_level, TerraNotFoundException, TerraAPIException, is_anvil_authenticated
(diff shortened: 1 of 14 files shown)


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/754529932

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -21,14 +22,38 @@
...
+        r = is_google_authenticated(local_user)
+        self.assertFalse(r)
+
+        mock_social_auth_key.__bool__.return_value = False
         r = is_google_authenticated(user)
         self.assertFalse(r)
 
+    @mock.patch('seqr.views.utils.terra_api_utils.anvil_enabled')
+    @mock.patch('seqr.views.utils.terra_api_utils.is_google_authenticated')
+    def test_is_anvil_authenticated(self, mock_is_google_authenticated, mock_anvil_enabled):
```
ShifaSZ, line 41: Should it also add a test for a local logged in user like `test_is_google_authenticated`?

ShifaSZ (changes requested):

seqr/views/utils/terra_api_utils_tests.py
hanars: no because `is_google_authenticated` is mocked and thats is where the local user check occurs. That case is tested explicitly in the test for that function, and that function is mocked and not tested here

ShifaSZ (approved):
Awesome work.

PR #1622, opened 2021-01-15
move media images to repo

hanars:
Moves static images referenced in seqr into the static files. They are currently on a mounted disk on the seqr pod, but this means they aren't available for local installs.


ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/754690248

PR #1625, opened 2021-01-19
make new folder for static assets

hanars:
Using the `ui/dist` folder as our static root folder does not work with keeping static images in the repo, because when we rebuild the assets it deletes everything in that folder including the images. This makes a new `static` folder for our static assets, and updates the settings so the `collectstatic` command can find `ui/dist`. The process for building assets does not change as a result of this PR

The only changes that needs reviewing is the changes to settings.py, everything else is just moving the assets to a new folder and rebuilding the js


settings.py
```diff
@@ -235,7 +236,7 @@
         'http://localhost:3000',
         'http://localhost:8000',
     )
-    STATICFILES_DIRS = [STATIC_ROOT]
+    STATICFILES_DIRS.append(STATIC_ROOT)
```
ShifaSZ, line 239: It seems that the first element of `STATICFILES_DIRS` (which is 'ui/dist') doesn't have a full path while the second one (which is `STATIC_ROOT`) has. Is that correct?

ShifaSZ (changes requested):
There is a question that needs to be answered.

settings.py
hanars: yes, it works fine that way

ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/755230254

PR #1628, opened 2021-01-21
Depenedency updates

hanars:
Bumps some of our dependencies that synk has flagged as having vulnerabilities. For the javascript packages, we do not need to update the version for any of our direct depndencies so there are no changes to package.json, but this updates the locked versions of several vulnerable downstream dependencies to their latest possible versions


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/755455892

ShifaSZ (approved):
Looks good.

PR #1630, opened 2021-01-22
Added frontend page for loading data from workspace.

ShifaSZ:
This is a draft PR for review.

The layout of the page looks like the picture below.
image


hanars:
Some formatting things:

- Add a title to the top of the form. It should not be the standard seqr page header, just a header tag on the page stating something like "Load AnVIL data to seqr". Ideally, include the name of the workspace in here
- put a Segment with a border around the content for the file upload to make it clearer how the content is grouped
- Move the comment about "The Individual IDs need to math the VCF IDs" into the description for the Individual ID column
- Add a checkbox that has the text "By submitting this form I agree to grant seqr access to the data in the associated workspace"and then the form should fail validation if the user does not select this

ui/pages/Project/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,41 @@
...
+import { connect } from 'react-redux'
+
+import ReduxFormWrapper, { configuredField } from 'shared/components/form/ReduxFormWrapper'
+import { BaseBulkContent } from './edit-families-and-individuals/BulkEditForm'
+import { getEntityExportConfig } from '../selectors'
+
+const FILE_FIELD_NAME = 'uploadedFile'
+
+const mapStateToProps = (state, ownProps) => ({
+  blankExportConfig: getEntityExportConfig({ name: ownProps.workspace }, [], null, 'template', ownProps.requiredFields.concat(ownProps.optionalFields)),
```
hanars, line 12: so if you dive into the code, `getEntityExportConfig` return a dictionary as follows:
```
 getEntityExportConfig = (project, rawData, tableName, fileName, fields) => ({
  filename: `${project.name.replace(' ', '_').toLowerCase()}_${tableName ? `${toSnakecase(tableName)}_` : ''}${fileName}`,
  rawData,
  headers: fields.map(config => config.header),
  processRow: family => fields.map((config) => {
    const val = family[config.field]
    return config.format ? config.format(val) : val
  }),
})
```
looking at this code, you see that its essentially all about how to get and process/format the project data. Since this is for a blank template, and its fine to use a basic hardcoded filename since this is empty data,  we don't need this to be a connected component that calls this selector at all. We should instead use the following constant:
```
const BLANK_EXPORT = {
  filename: 'individuals_template',
  rawData: [],
  headers: [... INDIVIDUAL_ID_EXPORT_DATA, ... INDIVIDUAL_CORE_EXPORT_DATA].map(config => config.header), // or whatever our field constants are
  processRow: val => val,
}
```

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -22,6 +22,7 @@ const Container = styled.div`
   textAlign: left;
   width: 100%;
   padding: 5px 15px 5px 35px;
+  border-bottom: solid gray 1px;
```
hanars, line 25: I don't think we want to change this everywhere

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -43,9 +44,13 @@ const TableCell = styled(Table.Cell)`
   vertical-align: top;
 `
 
+const TitleTableCell = styled(TableCell)`
```
hanars, line 47: I am concerned that you are unconditonally changing the formatting for this everywhere this component is used. Why are these changes neccesssary?

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -62,21 +67,21 @@ const FAM_UPLOAD_FORMATS = [].concat(BASE_UPLOAD_FORMATS)
...
-              <TableCell>
+              <TitleTableCell>
                 <BoldText>{title}</BoldText> ({formatLinks ? formatLinks.map(
                   ({ href, linkExt }, i) => <span key={linkExt}>{i > 0 && ' / '}<a href={href} target="_blank">.{linkExt}</a></span>)
                   : `.${ext}`})
-              </TableCell>
+              </TitleTableCell>
               <TableCell>
-                {ext &&
+                {ext && project ?
```
hanars, line 84: Instead of changing this out with a ternary based on project, just makes this conditional on whether or not `exportConfig` is defined

```
{ext &&
<span>
    download &nbsp;
    {blankExportConfig &&
    <span>
      template: <FileLink data={blankExportConfig} ext={ext} linkContent="blank" /> {exportConfig &&or&nbsp;}
    </span>
    }
   {exportConfig &&
    <FileLink data={exportConfig} ext={ext} linkContent="current individuals" />
   }
</span>
}
```

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -112,21 +124,21 @@ const BaseBulkContent = React.memo(({ actionDescription, details, project, name,
...
                 <TableCell>{field.description}</TableCell>
               </TableRow>,
             )}
           </Table.Body>
         </StyledTable>
       </div>
-      {details && <div><br />{details}</div>}
-      <br />
     </Container>
+    {details && <div><br />{details}</div>}
```
hanars, line 135: As I mentioned, the detail you added belongs in a field description so you should take this out

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -112,21 +124,21 @@ const BaseBulkContent = React.memo(({ actionDescription, details, project, name,
...
     <br />
     <FileUploadField
       clearTimeOut={0}
       dropzoneLabel="Click here to upload a table, or drag-drop it into this box"
-      url={`/api/project/${project.projectGuid}/upload_${name}_table`}
+      url={project ? `/api/project/${project.projectGuid}/upload_${name}_table`
```
hanars, line 140: I would rather have this as an explicitly passed optional url prop rather than hardcoding this somewhat unrelated url in here. So ```url={url || `/api/project/${project.projectGuid}/upload_${name}_table`}```

ui/shared/utils/constants.js
```diff
@@ -291,8 +291,8 @@ export const INDIVIDUAL_FIELD_NOTES = 'notes'
 export const INDIVIDUAL_FIELD_PROBAND_RELATIONSHIP = 'probandRelationship'
 
 export const INDIVIDUAL_FIELD_CONFIGS = {
-  [FAMILY_FIELD_ID]: { label: 'Family ID' },
-  [INDIVIDUAL_FIELD_ID]: { label: 'Individual ID' },
+  [FAMILY_FIELD_ID]: { label: 'Family ID', description: 'Family ID of the individual' },
+  [INDIVIDUAL_FIELD_ID]: { label: 'Individual ID', description: 'ID of the Individual' },
```
hanars, line 295: I do not want this changing on the other pages. You need to create custom constants to pass to the form, not change the existing constants. So for example in the `LoadWorkspaceDataForm` you might define something like

```
const FIELD_DESCRIPTIONS = {
  [FAMILY_FIELD_ID]:  'Family ID of the individual',
  [INDIVIDUAL_FIELD_ID]: 'ID of the Individual', //change to include that should match the VCF ID
}
const REQUIRED_FIELDS = INDIVIDUAL_ID_EXPORT_DATA.map(config => (
  {  ...config, description: FIELD_DESCRIPTIONS[config.field] })
```

hanars (changes requested):
This is a really awesome start!

seqr/urls.py
```diff
@@ -269,6 +271,7 @@
     'matchmaker/v1/match': external_api.mme_match_proxy,
     'matchmaker/v1/metrics': external_api.mme_metrics_proxy,
 
+    'anvil_workspace/(?P<workspace_name>[^/]+)': anvil_workspace_page,
```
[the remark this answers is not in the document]
ShifaSZ: Can I add the endpoint to the `urlpatterns` like the `login` and `logout` endpoints? Users have to provide both `name` and `namespace` for the workspace. We have to options to do that 1) use the Django URL parameter like `^anvil_workspace/(?P<namespace>[^/]+)/(?P<name>[^/]+)`; 2) use queries in the URL such as `anvil_workspace/?namespace=user_provided&name=user_provided`. Which one do you prefer?

ui/pages/Project/components/LoadWorkspaceDataForm.jsx
ShifaSZ: Sounds great.

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
ShifaSZ: Sounds good.

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
ShifaSZ: Cool

seqr/urls.py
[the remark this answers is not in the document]
hanars: > Can I add the endpoint to the `urlpatterns` like the `login` and `logout` endpoints? 

Yes, although it would be more straightforward to add it to the initial `urlpatterns` list like the `status_view`

> Users have to provide both `name` and `namespace` for the workspace. We have to options to do that 1) use the Django URL parameter like `^anvil_workspace/(?P<namespace>[^/]+)/(?P<name>[^/]+)`; 2) use queries in the URL such as `anvil_workspace/?namespace=user_provided&name=user_provided`. Which one do you prefer?

I prefer option 1, as I think it is more explicit and requires less validation (no need to check that all the required query params are there). You should also then make sure the `/create_project_from_workspace` page also has the same url pattern with both the namespace and name in the url

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
ShifaSZ: Basically, it adds styling to the table cell width. Without the styling, the table looks weird as below.
The bulk-edit-individual modal:
image
The new create-project-from-anvil form:
image

hanars: Oh I see, and yeah that is something we would want to fix everywhere. I think a better way of fixing it, though, is to instead of having 3 different Table rendered here, we could use one big table, and then the columns would all be aligned properly on every row. You can use the `colSpan` property on a cell to make it use multiple columns, so you can do that for the description text rows that can use the full width

ShifaSZ: Gotcha, but I maybe do it a little later.

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -24,6 +24,11 @@ const Container = styled.div`
   padding: 5px 15px 5px 35px;
 `
 
+const StyledSpacer = styled.div`
+  padding: 5px 15px 5px 35px;
+  border-top: solid gray 1px;
+`
+
```
ShifaSZ, line 31: Ignore this for now. To be updated.

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
hanars: I actually ended up needing to do this for an unrelated PR that updates the bulk upload for IGV, so if you want an example you can see that here

https://github.com/broadinstitute/seqr/commit/0baa4a95a3fa4e6c9f80f2cb322b901b6aeaa91a#diff-b7544cdb2ed88c5c4972f5e29449806f151c89bafbfe26c3f3b0852dea4f9ad8R93

ShifaSZ:
> Some formatting things:
> 
> * Add a title to the top of the form. It should not be the standard seqr page header, just a header tag on the page stating something like "Load AnVIL data to seqr". Ideally, include the name of the workspace in here
> * put a Segment with a border around the content for the file upload to make it clearer how the content is grouped
> * Move the comment about "The Individual IDs need to math the VCF IDs" into the description for the Individual ID column
> * Add a checkbox that has the text "By submitting this form I agree to grant seqr access to the data in the associated workspace"and then the form should fail validation if the user does not select this

1) for the title, it seems not so simple to get rid of the seqr standard header and footer. And I think it is reasonable to have the header so that the users can easily go to the dashboard if they don't want to continue loading data. I added the line "Load data to seqr from AnVIL Workspace <namespace>/<name>" as the local title.
2) A border is added. I'm not sure if it is as your expectation.
3) The comment for `Individual IDs` is added.
4) A checkbox is added. But it seems that it should be associated with the `submit` button. If not checked, the submit button should be gray. I haven't done it yet.

The latest UI looks like the below picture.
image

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -62,80 +49,93 @@ const FAM_UPLOAD_FORMATS = [].concat(BASE_UPLOAD_FORMATS)
...
-              <TableCell>
-                <BoldText>{title}</BoldText> ({formatLinks ? formatLinks.map(
-                  ({ href, linkExt }, i) => <span key={linkExt}>{i > 0 && ' / '}<a href={href} target="_blank">.{linkExt}</a></span>)
-                  : `.${ext}`})
-              </TableCell>
-              <TableCell>
-                {ext &&
+export const BaseBulkContent = React.memo(({ url, actionDescription, details, project, name, requiredFields, optionalFields, uploadFormats, exportConfig, blankExportConfig }) =>
+  <StyledForm>
+    <NoBorderTable basic="very" compact="very">
```
hanars, line 54: Instead of putting this Table in a container that adds a margin and then making it very compact, maybe try changing `compact="very"` to plain `compact` or taking out the `compact` option all together

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -62,80 +49,93 @@ const FAM_UPLOAD_FORMATS = [].concat(BASE_UPLOAD_FORMATS)
...
-        </StyledTable>
-      </div>
-      {details && <div><br />{details}</div>}
-      <br />
-    </Container>
-    <br />
+                  }
+                  {exportConfig &&
+                  <span>
+                    or <FileLink data={exportConfig} ext={ext} linkContent="current individuals" />
```
hanars, line 80: The "or" needs to only be shown if you have both `blankExportConfig` and `exportConfig`

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -62,80 +49,93 @@ const FAM_UPLOAD_FORMATS = [].concat(BASE_UPLOAD_FORMATS)
...
+        )}
+        <Table.Row><Table.Cell /></Table.Row>
+        <Table.Row>
+          <Table.Cell colSpan={2}>
+            The table must have a header row with the following column names.
+          </Table.Cell>
+        </Table.Row>
+        <Table.Row>
+          <Table.Cell colSpan={2}>
+            <BoldText>Required Columns:</BoldText>
```
hanars, line 96: You can swap out all the `<Table.Cell><BoldText>...</BoldText></Table.Cell>` for  `<Table.HeaderCell>...</Table.HeaderCell>`

ui/shared/utils/constants.js
```diff
@@ -41,9 +41,13 @@ export const GENOME_VERSION_DISPLAY_LOOKUP = {
 
 // PROJECT FIELDS
 
+export const FILE_FIELD_NAME = 'uploadedFile'
+
+export const PROJECT_DESC_FIELDS = { name: 'description', label: 'Project Description', placeholder: 'Description' }
```
hanars, line 46: Using plural for `FIELDS` is confusing because this is only one field, so I would rename it to `PROJECT_DESCRIPTION_FIELD`

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import { FILE_FIELD_NAME, PROJECT_DESC_FIELDS, FAMILY_FIELD_ID, INDIVIDUAL_FIELD_ID } from 'shared/utils/constants'
+import { BaseBulkContent, BASE_UPLOAD_FORMATS } from 'pages/Project/components/edit-families-and-individuals/BulkEditForm'
+import { INDIVIDUAL_CORE_EXPORT_DATA, INDIVIDUAL_ID_EXPORT_DATA } from 'pages/Project/constants'
```
hanars, line 7: Our general policy is to not import component and constants in between different `pages` folders, and we instead move them to the `shared` folder. However, that adds a bunch of stuff to the diff and makes the PR harder to read. I thin what makes sense is to leave this as is for now, and then immediately after this PR is merged put up a second PR to move these. I would say `BaseBulkContent` would move to its own file in `shared/components/form` and the constant can just move to the main `shared/utils/constants`

hanars (changes requested):
Awesome work! This is getting really close

Can you also make sure to resolve these codacy issues: https://app.codacy.com/gh/broadinstitute/seqr/pullRequest?prid=6839237

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
ShifaSZ: Good point.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -0,0 +1,83 @@
...
+    parses the records, but doesn't save them in the database. Instead, it saves them to
+    a temporary file and sends a 'uploadedFileId' representing this file back to the client. If/when the
+    client then wants to 'apply' this table, it can send the uploadedFileId to the
+    save_individuals_table(..) handler to actually save the data in the database.
+
+    Args:
+        request (object): Django request object
+    """
+
+    uploaded_file_id, filename, json_records = save_uploaded_file(request)
```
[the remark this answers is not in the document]
ShifaSZ: Okay. That's good.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -0,0 +1,83 @@
...
+        'warnings': [],
+        'info': ['File {} uploaded.'.format(filename)],
+    }
+    logger.info(response)
+    return create_json_response(response)
+
+
+@login_required(login_url=API_LOGIN_REQUIRED_URL)
+def create_project_from_workspace(request, namespace, name):
+    """This API implements these features: 1) Validate that the current user has logged in through google and has one
```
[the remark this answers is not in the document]
ShifaSZ: Agree. I will move them into the function body when I implement them.

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
ShifaSZ: Okay.

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,101 @@
...
+  component: UploadPedigreeField,
+}
+
+const AGREE_CHECKBOX = {
+  name: 'agreeCheckbox',
+  component: BooleanCheckbox,
+  label: 'By submitting this form I agree to grant seqr access to the data in the associated workspace',
+}
+
+const FORM_FIELDS = [PROJECT_DESC_FIELDS, UPLOAD_PEDIGREE_FIELD, AGREE_CHECKBOX]
```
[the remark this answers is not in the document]
ShifaSZ: Okay. BTW. Should the `FORM_FIELDS` be moved to the constants file? If yes, I may change it to `LOAD_DATA_FORM_FIELDS`.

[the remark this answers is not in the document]
hanars: no, as they are only used in this file. The constants file is only if they need to be shared elsewhere

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,75 @@
...
+      url="/api/upload_temp_file"
+    />
+  </Segment>,
+)
+
+const UPLOAD_PEDIGREE_FIELD = {
+  name: FILE_FIELD_NAME,
+  label: 'Upload Pedigree Data',
+  component: UploadPedigreeField,
+}
```
ShifaSZ, line 43: The `label` is not displayed. I don't know why yet.

hanars: Oh hmm yeah it looks like you do nned to explicitly add the label to the component the way you originally were instead of passing it in as a prop. But you will want to add it in a `<label>` tag, not as a custom styled component. If that throws a warning in the js you can disable it they way we do here: https://github.com/broadinstitute/seqr/blob/master/ui/shared/components/form/Inputs.jsx#L300

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,87 @@
...
+  filename: 'individuals_template',
+  rawData: [],
+  headers: [...INDIVIDUAL_ID_EXPORT_DATA, ...INDIVIDUAL_CORE_EXPORT_DATA].map(config => config.header),
+  processRow: val => val,
+}
+
+const UploadPedigreeField = React.memo(() =>
+  <div>
+    {/* eslint-disable-next-line jsx-a11y/label-has-for */}
+    <StyledLabel>Upload Pedigree Data</StyledLabel>
```
ShifaSZ, line 36: To make the label style be consistent with other fields, I add custom styling to it. Is it okay or not?

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,87 @@
...
+      <BaseBulkContent
+        blankExportConfig={BLANK_EXPORT}
+        requiredFields={REQUIRED_FIELDS}
+        optionalFields={INDIVIDUAL_CORE_EXPORT_DATA}
+        uploadFormats={BASE_UPLOAD_FORMATS}
+        actionDescription="load individual data from an AnVIL workspace to a new seqr project"
+        url="/api/upload_temp_file"
+      />
+    </Segment>
+    <VerticalSpacer height={1} />
```
ShifaSZ, line 47: I'm not sure if there any other better way to add spacing between these two fields.

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
hanars: Labels should render properly inside forms, and adding custom styling to get things to look like default semantic ui usually means we are doing something wrong. My guess is that the issue is the exra nested div here. If you inspect the rendered html, my guess is what you see here is something like 
```
<div class="field">
  <div>
    <label>
```
and semantic ui is expecting 
```
<div class="field">
  <label>
```
The way to fix this here is is to change this component to a list instead of wrapping everything in an extra div. Remmeber with lists you always need to add a `key` prop to each element. So this function would return 
```
[
  <label key="label>Upload Pedigree Data</label>,
  <Segment key="uploadForm">....</Segment>,
]
```

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
hanars: fields should always have a 10px margin under them by default so not entirely sure whats happening here. See if the change above fixes it

hanars: also, if you are having trouble getting things to look right, feel free to set up a meeting with me and we can debug the styling issues together

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,78 @@
...
+        actionDescription="load individual data from an AnVIL workspace to a new seqr project"
+        url="/api/upload_temp_file"
+      />
+    </Segment>,
+  ],
+)
+
+const UPLOAD_PEDIGREE_FIELD = {
+  name: FILE_FIELD_NAME,
+  label: 'Upload Pedigree Data',
```
hanars, line 44: you can take this label field out since we ended up not using it

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -0,0 +1,78 @@
...
+  name: 'agreeCheckbox',
+  component: BooleanCheckbox,
+  label: 'By submitting this form I agree to grant seqr access to the data in the associated workspace',
+}
+
+const FORM_FIELDS = [PROJECT_DESC_FIELD, UPLOAD_PEDIGREE_FIELD, GENOME_VERSION_FIELD, AGREE_CHECKBOX]
+
+const LoadWorkspaceDataForm = React.memo(({ namespace, name }) =>
+  <div>
+    <Header className="large center aligned">Load data to seqr from AnVIL Workspace &quot;{namespace}/{name}&quot;</Header>
```
hanars, line 58: please read the documentation for components you are adding (https://react.semantic-ui.com/elements/header/) or look at other places in the code to see how they are done. If you still can't figure it out, ask me for help. We NEVER pass `className` to our semantic UI components. We pass the props we want to the component  and let them compute the classes. If you read the documentation it will tell you what props are available. This is an important thing for you to learn to do so I am not going to tell you which props you want ehre. If you can't figure it out from the documentation on your own, set up a meeting with me and I will show you how to do it for next time

hanars (changes requested):

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
ShifaSZ: okay

ui/pages/Login/components/Login.jsx
```diff
@@ -35,7 +36,7 @@ const Login = ({ onSubmit, googleLoginEnabled, location }) =>
       <Message visible warning>
         No existing seqr account for the Google user.
       </Message> }
-      <a href="/login/google-oauth2">Sign in with Google</a>
+      <a href={'/login/google-oauth2?next='.concat(`${queryString.parse(location.search).next || '/'}`)}>Sign in with Google</a>
```
ShifaSZ, line 39: I updated this to fix the issue of redirecting to the next page when logging in with social auth.

seqr/urls.py
```diff
@@ -269,10 +271,15 @@
     'matchmaker/v1/match': external_api.mme_match_proxy,
     'matchmaker/v1/metrics': external_api.mme_metrics_proxy,
 
+    'create_project_from_workspace/submit/(?P<namespace>[^/]+)/(?P<name>[^/]+)': create_project_from_workspace,
 }
 
 urlpatterns = [url('^status', status_view)]
 
+# anvil workspace
+anvil_workspace_url = 'anvil_workspace/(?P<namespace>[^/]+)/(?P<name>[^/]+)'
```
hanars, line 280: messaged in slack, but just to record it on the PR as well the AnVIL team requested we change the url to `'workspace/(?P<namespace>[^/]+)/(?P<name>[^/]+)'`

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.test.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { shallow, configure } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
+import LoadWorkspaceDataForm from './LoadWorkspaceDataForm'
+
+configure({ adapter: new Adapter() })
+
+test('shallow-render without crashing', () => {
+  shallow(<LoadWorkspaceDataForm namespace="test_namespace" name="test_name" />)
```
hanars, line 9: looks good!

seqr/urls.py
ShifaSZ: It belongs to the backend. I updated it in the backend branch.

ui/pages/Login/components/Login.jsx
hanars: you shouldn't provide any empty next param, if there isn't one. Also, theres no harm in proxying all the query parameters, so it would be a lot easier to just use the whole unparsed `location.search`:
```
`/login/google-oauth2${location.search}`
```

hanars (changes requested):

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
hanars: reminder that once this PR is merged to dev you should put up another PR  to move the necessary components and constants to the shared folder. Don't move the whole files, just the shared parts

hanars (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/758703009

codacy doesn't seem to be reporting any errors, so we can ignore the reported failure here

PR #1631, opened 2021-01-22
Igv controlled portal

hanars:
We were using the global redux state to control whether the IGV viewer was open and what was in it, and then had separate components for the button to open it and for the viewer itself. This changes it so there is a single component with its own state that manages both the viewer and the button, which then renders any layout component to show those components in the correct relative positions. 

This change is necessary as a precursor to some updates I need to make in IGV, as there will be more state needed for that so I wanted to get the state management cleaned up before I started on that


ui/shared/components/panel/FamilyReads.jsx
```diff
@@ -0,0 +1,198 @@
...
+import { Segment, Icon } from 'semantic-ui-react'
+
+import { getIndividualsByGuid, getIGVSamplesByFamily, getFamiliesByGuid } from 'redux/selectors'
+import PedigreeIcon from '../icons/PedigreeIcon'
+import IGV from '../graph/IGV'
+import { ButtonLink } from '../StyledComponents'
+import { VerticalSpacer } from '../Spacers'
+import { getLocus } from './variants/Annotations'
+
+const CRAM_PROXY_TRACK_OPTIONS = {
```
hanars, line 14: The top of this file is directly copy-pasted form the old FamlyVariantReads component, so I didn't change anything and it doesn't need review

ui/shared/components/panel/FamilyReads.jsx
```diff
@@ -0,0 +1,198 @@
...
+    showIdeogram: true,
+    showNavigation: true,
+    showRuler: true,
+    showCenterGuide: true,
+    showCursorTrackingGuide: true,
+    showCommandBar: true,
+  }
+}
+
+const ReadIconButton = props => <ButtonLink icon="options" content="SHOW READS" {...props} />
```
hanars, line 85: Start review for this file here

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/756127232

ui/shared/components/panel/FamilyReads.jsx
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      openFamily: null,
+    }
+  }
+
+  showReads = familyGuid => () => {
```
ShifaSZ, line 153: It can be more descriptive by changing this `showReads` to `getShowReadsSetter` or something like that.

ShifaSZ (changes requested):
Looks good. One comment needs to be responded to.

ui/shared/components/panel/FamilyReads.jsx
hanars: This is a function that shows the reads for the bound family. Using the words "getter" and "setter" is really not a javascript convention and changing to that would be confusing and would be a complete antipattern for how we name functions like this everywhere else. See https://github.com/broadinstitute/seqr/blob/master/ui/shared/components/table/DataTable.jsx#L82

ShifaSZ (approved):

PR #1633, opened 2021-01-25
fix memory issue

hanars:


seqr/utils/elasticsearch/es_search.py
```diff
@@ -812,7 +812,8 @@ def _filter_invalid_family_compound_hets(self, gene_id, family_compound_het_pair
                     continue
 
                 if self._allowed_consequences and self._allowed_consequences_secondary:
-                    consequences = variant_1['gene_consequences'].get(gene_id, [])
+                    # Make a copy of lists to prevent blowing up memory usage
+                    consequences = [] + variant_1['gene_consequences'].get(gene_id, [])
```
ShifaSZ, line 816: Is it a bug of the Python? Assigning a new value to `consequences` should make its old contents be recycled.

hanars: Its not a bug per say, it just causes a list that remains in memory to be updated so it causes memory to be really big. See the linked issue for detail

ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/756128248

PR #1636, opened 2021-01-26
Igv update

hanars:
Currently in seqr, we show bams/crams in IGV. We want to add the ability to show RNAseq data and SVs in IGV as well. This PR:

- Updates the `IgvSample` model to include a mandatory sample_type field to determine what sort of sample something is, and an optional sample_id field (as some samples will have different IDs from their individuals) 
- Updates the `IgvSample` to allow multiple samples per individual as long as they all have different sample types (previously it was restricted to one sample per individual)
- Updates the UI and apis for bulk uploading the IGV sample data to allow for the new sample types
- Updates the js code for generating the IGV tracks to show these new data types properly


hanars:
codacy error is okay to ignore

ShifaSZ (approved):
All look great.

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/756450117

Issue #1627, opened 2021-01-20
Elasticsearch 7 upgrade help and support


tommyli:
**Describe the bug**

This is not a bug report but more a support/help request.  We at MCRI recently merged latest upstream (this repo) into our forked repo (up to inclusive commit d86fb918e40d8ce848559512ff6a9ba593a917c9).
Last time we did this was around 4 months ago.  A major change within these 4 months is the upgrade from Elasticsearch 6 to 7 as described [here](https://github.com/broadinstitute/seqr/pull/1452).  So far our merging and testing has been pretty good.  Thanks very much for all your hard work in the ongoing development and improvement to Seqr and making it all freely available!

We had to make a couple of changes in the `seqr.utils.elasticsearch` modules to ensure backwards compatibility with our ES 6 instance.  We do have plans to upgrade to ES 7 as we feel matching upstream as close as possible is important for maintainability.  However, we may do that separately.  We have a few questions regarding using latest Seqr codebase with ES 6 and upgrading to ES 7 in general.

* Can you think of any major problems with running latest Seqr codebase on ES 6 cluster?  I've read through the changes in [breaking_70_search_changes](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/breaking-changes-7.0.html#breaking_70_search_changes)
  and cannot identify anything major.  More important is that I've done some manual testing to ensure latest codebase can query our MCRI Elasticsearch 6 cluster fine.
* The couple of changes we made makes Seqr backwards compatible on ES 6, would you like a PR for this?
* We plan to follow [rolling upgrade instructions](https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html)
  to upgrade our ES 6 (6.3) cluster to match upstream (7.8).  Were there anything to watch for when upgrading?

hanars:
> * Can you think of any major problems with running latest Seqr codebase on ES 6 cluster?  I've read through the changes in [breaking_70_search_changes](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/breaking-changes-7.0.html#breaking_70_search_changes)
>   and cannot identify anything major.  More important is that I've done some manual testing to ensure latest codebase can query our MCRI Elasticsearch 6 cluster fine.

[...]

PR #1639, opened 2021-01-27
feat (#1627): Ensure es client code is backwards compatible with ES6

tommyli:
As discussed in #1627 - this PR should not be accepted but raising this as a reference for other users on the changes required to work with Elasticsearch 6.

(cherry picked from commit 7ae922efacd9784fbfedaaac1d5615748066f9f6)


hanars:
Thanks for sharing!

PR #1640, opened 2021-01-27
feat: Parameterised repo and branch when building Seqr image

tommyli:
This change allows forked repos to easily override the repository and branch when building their own images using [docker build-arg](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

Overriding is optional and default values are the same as original.

E.g.:

```bash
docker-compose build \
  --build-arg "SEQR_REPO=https://github.com/ssadedin/seqr" \
  --build-arg "SEQR_GIT_BRANCH=mcri/master"
```

(cherry picked from commit 8c66148a8aa7f57e13de9e5320c2211cda80d6d3)


hanars (approved):
thanks, this is really useful!

PR #1652, opened 2021-01-29
better error handling for terra errors

hanars:
In dev I noticed that when my token refresh failed the API returned an unhandled 500 error and the ui did not show any sort of useful error message. This updates the middleware to correctly handle these errors and return the appropriate response code and json errors to the client


hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/756771356

ShifaSZ (approved):
It is good that we have a more consistent AnVIL exception handling now.

PR #1657, opened 2021-02-01
do not show last tagged sort on search page

hanars:
Similar to "family" sort, "last tagged" sort should only be shown on the saved variant page, not the search page


ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/757110979

PR #1660, opened 2021-02-01
remove duplicate xpos fields

hanars:
In the seqr database we have a field named `xpos_start`, but we added a special django field that lets us reference it in the code as either `xpos_start` or `xpos`, and we basically only use the aliased `xpos` fied. This is unnecessarily confusing and doesn't add any value, so this renames `xpos_start` to `xpos` in the database and removes the confusing alias


ShifaSZ (approved):

hanars:
passing tests: https://travis-ci.org/github/broadinstitute/seqr/builds/757151858

PR #1663, opened 2021-02-02
Automate es snapshots

sjahl:
Here's a start to turning on our elasticsearch snapshots. I'm having trouble determining the best way to automate some of the steps here in our current servctl setup -- we have kind of a chicken-and-egg problem with the service account and its key. We create the service account as a step in servctl, and then the key for it needs to be added to our secrets, which we can't really do from servctl. I'm open to suggestions on how to handle this scenario...

We *could* just create the service account and add the key to secrets manually before deploying, making the servctl serviceaccount creation a no-op. This is easy way out, but I feel like it has the potential to bite us if we deploy a fresh elasticsearch cluster where we haven't configured the secrets yet.

The last few remaining steps involve making curl requests against the elasticsearch API, which need to be authenticated. These requests 1) configure which bucket to send snapshots to and which creds to use, and 2) set up the schedule for automating the backups. Do we have any prior art for making elasticsearch API/configuration calls like this from servctl?


hanars:
> we have kind of a chicken-and-egg problem with the service account and its key. We create the service account as a step in servctl, and then the key for it needs to be added to our secrets, which we can't really do from servctl. I'm open to suggestions on how to handle this scenario...

What we did for the elasticsearch service accounts was we didn't use the auto-generated service account at all, and instead defined a secret that defines the service accounts and roles we want, that the elasticsearch deployment then picks up. So our elasticsearch instance will always have the same 3 service accounts with the same auth credentials for a seqr user, a pipeline user, and a kibana user, and then it also has an auto-generated account that changes every time we redeploy and we never use. Could something like that work here?


> We _could_ just create the service account and add the key to secrets manually before deploying, making the servctl serviceaccount creation a no-op. This is easy way out, but I feel like it has the potential to bite us if we deploy a fresh elasticsearch cluster where we haven't configured the secrets yet.

Like I mentioned above, our seqr connection to elasticsearch is 100% dependent on elasticsearch having the correct secrets configured. In fact, almost all of our deployments depend on secrets being properly configured. I don't see this as an issue, at least as long as we use our current deployment system. Also, If you deploy elasticsearch with an incorrect/missing secret, and then deploy or redeploy that secret, ECK will automatically pick up and apply the change.
TL;DR The idea that automated snapshots won't run on a brand new cluster that we haven't properly initialized is fine with me

> The last few remaining steps involve making curl requests against the elasticsearch API, which need to be authenticated. These requests 1) configure which bucket to send snapshots to and which creds to use, and 2) set up the schedule for automating the backups. Do we have any prior art for making elasticsearch API/configuration calls like this from servctl?

We haven't really done this sort fo thing before. The closest I can think of is in the seqr docker image we have a line that whenever the docker image is pulled it runs a post-deploy entyrpoint.sh script that does a bunch of set up, but honestly there are a lot of issues with this approach and I'm not sure we want to replicate it.

 I know it says this is an older approach, but do we maybe just want to use a kubernetes CronJob instead of the anpshot API? https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-snapshots.html#k8s_periodic_snapshots_with_a_cronjob
Alternatively, maybe we could add some sort of container lifecycle hook to the kubernetes config to run after it starts up? https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/

deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
```diff
@@ -18,6 +18,10 @@ spec:
     tls:
       selfSignedCertificate:
         disabled: true
+  {% if ES_CONFIGURE_SNAPSHOTS %}
+  secureSettings:
+    - secretName: es-snapshot-gcs-secrets
```
hanars, line 23: I have found that a lot of ECK's auto-magic is really heavily dependent on naming. The documentation names this secret `gcs-credentials`  but you've named it something else, and I don't see this secret being referenced anywhere else in your PR. Just confirming that this being used properly

sjahl:
OK, I'll move forward with the assumption that the key needs to be added to secrets to get things set up. We should be mostly fine here, since a lot of the logic is gated behind a boolean to enable the snapshots.

Using a CronJob would save us from one curl request that configures the schedule, but it won't do the other part of the configuration that sets up the bucket/repository. So I don't think it helps us any. I'm thinking this is just going to be a shell script that we run when we want to configure things (it only needs to be run once as far as I can tell).

deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
sjahl: Yes, this should be correct. What matters isn't the name of the containing secret, but the name of the *data* inside the secret. In this case, we're ending up with a key inside of `es-snapshot-gcs-secrets` called `gcs.client.default.credentials_file`, which tells elasticsearch that it's the key for a GCS client called `default`. When configuring the repository, we set the name of the client: https://github.com/broadinstitute/seqr/pull/1663/files#diff-0c198e3b884d6600c5301859c8677080b8878c0582f4f03075733d3510e9939aR12

see https://www.elastic.co/guide/en/elasticsearch/plugins/master/repository-gcs-usage.html#repository-gcs-using-service-account

sjahl: also re: the secret not being mentioned anywhere else in the PR: `es-snapshot-gcs-secrets` is the name of the secret as generated by `deploy_secrets`: https://github.com/broadinstitute/seqr/pull/1663/files#diff-ab443a9f2f134fc3a61c20f3a409e61afa6e604018632336eee9ac36795b2e9bR44

hanars:
> I'm thinking this is just going to be a shell script that we run when we want to configure things (it only needs to be run once as far as I can tell).

For this sort of thing, it might be worth embedding the script call in the deploy function in servctl

sjahl:
OK -- I think this is good to go. End result is that the `deploy_elasticsearch_shapshot_infra` command copies the config shell script to the seqr pod and executes it at the end.

To ship this to prod, we need to:

1) create the serviceaccount, generate a key and add it to secrets
2) deploy elasticsearch -- This adds the `secureSettings` do the Elasticsearch k8s object. In my testing in dev, this caused all the pods in the ES cluster to terminate and redeploy (one by one)
3) deploy seqr to pick up the new `$ES_SNAPSHOTS_BUCKET` envvar
4) deploy `elasticsearch-snapshot-infra`

deploy/kubernetes/gcloud-dev-settings.yaml
```diff
@@ -32,6 +32,10 @@ ES_DATA_DISK_TYPE: pd-standard
 ES_DATA_DISK_SIZE: 10
 ES_POD_VOLUME_SIZE: '1Gi'
 
+ES_CONFIGURE_SNAPSHOTS: true
```
hanars, line 35: I think generally we want this turned off for dev, so I would say we want to switch this before merging

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -40,6 +42,7 @@
 
 SECRETS = {
     'elasticsearch': ['users', 'users_roles', 'roles.yml'],
+    'es-snapshot-gcs': ['{deploy_to}/gcs.client.default.credentials_file'],
```
hanars, line 45: shouldn't our credentials for accessing google cloud be the same in both environments?

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -182,6 +185,29 @@ def _set_elasticsearch_kubernetes_resources():
...
+
+        # create the bucket
+        run("gsutil mb -p seqr-project -c STANDARD -l US-CENTRAL1 gs://%(ES_SNAPSHOTS_BUCKET)s" % settings, errors_to_ignore=["already exists"])
+
+        # check if the serviceaccount exists
+        try:
+            run("gcloud iam service-accounts describe %(ES_SNAPSHOTS_ACCOUNT_NAME)s@seqr-project.iam.gserviceaccount.com" % settings)
+        except RuntimeError:
+            # service account does not exist, create it now
+            run("gcloud iam service-accounts create %(ES_SNAPSHOTS_ACCOUNT_NAME)s --display-name %(ES_SNAPSHOTS_ACCOUNT_NAME)s" % settings)
```
hanars, line 201: our general approach everywhere else in servctl is to just run things with errors_to_ignore, rather than explicitly try-catching if things exist. This is okay to leave as is, but I was wondering if there was a reason you picked a different approach here

hanars:
A couple of questions about the deployment plan:

- The first step you list is creating the service account, but there is code to do this in the deploy function for `elasticsearch-snapshot-infra` which you have as the last step. Why do we need to do this manually first?
- Why do we need an environment variable set in the seqr pod at all? I see that you added it as an environment variable, but in general we only add environment variables for things that either the seqr app itself or the seqr deployment/ cron jobs need.

deploy/servctl_utils/deploy_command_utils.py
hanars: Also, given that this secret is referenced in our elasticsearch yml file and is used by ECK, is there a reason you chose to make this its own secret and not part of the main elasticsearch secret?

sjahl: No, the service account we're creating only grants access to a particular GCS bucket -- we don't want the dev ES cluster to have storage.admin on the prod cluster's snapshot bucket.

deploy/servctl_utils/deploy_command_utils.py
sjahl: This was just to avoid running the create command multiple times, as I wasn't sure how idempotent it was. Testing now, it just fails if the acct already exists, so I can switch to more familiar logic.

hanars:
If ECK restarts the cluster, or if we redeploy the cluster fully by deleting and rebuilding instead of updating in place, my understanding is these snapshots would be disabled, and we would have to manually re-run the servctl command to actually schedule the backups. If I am wrong about that feel free to ignore the rest of this comment 

I see your point about the fact that using a cron job would "only' save us sending the curl command and wouldn't handle setting up the bucket and service account. But moving to an entirely new google project, where we would need a new service account and bucket, is something we essentially never do and of we did we'd have to be really careful about, so having that be a one-time servctl command is fine with me. But fully redeploying elasticsearch is something we do plan to do with some regularity, and certainly ECK restarts things all the time without telling us. So to me, having kubernetes managing the scheduling seems a lot safer and more maintainable

sjahl:
> A couple of questions about the deployment plan:
> 
>     * The first step you list is creating the service account, but there is code to do this in the deploy function for `elasticsearch-snapshot-infra` which you have as the last step. Why do we need to do this manually first?

This is precisely the chicken-and-egg scenario I mentioned. The creation of the service account is scripted here, but we can't automate adding the secret. So this step needs to be done first, then we manually generate the secret and add it to our secrets, then do the rest of the steps. In the final snapshot infra deploy, the account will already exist, and it's a no-op. For the most part, the service account creation steps in deployctl are there so we don't forget how to do it. I can take them out and document them somewhere instead, if that's preferable.

>     * Why do we need an environment variable set in the seqr pod at all? I see that you added it as an environment variable, but in general we only add environment variables for things that either the seqr app itself or the seqr deployment/ cron jobs need.

Since we run the final script that configures the snapshots from the seqr pod, it needs to know the name of the snapshot bucket. This pod seemed like the most convenient place to run this, since it had all the other credentials required.

sjahl:
> If ECK restarts the cluster, or if we redeploy the cluster fully by deleting and rebuilding instead of updating in place, my understanding is these snapshots would be disabled, and we would have to manually re-run the servctl command to actually schedule the backups. If I am wrong about that feel free to ignore the rest of this comment
> 
> I see your point about the fact that using a cron job would "only' save us sending the curl command and wouldn't handle setting up the bucket and service account. But moving to an entirely new google project, where we would need a new service account and bucket, is something we essentially never do and of we did we'd have to be really careful about, so having that be a one-time servctl command is fine with me. But fully redeploying elasticsearch is something we do plan to do with some regularity, and certainly ECK restarts things all the time without telling us. So to me, having kubernetes managing the scheduling seems a lot safer and more maintainable

I can't find any docs on how persistent the snapshot policy and repository settings are... My expectation is that they'd be stored in an index or something, so redeploying the cluster shouldn't delete the policy. But without docs from elastic, I have no idea 🤷

deploy/servctl_utils/deploy_command_utils.py
sjahl: > Also, given that this secret is referenced in our elasticsearch yml file and is used by ECK, is there a reason you chose to make this its own secret and not part of the main elasticsearch secret?

I wanted to avoid rocking the elasticsearch boat, since our deployment methodology deletes all the secrets and re-creates them. But this was before I realized that it was going to restart the cluster to add the `secretSetting`... so I'm happy to move this into the elasticsearch-secrets namespace, given that the boat is already rocking 😄

sjahl:
> I can't find any docs on how persistent the snapshot policy and repository settings are... My expectation is that they'd be stored in an index or something, so redeploying the cluster shouldn't delete the policy. But without docs from elastic, I have no idea 🤷

My understanding of the code that implements SLMs is that the configuration is stored inside one of the xpack indices, which I think we frequently/sometimes blow away when we redeploy ES. You are correct that it may be better to spin up a cronjob for this, so I'll add that approach. One of the reasons that I avoided the CronJob to begin with was that kubernetes cronjobs are notoriously unreliable (they only went 'stable' in k8s v1.20, which we're likely several months away from in GKE). But -- since the snapshots can always be run by hand trivially, we can see how well the cronjob works and make adjustments if necessary.

deploy/servctl_utils/deploy_command_utils.py
hanars: Yeah I totally get the no-boat-rocking thing, but sadly that does seem necessary here so 🤞

hanars:
> So this step needs to be done first, then we manually generate the secret and add it to our secrets, then do the rest of the steps. In the final snapshot infra deploy, the account will already exist, and it's a no-op. For the most part, the service account creation steps in deployctl are there so we don't forget how to do it. I can take them out and document them somewhere instead, if that's preferable.

Oh yeah you did explain that already, my bad. I think leaving the code as is is fine then, since having it documented in the code is helpful and it doesn't hurt to run to no-op

> Since we run the final script that configures the snapshots from the seqr pod, it needs to know the name of the snapshot bucket. This pod seemed like the most convenient place to run this, since it had all the other credentials required.

Ah, so this is like why we run all of our postgres backups out of the seqr pod 🙃 . I feel like this is maybe moot with the cron approach?

hanars:
> One of the reasons that I avoided the CronJob to begin with was that kubernetes cronjobs are notoriously unreliable (they only went 'stable' in k8s v1.20, which we're likely several months away from in GKE). But -- since the snapshots can always be run by hand trivially, we can see how well the cronjob works and make adjustments if necessary.

We could also try adding the script to a lifecycle hook in the elastcisearch kubernetes configuration? I'm less confident about this approach tbh, but figured it was worth mentioning

sjahl:
> We could also try adding the script to a lifecycle hook in the elastcisearch kubernetes configuration? I'm less confident about this approach tbh, but figured it was worth mentioning

I'll look into this. One of the things that we lose by using a CronJob is the ability to define a retention policy, which automatically cleans up old shapshots for us. The CronJob is more of an imperative "take a snapshot when this container runs". The built in lifecycle has a little more depth ("take snapshots monthly, and remove old snaps after 90days") Not super critical, but it's nice to have so that we don't have to clean up old snapshots on our own.

deploy/servctl_utils/deploy_command_utils.py
sjahl: Just following up on this, the snapshot secrets *must* remain separate. Because of the way ECK auto-injects the `secureSettings` into the cluster trust store, it tries to also inject all the other items in the elasticsearch-secrets as well -- not all of which are valid entries in the trust store. Result is crashy elasticsearch :/

sjahl:
@hanars I've got alerting set up for the cronjobs. In the case of a failure, log messages from the cronjob have a `jsonPayload.error` field; the field is absent in successful cases. The alert notifies our slack channel if more than zero of those messages appears.

I did additional testing to see what happens in cases where the snapshot repository is missing or the serviceAccount lacks permissions to write to the bucket. Both cases result in an error that fires the alert.

hanars:
Looks good to me! These are the steps you outlined before for deploying to prod. Is this still accurate?

> 1. create the serviceaccount, generate a key and add it to secrets
> 2. deploy elasticsearch -- This adds the `secureSettings` do the Elasticsearch k8s object. In my testing in dev, this caused all the pods in the ES cluster to terminate and redeploy (one by one)
> 3. deploy seqr to pick up the new `$ES_SNAPSHOTS_BUCKET` envvar
> 4. deploy `elasticsearch-snapshot-infra`

sjahl:
Ah, some things are different, since we no longer run commands in the seqr pod, and i moved some tasks. It should be:

1) deploy `elasticsearch-snapshot-infra` (creates the bucket, and IAM serviceAccount)
2) generate service account key and add it to secrets (this process is documented in our secrets),  and deploy `secrets`
3) deploy `elasticsearch` to get the new secureSettings, which causes a rolling restart of the ES cluster pods.
4) deploy `elasticsearch-snapshot-config`

hanars:
Okay, so it seems to me that 1, 2, and 4 are pretty non-disruptive and can be done whenever. For 3, we may want to do that off hours

sjahl:
@hanars I think this is good to merge?

hanars (approved):

hanars:
yep this looks good to me! We can coordinate the deployment in slack

PR #1673, opened 2021-02-16
Clean up shared components in the Project page.

ShifaSZ:


ui/shared/utils/constants.js
```diff
@@ -55,6 +55,17 @@ export const PROJECT_FIELDS = [
   GENOME_VERSION_FIELD,
 ]
 
+export const BASE_UPLOAD_FORMATS = [
```
hanars, line 58: variable names that make sense scoped within a file often don't make sense when they are exported. Base of what? This should be renamed `FILE_FORMATS`

hanars (changes requested):

ui/shared/utils/constants.js
ShifaSZ: okay

requirements-dev.txt
```diff
@@ -5,4 +5,4 @@ django-debug-toolbar-request-history==0.0.11    # for debugging database queries
 mock                                            # mock objects for unit tests
 responses==0.10.15                               # mock HTTP responses for unit tests
 urllib3-mock==0.3.3                             # mock urllib3 for tests
-Jinja2==2.10.1                                  # jinja templates are used by servctl k8s deployment scripts
+Jinja2==2.11.3                                  # jinja templates are used by servctl k8s deployment scripts
```
ShifaSZ, line 8: There was a Snyk error as below:

Vulnerable module: | jinja2
-- | --
Introduced through: | jinja2@2.10.1
Exploit maturity: | Proof of concept
Fixed in: | 2.11.3

Is it okay to make this change?

hanars: no and you should not be changing python dependencies at all as part of this

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
```diff
@@ -1,143 +1,33 @@
...
 import { connect } from 'react-redux'
-import { Table } from 'semantic-ui-react'
 
 import { getCurrentProject, getUser } from 'redux/selectors'
-import { FileLink } from 'shared/components/buttons/ExportTableButton'
-import FileUploadField from 'shared/components/form/XHRUploaderField'
 import ReduxFormWrapper from 'shared/components/form/ReduxFormWrapper'
-import { NoBorderTable } from 'shared/components/StyledComponents'
-import { INDIVIDUAL_HPO_EXPORT_DATA, FILE_FIELD_NAME } from 'shared/utils/constants'
+import BaseBulkContent from 'shared/components/form/BulkUploadForm'
```
hanars, line 7: you should import this using the filename unless there is a very good reason not to, so this should be `BulkUploadForm`

hanars (changes requested):

ui/pages/Project/components/edit-families-and-individuals/BulkEditForm.jsx
ShifaSZ: `BaseBulkContent` is going to be connected in this file. So, it is a real `base` component. Don't you think it is a good reason to keep the name in this file?

hanars: no, I think its confusing to rename an imported component

hanars (approved):

PR #1679, opened 2021-02-18
Add data bucket path for data loading from AnVIL

ShifaSZ:


seqr/views/apis/anvil_workspace_api.py
```diff
@@ -79,6 +80,11 @@ def create_project_from_workspace(request, namespace, name):
     # Add the seqr service account to the corresponding AnVIL workspace
     add_service_account(request.user, namespace, name)
 
+    # Validate the data path
+    if not request_json['dataPath'].startswith("gs://") or not does_file_exist(request_json['dataPath']):
+        error = 'Data path is not valid.'
```
hanars, line 85: I think it would be helpful to users if you specify in the error what the data path is. So something like `'File "{}" not found'.format(request_json['dataPath'])`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -79,6 +80,11 @@ def create_project_from_workspace(request, namespace, name):
     # Add the seqr service account to the corresponding AnVIL workspace
     add_service_account(request.user, namespace, name)
 
+    # Validate the data path
+    if not request_json['dataPath'].startswith("gs://") or not does_file_exist(request_json['dataPath']):
```
hanars, line 84: So all of the files for a given workspace are in one google bucket, the name of which is shown on the main workspace page. But when you actually look at the files in the workspace, it just shows you the relative paths. I think expecting users to go and find the root google bucket name is a bit too confusing, and also not necessary as we should be able to programmatically get the bucket associated with the workspace. Can you change this so we get the google bucket from anvil and then only have the users provide the file path, and then validate against the whole thing?

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -62,7 +62,14 @@ const AGREE_CHECKBOX = {
   validate: validators.required,
 }
 
-const FORM_FIELDS = [PROJECT_DESC_FIELD, UPLOAD_PEDIGREE_FIELD, GENOME_VERSION_FIELD, AGREE_CHECKBOX]
+const DATA_BUCK_FIELD = {
+  name: 'dataPath',
+  label: 'Google Path to the VCF files',
```
hanars, line 67: This should be "Path to the Joint Called VCF". I also think adding a `labelHelp` here would be helpful, as that will render some additional help text on hover next to the label. I think it should say something like "File path for a joint called VCF available in the workspace files. If The VCF is split, provide the path to the directory containing the split VCF"

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -93,6 +100,14 @@ const LoadWorkspaceDataForm = React.memo(({ namespace, name }) =>
       size="small"
       fields={FORM_FIELDS}
     />
+    <p>
+      Need help? please submit &nbsp;
+      <a href="https://github.com/broadinstitute/seqr/issues">GitHub Issues</a>
+      , &nbsp; or &nbsp;
+      <a href="https://mail.google.com/mail/?view=cm&amp;fs=1&amp;tf=1&amp;to=seqr@broadinstitute.org" target="_blank">
```
hanars, line 107: you should use mailto for email links: `href="mailto:seqr@broadinstitute.org"`

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -93,6 +100,14 @@ const LoadWorkspaceDataForm = React.memo(({ namespace, name }) =>
       size="small"
       fields={FORM_FIELDS}
     />
+    <p>
+      Need help? please submit &nbsp;
+      <a href="https://github.com/broadinstitute/seqr/issues">GitHub Issues</a>
```
hanars, line 105: We want to direct them to the page for submitting a new bug report, not the page for the list of issues: `https://github.com/broadinstitute/seqr/issues/new?labels=bug&template=bug_report.md`

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: Good point. I should be able to get the bucket for the workspace.

seqr/views/utils/terra_api_utils.py
```diff
@@ -162,6 +162,12 @@ def list_anvil_workspaces(user):
     return r
 
 
+def user_get_workspace_bucket(user, workspace_namespace, workspace_name):
+    path = "api/workspaces/{0}/{1}?fields=workspace.bucketName".format(workspace_namespace, workspace_name)
+    r = _user_anvil_call('get', path, user)
+    return r['workspace']['bucketName']
```
ShifaSZ, line 168: The unit test for this function is to be added.

hanars: So I though I have is that when you call `check_workspace_perm`, that ends up calling `user_get_workspace_access_level` which calls this exact endpoint but with a different `fields` query param. What if `check_workspace_perm` and `has_workspace_perm` has a keyword arg called `get_workspace_meta` and if it was True both of those functions would pass an extra `fields` argument to `user_get_workspace_access_level` to return extra data from that request and then we would only need to query the bucket once. So something like this:
```
def check_workspace_perm(user, permission_level, namespace, name, can_share=False, get_workspace_meta=False):
    workspace_perm = has_workspace_perm(user, permission_level, namespace, name, can_share, get_workspace_meta)
    if workspace_perm:
        return workspace_perm
   # rest of the function stays the same

...

def has_workspace_perm(user, permission_level, namespace, name, can_share=False, get_workspace_meta=False):
    workspace_permission = user_get_workspace_access_level(user, namespace, name, fields=['workspace.bucketName'] if get_workspace_meta else None)
    ...
    if not has_permission:
       return False
    return workspace_permission['workspace'] if get_workspace_meta else True

...

def user_get_workspace_access_level(user, workspace_namespace, workspace_name, fields=None):
    query_fields = ['accessLevel', 'canShare'] + (fields or []) 
    path = "api/workspaces/{0}/{1}?fields={2}".format(workspace_namespace, workspace_name, ','.join(query_fields))
    # rest of the function stays the same

```

ShifaSZ: Sound good!

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -119,8 +123,18 @@ def test_create_project_from_workspace(self, mock_email, mock_add_service_accoun
         self.assertEqual(response.json()['error'], 'Failed to grant seqr service account access to the workspace {}/{}'
                               .format(TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME))
 
-        # Test valid operation
+        # Test bad data path
         mock_add_service_account.reset_mock(side_effect=True)
+        mock_get_bucket.return_value = 'test_bucket'
```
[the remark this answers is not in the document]
ShifaSZ: I still need to test the parameters while calling the mocked function. So, does it still need `mock_get_bucket` in the argument list?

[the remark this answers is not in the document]
hanars: oh I forgot about that. You are right that you need to test what parameters it is called with and you should leave this code as-is

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -69,6 +70,14 @@ def create_project_from_workspace(request, namespace, name):
         error = 'Must agree to grant seqr access to the data in the associated workspace.'
         return create_json_response({'error': error}, status=400, reason=error)
 
+    # Validate the data path
+    bucket_name = user_get_workspace_bucket(request.user, namespace, name)
+    slash = '' if request_json['dataPath'].startswith('/') else '/'
+    data_path = 'gs://{bucket}{slash}{path}'.format(bucket=bucket_name, slash=slash, path=request_json['dataPath'])
+    if not does_file_exist(data_path):
+        error = 'Data path {} is not valid.'.format(data_path)
```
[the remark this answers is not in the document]
ShifaSZ: Okay. I'm going to update it.

seqr/views/utils/permissions_utils.py
```diff
@@ -76,22 +76,26 @@ def _map_anvil_seqr_permission(anvil_permission):
...
         return False
-    return has_workspace_perm(user, permission_level, project.workspace_namespace, project.workspace_name)
+    return bool(has_workspace_perm(user, permission_level, project.workspace_namespace, project.workspace_name))
 
 
-def has_workspace_perm(user, permission_level, namespace, name, can_share=False):
-    workspace_permission = user_get_workspace_access_level(user, namespace, name)
+def has_workspace_perm(user, permission_level, namespace, name, can_share=False, meta_fields=None):
+    workspace_permission = user_get_workspace_access_level(user, namespace, name, meta_fields) if meta_fields else \
+        user_get_workspace_access_level(user, namespace, name)
```
ShifaSZ, line 84: I call `user_get_workspace_access_level()` with two different parameter lists because I don't want to change the test assertions about the function's parameters. There are many of these assertions which could be a distraction to review. After your reviewing, I could change it to `workspace_permission = user_get_workspace_access_level(user, namespace, name, meta_fields)` (without following `if`) and update the assertions.

seqr/views/utils/terra_api_utils_tests.py
```diff
@@ -203,11 +203,8 @@ def test_user_get_workspace_access_level(self, mock_time, mock_redis, mock_logge
 
     @responses.activate
     @mock.patch('seqr.views.utils.terra_api_utils.SERVICE_ACCOUNT_FOR_ANVIL', TEST_SERVICE_ACCOUNT)
-    @mock.patch('seqr.utils.redis_utils.redis.StrictRedis')
     @mock.patch('seqr.views.utils.terra_api_utils.time')
-    def test_add_service_account(self, mock_time, mock_redis):
-        mock_redis.return_value.get.return_value = None
-
+    def test_add_service_account(self, mock_time):
```
ShifaSZ, line 207: The mock of the `redis` is never used in this test. I didn't take them out in last PR.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -84,20 +86,33 @@ def test_create_project_from_workspace(self, mock_email, mock_add_service_accoun
...
+        _ = self.client.post(url)
         mock_utils_logger.warning.assert_called_with('test_user_collaborator does not have sufficient permissions for workspace {}/{}'
                                                .format(TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME))
 
         # Test missing required fields in the request body
+        self.login_manager()
+        self.mock_get_ws_access_level.return_value = {
+            'accessLevel': 'WRITER',
+            'canShare': True,
+            'workspace': {'bucketName': 'test_bucket'}
```
hanars, line 104: I would strongly prefer you just use the main mocked fixtures for all the test to ensure consistent behavior. Theres no harm in justadding this to the test fixture data

seqr/views/utils/permissions_utils.py
```diff
@@ -76,22 +76,26 @@ def _map_anvil_seqr_permission(anvil_permission):
 def anvil_has_perm(user, permission_level, project):
     if not project_has_anvil(project):
         return False
-    return has_workspace_perm(user, permission_level, project.workspace_namespace, project.workspace_name)
+    return bool(has_workspace_perm(user, permission_level, project.workspace_namespace, project.workspace_name))
```
hanars, line 79: The signature for `has_workspace_perm`should not change, if you don't pass `meta_fields` it needs to continue to return a bool

seqr/views/utils/permissions_utils.py
hanars: a cleaner way to do this that will not change how the existing calls are made is 
```
kwargs = {}
if meta_fields:
    kwargs['meta_fields'] = meta_fields
workspace_permission = user_get_workspace_access_level(user, namespace, name, **kwargs)
```

seqr/views/utils/permissions_utils.py
```diff
@@ -76,22 +76,26 @@ def _map_anvil_seqr_permission(anvil_permission):
...
+        user_get_workspace_access_level(user, namespace, name)
     if not workspace_permission:
         return False
     if can_share and not workspace_permission.get(CAN_SHARE_PERM):
         return False
     permission = _map_anvil_seqr_permission(workspace_permission)
-    return True if permission == CAN_EDIT else permission == permission_level
+    if permission != CAN_EDIT and permission != permission_level:
+        return False
+    return workspace_permission
```
hanars, line 92: should be `return workspace_permission if meta_fields else True`

hanars (changes requested):

seqr/views/utils/permissions_utils.py
ShifaSZ: I see what you mean.

hanars (approved):

PR #1681, opened 2021-02-19
Exome/genome combined projects

hanars:
Code tweaks to allow us to load multiple sample types (i.e both exomes and genomes) to a single seqr project


seqr/views/apis/dataset_api_tests.py
```diff
@@ -15,6 +15,7 @@
 PROJECT_GUID = 'R0001_1kg'
 INDEX_NAME = 'test_index'
 SV_INDEX_NAME = 'test_new_sv_index'
+NEW_INDEX_NAME = 'test_new_index'
```
ShifaSZ, line 18: Would it be more descriptive to rename `NEW_INDEX_NAME` to `NEW_SAMPLE_TYPE_INDEX_NAME` or a similar name else?

ShifaSZ (changes requested):
Looks mostly good to me. There is one comment.

seqr/views/apis/dataset_api_tests.py
hanars: done

ShifaSZ (approved):

PR #1682, opened 2021-02-22
Anvil UI tweaks

hanars:


ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -128,7 +128,25 @@ const ProjectOverview = React.memo((
...
+        <div>
+          No Datasets Loaded <br />
+          <i>Where is my data?</i> <Popup
+            trigger={<HelpIcon />}
+            hoverable
+            content={
+              <div>
+                Loading data from AnVIL to seqr is a slow process, and generally takes a week.
+                If you have been waiting longer than this for your data, please reach
+                out to <a href="mailto:seqr@broadinstitute.org">seqr@broadinstitute.org</a>
```
ShifaSZ, line 143: Should this message be shown only when AnVIL auth is enabled?

ShifaSZ (changes requested):
Good tweaks for the login redirecting and login UI. One comment for a popup message.

ui/pages/Project/components/ProjectOverview.jsx
hanars: hmm thinking about this we should actually only show it if that specific project has an anvil workspace associated with it. Thanks for catching this

ShifaSZ (approved):

PR #1684, opened 2021-02-23
nm package bumps

hanars:
Bumps the versions of a couple js dependencies to resolve some package vulnerabilities. Also switches us from using the semantic-ui package to the semantic-ui-css package as all we were using it for was the css import and this allows us to remove a ton of downstream dependencies and vulnerabilities. This is actually the recommended way of using semantic ui with the semantic-ui-react library we use


ShifaSZ (approved):
No issue is found.

PR #1686, opened 2021-02-25
use hail.hadoop_is_file instead of gsutil to check for missing cram paths more efficiently. Delete missing paths from database.

bw2:


bw2:
Need to update unittest

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -3,30 +3,47 @@
...
+        ) if args else IgvSample.objects.all()).prefetch_related('individual', 'individual__family__project')
 
-        failed = []
-        for sample in samples:
-            try:
-                validate_alignment_dataset_path(sample.file_path)
-            except Exception as e:
+        missing_counter = collections.defaultdict(int)
+        for sample in tqdm.tqdm(samples, unit=" samples"):
+            if sample.file_path and sample.file_path.startswith("gs://") and not hl.hadoop_is_file(sample.file_path):
```
hanars, line 35: this should really be done at the database query level - rather than pulling everything out of the database and then checking you could do `filter(file_path__startswith='gs://')`

hanars (changes requested):

seqr/management/commands/check_bam_cram_paths.py
bw2: oh yeah

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -3,30 +3,49 @@
...
+            if not hl.hadoop_is_file(sample.file_path):
                 individual_id = sample.individual.individual_id
-                failed.append(individual_id)
-                logger.info('Error at {} (Individual: {}): {} '.format(sample.file_path, individual_id, str(e)))
+                project = sample.individual.family.project.name
+                missing_counter[project] += 1
+                logger.info('Individual: {}  file not found: {}'.format(individual_id, sample.file_path))
+                if not options.get('dry_run'):
+                    sample.file_path = ""
+                    sample.save()
```
hanars, line 44: so my only thought would be if we thought this was going to be a bunch of samples, you could save a list of the ids of the models that need to be updated and then do it as a bulk update at the end instead of potentially hundreds of database saves - `Sample.objects.filter(in__in=missing_sample_ids).update(file_path='')`

bw2: good call

bw2:
Failing to install hail/pyspark when running tests. Do you have any suggestions?

hanars:
> Failing to install hail/pyspark when running tests. Do you have any suggestions?

no we've never installed hail before. Maybe ask the pipeline team how they do that for unit tests? Or look in hails documentation? You can also view the detailed output for why the install is failing in the detail for the job

We are also going to mock hail everywhere its used in our tests, so in a pinch you could update the command for installing dependencies to proceed if it fails to install hail without failing the test suite and the tests should be okay

bw2:
I found I can fix the hail install based on  Nick's reply @ https://discuss.hail.is/t/trouble-installing-hail/1520

Now tests pass in python3.7 but not 3.9. Based on   
https://hail.zulipchat.com/#narrow/stream/123010-Hail-0.2E2.20support/topic/.40mock.2Epatch(.22hail.2Ehadoop_is_file.22.2C.20lambda.20p.3A.20False)
hail will support 3.9 at the end of March. Not sure what you prefer to do in the meantime.

hanars:
> I found I can fix the hail install based on Nick's reply @ https://discuss.hail.is/t/trouble-installing-hail/1520
> 
> Now tests pass in python3.7 but not 3.9. Based on
> https://hail.zulipchat.com/#narrow/stream/123010-Hail-0.2E2.20support/topic/.40mock.2Epatch(.22hail.2Ehadoop_is_file.22.2C.20lambda.20p.3A.20False)
> hail will support 3.9 at the end of March. Not sure what you prefer to do in the meantime.

@sjahl the idea behind the 3.9 tests was supposed to be that we don't fail the test suite when those tests fail, its just a way of knowing how far off we are from being able to upgrade our version and try to keep us from making and changes that will make upgrading harder. But I think this is the exact sort of case where we want to know that we made a breaking change but are okay ignoring it for now. Is there a way we can change the way the test suite is reporting back? I don't really like the idea of training ourselves to ignore big red test failures

sjahl:
The continue-on-error flag that we have affects whether the workflow as a whole passes (you can see that this workflow run is in a "Passed" state / has a green check mark if you look at all the runs in https://github.com/broadinstitute/seqr/actions). It's unfortunate that this overall workflow status isn't communicated in the PR view, and instead we get red X's everywhere 😞 

Lots of people have complained about this: https://github.com/actions/toolkit/issues/399

From what I can tell, github doesn't have anything like a 'warning' state that we can optionally return. Possible workarounds that I see, in no particular order, are:

- We can designate some checks as required for merging (https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule). I can't see the settings page for this repo, so I can't confirm how the python matrix shows up there, and if it allows us to only designate one of the versions as required. The goal here would be to not allow merging if anything except the python 3.9 test fails. We still have red X marks, but the ones that matter are much harder to ignore.
- We can move the 3.9 check into its own workflow that runs after merge and doesn't impact the PR. We'd then put a separate 3.9 "future" test status badge in the readme. Only danger I can see with this option is that depending on other traffic in the repo, it can become less clear which change caused the test failure. We'd also still have red X marks in this scenario, but fewer of them on PRs
- We can forego testing 3.9. Small downside is that we lose some visibility into future breakage and exactly what we changed to break it. It can always be added back into the matrix as a required check when we're ready to start testing that version. Given that our upstream dependencies don't support 3.9, maybe it doesn't make sense to test it.

hanars:
I think option 2 or 3 would be best. I think that losing the link to the exact code change that started the tests failing is okay for us in the long run, as it should still be possible to fix the test

seqr/management/tests/check_bam_cram_paths_tests.py
```diff
@@ -3,32 +3,43 @@
...
-            mock.call('Error at /readviz/NA19675.cram (Individual: NA19675_1): Error accessing "/readviz/NA19675.cram" '),
+            mock.call('Individual: NA19675_1  file not found: gs://missing-bucket/missing_file'),
             mock.call('---- DONE ----'),
             mock.call('Checked 1 samples'),
-            mock.call('1 failed samples: NA19675_1'),
+            mock.call('1 files not found:'),
+            mock.call('   1 in 1kg project nåme with uniçøde'),
         ]
         mock_logger.info.assert_has_calls(calls)
-        mock_validate_path.assert_called_with("/readviz/NA19675.cram")
```
hanars, line 34: Its important to test that the mock hail check is being called the way we expect

seqr/management/tests/check_bam_cram_paths_tests.py
```diff
@@ -3,32 +3,43 @@
...
+from seqr.models import IgvSample
 
+
+@mock.patch('hail.hadoop_is_file', lambda path: False)
 class CheckBamCramPathsTest(TestCase):
     fixtures = ['users', '1kg_project']
 
+    def setUp(self):
+        existing_sample = IgvSample.objects.first()
+        IgvSample.objects.create(
```
hanars, line 15: Why not just update the existing igv sample to have gs file path so you can use it, or even add a new fixture. Setting up test data like this is is a real last resort and an antipattern

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -1,32 +1,51 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
+    help = 'Checks all gs:// bam or cram paths and, if a file no longer exists, deletes the path from the database'
 
     def add_arguments(self, parser):
-        parser.add_argument('args', nargs='*')
+        parser.add_argument(
```
hanars, line 17: you need to add a test for the command with the dry run argument

hanars (changes requested):

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -1,32 +1,51 @@
...
-                failed.append(individual_id)
-                logger.info('Error at {} (Individual: {}): {} '.format(sample.file_path, individual_id, str(e)))
+                project = sample.individual.family.project.name
+                missing_counter[project] += 1
+                logger.info('Individual: {}  file not found: {}'.format(individual_id, sample.file_path))
+                if not options.get('dry_run'):
+                    guids_of_samples_with_missing_file.add(sample.guid)
+
+        if len(guids_of_samples_with_missing_file) > 0:
+            IgvSample.objects.filter(guid__in=guids_of_samples_with_missing_file).update(file_path='')
```
hanars, line 44: Sorry, I forgot that we need to switch everything over to use our bulk helpers that log everything to compliance standards. This actually should be  `IgvSample.bulk_update(user=None, update_json={'file_path': ''}, guid__in=guids_of_samples_with_missing_file)`

seqr/management/tests/check_bam_cram_paths_tests.py
bw2: I think the db should start with 1 record that needs to be updated and 1 that doesn't, and to make sure it's only modifying the right one. I agree adding hundreds of records this way would be suboptimal, but adding 1 or 2 seems ok, unless there's a better way to add unittest-specific records.

hanars: I don't see any harm in adding a record to the fixtures file so we can start with 2 records in the configuration we want

bw2:
Reverting. "low coupling, high cohesion" => unittests should have their own db records.

seqr/management/tests/check_bam_cram_paths_tests.py
```diff
@@ -3,32 +3,53 @@
...
+        )
+
+    @mock.patch('hail.hadoop_is_file')
     @mock.patch('seqr.management.commands.check_bam_cram_paths.logger')
-    @mock.patch('seqr.views.utils.dataset_utils.validate_alignment_dataset_path')
-    def test_command(self, mock_validate_path, mock_logger):
+    def test_command_with_project(self, mock_logger, mock_hadoop_is_file):
+        mock_hadoop_is_file.return_value = False
         call_command('check_bam_cram_paths', '1kg project n\u00e5me with uni\u00e7\u00f8de')
+        self._check_results(1, mock_logger, mock_hadoop_is_file)
```
hanars, line 25: can you also add a test that calling it with a different project doesn't check this sample?

hanars:
@sjahl can we implement whatever change we decided on for the 3.9 tests. I'd rather not keep this blocked for too long

sjahl:
Done. I've left the matrix in place, with one version of python. It doesn't hurt anything, and will make it easier to add new versions later when we're ready to test them.

hanars (approved):

PR #1691, opened 2021-02-26
Moved parsing pedigree file to the end of validation.

ShifaSZ:
Re-order the validations of the pedigree file and the data path.
And, get the `SERVICE_ACCOUNT_FOR_ANVIL` setting only for AnVIL mode.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -359,7 +359,7 @@ if TERRA_API_ROOT_URL:
     SERVICE_ACCOUNT_FOR_ANVIL = subprocess.run(['gcloud auth list --filter=status:ACTIVE --format="value(account)"'],
                                                capture_output=True, text=True, shell=True).stdout.split('\n')[0]
     if not SERVICE_ACCOUNT_FOR_ANVIL:
-        raise Exception('Login Google service account with command "$ gcloud auth activate-service-account" before running seqr')
+        raise Exception('Error starting seqr - gcloud auth is not properly configured')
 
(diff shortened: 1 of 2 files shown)


settings.py
```diff
@@ -355,7 +353,14 @@
 TERRA_PERMS_CACHE_EXPIRE_SECONDS = os.environ.get('TERRA_PERMS_CACHE_EXPIRE_SECONDS', 60)
 TERRA_WORKSPACE_CACHE_EXPIRE_SECONDS = os.environ.get('TERRA_WORKSPACE_CACHE_EXPIRE_SECONDS', 300)
 
+SERVICE_ACCOUNT_FOR_ANVIL = None
+
 if TERRA_API_ROOT_URL:
+    SERVICE_ACCOUNT_FOR_ANVIL = subprocess.run(['gcloud auth list --filter=status:ACTIVE --format="value(account)"'],
+                                               capture_output=True, text=True, shell=True).stdout.split('\n')[0]
+    if not SERVICE_ACCOUNT_FOR_ANVIL:
+        raise Exception('Login Google service account with command "$ gcloud auth activate-service-account" before running seqr')
```
hanars, line 362: We should NEVER manually run this command. If seqr is deployed without the google account being authenticated, it means something has gone horribly wrong with our docker image. Pease replace this message with something like. "Error starting seqr - gcloud auth is not properly configured"

hanars (changes requested):

hanars (approved):

PR #1694, opened 2021-03-02
limit download variants to under max variants

hanars:
Currently this kind of search fails with an opaque Internal Server Error, so adding error handling to provide better user experience


ShifaSZ (approved):
Looks good to me.

PR #1701, opened 2021-03-04
Fix airtable access

hanars:
There are endpoints that need to check that  user has ever logged in with google (i.e has a valid google account) but recent changes added a check for access token. Since AnVIL isn't live yet, obviously no one has an access token set and this is making it impossible for users to access these pages. This change the is_google_authenticated helper back to its original purpose, and beefs up the is_anvil_authenticated helper and uses that in more places

### DIFF
--- a/seqr/views/utils/terra_api_utils.py
+++ b/seqr/views/utils/terra_api_utils.py
@@ -89,7 +89,7 @@ def _get_call_args(path, headers=None, root_url=None):
 
 def _safe_get_social(user):
     if not google_auth_enabled() or not hasattr(user, 'social_auth'):
-        return False
+        return None
 
     social = user.social_auth.filter(provider=SOCIAL_AUTH_PROVIDER)
     return social.first() if social else None
--- a/seqr/views/utils/terra_api_utils.py
+++ b/seqr/views/utils/terra_api_utils.py
@@ -55,20 +55,10 @@ def anvil_enabled():
 
 
 def is_google_authenticated(user):
-    if not google_auth_enabled() or not hasattr(user, 'social_auth'):
-        return False
-
-    social = _safe_get_social(user)
-    if social and social.extra_data:
-        return social.extra_data.get('access_token', '') != ''
-
-    return False
+    bool(_safe_get_social(user))
 
 
 def remove_token(user):
-    if not google_auth_enabled():
-        return
-
     social = _safe_get_social(user)
     if social and social.extra_data:
         social.extra_data.pop('access_token', None)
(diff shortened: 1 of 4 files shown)


seqr/views/utils/terra_api_utils.py
```diff
@@ -91,6 +88,9 @@ def _get_call_args(path, headers=None, root_url=None):
 
 
 def _safe_get_social(user):
+    if not google_auth_enabled() or not hasattr(user, 'social_auth'):
+        return False
```
ShifaSZ, line 92: Return `None` seems better than `False` since the purpose of the helper function is to get the social auth information.

seqr/views/utils/terra_api_utils.py
```diff
@@ -77,7 +67,14 @@ def remove_token(user):
 
 
 def is_anvil_authenticated(user):
-    return anvil_enabled() and is_google_authenticated(user)
+    if not anvil_enabled():
+        return False
+
+    social = _safe_get_social(user)
+    if social and social.extra_data:
+        return social.extra_data.get('access_token', '') != ''
```
ShifaSZ, line 75: I've just recognized that it would be even better to add a token expiration test like:`social.extra_data['auth_time'] + social.extra_data['expires'] - 10) <= int(time.time())`. And, we could define a helper function for the test and re-use it with `_get_social_access_token()`.

seqr/views/utils/terra_api_utils.py
```diff
@@ -55,20 +55,10 @@ def anvil_enabled():
...
 def is_google_authenticated(user):
-    if not google_auth_enabled() or not hasattr(user, 'social_auth'):
-        return False
-
-    social = _safe_get_social(user)
-    if social and social.extra_data:
-        return social.extra_data.get('access_token', '') != ''
-
-    return False
+    return bool(_safe_get_social(user))
```
ShifaSZ, line 58: Does `is_google_authenticated(user)` need to ensure the `user` has logged in? If yes, we may still need to do the token information test.

ShifaSZ (changes requested):
It is nice to re-organize the user authentication status tests. Some thoughts are added.

seqr/views/utils/terra_api_utils.py
hanars: this is out of scope of this PR. I am happy to have a conversation offline about why I think that is a bad idea

seqr/views/utils/terra_api_utils.py
hanars: no, it needs to test that they have ever logged in with google, as I stated in the PR description

hanars: also, again, only with anvil enabled is there ever a token. Currently, logged in users do not have a token. Adding this sort of thing is out of the scope of this PR

ShifaSZ: Oh, yes. No token is needed for local google login.

ShifaSZ (approved):

PR #1702, opened 2021-03-04
Clear demo tags

hanars:
Add a command to clear tags from demo projects. This type of request happens with some regularity so we should have an actual command for this, instead of me bulk deleting things in the db

### DIFF
--- a/seqr/management/tests/clear_project_tags_tests.py
+++ b/seqr/management/tests/clear_project_tags_tests.py
@@ -31,3 +31,9 @@ class TransferFamiliesTest(TestCase):
 
         self.assertEqual(SavedVariant.objects.filter(family__project__guid='R0003_test').count(), 0)
         self.assertEqual(VariantTag.objects.filter(saved_variants__family__project__guid='R0003_test').count(), 0)
+
+        # Test only demo projects can be deleted
+        mock_input.side_effect = 'y'
+        call_command('clear_project_tags', '1kg')
+        mock_input.assert_called_with('Are you sure you want to clear the tags for the following 0 projects (y/n): \n')
+        mock_logger.assert_called_with('Deleted 0 entities:')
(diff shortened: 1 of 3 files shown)


seqr/management/tests/clear_project_tags_tests.py
```diff
@@ -0,0 +1,39 @@
...
+            'Are you sure you want to clear the tags for the following 1 projects (y/n): Test Reprocessed Project\n')
+        mock_logger.assert_has_calls([
+            mock.call('Deleted 5 entities:'),
+            mock.call('    VariantTag_saved_variants: 2'),
+            mock.call('    VariantTag: 1'),
+            mock.call('    SavedVariant: 2'),
+        ], any_order=True)
+
+        self.assertEqual(SavedVariant.objects.filter(family__project__guid='R0003_test').count(), 0)
+        self.assertEqual(VariantTag.objects.filter(saved_variants__family__project__guid='R0003_test').count(), 0)
```
ShifaSZ, line 33: Do you also want to test the `VariantNote` and `VariantFunctionalData` results?

seqr/management/tests/clear_project_tags_tests.py
```diff
@@ -0,0 +1,39 @@
...
+
+        # Test only demo projects can be deleted
+        mock_input.side_effect = 'y'
+        call_command('clear_project_tags', '1kg')
+        mock_input.assert_called_with('Are you sure you want to clear the tags for the following 0 projects (y/n): \n')
+        mock_logger.assert_called_with('Deleted 0 entities:')
```
ShifaSZ, line 39: Since entities have been deleted in the previous tests, it will always delete 0 entities. Maybe move these lines before the success deletion test.

ShifaSZ (changes requested):

seqr/management/tests/clear_project_tags_tests.py
hanars: no because we don't have fixture data for them

ShifaSZ (approved):

PR #1712, opened 2021-03-10
Single SV search fix

hanars:
Updates how the "seqr" search works for SVs to restrict results to those that both start and end within a 20% margin of the start and end of the given  SV, rather than the current search will return any overlapping SV for the whole region.See discussion in the ticket for motivation


ShifaSZ (approved):
Looks good to me.

PR #1714, opened 2021-03-10
Fix refresh token

hanars:


settings.py
```diff
@@ -363,6 +363,7 @@
 
     SOCIAL_AUTH_GOOGLE_OAUTH2_AUTH_EXTRA_ARGUMENTS = {
         'access_type': 'offline',  # to make the access_token can be refreshed after expired (expiration time is 1 hour)
+        'approval_prompt': 'auto', # required for successful token refresh
```
ShifaSZ, line 366: It is required only when we want to refresh the refresh_token. It should not be required for refreshing the access_token.

ShifaSZ (approved):
There is one comment but it shouldn't matter.

PR #1726, opened 2021-03-16
List project optimization

hanars:
The database query that was causing postgres to run out of memory was constructed by getting the project models supported for anvil projects and then annotating them, which resulted in a very large query that grouped, annotated, and distinct-ed every project in the database. This instead gets an explicit list of projects a user has permission for, and then makes a query and annotation on only that subset of projects. 

It turns out that with this change we weren't really using the list of project models a user had access to, we were primarily using that list for further filtering, so I added caching for the project guids and updated the code to expect a list of guids a user has access to,  instead of a list of models


seqr/views/apis/dashboard_api.py
```diff
@@ -40,10 +40,11 @@ def dashboard_page_data(request):
 
 
 def _get_projects_json(user):
-    projects = get_projects_user_can_view(user)
-    if not projects:
+    project_guids = get_project_guids_user_can_view(user)
+    if not project_guids:
         return {}
 
+    projects = Project.objects.filter(guid__in=project_guids)
```
hanars, line 47: this is the change that actually fixes the out of memory error. Before, `projects` here in the anvil-only case was 
```
projects = Project.objects.filter(can_view_group__user=user)
...
projects = (projects | _get_analyst_projects())
...
anvil_permitted_projects = Project.objects.annotate(
            workspace = Concat('workspace_namespace', Value('/'), 'workspace_name')).filter(workspace__in=workspaces)
return (anvil_permitted_projects | projects).distinct()
```

calling annotate for all the counts on this query was causing everything to blow up. While this change technically can mean there are 2 queries occurring instead of only one (although frequently the first query will be cached), the huge annotation step is being called on a much smaller and better scoped query

seqr/views/apis/users_api.py
```diff
@@ -24,7 +24,7 @@
 @login_required(login_url=API_LOGIN_REQUIRED_URL)
 def get_all_collaborator_options(request):
     collaborators = set()
-    for project in get_projects_user_can_view(request.user):
+    for project in get_local_access_projects(request.user):
```
hanars, line 27: this endpoint is only ever used for giving a list of users who can be added to a project as local collaborators. Since this functionality is completely disabled for projects that have workspaces associated with them, there is no harm in restricting the user list here to only local projects.

seqr/views/utils/permissions_utils.py
```diff
@@ -150,21 +152,37 @@ def _get_analyst_projects():
...
+        projects = get_local_access_projects(user)
+
+    project_guids = [p.guid for p in projects.only('guid')]
+    if is_anvil_authenticated(user) and not is_data_manager:
+        workspaces = ['/'.join([ws['workspace']['namespace'], ws['workspace']['name']]) for ws in
+                      list_anvil_workspaces(user)]
+        project_guids += [p.guid for p in Project.objects.filter(workspace_name__isnull=False).exclude(
+            workspace_name='').exclude(guid__in=project_guids).annotate(
+            workspace=Concat('workspace_namespace', Value('/'), 'workspace_name')).filter(
+            workspace__in=workspaces).only('guid')]
```
ShifaSZ, line 174: Doing filtering first then annotating will help for now when most of the projects haven't been associated with anvil workspaces. But, in the future, when more projects have workspaces configured, the annotation could cost more performance loss. I still prefer to add a field for combining the namespace and the name of the workspace. But, we could leave it as is and change it later based on further performance evaluation.

seqr/views/utils/permissions_utils.py
```diff
@@ -150,21 +152,37 @@ def _get_analyst_projects():
...
 
 
-def get_projects_user_can_view(user):
-    if user_is_data_manager(user):
-        return Project.objects.all()
+def get_project_guids_user_can_view(user):
+    cache_key = 'projects__{}'.format(user)
+    project_guids = safe_redis_get_json(cache_key)
+    if project_guids is not None:
+        return project_guids
```
ShifaSZ, line 159: How about caching the project guids permitted by AnVIL only. It seems doesn't make sense for the location permissions won't take effect instantly.

ShifaSZ (changes requested):

seqr/views/utils/permissions_utils.py
hanars: I disagree. The performance boost for caching all guids is more than worth a small lag time in local project permissions, especially since we are actively working to get rid of local project permissions

seqr/views/utils/permissions_utils.py
hanars: Changing this now seems like a seriously premature optimization. This type of annotation is still pretty quick until you get up into the thousands of projects probably. The issue we were running into with the annotations today is because those annotations include joins with other tables, which can get slow. I will leave this as is, and we can revisit this idea in the future if we start to notice performance issues

seqr/views/utils/permissions_utils.py
ShifaSZ: Okay

seqr/views/utils/permissions_utils.py
ShifaSZ: Okay

ShifaSZ (approved):

PR #1739, opened 2021-03-18
tweak logs for better debugging

hanars:
We get a lot of noisy error alerts for dev that we don't want, so this disables them. Also, stops logging a null traceback which makes logs harder to read

### DIFF
--- a/seqr/utils/logging_utils.py
+++ b/seqr/utils/logging_utils.py
@@ -1,6 +1,8 @@
 import json
 import logging
 
+from settings import DEPLOYMENT_TYPE
+
 class JsonLogFormatter(logging.Formatter):
 
     def usesTime(self):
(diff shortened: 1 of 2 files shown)


seqr/utils/logging_utils.py
```diff
@@ -22,10 +24,10 @@ def formatMessage(self, record):
         if hasattr(record, 'db_update'):
             log_json['dbUpdate'] = record.db_update
 
-        if hasattr(record, 'traceback'):
+        if getattr(record, 'traceback', None):
             log_json['traceback'] = record.traceback
 
-        if record.levelname == 'ERROR':
+        if record.levelname == 'ERROR' and DEPLOYMENT_TYPE != 'dev':
```
ShifaSZ, line 30: Does `DEPLOYMENT_TYPE != 'dev'` mean `DEPLOYMENT_TYPE == 'prod'?  Is there any other valid type which logging on google?

ShifaSZ (changes requested):
There is one question.

seqr/utils/logging_utils.py
hanars: right now theres only dev and prod, but I thought it was better to alert on log errors by default and explicitly exclude the cases we don't want to be alerted (i.e for dev), rather than explicitly include what we do want. This means that if in the future we change the name of the deployment there is no risk of losing our error reporting - if we make a breaking change it would cause us to get more logs than we want rather than less, which is a safer problem to have

PR #1740, opened 2021-03-18
Admin logging

hanars:
For compliance purposes, we need a record of all changes to the database made through the django admin. It turns out that the admin by default writes all changes (create, update, and delete) to a table in the database with all the necessary metadata we need. This PR makes that table viewable in the admin itself so we can easily see what changes were made if we are ever asked about it.  I also tweaked some of the existing admin display while I was in there


ShifaSZ (approved):
Can't find any issues.

PR #1752, opened 2021-03-19
correctly handle missing gene sensitivity

hanars:
Fixes a bug that causes saved variant pages to error and not display anything


ShifaSZ (commented):
Good to me.

PR #1755, opened 2021-03-22
install wheel before installing dependencies

hanars:
Adding hail as a dependency breaks the requirements installation in docker build without first downloading wheel


hanars:
we can ignore codacy for the docker file

sjahl (approved):
This looks OK to me. Only suggestion I might have is to put both the pip installs in the same `RUN` command to avoid creating another layer in the docker image (i.e. `RUN pip install wheel && pip install -r requirements.txt`). But, feel free to take or leave that suggestion, as this is hardly the worst offender in this Dockerfile :)

PR #1756, opened 2021-03-22
fix bug when creating partially saved compound het variant

hanars:


seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -431,7 +431,11 @@ def test_create_partially_saved_compound_het_variant_note(self):
...
         request_body = {
-            'variant': [COMPOUND_HET_5_JSON, {'variantId': '21-3343353-GAGA-G', 'xpos': 21003343353, 'ref': 'GAGA', 'alt': 'G'}],
+            'variant': [COMPOUND_HET_5_JSON, {
+                'variantId': '21-3343353-GAGA-G', 'xpos': 21003343353, 'ref': 'GAGA', 'alt': 'G',
+                'variantGuid': 'SV0000001_2103343353_r0390_100',
+                'tagGuids': ['VT1708633_2103343353_r0390_100', 'VT1726961_2103343353_r0390_100'], 'noteGuids': []},
+            ],
             'note': 'one_saved_one_not_saved_compount_hets_note',
             'submitToClinvar': True,
             'familyGuid': 'F000001_1',
```
ShifaSZ, line 441: It will be better to add a test for not updating an existing saved variant.

ShifaSZ (changes requested):

seqr/views/apis/saved_variant_api_tests.py
hanars: it is testing that, with the change to this test data without the corresponding actual code fix this test gives a 500 error identical to the one reported in the ticket

ShifaSZ (approved):

PR #1762, opened 2021-03-24
Create Mnv/ complex SV

hanars:
Creates the ability to associate existing tagged variants with newly created manual variants. For SNVs this is used for multinucleotide variants, for SVs this will be used for complex SVs. Requires adding support for more than 2 "paired" variants (currently we only pair compound hets, which are always exactly 2). See linked ticket for details


hanars:
note: hold off on merging until after the 3/25 deploy

ShifaSZ (approved):
Can't find any issues.

PR #1768, opened 2021-03-30
add better log detail for failed pedigree generation

hanars:


seqr/utils/logging_utils.py
```diff
@@ -27,6 +27,9 @@ def formatMessage(self, record):
         if getattr(record, 'traceback', None):
             log_json['traceback'] = record.traceback
 
+        if getattr(record, 'detail', None):
+            log_json['detail'] = record.detail
+
```
ShifaSZ, line 32: Are there any tests that cover these?

ShifaSZ (changes requested):

seqr/utils/logging_utils.py
ShifaSZ: Okay

ShifaSZ (approved):

PR #1772, opened 2021-03-31
Fail unaccepted policies

hanars:
For security reasons, users should not be able to access any API endpoints without first accepting the seqr policies


ui/pages/Login/components/AcceptPolicies.jsx
```diff
@@ -0,0 +1,46 @@
...
+]
+
+const AcceptPolicies = React.memo(({ onSubmit }) =>
+  <UserFormLayout
+    header="Seqr Policies"
+    subheader="Before continuing to use seqr, please read and accept our policies"
+    onSubmit={onSubmit}
+    form="acceptPolicies"
+    fields={POLICY_FORM_FIELDS}
+  />,
```
ShifaSZ, line 35: The `UserFormLayout` component has a header bar with some buttons on it. It seems that it is better to disable these buttons except `loggout`.

ShifaSZ (changes requested):

seqr/views/apis/users_api_tests.py
```diff
@@ -390,12 +393,12 @@ def test_set_password(self):
         self.mock_list_workspaces.assert_not_called()
         self.mock_get_ws_acl.assert_not_called()
 
-    def test_forgot_password(self, *args):
-        super(MixUsersAPITest, self).test_forgot_password(*args)
+    def test_forgot_password(self, *args, **kwargs):
+        super(MixUsersAPITest, self).test_forgot_password(*args, **kwargs)
```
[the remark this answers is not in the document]
hanars: yes I updated codacy to use the python 3 pylint instead of python 2 pylint, and the arguments-differ error now goes away if you use `(*args, **kwargs)` notation

seqr/fixtures/users.json
```diff
@@ -171,22 +171,112 @@
...
     "user_id": 11,
-    "privacy_version": 0.7,
-    "tos_version": 0.3
+    "privacy_version": 2.1,
+    "tos_version": 1.3
+  }
+},
+{
+  "model": "seqr.userpolicy",
+  "pk": 1,
```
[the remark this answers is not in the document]
hanars: yes, I will fix it

ui/pages/Login/components/AcceptPolicies.jsx
hanars: so actually, this header has no concept of which page it is on, so changing this is incredibly difficult. If a user clicks any of those links and tries to view any data it will just redirect them back to the accept policies page, so I don't think leaving this as-is will cause any harm

ShifaSZ: Okay

ShifaSZ (approved):

PR #1778, opened 2021-04-07
Update markdown renderer

hanars:
The library we were using for rendering markdown hasn't been updated in 4 years and is insecure, so switching to something thats more current


ShifaSZ (approved):
Looks good to me

PR #1781, opened 2021-04-09
Add Oauth column to users table

hanars:
This will allow us to easily monitor the progress of getting all our users to log in with oauth


ShifaSZ (approved):
Looks good.

PR #1786, opened 2021-04-09
correctly handle django 404 errors

hanars:


seqr/utils/middleware.py
```diff
@@ -20,6 +21,7 @@
 EXCEPTION_ERROR_MAP = {
     PermissionDenied: 403,
     ObjectDoesNotExist: 404,
+    Http404: 404,
```
ShifaSZ, line 24: Are there any specific cases that raise ObjectDoesNotExist or Http404 exception? I can't find any in seqr.

ShifaSZ (changes requested):
There is a question I would like to be answered.

seqr/utils/middleware.py
hanars: these are django builtin errors, so while we never reaise them ourselves they happen all the time in the django library code. `Http404` happens for example whenever we request an image file path that doesn't exist, which is the issue that this PR is fixing, and ObjectDoesNotExist is raised if you do a get query for something that doesn't exist

ShifaSZ (approved):

PR #1790, opened 2021-04-12
run check_bam_cram_paths at 2am daily

bw2:


hanars:
the output of cron jobs doesn't go anywhere, and I don't really like the idea of silently resetting cram paths and not telling anyone about it. When last we spoke about this we had discussed adding some sort of email/ slack alerting to let the PM team know which paths were broken/ removed. I thought that was going to be done before we added a cron job

bw2:
is there any code in seqr that already does this for other things?

hanars:
theres code for emailing people and for slacking updates, with helper functions in `communication_utils.py`. Matchmaker does a lot of slack updates for example, and when someone uploads one of the merged pedigree files it emails all PMs. And theres a `resend_welcome_email` manage command that sends an email from a manage command. I think I have a slight preference for slack over email, but its a a weak preference

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -49,3 +53,16 @@ def handle(self, *args, **options):
...
+                SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL,
+                'Found {} broken bam/cram path(s)'.format(sum(missing_counter.values())),
+            )
+            slack_message = ""
+            for project_name, missing_paths_list in project_name_to_missing_paths.items():
+                slack_message += "In project {}\n".format(project_name)
+                slack_message += "\n".join([
+                    "  {}   {}".format(individual_id, path) for individual_id, path in missing_paths_list
+                ])
+                safe_post_to_slack(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message)
```
hanars, line 68: Is there a reason to do this as separate slacks for each project instead of one big post? My concern is sometimes we get errors when we send too many slack messages in a row where a couple fail to go through because it gets overloaded, and if we only send one message for the job that won't be an issue

settings.py
```diff
@@ -316,6 +316,8 @@
 MME_SLACK_MATCH_NOTIFICATION_CHANNEL = 'matchmaker_matches'
 MME_SLACK_SEQR_MATCH_NOTIFICATION_CHANNEL = 'matchmaker_seqr_match'
 
+SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL = 'seqr_loading_pipeline'
```
hanars, line 319: not sure this is a good channel for these alerts, since I think we are trying to move responsibility for managing IGV issues to the PM team. Can you post in the channel asking people there if they think this would be a good channel for them/ which channel they would recommend?

hanars (changes requested):

deploy/docker/seqr/entrypoint.sh
```diff
@@ -72,6 +72,7 @@ if [ $ENABLE_DATABASE_BACKUPS ]; then
 0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_omim --omim-key $OMIM_KEY >> /proc/1/fd/1 2>&1
 0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_human_phenotype_ontology >> /proc/1/fd/1 2>&1
 0 12 * * 1 /usr/local/bin/python /seqr/manage.py detect_inactive_privileged_users >> /proc/1/fd/1 2>&1
+0 2 * * * /usr/local/bin/python /seqr/manage.py check_bam_cram_paths >> /var/log/cron.log 2>&1
```
hanars, line 75: I *just* switched piping the output from the cron jobs to `/proc/1/fd/1` instead of `/var/log/cron.log` as that will allow these logs to get picked up by stackdriver so I can get alerted on failures and check that things are running as expected etc.

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -49,3 +53,12 @@ def handle(self, *args, **options):
...
                 logger.info('   {} in {}'.format(c, project_name))
+
+            # post to slack
+            slack_message = 'Found {} broken bam/cram path(s)\n'.format(sum(missing_counter.values()))
+            for project_name, missing_paths_list in project_name_to_missing_paths.items():
+                slack_message += "\nIn project {}:\n".format(project_name)
+                slack_message += "\n".join([
+                    "  {}   {}".format(individual_id, path) for individual_id, path in missing_paths_list
+                ])
+            safe_post_to_slack(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message)
```
hanars, line 64: can you update the unit test to test the slack notifications

hanars (changes requested):

seqr/management/commands/check_bam_cram_paths.py
bw2: by mocking `safe_post_to_slack`?

hanars: yep!

PR #1791, opened 2021-04-12
Transport error formatting

hanars:


seqr/utils/middleware.py
```diff
@@ -62,6 +76,9 @@ def process_exception(request, exception):
         if DEBUG or status == 500:
             traceback_message = traceback.format_exc()
             exception_json['traceback'] = traceback_message
+        detail = getattr(exception, 'info', None)
+        if isinstance(detail, dict):
+            exception_json['detail'] = detail
```
ShifaSZ, line 81: I'm curious that since the details in `exception.info` have been put into `exception_josn['error']`, does it still necessary to add `exception_json['detail']`?

ShifaSZ (changes requested):
I have a question.

seqr/utils/middleware.py
hanars: yes this works

ShifaSZ (approved):

PR #1797, opened 2021-04-16
use kubernetes django secret key by default

hanars:
django uses a `SECRET_KEY` setting that is used as part of the hashing to determining whether or not a session is valid. Users have been complaining recently of being frequently logged out of seqr and there are a ton of "Session data corrupted" warning logs. Django makes it clear that resetting the `SECRET_KEY` invalidates all active sessions, so you are supposed to keep this a constant. We have a `django_key` secret in our deployment that gets assigned to the `DJANGO_KEY` environment variable which will persist across seqr deployments. However, for some reason seqr defaults to getting the key from a file, and if that file is missing it generates a new key, only using the env variable if that fails. While we need to keep the file generation method for local users (who will need a key generated the first time they deploy seqr) we should start off by using the environment variable corresponding to the kubernetes secret

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -183,17 +183,19 @@ AUTHENTICATION_BACKENDS = (
 )
 
 # set the secret key
-SECRET_FILE = os.path.join(BASE_DIR, 'django_key')
-try:
-    SECRET_KEY = open(SECRET_FILE).read().strip()
-except IOError:
+SECRET_KEY = os.environ.get('DJANGO_KEY')
+if not SECRET_KEY:
+    SECRET_FILE = os.path.join(BASE_DIR, 'django_key')
     try:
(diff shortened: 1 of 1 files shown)


settings.py
```diff
@@ -183,17 +183,15 @@
...
 # set the secret key
-SECRET_FILE = os.path.join(BASE_DIR, 'django_key')
-try:
-    SECRET_KEY = open(SECRET_FILE).read().strip()
-except IOError:
+SECRET_KEY = os.environ.get('DJANGO_KEY')
+if not SECRET_KEY:
+    SECRET_FILE = os.path.join(BASE_DIR, 'django_key')
     try:
+        SECRET_KEY = open(SECRET_FILE).read().strip()
```
sjahl, line 190: Do we need to close this file handle?

hanars: fixed it!

sjahl (approved):

PR #1800, opened 2021-04-16
only send necessary update fields

hanars:
editing large projects is getting a 413 from nginx because we are sending the whole project json which is very large. Instead, we should only send the fields that actually might have been edited


ShifaSZ (approved):
Looks good.

PR #1801, opened 2021-04-16
Better permission denied error

hanars:
Instead of redirecting logged in users to the login page when they try to access something they don't have permission for, return a 403 error instead

### DIFF
--- a/seqr/views/utils/permissions_utils.py
+++ b/seqr/views/utils/permissions_utils.py
@@ -34,14 +34,17 @@ def _has_current_policies(user):
 
 
 # User access decorators
-_current_policies_required = user_passes_test(_has_current_policies, login_url=API_POLICY_REQUIRED_URL)
+def _check_active(user):
+    if not user.is_active:
+        raise PermissionDenied('User is no longer active')
+    return True
 
-def _active_required(view_func, login_url=API_LOGIN_REQUIRED_URL):
-    return user_passes_test(lambda user: user.is_active, login_url=login_url)(view_func)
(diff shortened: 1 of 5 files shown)


seqr/views/utils/permissions_utils.py
```diff
@@ -34,14 +34,19 @@ def _has_current_policies(user):
 
 
 # User access decorators
+def _require_permission(test_func, error='User has insufficient permission'):
+    def test_user(user):
+        if not test_func(user):
+            raise PermissionDenied(error)
+        return True
+    return test_user
```
ShifaSZ, line 42: It looks like renaming `test_func` to `permission_test_func` will be more descriptive. It can remind that not to use this function for other test condition. Otherwise, it could raise `PermissionDenied` for a condition unrelated to permission.

seqr/views/utils/permissions_utils.py
```diff
@@ -51,16 +56,15 @@ def login_and_policies_required(view_func):
 
 def _user_has_policies_and_passes_test(test_func):
     def decorator(view_func):
-        return _active_required(_current_policies_required(user_passes_test(test_func, login_url=API_LOGIN_REQUIRED_URL)(view_func)))
+        return login_and_policies_required(user_passes_test(_require_permission(test_func))(view_func))
```
ShifaSZ, line 59: So thoughts as above, `test_func` is better to be renamed as `permission_test_func`.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #1812, opened 2021-04-26
update landing page

hanars:


ui/pages/Public/LandingPage.jsx
```diff
@@ -28,46 +38,50 @@ export default () =>
...
+    <PageSegment secondary>
+      <List bulleted>
+        <List.Item>
+          If you are interested in collaborating with our group, please &nbsp;
+          <Anchor href="mailto:seqr@broadinstitute.org">contact us</Anchor>
+        </List.Item>
+        <List.Item>
+          Please use the <Anchor href="http://github.com/broadinstitute/seqr/issues">github issues page</Anchor> to
+          submit bug reports or feature requests
+        </List.Item>
```
ShifaSZ, line 78: It seems that a list item is missing in the action list. You can find it at https://docs.google.com/document/d/1lCis6ze8FYGqI6AB9Fm4tqOrSJg3NMp_ddAiwOyxuVM/edit?disco=AAAAMEh6HVs

ShifaSZ (changes requested):
The updates look good except for one list item.

ui/pages/Public/LandingPage.jsx
hanars: that list item is no longer accurate so I intentionally removed it. That document is the proposal, this PR is what we actually want

ShifaSZ (approved):

PR #1814, opened 2021-04-27
Push users to use google login

hanars:


ShifaSZ (approved):
Cool! I like the new login design.

PR #1815, opened 2021-04-27
add returned variant tag type

hanars:


ShifaSZ:
I don't see any real code changes except the fixture data for testing. How is the new variant tag type added?

hanars:
> I don't see any real code changes except the fixture data for testing. How is the new variant tag type added?

By running `./manage.py loaddata new_variant_tag_types` which is our standard way of adding new tag types to the database

ShifaSZ (approved):

PR #1817, opened 2021-04-28
add external solve status

hanars:


seqr/models.py
```diff
@@ -241,6 +241,7 @@ class Family(ModelWithGUID):
         ('S_kgfp', 'Solved - known gene for phenotype'),
         ('S_kgdp', 'Solved - gene linked to different phenotype'),
         ('S_ng', 'Solved - novel gene'),
+        ('ES', 'External solve'),
```
ShifaSZ, line 244: To be consistent with other choices, should `ES` be `Es`?

ShifaSZ (changes requested):
Looks good except for one sugguestion.

seqr/models.py
hanars: Using capitals is more consistent with every other model field. The lower case in this list are actually the antipattern here and I would get rid of them if it wasn't such a pain

ShifaSZ (approved):

PR #1819, opened 2021-04-28
Dashboard download client side

hanars:
Currently we have a server-side endpoint to download the project dashboard data, but moving the download code to be based off of the client side table is a) more consistent with how almost every download in seqr works and b) easier to maintain and less fragile


ShifaSZ (approved):
Awesome work.

PR #1826, opened 2021-04-30
fix allele display

hanars:


ShifaSZ (approved):
Can't find any issues.

PR #1827, opened 2021-04-30
Fix local file IGV

hanars:


ShifaSZ:
The issue connected hasn't been solved yet. How can this PR help the issue?

hanars:
> The issue connected hasn't been solved yet. How can this PR help the issue?

I don't now how to respond to this. This code change fixes IGV so it works locally, which is the problem reported in the linked issue. The linked issue is not going to be marked as "solved" until this code is merged to master and I get confirmation that the fix works for them, but that doesn't somehow mean the PR doesn't work... we never close out tickets while a PR is open...

seqr/utils/file_utils.py
```diff
@@ -34,19 +38,20 @@ def file_iter(file_path, byte_range=None, raw_content=False):
...
-                f.seek(byte_range[0])
-                for line in f:
-                    if f.tell() < byte_range[1]:
-                        yield line
-                    else:
-                        break
-            else:
-                for line in f:
-                    yield line
+            for line in f:
```
bw2, line 53: This seems like a suboptimal way to read binary files. Depending on the file, new-line chars may be very unevenly distributed in the file, so the "line" size will be very uneven. Reading a fixed number of bytes at a time should provide more consistent i/o speed.

hanars: this hasn't changed as part of this PR though so I'm leaving it as-is for now

ShifaSZ: It does return `lines` with large size differences. But, they will be used by `StreamingHttpResponse` as a byte stream in seqr's case. So, it seems doesn't matter.

seqr/views/apis/igv_api_tests.py
```diff
@@ -37,27 +37,27 @@ def test_proxy_google_to_igv(self, mock_subprocess):
...
-        response = self.client.get(url, HTTP_RANGE='bytes=100-200')
+        response = self.client.get(url, HTTP_RANGE='bytes=100-250')
         self.assertEqual(response.status_code, 206)
-        self.assertListEqual([val for val in response.streaming_content], STREAMING_READS_CONTENT[:2])
-        mock_open.assert_called_with('/project_A/sample_1.bai', 'rb')
-        mock_file.seek.assert_called_with(100)
+        self.assertListEqual([val for val in response.streaming_content], STREAMING_READS_CONTENT)
+        mock_subprocess.assert_called_with(
+            'dd skip=100 count=150 bs=1 if=/project_A/sample_1.bai',
+            stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
```
ShifaSZ, line 53: add an assertion for `mock_subprocess.stdout.__iter__` is called?

ShifaSZ (changes requested):
Awesome fixes for reading IVG files. I have a suggestion for the unit test.

seqr/views/apis/igv_api_tests.py
hanars: I want this test to be consistent with the test above, which also calls subprocess. The mocked return value and checking `response.streaming_content` is already checking that it is being correctly iterated over so adding that line doesn't actually add anything

ShifaSZ (approved):

PR #1835, opened 2021-05-04
Inactive user detection

hanars:


deploy/docker/seqr/entrypoint.sh
```diff
@@ -67,10 +67,11 @@ fi
...
-0 */4 * * * /usr/local/bin/python /seqr/manage.py run_postgres_database_backup --bucket $DATABASE_BACKUP_BUCKET --postgres-host $POSTGRES_SERVICE_HOSTNAME --deployment-type $DEPLOYMENT_TYPE >> /var/log/cron.log 2>&1
-0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_omim --omim-key $OMIM_KEY >> /var/log/cron.log 2>&1
-0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_human_phenotype_ontology >> /var/log/cron.log 2>&1
+0 0 * * * /usr/local/bin/python /seqr/manage.py run_settings_backup --bucket $DATABASE_BACKUP_BUCKET --deployment-type $DEPLOYMENT_TYPE >> /proc/1/fd/1 2>&1
+0 */4 * * * /usr/local/bin/python /seqr/manage.py run_postgres_database_backup --bucket $DATABASE_BACKUP_BUCKET --postgres-host $POSTGRES_SERVICE_HOSTNAME --deployment-type $DEPLOYMENT_TYPE >> /proc/1/fd/1 2>&1
+0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_omim --omim-key $OMIM_KEY >> /proc/1/fd/1 2>&1
+0 0 * * 0 /usr/local/bin/python /seqr/manage.py update_human_phenotype_ontology >> /proc/1/fd/1 2>&1
+0 12 * * 1 /usr/local/bin/python /seqr/manage.py detect_inactive_privileged_users >> /proc/1/fd/1 2>&1
```
ShifaSZ, line 74: Is there any reason to make these changes?

seqr/management/commands/detect_inactive_privileged_users.py
```diff
@@ -0,0 +1,59 @@
...
+WARNING_TEMPLATE = """
+Hi there {} - 
+
+You have not logged in to seqr in {} days. Unless you log in within the next {} days, your account will be deactivated.
+"""
+
+DEACTIVATE_TEMPLATE = """
+Hi there {} - 
+
+You have not logged in to seqr in {} days, and therefore your account has been deactivated.
```
ShifaSZ, line 19: Don't you think using named placeholders is better?

ShifaSZ (changes requested):
Overall looks good. But I have two questions.

deploy/docker/seqr/entrypoint.sh
hanars: yes it adds a cron job so this manage command gets run once a week and redirects the output to stackdriver so we can actually see the cron logs

seqr/management/commands/detect_inactive_privileged_users.py
hanars: not really, only when theres a lot of variables or the usage is confusing. In simple cases I think named placeholders just add unnecessary words.  If you think its confusing without named placeholders here I can change it, but I think its fine as is

ShifaSZ (approved):

PR #1836, opened 2021-05-05
Changed the way to get samples from an ES index.

ShifaSZ:
Currently, seqr gets samples from the `samples_num_alt_1` field for `VARIANT` indices and from the `samples` field for `SV` indices. To make it ready for adding genome SVs which don't have the `samples` field, this PR change it to get samples from the first field in the list of ['samples', 'samples_num_alt_1'] and it exists in the index.

### DIFF
--- a/seqr/views/apis/dataset_api_tests.py
+++ b/seqr/views/apis/dataset_api_tests.py
@@ -14,6 +14,13 @@ INDEX_NAME = 'test_index'
 SV_INDEX_NAME = 'test_new_sv_index'
 NEW_SAMPLE_TYPE_INDEX_NAME = 'test_new_index'
 ADD_DATASET_PAYLOAD = json.dumps({'elasticsearchIndex': INDEX_NAME, 'datasetType': 'VARIANTS'})
+MAPPING_PROPS_NO_SAMPLES = {
+    "samples_num_alt_1": {"type": "keyword"},
+}
+MAPPING_PROPS_WITH_SAMPLES = {
+    "samples": {"type": "keyword"},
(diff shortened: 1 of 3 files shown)


seqr/views/apis/dataset_api_tests.py
```diff
@@ -14,6 +14,13 @@
 SV_INDEX_NAME = 'test_new_sv_index'
 NEW_SAMPLE_TYPE_INDEX_NAME = 'test_new_index'
 ADD_DATASET_PAYLOAD = json.dumps({'elasticsearchIndex': INDEX_NAME, 'datasetType': 'VARIANTS'})
+MAPPING_PROPS_NO_SAMPLES = {
+    "samples_num_alt_1": {"type": "keyword"},
+}
+MAPPING_PROPS_WITH_SAMPLES = {
+    "samples": {"type": "keyword"},
+    "samples_num_alt_1": {"type": "keyword"},
```
hanars, line 22: you will never have an index with both of these fields, so this is not realistic data. you should just have `samples` here

hanars (changes requested):
one small change for the unit test but otherwise looks great!

ui/shared/utils/constants.js
```diff
@@ -550,6 +550,42 @@ const ORDERED_VEP_CONSEQUENCES = [
     value: 'DUP',
     group: VEP_GROUP_SV,
   },
+  {
```
hanars, line 553: I don't think you meant to include these changes as part of this PR

hanars (changes requested):

PR #1846, opened 2021-05-13
Add structural variant types to search annotation filters.

ShifaSZ:


ui/shared/utils/constants.js
```diff
@@ -550,6 +550,42 @@ const ORDERED_VEP_CONSEQUENCES = [
     value: 'DUP',
     group: VEP_GROUP_SV,
   },
+  {
+    description: 'A translocation variant',
+    text: 'Translocation',
+    value: 'BND',
```
hanars, line 556: `BND` seems like a weird abbreviation for a translocation, so just double checking that this isn't a typo

ui/shared/utils/constants.js
```diff
@@ -550,6 +550,42 @@ const ORDERED_VEP_CONSEQUENCES = [
...
+    group: VEP_GROUP_SV,
+  },
+  {
+    description: 'A copy number polymorphism variant',
+    text: 'Copy Number Polymorphism',
+    value: 'CNV',
+    group: VEP_GROUP_SV,
+  },
+  {
+    description: 'A Complex SV',
```
hanars, line 566: should probably expand "SV" to "structural variant" in the description

ui/shared/utils/constants.js
```diff
@@ -550,6 +550,42 @@ const ORDERED_VEP_CONSEQUENCES = [
...
+    text: 'Complex SV',
+    value: 'CPX',
+    group: VEP_GROUP_SV,
+  },
+  {
+    description: 'A reciprocal chromosomal translocation',
+    text: 'Reciprocal chromosomal translocation',
```
hanars, line 573: this is a very long name to show - is this actually the best human readable name for this type of variant or is there something shorter we can use?

hanars (commented):
a couple small questions before we merge. Also, if you don't mind I think it would be helpful to me if for the UI changes for the genome SVs you added a screenshot to the PR description

ui/shared/utils/constants.js
ShifaSZ: It is from the VCF. I checked online and couldn't find why `BND` stands for a translocation. But it is not a type.

ui/shared/utils/constants.js
ShifaSZ: I found a [literature](https://www.sciencedirect.com/topics/biochemistry-genetics-and-molecular-biology/reciprocal-chromosome-translocation) uses `RT` as the abbreviation for `Reciprocal Translocation`.  Not sure if it can be used for `Reciprocal chromosomal translocation`.

hanars: I'll ask Anne

hanars: Confirmed `Reciprocal Translocation` is the same, so lets use that for the text and leave the long form in the description

hanars:
I would still like to see a screenshot before I approve

ShifaSZ:
Current screenshot looks like below figure: 
image

hanars (approved):

PR #1858, opened 2021-05-19
Added gnomAD genome sv frequency search filter.

ShifaSZ:
The frequency search filters have been changed to the layout below.
image

The PR is still a preliminary one because I'm not sure if the updates to the backend are correct or not.


seqr/utils/elasticsearch/constants.py
```diff
@@ -99,6 +99,9 @@
     'gnomad_genomes': {
         'filter_AF': ['gnomad_genomes_FAF_AF', 'gnomad_genomes_AF_POPMAX_OR_GLOBAL'],
     },
+    'gnomad_genome_svs': {
+        'filter_AF': ['gnomad_svs_AF'],
```
hanars, line 103: `filter_AF` only is relevant if we want to use a different field for filtering than for displaying AF, which is not the case here. You want to change this config to just `'gnomad_svs': {}`

ui/shared/components/panel/search/constants.js
```diff
@@ -324,6 +324,12 @@ export const FREQUENCIES = [
     homHemi: false,
     labelHelp: 'Filter by allele count (AC) or allele frequency (AF) in TOPMed',
   },
+  {
+    name: 'gnomad_genome_svs',
```
hanars, line 328: as mentioned above, this should just be `gnomad_svs`. This  data in gnomad isn't genome specific so you should take out the word genome from the description as well

ui/shared/components/panel/search/constants.js
```diff
@@ -324,6 +324,12 @@ export const FREQUENCIES = [
...
+    label: 'gnomAD genome SVs',
+    homHemi: false,
+    labelHelp: 'Filter by allele count (AC) among gnomAD genomes, or by the allele frequency (for biallelic sites) or copy-state frequency (for multiallelic sites) of an overlapping event in gnomad',
```
hanars, line 331: Use instead:
"Filter by site frequency (AF) among gnomad SVs"

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -186,9 +186,15 @@ const PanelContent = React.memo(({ name, fields, fieldProps, helpText, fieldLayo
       {helpText && <i>{helpText} <VerticalSpacer height={20} /></i>}
       <Form.Group widths="equal">
         <Form.Field width={2} />
-        {fieldLayout ? fieldLayout(fieldComponents) : fieldComponents}
+        {fieldLayout ? fieldLayout(fieldComponents) : fieldComponents.slice(0, 6)}
         <Form.Field width={2} />
       </Form.Group>
+      {fieldComponents.length > 6 && !fieldLayout &&
```
hanars, line 192: don't modify the default behavior here. Instead, pass `fieldLayout` to the frequency field and explicitly style the layout. `fieldLayout` if a function that will be called with list of `fieldComponents` and then returns the layout you want to render

hanars (changes requested):
The callset AF filters should be aligned to the right

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
ShifaSZ: New layout:
image

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -122,6 +122,12 @@ export const annotationFieldLayout = (annotationGroups, hideOther) => fieldCompo
   ) : null,
 ].filter(fields => fields)
 
+const freqFieldLayout = fieldComponents => [
+  <Grid.Row className="field">{[fieldComponents[0], fieldComponents[6]]}</Grid.Row>,
+  <Grid.Row className="field">{[fieldComponents[1], fieldComponents[7]]}</Grid.Row>,
+  ...fieldComponents.slice(2, 6),
```
hanars, line 128: This is a little confusing to read. It looks like it should be putting components 0 and 6 on the top line, then 1 and 7 on the line after that, and then 2-5 on a third row. But based on the screenshot, it looks like 0-5 are on the first row and then 6 and 7 are on the second row, so I'm not sure what happening here. can you restructure this or, if you really can't, add comments to explain how this is expected to render? 

also, in general, if you are rendering lists of components you should add a unique`key` prop to each component in the list. You wouldn't need to do anything to the `...fieldComponents.slice(2, 6)` as they already have keys, but for the new row components you would need to add them. 

also, per my last review, the two "callset" filters should be aligned to the right

hanars (changes requested):

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
ShifaSZ: I have two choices to layout the components into two rows: two rows with multiple components in each row, or one row with multiple elements some of which have two rows.  The first option has difficulties in managing the widths of the components. The second option can utilize the option of `widths="equal"` of the parent component (which is `Form.Group`) to ensure all elements have the same width. Components 0 and 6 are put in the first column and components 1 and 7 in the second column.

ShifaSZ: New layout:
image

hanars: I see. So what you actually want to do is put the entire layout inside a Field so it uses the whole width available, and then add nested form groups to control their width and do equal layout
```
<Form.Field>
    <Form.Group widths="equal">
      {fieldComponents.slice(0, MAX_COMPONENTS_PER_ROW)}
    </Form.Group>,
    <Form.Group widths="equal">
      // add empty fields to pad out the second row so the "equal" widths are the same
      {[...Array(fieldComponents.length - MAX_COMPONENTS_PER_ROW)].map(() => <Form.Field />)}
      {fieldComponents.slice(MAX_COMPONENTS_PER_ROW)}
    </Form.Group>
</Form.Field>
```

ShifaSZ: That works but may not be perfect because the parent `Form.Group` in `PanelContent` has another two stuffing `Form.Field` components at both sides of the row. These stuff fields occupy bigger space when the `fieldComponents` are encapsulated into one component.
You can see the result as below:
image
If you think the size of the stuffing fields is acceptable, I can go with the code you proposed.

hanars: those fields on the side are there intentionally for padding, the code I proposed I tried locally and looks the way I want it to

hanars (approved):

PR #1861, opened 2021-05-19
Fix igv CSP/cross origin issues

hanars:


seqr/views/apis/igv_api_tests.py
```diff
@@ -169,13 +169,26 @@ def test_add_alignment_sample(self, mock_local_file_exists, mock_subprocess):
...
     @responses.activate
     def test_igv_genomes_proxyy(self):
-        url = reverse(igv_genomes_proxy)
+        url_path = 'org.genomes/foo?guery=true'
+        url = reverse(igv_genomes_proxy, args=[url_path])
 
         expected_body = {'genes': ['GENE1', 'GENE2']}
         responses.add(
-            responses.GET, 'https://s3.amazonaws.com/igv.org.genomes/genomes.json', status=200,
+            responses.GET, 'https://s3.amazonaws.com/igv.org.genomes/foo?guery=true', match_querystring=True,
```
ShifaSZ, line 177: Do you mean `query` in `foo?guery=true`?

ShifaSZ (changes requested):
It is orverall good to my knowledge. There could be a nitpick question.

seqr/views/apis/igv_api_tests.py
hanars: yeah good point

ShifaSZ (approved):

PR #1869, opened 2021-05-20
validate end of interval

hanars:


seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1260,10 +1260,10 @@ def test_filtered_get_es_variants(self):
         results_model = VariantSearchResults.objects.create(variant_search=search_model)
 
         # Test invalid locations
-        search_model.search['locus'] = {'rawItems': 'chr27:1234-5678, ENSG00012345', 'rawVariantItems': 'chr2-A-C'}
+        search_model.search['locus'] = {'rawItems': 'chr27:1234-5678,2:40-400000000, ENSG00012345', 'rawVariantItems': 'chr2-A-C'}
         with self.assertRaises(InvalidSearchException) as cm:
             get_es_variants(results_model, sort='cadd', num_results=2)
-        self.assertEqual(str(cm.exception), 'Invalid genes/intervals: chr27:1234-5678, ENSG00012345')
+        self.assertEqual(str(cm.exception), 'Invalid genes/intervals: chr27:1234-5678, chr2:40-400000000, ENSG00012345')
```
ShifaSZ, line 1266: How about also add a test for an invalid chromosome?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_utils_tests.py
ShifaSZ: Okay. I thought the exception message should have something like `Invalid chromosome` but actually, that exception is caught internally.

ShifaSZ (approved):

PR #1876, opened 2021-05-21
Excluded tag reason

hanars:
This code updates the models so its possible to edit tag types in the django admin, which is the preferred way of making database changes. After this is released, I will use the admin to update the metadata_title for this tag type


ShifaSZ (approved):
Can't find any issues.

PR #1877, opened 2021-05-21
do not fail uptime check for kibana or redis

hanars:


seqr/views/status.py
```diff
@@ -26,7 +27,7 @@ def status_view(request):
     try:
         redis.StrictRedis(host=REDIS_SERVICE_HOSTNAME, socket_connect_timeout=3).ping()
     except Exception as e:
-        dependent_services_ok = False
+        secondary_services_ok = False
```
ShifaSZ, line 30: Is Redis also a secondary service? It looks like a dependent one.

seqr/views/status.py
```diff
@@ -43,11 +44,11 @@ def status_view(request):
         if resp.status >= 400:
             raise ValueError('Error {}: {}'.format(resp.status, resp.reason))
     except Exception as e:
-        dependent_services_ok = False
+        secondary_services_ok = False
         logger.error('Kibana connection error: {}'.format(str(e)))
```
ShifaSZ, line 48: Don't you think it is better to change the log from an `error` to a `warning`?

ShifaSZ (changes requested):

seqr/views/status.py
hanars: no I need to be alerted if these services are down, its really bad if it happens

seqr/views/status.py
hanars: seqr runs fine if redis is down, just a bit slower

ShifaSZ (approved):

PR #1893, opened 2021-06-02
fix predictor sorts for keyword types

hanars:
Fixes bugs introduced in https://github.com/broadinstitute/seqr/pull/1867/files


ShifaSZ (approved):
Can't find any issues.

PR #1898, opened 2021-06-03
Managed postgres

hanars:
The switches seqr from using postgres deployed off a docker image in a kubernetes pod to a managed google Cloud SQL database. This has several advantages, including automated backups, automated restoration if there are any outages, automated minor version and security patches, etc. 

The big changes in this PR are:
- changing the `deploy_postgres` helper to deploy the cloud SQL instance
- changing the seqr kubernetes config to connect to the cloud SQL instance
- deleting all the docker/ kubernetes code for potgres
- cleaning up kubernetes helpers for postgres (i.e. deleting and port forwarding)
- changing how local installs set up their reference data. Currently, we store a pg_dump of our reference database in a public bucket, this changes it so that local installs actually download all the latest reference data from their sources

Note that this PR won't be merged until after I get back from my vacation next week, but I'm putting it up now so I can iterate on review feedback before I go


deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -250,25 +250,52 @@ def deploy_linkerd(settings):
...
+        '--assign-ip',
+        '--backup',
+        '--maintenance-release-channel=production', '--maintenance-window-day=SUN', '--maintenance-window-hour=5',
+        '--require-ssl',
+        '--retained-backups-count=30',
+        '--storage-auto-increase',
+    ]), errors_to_ignore=['already exists'])
+
+    seqr_db_backup = settings.get('RESTORE_SEQR_DB_FROM_BACKUP')
+    reference_data_db_backup = settings.get('RESTORE_REFERENCE_DB_FROM_BACKUP')
```
sjahl, line 274: If I'm understanding correctly, these variables affect whether or not we perform a restore operation. I think maybe it might make sense to name the variables with a `_restore` suffix? My brain jumps to the conclusion that we're performing a backup with the current prefix. Not a huge deal, though.

hanars: the variables contain the path to the backup file if we want to do a restore, and are null if we shouldn't do a restore. I agree that the current combination of settings variables called `RESTORE_X_DB_FROMBACKUP` being assigned to local variables called `X_db_backup` is confusing, but I think changing the local variable name would make the below code of `'gcloud sql import sql', sql_instance_name, X_db_backup_restore, '--database=X_db',` a little weird looking. But if you think thats still readable I'm happy to change it

ShifaSZ:
Don't you want to add the `CachedOmimReferenceDataHandler` in a separate PR? I think it will be cleaner.

reference_data/management/commands/update_omim.py
```diff
@@ -64,18 +64,41 @@
...
 class OmimReferenceDataHandler(ReferenceDataHandler):
 
     model_cls = Omim
     url = "https://data.omim.org/downloads/{omim_key}/genemap2.txt"
 
-    def __init__(self, omim_key=None, **kwargs):
+    def __init__(self, omim_key=None, cache_parsed_records=False, **kwargs):
+        """Init OMIM handler."""
         if not omim_key:
             raise CommandError("omim_key is required")
```
ShifaSZ, line 97: Since omin_key is required, a `None` default value is not necessary.

reference_data/management/tests/update_all_reference_data_tests.py
```diff
@@ -111,10 +115,30 @@ def test_skip_all_update_reference_data_command(self):
...
         self.mock_logger.info.assert_called_with("Done")
 
+    def test_cached_omim_update_reference_data_command(self):
+        self.mock_cached_omim.return_value = 'cached_omim'
+
+        call_command(
+            'update_all_reference_data', '--use-cached-omim', *SKIP_ARGS)
+
+        self.mock_cached_omim.assert_called_with()
+        self.mock_update_records.assert_called_with('cached_omim')
```
ShifaSZ, line 130: Do you want to add tests for the cached data?

ShifaSZ (changes requested):
There is a minor recommendation.

hanars:
> Don't you want to add the `CachedOmimReferenceDataHandler` in a separate PR? I think it will be cleaner.

No I need this to all go in as one PR as its necessary for local installs to work

reference_data/management/commands/update_omim.py
hanars: the signature for the class inheritance requires this to be a keyword arg and not a regular arg, and thus a default is required

ShifaSZ (approved):

PR #1901, opened 2021-06-04
Update the gnomAD link for genome SVs

ShifaSZ:


seqr/utils/elasticsearch/constants.py
```diff
@@ -108,6 +108,7 @@
     'AN': {},
     'Hom': {},
     'Hemi': {},
+    'ID': {'format_value': str, 'default_value': None},
```
ShifaSZ, line 111: There is a field `sv_callset_het` for genome SVs which is still not been loaded into the search response. Should I add one more field to `POPULATION_FIELD_CONFIGS`?

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -56,7 +62,7 @@ const FreqSummary = React.memo((props) => {
   const { field, fieldTitle, variant, urls, queryParams, acDisplay, titleContainer, precision = 2 } = props
   const { populations = {}, chrom } = variant
   const population = populations[field] || {}
-  if (population.af === null || population.af === undefined) {
+  if (population.af === null || population.af === undefined) { //When population is "gnomad_svs", how can I remove the frequency for null "gnomad_svs_AF"?
```
ShifaSZ, line 65: When `gnomad_svs_AF` is null, it is parsed as `0` after search. Therefore, the `population.af` is `0`, neither `null` nor `undefined`. So, it always displays `gnomAD SVs 0.0` which doesn't look right. Should I remove the display when `gnomad_svs_AF` is null or doesn't exist?

hanars: so it should only be 0 for indices that have that field populated for some variants, as based on the index mapping. So in an exome SV index or a SNV index, this should be left out entirely/ null. However, if there are any variants that have this populated (like in a genome SV index) than we do expect this to be 0! The fact that some variants in the index are annotated with gnomad SVs means we tried to annotatate everything in that index, and if we tried and there wasn't anything that means its not in gnomad, and has an AF of 0. In other words, what you are seeing is the expected behavior

seqr/views/utils/test_utils.py
```diff
@@ -853,14 +853,14 @@ def call_request_json(self, index=-1):
...
-        'gnomad_svs': {'ac': 0, 'af': 0, 'an': 0, 'filter_af': None, 'hemi': 0, 'hom': 0},
+        'callset': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'g1k': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'gnomad_genomes': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'exac': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'gnomad_exomes': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'topmed': {'an': None, 'ac': None, 'hom': None, 'af': None, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'sv_callset': {'an': 10088, 'ac': 7, 'hom': None, 'af': 0.000693825, 'hemi': None, 'filter_af': None, 'het': None, 'id': None},
+        'gnomad_svs': {'ac': 0, 'af': 0, 'an': 0, 'filter_af': None, 'hemi': 0, 'hom': 0, 'het': 0, 'id': None},
```
hanars, line 863: lets add an ID to the test data so we can test that it properly is handling the string value (I am very confident that this works now, but want to add the test so we don't ever break it by accident in the future)

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -19,7 +19,13 @@ const FreqLink = React.memo(({ urls, value, displayValue, variant, queryParams,
     genomeVersion = variant.liftedOverGenomeVersion
   }
 
-  const path = getPath({ chrom, pos, genomeVersion, variant, value })
+  let path
```
hanars, line 22: This would change the link for ALL populations instead of just `gnomad_svs`. You wouldn't see that bug yet since currently gnomad svs are the only frequencies with links for genomne SVs, but if we ever did add another population to these variants the links would all be wrong. `FreqLink` is created within a population-specific `FreqSummary`, and you need to make the change at that point and not here so that this change will apply only to the needed population. Currently in `FreqSummary` we use `population.af` as the `value` and `population.filter_af` as the `displayValue`. I think you should change the logic so it can use `population.id` as the value if it exists and `population.af` as the display value if need be. So inside `FreqSummary` you would change the following from what it currently is
```
const value = population.af > 0 ? population.af.toPrecision(precision) : '0.0'
const filterValue = population.filter_af > 0 ? population.filter_af.toPrecision(precision) : null
```
to 
```
const afValue = population.af > 0 ? population.af.toPrecision(precision) : '0.0'
const value = population.id ? population.id.replace('gnomAD-SV_v2.1_', '') : afValue
const displayValue = population.filter_af > 0 ? population.filter_af.toPrecision(precision) : afValue
```
Then you also need to change the implementation of `getFreqLinkPath` to use the `value` plain if its a string and not a number:
```
const floatValue = parseFloat(value, 10)
const isRegion = floatValue <= 0
let coords
if (isNaN(floatValue)) {
  coords = value
} else if (isRegion) {
  ...
```

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
```diff
@@ -108,6 +108,8 @@
     'AN': {},
     'Hom': {},
     'Hemi': {},
+    'Het': {},
```
hanars, line 111: you've added this to the parsed data but it doesn't look like we are displaying it anywhere?

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
ShifaSZ: Ooh... I thought it could be display automatically by the frontend.
Updated now, and it is displayed as below:
image

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -249,7 +249,7 @@ const BaseVariantGene = React.memo(({ geneId, gene, variant, compact, showInline
 
   let summaryDetail
   if (compact) {
-    summaryDetail = showInlineDetails ? [geneDetails, geneConsequence] : geneConsequence
+    summaryDetail = showInlineDetails ? <span>{geneDetails} {geneConsequence}</span> : geneConsequence
```
ShifaSZ, line 252: It is changed to remove a warning for requiring keys in the list.

seqr/views/utils/test_utils.py
ShifaSZ: Just see the comment. I will add the test.

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Sounds a good solution. I would like to learn what's different between `filter_af` and `af` when we meet next time. :)

hanars: sure! filter_af is only populated for gnomad in the latest release. Their documentation for it is here: https://blog.limbus-medtec.com/how-to-use-gnomad-v2-1-for-variant-filtering-d7d2a7ee710a

Essentially, they create a different allele frequency calculation that takes into account subpopulations and context, which differs slightly from the uncorrected allele frequency. I don't really understand all the math and biology involved, but the summary is if there is a filter_af available its always better to filter variants based on that than on the plain af. Because of that, we want to display that value in seqr because thats what the person actually searched on. So for example, if a variant had an af of .02 and a filtering af of 0.009 and somebody applied an AF filter of .01, the variant would be returned because  the filtering af is below the threshold. But if we then showed the regular af, the person would see a variant with af of 0.02, which is above the threshhold they searched for, and think search was broken. So we show filter_Af if we have it, and then we show the plain af if you hover the frequency

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -72,15 +77,18 @@ const FreqSummary = React.memo((props) => {
               urls={urls}
               queryParams={queryParams}
               value={value}
-              displayValue={filterValue}
+              displayValue={displayValue}
               variant={variant}
               getPath={getFreqLinkPath}
-            /> : (filterValue || value)
+            /> : (displayValue || value)
```
hanars, line 83: since `displayValue` is now always set we don't need the `|| value` anymore

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -40,9 +40,13 @@ FreqLink.propTypes = {
 }
 
 const getFreqLinkPath = ({ chrom, pos, variant, value }) => {
-  const isRegion = parseFloat(value, 10) <= 0
+  const floatValue = parseFloat(value, 10)
+  const isRegion = floatValue <= 0
   let coords
-  if (isRegion) {
+  // eslint-disable-next-line no-restricted-globals
```
hanars, line 46: From this article (https://stackoverflow.com/questions/46677774/eslint-unexpected-use-of-isnan) it looks like if we use `Number.isNaN` this will work without throwing a linter warning. Suppressing linter warnings is totally okay to do, but if there is an easy workaround like this one its better to use it. But if that doesn't actually work than you should leave this as is

hanars (changes requested):

hanars (approved):

PR #1906, opened 2021-06-15
Added discovery_notes column.

ShifaSZ:
It seems that the "discovery_note" logic should work for both UI and AnVIL Report. The changes in this draft PR are for AnVIL Report only.

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -344,13 +344,20 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
                 potential_com_het_gene_variants[gene_id].append(variant)
         for gene_id, comp_het_variants in potential_com_het_gene_variants.items():
             if len(comp_het_variants) > 1:
+                discovery_notes = ''
                 if len(comp_het_variants) > 2:
-                    variant['discovery_notes'] = 'The following variants are part of the multinucleotide variant {}: {}'.format(
-                        variant['variantId'], ', '.join([v['variantId'] for v in comp_het_variants ]))
-                else:
-                    variant['inheritance_models'] = {'AR-comphet'}
+                    if variant.get('svType'):
+                        discovery_notes = 'The following variants are part of the complex structural variant {}: {}'.format(
+                            variant['svName'], ', '.join([v['svName'] for v in comp_het_variants]))
+                    else:
+                        discovery_notes = 'The following variants are part of the multinucleotide variant {}: {}'.format(
+                            variant['variantId'], ', '.join([v['variantId'] for v in comp_het_variants]))
                 main_gene_ids = set()
                 for variant in comp_het_variants:
+                    if discovery_notes:
+                        variant['discovery_notes'] = discovery_notes
+                    else:
+                        variant['inheritance_models'] = {'AR-comphet'}
                     if variant['main_transcript']:
                         main_gene_ids.add(variant['main_transcript']['geneId'])
                     else:
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -331,6 +331,7 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
     gene_ids = set()
     for family_guid, saved_variants in saved_variants_by_family.items():
         potential_com_het_gene_variants = defaultdict(list)
+        potential_mnvs = defaultdict(lambda: defaultdict(list))
         for variant in saved_variants:
             variant['main_transcript'] = _get_variant_main_transcript(variant)
             if variant['main_transcript']:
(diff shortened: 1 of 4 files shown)


seqr/views/apis/report_api.py
```diff
@@ -344,9 +344,13 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
                 potential_com_het_gene_variants[gene_id].append(variant)
         for gene_id, comp_het_variants in potential_com_het_gene_variants.items():
             if len(comp_het_variants) > 1:
+                if len(comp_het_variants) > 2:
+                    variant['discovery_notes'] = 'The following variants are part of the multinucleotide variant {}: {}'.format(
+                        variant['variantId'], ', '.join([v['variantId'] for v in comp_het_variants ]))
+                else:
+                    variant['inheritance_models'] = {'AR-comphet'}
```
hanars, line 351: this used to be inside the `for variant in comp_het_variants:` loop. This change needs to be inside that loop as well. Since the discovery note is the same for all the variants, I think I would still create the note here, and then set it below in the loop
```
discovery_notes = None
if len(comp_het_variants) > 2:
    discovery_notes = 'The following ...'.format(...)
...
for variant in comp_het_variants:
    if discovery_notes:
        variant['discovery_notes'] = discovery_notes
    else:
        variant['inheritance_models'] = {'AR-comphet'}
```

hanars (changes requested):

seqr/views/apis/report_api.py
ShifaSZ: In the code above, I think the variants in `comp_het_variants` belong to a gene. Do they also belong to an MNV? I'm trying to find an MNV in the `comp_het_variants` as the parent variant and the rest of variants as the nested variants of the MNV. But when I ran the code with project `R0384_rare_genomes_project_gen`, I got a `comp_het_variants` with these variants: [19-1912633-G-T, 19-1912632-GC-TT, 19-1912632-GC-TT]. It seems missing `19-1912634-C-T` and duplicating `19-1912632-GC-TT`. Does it have the right variants?

ShifaSZ: Or, should I check the tags for the variants in the `comp_het_variants` to find MNVs?

hanars: Ah I see the problem - the issue is my instructions to you were wrong. Sadly, we can't just repurpose `comp_het_variants` here. What we need to do is where we define `potential_com_het_gene_variants` we also need to define another dict for `potential_mnvs`, which will be defined as variants in the same genes with the same tag, and then explicitly check those. So it will be something like
```
potential_com_het_gene_variants = defaultdict(list)
potential_mnvs = defaultdict(lambda: defaultdict(list))
for variant in saved_variants:
    variant['main_transcript'] = _get_variant_main_transcript(variant)
    ...
    for gene_id in variant_json['transcripts'].keys():
        for tag in variant_json['discovery_tag_names']:
            potential_mnvs[gene_id][tag].append(variant_json)
mnv_genes = set()
for gene_id, variants_by_tag in potential_mnvs.items():
    mnvs = next((variants for variants in variants_by_tag.values() if len(variants) > 2), None)
    if not mnvs:
        continue
    mnv_genes.add(gene_id)
    discovery_note = <construct note from mnvs>
    for variant in mnvs:
        variant['discovery_note'] = discovery_note
for gene_id, comp_het_variants in potential_com_het_gene_variants.items():
    if gene_id in mnv_genes:
        continue
    if len(comp_het_variants) > 1:
    ...
```

hanars: let me know if you would rather meet to discuss

ShifaSZ: Thanks for your detailed instructions. I prefer to do it by myself first and then we meet later so that I can understand better the logic among the data. Maybe we can meet sometime next week.

seqr/views/apis/report_api.py
```diff
@@ -759,6 +781,9 @@ def _get_saved_discovery_variants_by_family(variant_filter, parse_json=False):
         if parse_json:
             parsed_variant = variant_by_guid[saved_variant.guid]
             parsed_variant['discovery_tag_names'] = {vt.variant_tag_type.name for vt in saved_variant.discovery_tags}
+            parsed_variant['discovery_tag_guids_by_name'] = {
```
hanars, line 784: so thinking about this more, this is actually some unneccessary duplication. Although its a slightly larger code change, I think we should actually replace `discovery_tag_names` with `discovery_tag_guids_by_name`. We would create it more cleanly than you do here, by doing
```
parsed_variant['discovery_tag_guids_by_name'] = {vt.variant_tag_type.name: vt.guid for vt in saved_variant.discovery_tags}
```
and then everywhere that currently uses `variant['discovery_tag_names']` would instead use `variant['discovery_tag_guids_by_name'].keys()`

seqr/views/apis/report_api.py
```diff
@@ -342,7 +343,27 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
             variant['inheritance_models'] = inheritance_models
             for gene_id in potential_compound_het_gene_ids:
                 potential_com_het_gene_variants[gene_id].append(variant)
+            for tag in variant['discovery_tag_names']:
+                for guid in variant.get('discovery_tag_guids_by_name', {}).get(tag):
+                    potential_mnvs[tag][guid].append(variant)
```
hanars, line 348: the `guid` will always have the same tag name, so theres no need to do the nested dict - you can just do a regular `defaultdict(list)` and then here leave out the tag check and do `potential_mnvs[guid].append(variant)`

seqr/views/apis/report_api.py
```diff
@@ -342,7 +343,27 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
...
+        mnv_genes = set()
+        for variants_by_tag_guid in potential_mnvs.values():
+            mnvs = next((variants for variants in variants_by_tag_guid.values() if len(variants) > 2), None)
+            if not mnvs:
+                continue
+            parent_mnv = next((v for v in mnvs if not v.get('populations')), mnvs[0])
+            nested_mnvs = [v for v in mnvs if v.get('populations')]
```
hanars, line 355: this will be cleaner if you instead explicitly check that it isn't the parent mnv (i.e. `v['variantId'] != parent_mnv['variantId']`)

seqr/views/apis/report_api.py
```diff
@@ -342,7 +343,27 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
...
+            mnv_genes |= {gene_id for variant in nested_mnvs for gene_id in variant['transcripts'].keys()}
+            discovery_notes = 'The following variants are part of the complex structural variant {}: {}'.format(
```
hanars, line 357: this is a bit hard to read, I think it would be better to use an additional helper method for getting the IDs:
```
def _get_nested_variant_name(variant):
    return _get_sv_name(variant) if variant.get('svType') else variant['variantId']
...
discovery_notes = 'The following variants are part of the {variant_type} variant {parent}: {nested}'.format(
    variant_type='complex structural' if parent_mnv.get('svType') else 'multinucleotide',
    parent=_get_nested_variant_name(parent_mnv), 
    nested=', '.join([_get_nested_variant_name(v) for v in nested_mnvs]))
)

```

seqr/views/apis/report_api_tests.py
```diff
@@ -375,15 +375,15 @@ def test_anvil_export(self, mock_google_authenticated,  mock_zip, mock_analyst_g
             'entity:discovery_id', '01-subject_id', '02-sample_id', '03-Gene', '04-Gene_Class',
             '05-inheritance_description', '06-Zygosity', '07-variant_genome_build', '08-Chrom', '09-Pos',
             '10-Ref', '11-Alt', '12-hgvsc', '13-hgvsp', '14-Transcript', '15-sv_name', '16-sv_type',
-            '17-significance']))
+            '17-significance', '18-discovery_notes']))
```
hanars, line 378: We need to add a test case for this, sadly. I know that involves adding more fixture data which is frustrating to test but it is important. I think its okay to add either a MNV OR a complex SV, I don't think we need both

hanars (changes requested):

seqr/views/apis/report_api_tests.py
ShifaSZ: Okay, I will add some tests.

ShifaSZ: A test has been added.

hanars (approved):

PR #1912, opened 2021-06-17
Delete indices button

hanars:


seqr/views/apis/data_manager_api.py
```diff
@@ -39,6 +39,27 @@ def elasticsearch_status(request):
             client, 'nodes', ['name', 'heap.percent'], filter_rows=lambda node: node['name'] in disk_status):
         disk_status[node.pop('name')].update(node)
 
+    indices, seqr_index_projects = _get_es_indices(client)
```
hanars, line 42: so the diff here is showing really weird, but essentially I pulled out a lot of logic into a `_get_es_indices` helper method, and then everything else showing in this block of green is actually unchanged, and all the red below showing those blocks is also unchanged. Other than pullng out the helper method, the only thing that changed in this file is I added the new `delete_index` method

seqr/views/apis/data_manager_api.py
```diff
@@ -72,26 +93,27 @@ def elasticsearch_status(request):
...
 
+@data_manager_required
+def delete_index(request):
+    index = json.loads(request.body)['index']
+    active_index_samples = Sample.objects.filter(is_active=True, elasticsearch_index=index)
+    if active_index_samples:
+        projects = {
+            sample.individual.family.project.name for sample in active_index_samples.select_related('individual__family__project')
+        }
+        return create_json_response({'error': 'Index "{}" is still used by: {}'.format(index, ', '.join(projects))}, status=403)
```
ShifaSZ, line 110: Is there a way that the users can delete the samples that using a specific index? In some cases, users may want to replace an old index with a new one. To delete the old index, the users have to delete the dataset for the old index.

seqr/views/apis/data_manager_api.py
ShifaSZ: Moving the definition of `_get_es_meta` to the location after `_get_es_indices` may help. But, anyway, they don't matter.

ShifaSZ (changes requested):
The PR looks good. I have just one question you may wanna think about.

seqr/views/apis/data_manager_api.py
hanars: The samples are records of what indices ever existed and are never deleted. They are marked inactive when they are no longer used. There is a different user interface for doing that, this is ONLY for deleting unused indices from elasticsearch

ShifaSZ (approved):

PR #1928, opened 2021-07-01
Disable username/pwd login if google login enabled

hanars:


ui/pages/Login/components/LoginError.jsx
```diff
@@ -0,0 +1,44 @@
...
+const LoginError = ({ location, match }) => (
+  <Segment textAlign="center" padded="very" basic>
+    <Message error compact>
+      {ERROR_MESSAGES[match.params.error] || `Unknown error occured: ${match.params.error}`}
+    </Message>
+    <div>
+      <Button as="a" href={`${GOOGLE_LOGIN_URL}${location.search}`} primary>
+        <Icon name="google" />
+        Sign in to seqr
+      </Button>
```
ShifaSZ, line 32: A little weird to have a Google sign-in button here. How about just a link to the `landing page` or `front page`?

ShifaSZ (changes requested):
A general question: is it okay to remove username/password login for local installation of seqr while enabling Google auth? If yes, the PR looks good.

seqr/views/apis/auth_api.py
```diff
@@ -4,20 +4,24 @@
...
-from seqr.views.utils.permissions_utils import user_is_data_manager
 from seqr.views.utils.terra_api_utils import google_auth_enabled, remove_token
+from settings import LOGIN_URL
 
 logger = SeqrLogger(__name__)
 
 
 def login_view(request):
+    if google_auth_enabled():
+        raise PermissionDenied('Username/ password authentication is disabled')
```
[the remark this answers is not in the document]
hanars: no, they will have to use either one or the other

ui/pages/Login/components/LoginError.jsx
hanars: so you only get to this page if you tried to log in and it failed. So in that case you want the ability to log in again. A common case would be you clicked the wrong google account by accident and now want to try again with the correct google account

ShifaSZ (approved):

PR #1935, opened 2021-07-06
Servctl rewrite

hanars:
seqr uses a complicated `servctl` script to help manage deployments. This large script can be very hard to read and maintain. The vast majority of the commands it supports are essentially helpers to format `kubectl` commands to run on the correct pods. At some point in the future, we will want to move our deployment configuration and management to something better designed for that purpose, like terraform. In the interim, I think it will be helpful to pull out the helper scripts that we actually want to use in the day-to-day maintenance of seqr into explicit helper scripts that we can run and easily maintain. After this PR, all that is left in `servctl` are the deploy and delete commands, which means once we migrate to different deployment system we can safely delete them without losing any helper functionality.

TL;DR this pulls a lot of the convoluted python functions out of `servctl` and into their own bash scripts


deploy/kubectl_helpers/restore_local_db.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -x -e
```
sjahl, line 3: You may want a `-u` here too, so that the script exits if any of the variables are undefined. The default will be to continue running the commands with empty strings in place of the empty variables. These may very well result in benign errors that are caught by the -e, but I feel like it's best to bail before trying to execute these gcloud and psql commands with partial inputs.

deploy/kubectl_helpers/utils/copy.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+DIR=$(dirname "$BASH_SOURCE")
+
+set -x -e
+
+DEPLOYMENT_TARGET=$1
+COMPONENT=$2
+FILE=$3
+TO_POD=$4
```
sjahl, line 10: is TO_POD expected to be a boolean here? It might be worth adding a comment to explain this a little bit (particularly since I think the default is going to end up true if you don't pass a $4, which could be a little confusing).

deploy/kubectl_helpers/utils/get_pod_info.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
```
sjahl, line 2: I would probably `set -u` here too. The `get` later on isn't doing anything destructive, but I think it'll help a bit with debugging and maintainability.

sjahl (commented):
Mostly looks good to me. I noted a few places that I thought `set -u` would be useful.

It might be helpful to have a readme in here somewhere that describes some of the invocation and examples for some of the common input variables; It would remove a little bit of guesswork when trying to re-learn how to run the scripts after not using them for a while.

One other thing to keep in mind, is that the version of bash that's installed by default on most of our developer macs is from 2007 (for licensing reasons). Not sure what (if any) checks codacy is running on this, but it's possible that it's expecting a much newer version of bash, and things that it thinks are OK in bash 5 might not be OK in 3.2 (or vice versa). Should be fine if we stick to POSIX compliant / sh syntax, but if it ends up being a problem, newer bash can be had from homebrew.

sjahl (approved):

PR #1938, opened 2021-07-09
Bump IGV.JS to v2.8.6

ShifaSZ:
Changes:
1. IGV version from 2.2.2 to 2.8.6
2. Add `indexURL` to the Gencode v27 track config. It seems that igv couldn't load the Gencode data with the index file.
3. Update CSP configurations for IGV.


settings.py
```diff
@@ -69,11 +69,12 @@
 
 CSP_INCLUDE_NONCE_IN = ['script-src', 'style-src-elem']
 CSP_FONT_SRC = ('https://fonts.gstatic.com', 'data:')
-CSP_CONNECT_SRC = ("'self'", 'https://gtexportal.org', 'https://storage.googleapis.com') # used by IGV
+CSP_CONNECT_SRC = ("'self'", 'https://gtexportal.org', 'https://storage.googleapis.com', 'https://igv.org', 'https://data.broadinstitute.org') # used by IGV
```
hanars, line 72: do we still need these new connect sources? I don't see them used anywhere so wanted to double check

ui/package.json
```diff
@@ -63,13 +63,14 @@
     "why-did-you-update": "^0.1.1"
   },
   "dependencies": {
+    "@types/react": "^17.0.14",
```
hanars, line 66: why did we add a new top level dependency?

settings.py
ShifaSZ: The `igv.org` must be still required. If we don't add it, there will be an error when opening an igv: `Refused to connect to 'https://igv.org/genomes/genomes.json?randomSeed=0.zx0ftapqh3r&version=2.8.6' because it violates the following Content Security Policy directive: "connect-src 'self' https://gtexportal.org https://storage.googleapis.com".`

The `data.broadinstitute.org` could be removed. It seems that it is for increasing an igv js use counter on a Broad's server. If it is refused to connect, it will generate an error but the error can be ignored. The error is `Refused to connect to 'https://data.broadinstitute.org/igv/projects/current/counter_igvjs.php?version=0' because it violates the following Content Security Policy directive: "connect-src 'self' https://gtexportal.org https://storage.googleapis.com".`

ui/package.json
ShifaSZ: The reason I made some changes to some un-related packages is that I upgraded my node.js to v14 once a time because a post said some node versions have a bug causing XHR requests to never ended. With node v14, when I run `npm list|grep igv`, it will report some errors:
```
npm ERR! peer dep missing: jsoneditor@^7.1.0, required by jsoneditor-react@3.0.0
npm ERR! peer dep missing: react@^0.14.0 || ^15.0.0, required by react-rangeslider@2.2.0
npm ERR! peer dep missing: @types/react@>=16, required by react-markdown@5.0.3
├─┬ igv@2.8.6
│ ├─┬ igv-ui@1.1.7 (git+https://github.com/igvteam/igv-ui.git#b7fddfda708d2339b3bba704f8e1aeccbcd695a4)
│ │ └── igv-utils@1.2.8 (github:igvteam/igv-utils#4443005457137b0d45d77165c2fc551265522490)
│ └── igv-utils@1.2.9 (git+https://github.com/igvteam/igv-utils.git#bd1197a1733cea730875c23ff3171748cc7449bb)
```
But when I downgrade node to v11 (original version), there is no error reported.
I believe the errors do exist. So I kept the changes. For now, there are two npm errors still there:
```
npm ERR! peer dep missing: jsoneditor@^7.1.0, required by jsoneditor-react@3.0.0
npm ERR! peer dep missing: react@^0.14.0 || ^15.0.0, required by react-rangeslider@2.2.0
```
I still keep `jsoneditor@^7.1.0` because the suggested version `jsoneditor@^9.0.0` has a security flaw. Maybe I should upgrade it to  `jsoneditor@>=9.1` which the security bug is fixed. For the other error, it can't be fixed because some other packages require `react@>=16`.

settings.py
hanars: But *why* are we trying to connect to `https://igv.org/genomes/genomes.json` in the first place? It specifies a bunch of default configurations so  you can initialize IGV with a genome version and it will know what configuration to use, but we use our own reference configuration so we shouldn't need to fetch the defaults at all. Maybe look in to the `loadDefaultGenomes` setting? Or look through the code to see if we actually need to fetch this?

And since it is just a counter, we should have the `data.broadinstitute.org` requests fail, the error is fine to ignore.

ui/package.json
hanars: Peer dependency errors can be ignored if they aren't actually causing issues. The `jsoneditor` and `react-rangeslider` have been there forever and there is no reason to try to fix them, and unless you ACTUALLY see an error when you try to edit markdown there is no reason to try to fix the peer dependency for `react-markdown`. Please undo all the unrelated changes, as testing that there aren't unintended consequences to these version changes is really hard to do. If you are worried about a dependency vulnerability in an unrelated npm package, that change belongs in its own PR, not lumped in to a different PR where I can't properly evaluate it. But also, I do dependency audits really frequently, so in general its not something you need to worry about

hanars (approved):

PR #1956, opened 2021-07-13
update mnv reporting

hanars:


seqr/views/apis/report_api.py
```diff
@@ -356,12 +356,16 @@ def _process_saved_variants(saved_variants_by_family, family_individual_affected
...
+            parent_name = _get_nested_variant_name(parent_mnv)
             discovery_notes = 'The following variants are part of the {variant_type} variant {parent}: {nested}'.format(
                 variant_type='complex structural' if parent_mnv.get('svType') else 'multinucleotide',
-                parent=_get_nested_variant_name(parent_mnv),
+                parent='{} ({})'.format(parent_name, ', '.join(parent_details)) if parent_details else parent_name,
                 nested=', '.join([_get_nested_variant_name(v) for v in nested_mnvs]))
-            for variant in mnvs:
+            for variant in nested_mnvs:
                 variant['discovery_notes'] = discovery_notes
+            saved_variants.remove(parent_mnv)
```
ShifaSZ, line 368: I think that parent MNVs are required to be removed from the AnVIL report but should be kept in the Saved Variants, right?

ShifaSZ (changes requested):

seqr/views/apis/report_api.py
hanars: `saved_variants` here is just a list of variants being included in the anvil report and sample metadata reports, where it should not be included. This does not actually remove it from the database

ShifaSZ (approved):

PR #1969, opened 2021-07-19
Changed the options for the IGV reference tracks to improve IGV performance.

ShifaSZ:
The changes are in the constant `REFERENCE_TRACKS` and then regenerated the static files.

### DIFF
--- a/ui/shared/components/panel/FamilyReads.jsx
+++ b/ui/shared/components/panel/FamilyReads.jsx
@@ -128,24 +128,25 @@ const REFERENCE_URLS = [
 
 const REFERENCE_TRACKS = [
   {
-    name: 'Gencode v27',
+    name: 'Gencode v32',
     indexPostfix: 'tbi',
     baseUrl: 'https://storage.googleapis.com/seqr-reference-data',
     path: {
-      37: 'GRCh37/gencode/gencode.v27lift37.annotation.sorted.gtf.gz',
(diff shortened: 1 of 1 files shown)


hanars:
In general, static assets should only be built right before a deployment, by me, as part of our release process. The only reason to have built assets in a PR is if you needed to test something on the dev server, and even then its best not to commit the built assets to the final branch if you don't have to. All it does is create a ton of merge conflicts

ui/shared/components/panel/FamilyReads.jsx
```diff
@@ -128,24 +128,25 @@ const REFERENCE_URLS = [
...
+    indexPostfix: 'tbi',
     baseUrl: `${BASE_REFERENCE_URL}/org.genomes`,
     path: {
-      37: 'hg19/ncbiRefGene.txt.gz',
-      38: 'hg38/ncbiRefGene.txt.gz',
+      37: 'hg19/refGene.sorted.txt.gz',
+      38: 'hg38/refGene.sorted.txt.gz',
     },
     format: 'refgene',
-    indexed: false,
```
hanars, line 148: haha WHY would I have done this 🤦

hanars (changes requested):
package.lock should not have changed in the PR, given that the package.json wasn't changed. In general, reading a huge diff with all the built assets is more effort than it is worth and this PR should only have the changes to the `REFERENCE_TRACKS` committed on it

hanars (approved):

PR #1970, opened 2021-07-20
handle social auth exceptions

hanars:


ShifaSZ (changes requested):
Is there any case that causes the exception? `AuthException` includes all errors because of either the client-side or the server-side. But 401 is for the client-side errors only. is '400' better?

hanars:
> Is there any case that causes the exception? `AuthException` includes all errors because of either the client-side or the server-side. But 401 is for the client-side errors only. is '400' better?

400 means user error, 401 means authentication error. So 401 is always the appropriate response when authentication fails

ShifaSZ (approved):
Can't be better.

PR #1998, opened 2021-07-27
Resize postgres instances, and parameterize the HA config flag

sjahl:
Still need to work out the process for converting/editing the existing databases, but I think these are the only servctl changes that are needed to implement what's suggested in https://github.com/broadinstitute/seqr-private/issues/1022


hanars:
this looks good to me! Might be easiest to actually make the changes in the cloud SQL UI in the GCP website ratehr than try to apply them via the gcloud cli, but either way we want the servctl code to match what we have deployed

sjahl:
I applied this change in dev and prod today. I'll leave this PR open for an actual github approval -- not sure if the green check marks are required for compliance :)

hanars (approved):
looks good to merge, especially since its already on master :)

PR #2000, opened 2021-07-28
Created custom annotation filters

MladenZeljic:
Created the custom annotation filters on Variant Search page, under "In Silico Filters" section
image

Based on the type of the option (numeric or non numeric), we have appropriate compare options (>,>=,<,<=,= - for numeric, is - for non numeric) and also there is the blank option which will clear the associated filter
image
image

Non numeric options are also searchable 
image


hanars:
seqr search relies on pagination and sorting in the elasticsearch query. Post-processing search results in the javascript to filter them is an unacceptable approach to search. To add new filter criteria to search, you need to update the POST body that is sent to the search api endpoint, and you need to update the query parsing server-side code to change the query criteria that is sent to elasticsearch. Please feel free to reach out and schedule a meeting with me if you need more guidance for implementing this

PR #2012, opened 2021-07-30
Changed the genotype and SV type info displays.

ShifaSZ:
Change the SV_TYPE (e.g. BND) displays to descriptive ones (e.g. Translation).  Show the SV_TYPE details if they exist in the hove-over messages.
Also, replace X/- notation with ref/alt.

### DIFF
--- a/ui/shared/components/panel/variants/Annotations.jsx
+++ b/ui/shared/components/panel/variants/Annotations.jsx
@@ -251,9 +251,9 @@ const Annotations = React.memo(({ variant }) => {
           trigger={
             <ButtonLink size={svType && 'big'}>
               {svType ? (SVTYPE_LOOKUP[svType] || svType) : mainTranscript.majorConsequence.replace(/_/g, ' ')}
-              {svType && svTypeDetail && SVTYPE_DETAILS[svType] &&
+              {svType && svTypeDetail &&
               <Popup trigger={<Icon name="info circle" size="small" corner="top right" />}
-                content={SVTYPE_DETAILS[svType][svTypeDetail]}
+                content={(SVTYPE_DETAILS[svType] || {})[svTypeDetail] || svTypeDetail}
                 position="top center"
               />}
             </ButtonLink>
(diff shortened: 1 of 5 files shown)


ui/shared/utils/constants.js
```diff
@@ -861,6 +861,31 @@ export const VEP_CONSEQUENCE_ORDER_LOOKUP = ORDERED_VEP_CONSEQUENCES.reduce((acc
   ({ ...acc, [consequence.value]: i }),
 {})
 
+export const SVTYPE_TEXT_MAPPING = ORDERED_VEP_CONSEQUENCES.reduce((acc, consequence) =>
```
hanars, line 864: this works, but I think it is more maintainable to make a list of just the sv types rather than filtering the whole consequence list here. So above, define 
```
SV_TYPES = [
  {
    description: 'A large duplication',
    text: 'Duplication',
    value: 'DUP',
  }, 
 ...
]
```
Then inside `ORDERED_VEP_CONSEQUENCES` add `...SV_TYPES.map(o => ({ ...o, group: VEP_GROUP_SV})` to add all the SVs to the main list of consequences, and then for the mapping do 
```
SV_TYPES.reduce((acc, { value, text }) = ({ ...acc, [value]: text }), {})
```
I might also rename `SVTYPE_TEXT_MAPPING` to `SVTYPE_LOOKUP` to be more consistent with the other constant names

hanars (changes requested):

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -250,8 +250,12 @@ const Annotations = React.memo(({ variant }) => {
           size="large"
           trigger={
             <ButtonLink size={svType && 'big'}>
-              {svType || mainTranscript.majorConsequence.replace(/_/g, ' ')}
-              {svType && svTypeDetail && `:${svTypeDetail}`}
+              {svType ? (SVTYPE_LOOKUP[svType] || svType) : mainTranscript.majorConsequence.replace(/_/g, ' ')}
+              {svType && svTypeDetail && SVTYPE_DETAILS[svType] &&
```
hanars, line 254: similar to above, if there is ever `svTypeDetail` that is somehow missing from `SVTYPE_DETAILS` we still want to show it, so this should only be conditional on `svTypeDetail` and then for the popup content we want to do something like `(SVTYPE_DETAILS[svType] || {})[svTypeDetail] || svTypeDetail`

hanars (changes requested):

hanars (approved):

PR #2015, opened 2021-08-02
add seqr link hover texts

hanars:


ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -255,9 +255,17 @@ const BaseVariantGene = React.memo(({ geneId, gene, variant, compact, showInline
...
-        <NavLink to={`/summary_data/saved_variants/ALL/${gene.geneId}`} target="_blank">seqr</NavLink>
+        <Popup
+          trigger={<NavLink to={`/summary_data/saved_variants/ALL/${gene.geneId}`} target="_blank">seqr</NavLink>}
+          content="Show all previously saved variants in this gene across all your seqr projects"
+          size="tiny"
+        />
         <HorizontalSpacer width={5} />|<HorizontalSpacer width={5} />
-        <SearchResultsLink location={gene.geneId} familyGuids={variant.familyGuids} />
+        <Popup
+          trigger={<SearchResultsLink location={gene.geneId} familyGuids={variant.familyGuids} />}
```
ShifaSZ, line 265: It seems to me that the properties `location` and `familyGuids` will never be used by `SearchResultsLink`. Can we remove them?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -255,9 +255,17 @@ const BaseVariantGene = React.memo(({ geneId, gene, variant, compact, showInline
       <GeneLinks>
         <a href={`https://decipher.sanger.ac.uk/gene/${gene.geneId}/overview/protein-genomic-info`} target="_blank">Decipher</a>
```
ShifaSZ, line 256: Can we also add a popup hint for `Decipher`? Otherwise, it seems weird that it has nothing hovering while the other two buttons have.

ShifaSZ (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
hanars: no, they are used in the `openSearchResults` function

ui/shared/components/panel/variants/VariantGene.jsx
hanars: no, because the analysts don't have any confusion about that link and did not ask for a hover

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay, I see.

ShifaSZ (approved):

PR #2020, opened 2021-08-03
Anvil accept policies

hanars:
Enforce that users accept policies in order to access anvil-specific pages

### DIFF
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -21,12 +21,15 @@ from seqr.views.utils.individual_utils import add_or_update_individuals_and_fami
 from seqr.utils.communication_utils import send_html_email
 from seqr.utils.file_utils import does_file_exist
 from seqr.utils.logging_utils import SeqrLogger
-from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm
+from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm, login_and_policies_required
 from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL
 
 logger = SeqrLogger(__name__)
 
 anvil_auth_required = user_passes_test(is_anvil_authenticated, login_url=GOOGLE_LOGIN_REQUIRED_URL)
+
+def anvil_auth_and_permission_required(view_func):
+    return login_and_policies_required(anvil_auth_required(view_func), login_url=GOOGLE_LOGIN_REQUIRED_URL)
 
 @anvil_auth_required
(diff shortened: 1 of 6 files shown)


seqr/views/apis/anvil_workspace_api.py
```diff
@@ -21,14 +21,21 @@
...
 
-@anvil_auth_required
+def anvil_auth_and_policies_required(wrapped_func=None, policy_url=API_POLICY_REQUIRED_URL):
+    def decorator(view_func):
+        return login_and_policies_required(anvil_auth_required(view_func), login_url=GOOGLE_LOGIN_REQUIRED_URL, policy_url=policy_url)
+    if wrapped_func:
+        return decorator(wrapped_func)
+    return decorator
+
+@anvil_auth_and_policies_required(policy_url=POLICY_REQUIRED_URL)
```
ShifaSZ, line 38: Does it work with two decorators rather than one cascading decorator? If yes, maybe two decorators could be more flexible.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -36,27 +36,14 @@
...
 
-    def _check_login_permissions(self, url, mock_logger):
-        # Test user doesn't login
-        response = self.client.post(url)
-        self.assertEqual(response.status_code, 302)
-        self.assertEqual(response.url, '/login/google-oauth2?next={}'.format(url))
+    def test_anvil_workspace_page(self, mock_logger):
+        # Requesting to load data from a workspace without an existing project
+        url = reverse(anvil_workspace_page, args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME])
+        collaborator_response = self.check_manager_login(url, login_redirect_url='/login/google-oauth2', policy_redirect_url='/accept_policies')
```
ShifaSZ, line 42: Cool. I finally got how the `check_...login()` test utilities work.

ShifaSZ (changes requested):
Looks good. But I have a question expected to be answered.

seqr/views/apis/anvil_workspace_api.py
hanars: thats really inconsistent with our style for decorators everywhere else, and is less readable. It makes it seem like the decorators do different things instead of the same thing

ShifaSZ (approved):

seqr/views/apis/anvil_workspace_api_tests.py
hanars: to be fair, we couldn't use the utility before because it checks that it fails for inactive users or users have not accepted the policies, and until this change those checks were failing

PR #2021, opened 2021-08-03
Added CN and call type miss match warning.

ShifaSZ:
The layout of an example miss matching case:
image

### DIFF
--- a/ui/shared/components/panel/variants/VariantIndividuals.jsx
+++ b/ui/shared/components/panel/variants/VariantIndividuals.jsx
@@ -117,13 +117,20 @@ const svGenotype = (genotype, isHemiX) => {
   )
 }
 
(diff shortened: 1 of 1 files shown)


ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -126,6 +126,13 @@ export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) =>
         content={<div><b>Warning:</b> {warning}</div>}
       />
     }
+    {cnSvtypeMissmatch &&
```
hanars, line 129: we don't want a second icon popup here, we want to use the `warning` prop for this

hanars (changes requested):

hanars (approved):

PR #2029, opened 2021-08-04
better svsize display

hanars:


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -225,6 +225,16 @@ const mapLocusListStateToProps = (state, ownProps) => ({
...
 const VariantLocusListLabels = connect(mapLocusListStateToProps)(BaseVariantLocusListLabels)
 
+const svSizeDisplay = (size) => {
+  if (size < 1000) {
+    return `${size}bp`
+  } else if (size < 1000000) {
+    // dividing by 1 removes trailing 0s
+    return `${((size) / 1000).toPrecision(3) / 1}kb`
+  }
+  return `${(size / 1000000).toFixed(2) / 1}mb`
```
ShifaSZ, line 235: `.toPrecision(3)` could lose some accuracies. It is better `.toFixed(3)`.

ShifaSZ (changes requested):

ui/shared/components/panel/variants/Annotations.jsx
hanars: This is an intentional decision. If it is a 3-digit kb (i.e 123.456) we don't want to show any decimal points and just show 123, but if its a 1-digit kb (i.e 1.23456) we want to show 2 decimal points so it would be 1.23 so `toPrecision` works perfectly here. The decimal point become meaningless for the larger kb values and just clutter the display. We use `toFixed` below because there we always care about decimals

ShifaSZ (approved):

Issue #157, opened 2015-10-28
In the downloadable table of mendelian variant search results, make genotype values clearer


bw2:
Make genotypes be like GCCAGGA/G instead of the current 0 for hom, 1 for het, 2 for hom-alt.

@evalkanas

PR #2035, opened 2021-08-06
Ensure no missing samples in the VCF

ShifaSZ:
I finally come up with a solution without using any 3rd-party tools. Just load the VCF header line and retrieve the sample IDs in the line, and then compare the individual IDs to those sample IDs to find out if there are any missing IDs.
The delay is around 4 seconds which includes two almost equal parts: one is to get the VCF file name, another is to get the samples from the VCF file. If the user provides a VCF file name directly in the `Path to the Joint Called VCF` field, then the delay will be cut in half.

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -63,26 +63,3 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
         if not raw_content:
             line = line.decode('utf-8')
         yield line
-
-
-def get_vcf_filename(vcf_path):
-    if vcf_path.endswith('.vcf') or vcf_path.endswith('.vcf.gz') or vcf_path.endswith('.vcf.bgz'):
-        return vcf_path
-    process = subprocess.Popen('gsutil ls ' + vcf_path, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
-    if process.wait() != 0:
-        return None
-    for line in process.stdout:
-        file = line.decode('utf-8').strip()
-        if file.endswith('.vcf') or file.endswith('.vcf.gz') or file.endswith('.vcf.bgz'):
-            return file
-    return None
-
-
-def get_vcf_samples(vcf_path):
-    vcf_filename = get_vcf_filename(vcf_path)
-    if not vcf_filename:
-        return {}
-    for line in file_iter(vcf_filename, byte_range=(0, 65536)):
-        if line.startswith('#CHROM'):
-            return set(line.rstrip().split('\tFORMAT\t', 2)[1].split('\t') if line.endswith('\n') else [])
-    return {}
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -64,3 +64,25 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
             line = line.decode('utf-8')
         yield line
 
+
+def get_vcf_filename(vcf_path):
+    if vcf_path.endswith('.vcf') or vcf_path.endswith('.vcf.gz') or vcf_path.endswith('.vcf.bgz'):
+        return vcf_path
+    process = subprocess.Popen('gsutil ls ' + vcf_path, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+    if process.wait() != 0:
+        return None
+    for line in process.stdout:
+        file = line.decode('utf-8').strip()
+        if file.endswith('.vcf') or file.endswith('.vcf.gz') or file.endswith('.vcf.bgz'):
+            return file
+    return None
+
+
+def get_vcf_samples(vcf_path):
+    vcf_filename = get_vcf_filename(vcf_path)
+    if not vcf_filename:
+        return {}
+    for line in file_iter(vcf_filename, byte_range=(0, 65536)):
+        if line.startswith('#CHROM'):
+            return set(line.rstrip().split('\tFORMAT\t', 2)[1].split('\t') if line.endswith('\n') else [])
+    return {}
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -1,6 +1,7 @@
 """APIs for management of projects related to AnVIL workspaces."""
 import json
 import time
+import subprocess
 
 from django.contrib.auth.decorators import user_passes_test
 from django.contrib.auth.models import User
@@ -19,7 +20,7 @@ from seqr.views.utils.terra_api_utils import add_service_account, has_service_ac
 from seqr.views.utils.pedigree_info_utils import parse_pedigree_table
 from seqr.views.utils.individual_utils import add_or_update_individuals_and_families
 from seqr.utils.communication_utils import send_html_email
-from seqr.utils.file_utils import does_file_exist
+from seqr.utils.file_utils import does_file_exist, get_vcf_samples
 from seqr.utils.logging_utils import SeqrLogger
 from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm, login_and_policies_required
 from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL, POLICY_REQUIRED_URL, API_POLICY_REQUIRED_URL
@@ -110,6 +111,13 @@ def create_project_from_workspace(request, namespace, name):
     if errors:
         return create_json_response({'errors': errors}, status=400)
 
+    # Validate the VCF to see if it contains all the required samples
+    samples = get_vcf_samples(data_path)
+    missing_samples = [record['individualId'] for record in pedigree_records if record['individualId'] not in samples]
+    if missing_samples:
+        return create_json_response(
+            {'errors': 'There are missing samples in the data: {}'.format(', '.join(missing_samples))}, status=400)
+
     # Create a new Project in seqr
     project_args = {
         'name': name,
@@ -142,6 +150,7 @@ def _wait_for_service_account_access(user, namespace, name):
         if has_service_account_access(user, namespace, name):
             return True
     raise TerraAPIException('Failed to grant seqr service account access to the workspace', 400)
+
 
 def _send_load_data_email(project, updated_individuals, data_path, sample_type, user):
     email_content = """
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -19,7 +19,7 @@ from seqr.views.utils.terra_api_utils import add_service_account, has_service_ac
 from seqr.views.utils.pedigree_info_utils import parse_pedigree_table
 from seqr.views.utils.individual_utils import add_or_update_individuals_and_families
 from seqr.utils.communication_utils import send_html_email
-from seqr.utils.file_utils import does_file_exist, get_vcf_samples
+from seqr.utils.file_utils import does_file_exist, file_iter
 from seqr.utils.logging_utils import SeqrLogger
 from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm, login_and_policies_required
 from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL, POLICY_REQUIRED_URL, API_POLICY_REQUIRED_URL
@@ -28,12 +28,26 @@ logger = SeqrLogger(__name__)
 
 anvil_auth_required = user_passes_test(is_anvil_authenticated, login_url=GOOGLE_LOGIN_REQUIRED_URL)
 
+BLOCK_SIZE = 65536
+
+
+def get_vcf_samples(vcf_filename):
+    byte_range = None if vcf_filename.endswith('.vcf') else (0, BLOCK_SIZE)
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            break
+        if line.startswith('#CHROM'):
+            return set(line.rstrip().split('FORMAT\t', 2)[1].split('\t') if line.endswith('\n') else [])
+    return {}
+
+
 def anvil_auth_and_policies_required(wrapped_func=None, policy_url=API_POLICY_REQUIRED_URL):
     def decorator(view_func):
         return login_and_policies_required(anvil_auth_required(view_func), login_url=GOOGLE_LOGIN_REQUIRED_URL, policy_url=policy_url)
     if wrapped_func:
         return decorator(wrapped_func)
     return decorator
+
 
 @anvil_auth_and_policies_required(policy_url=POLICY_REQUIRED_URL)
 def anvil_workspace_page(request, namespace, name):
@@ -94,11 +108,14 @@ def create_project_from_workspace(request, namespace, name):
     # Add the seqr service account to the corresponding AnVIL workspace
     added_account_to_workspace = add_service_account(request.user, namespace, name)
     if added_account_to_workspace:
-        _wait_for_service_account_access(request.user,namespace, name)
+        _wait_for_service_account_access(request.user, namespace, name)
 
     # Validate the data path
     bucket_name = workspace_meta['workspace']['bucketName']
     data_path = 'gs://{bucket}/{path}'.format(bucket=bucket_name.rstrip('/'), path=request_json['dataPath'].lstrip('/'))
+    if not (data_path.endswith('.vcf') or data_path.endswith('.vcf.gz') or data_path.endswith('.vcf.bgz')):
+        error = 'The data file must have a postfix of ".vcf", ".vcf.gz", or ".vcf.bgz".'
+        return create_json_response({'error': error}, status=400, reason=error)
     if not does_file_exist(data_path, user=request.user):
         error = 'Data file or path {} is not found.'.format(request_json['dataPath'])
         return create_json_response({'error': error}, status=400, reason=error)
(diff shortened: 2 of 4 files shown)


seqr/utils/file_utils.py
```diff
@@ -64,3 +64,25 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
...
+        if file.endswith('.vcf') or file.endswith('.vcf.gz') or file.endswith('.vcf.bgz'):
+            return file
+    return None
+
+
+def get_vcf_samples(vcf_path):
+    vcf_filename = get_vcf_filename(vcf_path)
+    if not vcf_filename:
+        return {}
+    for line in file_iter(vcf_filename, byte_range=(0, 65536)):
```
ShifaSZ, line 85: `65536` here is the max block size for BGZIP. It ensures the data downloaded include a completed gzip block.

seqr/utils/file_utils.py
```diff
@@ -64,3 +64,25 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
...
+    if vcf_path.endswith('.vcf') or vcf_path.endswith('.vcf.gz') or vcf_path.endswith('.vcf.bgz'):
+        return vcf_path
+    process = subprocess.Popen('gsutil ls ' + vcf_path, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+    if process.wait() != 0:
+        return None
+    for line in process.stdout:
+        file = line.decode('utf-8').strip()
+        if file.endswith('.vcf') or file.endswith('.vcf.gz') or file.endswith('.vcf.bgz'):
+            return file
+    return None
```
ShifaSZ, line 78: Must we do the mapping from the data path (`Path to the Joint Called VCF`) to VCF file name mapping? Can we ask the user to provide the VCF filename directly?

seqr/utils/file_utils.py
```diff
@@ -64,3 +64,25 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
...
+
+
+def get_vcf_samples(vcf_path):
+    vcf_filename = get_vcf_filename(vcf_path)
+    if not vcf_filename:
+        return {}
+    for line in file_iter(vcf_filename, byte_range=(0, 65536)):
+        if line.startswith('#CHROM'):
+            return set(line.rstrip().split('\tFORMAT\t', 2)[1].split('\t') if line.endswith('\n') else [])
```
hanars, line 87: @mike-w-wilson I'm not super familiar with VCF formats - can we safely assume that any VCF header (or at least any VCF that would work with our loading pipeline)) would include  a tab-separated list of sample IDs after the `FORHAT` header?

seqr/utils/file_utils.py
hanars: I'm really confused what filename we are even getting here? In the main code block, we have 
```
data_path = 'gs://{bucket}/{path}'.format(bucket=bucket_name.rstrip('/'), path=request_json['dataPath'].lstrip('/'))
if not does_file_exist(data_path, user=request.user):
    error = 'Data file or path {} is not found.'.format(request_json['dataPath'])
```
where `does_file_exist` is ultimately running a `gsutil ls` check on that full `data_path` and throws an error if it fails. So I don't get what we are checking here? In what scenario is the `file` that is returned from this function different from the `data_path` input into it?

seqr/utils/file_utils.py
hanars: We should use a constant for the hardcoded number with a comment explaining this

Also, this helper function doesn't belong in `file_utils` which includes really generic functions for interacting with files, it just belongs in anvil_workspace_api.py . If you wanted to do a generic helper function like `get_google_bucket_file_header` that could live in this file

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -110,6 +110,13 @@ def create_project_from_workspace(request, namespace, name):
     if errors:
         return create_json_response({'errors': errors}, status=400)
 
+    # Validate the VCF to see if it contains all the required samples
+    samples = get_vcf_samples(data_path)
+    missing_samples = [record['individualId'] for record in pedigree_records if record['individualId'] not in samples]
+    if missing_samples:
+        return create_json_response(
+            {'errors': 'There are missing samples in the data: {}'.format(', '.join(missing_samples))}, status=400)
```
hanars, line 118: A better error would be "The following samples are included in the pedigree file but are missing from the VCF: {}"

seqr/utils/file_utils.py
```diff
@@ -64,3 +64,25 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
...
+
+def get_vcf_filename(vcf_path):
+    if vcf_path.endswith('.vcf') or vcf_path.endswith('.vcf.gz') or vcf_path.endswith('.vcf.bgz'):
+        return vcf_path
+    process = subprocess.Popen('gsutil ls ' + vcf_path, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+    if process.wait() != 0:
+        return None
+    for line in process.stdout:
+        file = line.decode('utf-8').strip()
+        if file.endswith('.vcf') or file.endswith('.vcf.gz') or file.endswith('.vcf.bgz'):
```
hanars, line 76: if we are want to validate that a file has a specific file extension, we should be validating that well upstream of this check, we should be validating that before we even check if the file exists and should throw an explicit error stating that the file extension is not supported

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -110,6 +110,13 @@ def create_project_from_workspace(request, namespace, name):
     if errors:
         return create_json_response({'errors': errors}, status=400)
 
+    # Validate the VCF to see if it contains all the required samples
+    samples = get_vcf_samples(data_path)
```
hanars, line 114: `get_vcf_samples` returns empty in the scenario where we are not able to get a properly formatted header row and parse the sample IDs, and in that case we need to return an error that explains that instead of just giving an error that says all the samples are missing

seqr/utils/file_utils.py
hanars: Also, if the file is NOT gzipped, can we be confident that that will be enough bytes for the whole header?

hanars (commented):
This is a really clever and efficient approach! I have some implementation feedback but overall I love this idea

seqr/utils/file_utils.py
ShifaSZ: I would probably move the functions to `anvil_workspace_api.py`.

For non-gzipped files, I tried 1000 genome project data which contains 2500+ samples, and found that the header is within 64k bytes. I could add a loop to read more blocks. But I'm sure if it is overkilled or not.

seqr/utils/file_utils.py
mike-w-wilson: Yes, that is the spec put out by the official docs. All samples are in the header and all entires are tab-separated -- though I guess someone could technically change it to spaces but this does not conform to the VCF spec so I don't think we need to support that.
```
1 The VCF specification
VCF is a text file format (most likely stored in a compressed manner). It contains meta-information lines (prefixed
with “##”), a header line (prefixed with “#”), and data lines each containing information about a position in the
genome and genotype information on samples for each position (text fields separated by tabs). Zero length fields are
not allowed, a dot (“.”) must be used instead. In order to ensure interoperability across platforms, VCF compliant
implementations must support both LF (“\n”) and CR+LF (“\r\n”) newline conventions.
...
##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Read Depth">
##FORMAT=<ID=HQ,Number=2,Type=Integer,Description="Haplotype Quality">
#CHROM POS ID REF ALT QUAL FILTER INFO FORMAT NA00001 NA00002 NA00003
20 14370 rs6054257 G A 29 PASS NS=3;DP=14;AF=0.5;DB;H2 GT:GQ:DP:HQ 0|0:48:1:51,51 1|0:48:8:51,51 1/1:43:5:.,.
```

seqr/utils/file_utils.py
ShifaSZ: To my understanding, the path that the user provides is just the path to the VCF file, not including the filename. Especially, the `data_path` is different from the `filename` when the VCF is split into multiple files. In that case, the `data_path` is a directory, not a file. I know the VCF can be split per chromosome basis and we can use the samples in any one of the non-Y chromosome splits of the VCF. I'm not sure if there are any other ways for splitting VCF.

I could combine the path existing check and getting a VCF filename so that there would be only one `gsutil ls` would be run.

seqr/utils/file_utils.py
hanars: I would check if a file is gzipped and only include the range size if it is. If its not gzipped, it will only read one line at a time which shouldn't be an issue in terms of size

seqr/utils/file_utils.py
hanars: We do not support sharded VCFs for anvil loading so the split file thing is not an edge case we need to hand;e

seqr/utils/file_utils.py
hanars: I agree we do not have to support VCFs that do not conform to spec

seqr/utils/file_utils.py
ShifaSZ: If we don't support split VCF, should I also update the helper message on the data loading page as below?
image

seqr/utils/file_utils.py
ShifaSZ: The VCFs are on Google cloud. Is there any way to read them line by line?

hanars: `file_iter` reads them line by line already

hanars: alternatively, we could just require vcfs to be gzipped. By default the GATK joint calling workflow in AnVIL gzips things and theres no reason *not* to store a smaller version of the file

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: Changed to respond with an error if the `samples` is empty.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -140,17 +143,28 @@ def test_create_project_from_workspace(self, mock_file_exist, mock_email, mock_a
...
+        mock_file_iter.return_value = ['##fileformat=VCFv4.2\n', '#CHROM	POS	ID	REF	ALT	QUAL']  # incomplete header line
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json()['error'], 'Bad VCF file or no sample found in the data file.')
+
         # Test valid operation
         mock_sleep.reset_mock()
         mock_has_service_account.reset_mock()
         mock_add_service_account.return_value = False
-        mock_file_exist.return_value = True
```
ShifaSZ, line 153: Moved to line #157 above.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -94,11 +108,14 @@ def create_project_from_workspace(request, namespace, name):
...
     added_account_to_workspace = add_service_account(request.user, namespace, name)
     if added_account_to_workspace:
-        _wait_for_service_account_access(request.user,namespace, name)
+        _wait_for_service_account_access(request.user, namespace, name)
 
     # Validate the data path
     bucket_name = workspace_meta['workspace']['bucketName']
     data_path = 'gs://{bucket}/{path}'.format(bucket=bucket_name.rstrip('/'), path=request_json['dataPath'].lstrip('/'))
+    if not (data_path.endswith('.vcf') or data_path.endswith('.vcf.gz') or data_path.endswith('.vcf.bgz')):
+        error = 'The data file must have a postfix of ".vcf", ".vcf.gz", or ".vcf.bgz".'
```
hanars, line 117: postfix isn't really a word people know - I would say something like `Invalid VCF file format - file path must end with .vcf, .vcf.gz, or .vcf.bgz`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -110,6 +127,17 @@ def create_project_from_workspace(request, namespace, name):
     if errors:
         return create_json_response({'errors': errors}, status=400)
 
+    # Validate the VCF to see if it contains all the required samples
+    samples = get_vcf_samples(data_path)
+    if not samples:
+        return create_json_response(
+            {'error': 'Bad VCF file or no sample found in the data file.'}, status=400)
```
hanars, line 134: I don't like using the word "Bad" here. I would just say "No samples found in the provided VCF. This may be due to a malformed file"

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -140,17 +143,28 @@ def test_create_project_from_workspace(self, mock_file_exist, mock_email, mock_a
...
         self.assertEqual(mock_has_service_account.call_count, 1)
         self.assertEqual(mock_sleep.call_count, 1)
 
+        response = self.client.post(url, content_type='application/json',
+                                    data=json.dumps(REQUEST_BODY_BAD_DATA_PATH))
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json()['error'], 'The data file must have a postfix of ".vcf", ".vcf.gz", or ".vcf.bgz".')
+
+        mock_file_exist.return_value = True
+        mock_file_iter.return_value = ['##fileformat=VCFv4.2\n', '#CHROM	POS	ID	REF	ALT	QUAL']  # incomplete header line
```
hanars, line 158: `file_iter` will return a string representation of a line in the file, not a list of strings

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -28,13 +28,27 @@
...
+BLOCK_SIZE = 65536
+
+
+def get_vcf_samples(vcf_filename):
+    byte_range = None if vcf_filename.endswith('.vcf') else (0, BLOCK_SIZE)
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            break
+        if line.startswith('#CHROM'):
+            return set(line.rstrip().split('FORMAT\t', 2)[1].split('\t') if line.endswith('\n') else [])
```
hanars, line 40: I'm confused by the ` if line.endswith('\n')` check - what does having a newline have to do with anything here? Wouldn't a gzipped block end in the middle of a line some of the time?

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -81,7 +81,7 @@ const SAMPLE_TYPE_FIELD = {
 const DATA_BUCK_FIELD = {
   name: 'dataPath',
   label: 'Path to the Joint Called VCF',
-  labelHelp: 'File path for a joint called VCF available in the workspace "Files". If the VCF is split, provide the path to the directory containing the split VCF',
+  labelHelp: 'File path for a joint called VCF available in the AnVIL workspace\'s "Files" folder.',
```
hanars, line 84: "Files" isn't a folder so adding that word makes no sense

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: I think it is possible for a gzipped block to end in the middle of a line. If it happens then the `.endswith('\n')` checking can avoid using partial samples and let the user know something wrong.

hanars: I mean, if there are partial samples then the missing sample check will fail and alert the user, so I think this is less helpful

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I know that `file_iter()` returns an iteration of lines in the file. The list is iterable so that it can emulate the return value. Maybe adding `iter()` to the list is better but it seems not necessary.

hanars: oh yeah good point, this is fine to keep as is. I would say add a couple of "lines" after the header in the list just to test that iteration stops correctly

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
ShifaSZ: How about changing it to `File path for a joint called VCF available in the AnVIL workspace's storage bucket.`? Because `Files` itself doesn't make much sense even it is shown in the AnVIL UI.

hanars: Screen Shot 2021-08-09 at 4 39 12 PM

Its literally that pat of workspaces that is labeled "Files", and if the file is somewhere else in the workspace's storage it won't work. So no. Please do not change this to something that is less accurate

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -28,13 +28,28 @@
...
+
+
+def get_vcf_samples(vcf_filename):
+    byte_range = None if vcf_filename.endswith('.vcf') else (0, BLOCK_SIZE)
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            break
+        if line.startswith('#CHROM'):
+            header = line.rstrip().split('FORMAT\t', 2)
+            return set(header[1].split('\t')) if len(header) == 2 else {}
```
hanars, line 41: I like this better, its much more readable than it was before

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -86,7 +88,8 @@ def test_anvil_workspace_page(self, mock_logger):
     @mock.patch('seqr.views.apis.anvil_workspace_api.add_service_account')
     @mock.patch('seqr.utils.communication_utils.EmailMultiAlternatives')
     @mock.patch('seqr.views.apis.anvil_workspace_api.does_file_exist')
-    def test_create_project_from_workspace(self, mock_file_exist, mock_email, mock_add_service_account,
+    @mock.patch('seqr.views.apis.anvil_workspace_api.file_iter')
```
hanars, line 91: you need to add tests that `mock_file_iter` is called with the expected arguments, same as how we currenlty check `mock_file_exist`. We want at least one call with a gzipped file and at least one with a non-gzipped file

hanars (changes requested):

hanars (approved):

PR #2043, opened 2021-08-13
Updated for AnVIL loading pipeline improvement.

ShifaSZ:
It is a draft PR because the unit tests haven't been fully updated.

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -63,3 +63,15 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
         if not raw_content:
             line = line.decode('utf-8')
         yield line
+
+
+def save_data_to_gs(gs_path, data, user=None):
+    if not _is_google_bucket_file_path(gs_path):
+        raise Exception('A Google Storage path is expected')
+    if does_file_exist(gs_path, user=user):
+        raise Exception('Can\'t overwrite existing file {}.'.format(gs_path))
+    command = 'echo "{data}" | gsutil cp - {gs_path}'.format(data=data, gs_path=gs_path)
+    process = _run_command(command, user=user)
+    if process.wait() != 0:
+        errors = [line.decode('utf-8').strip() for line in process.stdout]
+        raise Exception('Run command failed: ' + ' '.join(errors))
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -68,7 +68,7 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
 def mv_file_to_gs(local_path, gs_path, user=None):
     if not _is_google_bucket_file_path(gs_path):
         raise Exception('A Google Storage path is expected.')
-    command = 'mv {}'.format(local_path, gs_path)
+    command = 'mv {}'.format(local_path)
     process = _run_gsutil_command(command, gs_path, user=user)
     if process.wait() != 0:
         errors = [line.decode('utf-8').strip() for line in process.stdout]
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -65,12 +65,10 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
         yield line
 
 
-def save_data_to_gs(gs_path, data, user=None):
+def mv_file_to_gs(local_path, gs_path, user=None):
     if not _is_google_bucket_file_path(gs_path):
         raise Exception('A Google Storage path is expected')
-    if does_file_exist(gs_path, user=user):
-        raise Exception('Can\'t overwrite existing file {}.'.format(gs_path))
-    command = 'echo "{data}" | gsutil cp - {gs_path}'.format(data=data, gs_path=gs_path)
+    command = 'gsutil mv {} {}'.format(local_path, gs_path)
     process = _run_command(command, user=user)
     if process.wait() != 0:
         errors = [line.decode('utf-8').strip() for line in process.stdout]
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -67,7 +67,7 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
 
 def mv_file_to_gs(local_path, gs_path, user=None):
     if not _is_google_bucket_file_path(gs_path):
-        raise Exception('A Google Storage path is expected')
+        raise Exception('A Google Storage path is expected.')
     command = 'gsutil mv {} {}'.format(local_path, gs_path)
     process = _run_command(command, user=user)
     if process.wait() != 0:
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -68,8 +68,8 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
 def mv_file_to_gs(local_path, gs_path, user=None):
     if not _is_google_bucket_file_path(gs_path):
         raise Exception('A Google Storage path is expected.')
-    command = 'gsutil mv {} {}'.format(local_path, gs_path)
-    process = _run_command(command, user=user)
+    command = 'mv {}'.format(local_path, gs_path)
+    process = _run_gsutil_command(command, gs_path, user=user)
     if process.wait() != 0:
         errors = [line.decode('utf-8').strip() for line in process.stdout]
         raise Exception('Run command failed: ' + ' '.join(errors))
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -165,9 +165,9 @@ def create_project_from_workspace(request, namespace, name):
 
     # Upload sample IDs to a file on Google Storage
     ids_path = '{}base/{guid}_ids.txt'.format(_get_loading_project_path(project, request_json['sampleType']), guid=project.guid)
-    sample_ids = '\n'.join(['s'] + [individual.individual_id for individual in updated_individuals])
+    sample_ids = [individual.individual_id for individual in updated_individuals]
     try:
-        temp_path = save_temp_data(sample_ids)
+        temp_path = save_temp_data('\n'.join(['s'] + sample_ids))
         mv_file_to_gs(temp_path, ids_path, user=request.user)
     except Exception as ee:
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -164,7 +164,7 @@ def create_project_from_workspace(request, namespace, name):
     )
 
     # Upload sample IDs to a file on Google Storage
-    ids_path = _get_loading_project_path(project, request_json['sampleType']) + 'base/{guid}_ids.txt'.format(guid=project.guid)
+    ids_path = '{}base/{guid}_ids.txt'.format(_get_loading_project_path(project, request_json['sampleType']), guid=project.guid)
     sample_ids = '\n'.join(['s'] + [individual.individual_id for individual in updated_individuals])
     try:
         temp_path = save_temp_data(sample_ids)
@@ -199,7 +199,7 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
     pipeline_dag = {
         "active_projects": [project.guid],
         "vcf_path": data_path,
-        "project_path": _get_loading_project_path(project, sample_type) + 'v1',
+        "project_path": '{}v1'.format(_get_loading_project_path(project, sample_type)),
         "projects_to_run": [project.guid],
     }
     message_content = """
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -224,7 +224,4 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
         dag=json.dumps(pipeline_dag, indent=4),
     )
 
-    try:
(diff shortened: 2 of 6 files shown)


seqr/views/apis/anvil_workspace_api.py
```diff
@@ -155,11 +154,19 @@ def create_project_from_workspace(request, namespace, name):
         project, individual_records=pedigree_records, user=request.user
     )
 
-    # Send an email to all seqr data managers
+    # Upload sample IDs to a file on Google Storage
+    ids_path = _get_loading_project_path(project, request_json['sampleType']) + '/{guid}_ids.txt'.format(guid=project.guid)
     try:
-        _send_load_data_email(project, updated_individuals, data_path, request_json['sampleType'], request.user)
+        save_data_to_gs(ids_path, '\n'.join(['s'] + [individual.individual_id for individual in updated_individuals]),
+                        user=request.user)
```
ShifaSZ, line 161: I found that the ID file is not possible to be pre-existing.  It is because that the new project GUID in the ID file path must be different from that of the existing projects. Therefore, I don't rename any existing ID file with a sequence number.

seqr/utils/file_utils.py
```diff
@@ -63,3 +63,15 @@ def _google_bucket_file_iter(gs_path, byte_range=None, raw_content=False, user=N
...
             line = line.decode('utf-8')
         yield line
+
+
+def save_data_to_gs(gs_path, data, user=None):
+    if not _is_google_bucket_file_path(gs_path):
+        raise Exception('A Google Storage path is expected')
+    if does_file_exist(gs_path, user=user):
+        raise Exception('Can\'t overwrite existing file {}.'.format(gs_path))
+    command = 'echo "{data}" | gsutil cp - {gs_path}'.format(data=data, gs_path=gs_path)
```
hanars, line 73: I feel like writing a temp file and then `mv`ing the actual file would be better than  piping data directly to the cp command

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -155,11 +154,19 @@ def create_project_from_workspace(request, namespace, name):
...
     except Exception as ee:
-        message = 'AnVIL loading request email exception: {}'.format(str(ee))
+        logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user)
```
hanars, line 163: we also need to log the sample IDs somewhere so we can manually create the file. We will be reporting success to the user here and they will expect us to be able to load their data. You can use the `detail` kwargs for `logger.error` to pass it to the log json

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: It makes sense.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -173,30 +187,44 @@ def _wait_for_service_account_access(user, namespace, name):
...
-        email_content,
-        subject='AnVIL data loading request',
-        to=sorted([dm.email for dm in User.objects.filter(is_staff=True, is_active=True)]),
-        attachments=[(
-            '{}_sample_ids.tsv'.format(project.guid),
-            '\n'.join([individual.individual_id for individual in updated_individuals])
-        )]
-    )
+    try:
+        safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, message_content)
```
hanars, line 228: `safe_post_to_slack` is already wrapped in a try-except so we don't need one here

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -83,16 +87,18 @@ def test_anvil_workspace_page(self, mock_logger):
...
     @mock.patch('seqr.views.apis.anvil_workspace_api.load_uploaded_file')
     @mock.patch('seqr.views.apis.anvil_workspace_api.has_service_account_access')
     @mock.patch('seqr.views.apis.anvil_workspace_api.add_service_account')
-    @mock.patch('seqr.utils.communication_utils.EmailMultiAlternatives')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.safe_post_to_slack')
     @mock.patch('seqr.views.apis.anvil_workspace_api.does_file_exist')
     @mock.patch('seqr.views.apis.anvil_workspace_api.file_iter')
-    def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mock_email, mock_add_service_account,
-                                           mock_has_service_account, mock_load_file, mock_api_logger, mock_sleep,
+    @mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs')
```
hanars, line 98: so `mv_file_to_gs` is not going to be tested anywhere because you mock it here. So far, we've never explicitly added tests for the file utils file, we test the functions where they are called. For example, if you look at igv_api_tests.py and data_manager_api_tests.py you see it actually them mocking subprocess and checking that the correct gsutil command is being run. We either need to do that here, or we need to add a test file and add an explicit test for this helper method

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -183,27 +190,35 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
-        """.format(namespace=TEST_WORKSPACE_NAMESPACE, name=TEST_NO_PROJECT_WORKSPACE_NAME, guid=project.guid)
-        mock_email.assert_called_with(
-            subject='AnVIL data loading request',
-            body=email_body.format(project_name=TEST_NO_PROJECT_WORKSPACE_NAME),
-            to=['test_data_manager@test.com', 'test_superuser@test.com'],
-            attachments=[('{}_sample_ids.tsv'.format(project.guid), 'NA19675\nNA19678\nHG00735')]
+        mock_time.sleep.assert_not_called()
+        mock_file_exist.assert_called_with(TEST_VCF_PATH, user=self.manager_user)
+        mock_file_iter.assert_called_with(TEST_VCF_PATH, byte_range=None)
+        mock_tempfile.NamedTemporaryFile.assert_called_with(mode='wb', delete=False)
```
hanars, line 196: might be a bit cleaner if you mock `tempfile.NamedTemporaryFile` instead of just `tempfile`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -183,27 +190,35 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
-            email_body.format(project_name=html_project_name), 'text/html')
-        mock_email.return_value.send.assert_called()
+        slack_message = """
+        *{user}* requested to load WES data (GRCh38) from AnVIL workspace *{namespace}/{name}* at 
+        {vcf_path} to seqr project <http://testserver/project/{guid}/project_page|*anvil-no-project-workspace1*> (guid: {guid})  
+  
+        The sample IDs to load have been uploaded to gs://seqr-datasets/v02/GRCh38/AnVIL_WES/{guid}/base/{guid}_ids.txt.  
+  
+        DAG for the loading pipeline:
+        ```{{
```
hanars, line 209: I think for slack you want the block content on a new line under the ```

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -183,27 +190,35 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
+        "{guid}"
+    ],
+    "vcf_path": "{vcf_path}",
+    "project_path": "gs://seqr-datasets/v02/GRCh38/AnVIL_WES/{guid}/v1",
+    "projects_to_run": [
+        "{guid}"
+    ]
+}}```
+        """.format(user=self.manager_user.email, namespace=TEST_WORKSPACE_NAMESPACE, name=TEST_NO_PROJECT_WORKSPACE_NAME,
+                   guid=project.guid, vcf_path=TEST_VCF_PATH)
```
hanars, line 220: in tests wherever possible its better to hardcode the strings rather than string format. Other than guid I think you should put the literal strings in the template instead of formatting them here

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -183,27 +190,35 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
+            user=self.manager_user
         )
-        html_project_name = '<a href="http://testserver/project/{guid}/project_page">{name}</a>'.format(
-                name=TEST_NO_PROJECT_WORKSPACE_NAME, guid=project.guid)
-        mock_email.return_value.attach_alternative.assert_called_with(
-            email_body.format(project_name=html_project_name), 'text/html')
-        mock_email.return_value.send.assert_called()
+        slack_message = """
+        *{user}* requested to load WES data (GRCh38) from AnVIL workspace *{namespace}/{name}* at 
+        {vcf_path} to seqr project <http://testserver/project/{guid}/project_page|*anvil-no-project-workspace1*> (guid: {guid})
```
hanars, line 204: I'm not familiar with the `<http://...|*content*>` notation - what does it do?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -212,12 +227,12 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
         self.assertEqual(response.reason_phrase, 'Project "{name}" for workspace "{namespace}/{name}" exists.'
                          .format(namespace=TEST_WORKSPACE_NAMESPACE, name=TEST_NO_PROJECT_WORKSPACE_NAME))
 
-        # Test sending email exception
+        # Test sending slack message exception
         url = reverse(create_project_from_workspace, args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME2])
-        mock_email.side_effect = Exception('Something wrong while sending email.')
+        mock_slack.side_effect = Exception('Something wrong while sending the slack message.')
```
hanars, line 232: If you are mocking `safe_post_to_slack` you don't need to test error handling, as that function handles its own exceptions. If you want to test exception handling you need to mock it at a lower level, mocking the redis client in the helper function itself. Or you can opt not to test error handling since this helper function is tested elsewhere. Its your call

hanars (changes requested):

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I added a test file to test the helper.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: The markdown always starts a new line after ``` even there is no newline after it.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: It is a slack message format for the URL. The texts after '|' will be displayed.  The asterisk ('*') pair is for bold text.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -212,14 +225,15 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
 
-        # Test sending email exception
+        # Test saving ID file exception
         url = reverse(create_project_from_workspace, args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME2])
-        mock_email.side_effect = Exception('Something wrong while sending email.')
+        mock_mv_file.side_effect = Exception('Something wrong while moving the ID file.')
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
         self.assertEqual(response.status_code, 200)
-        mock_api_logger.error.assert_called_with('AnVIL loading request email exception: Something wrong while sending email.', self.manager_user)
+        mock_api_logger.error.has_called_with('Uploading sample IDs to Google Storage failed. Errors: Something wrong while moving the ID file.',
```
hanars, line 233: `has_called_with` is not a thing, this isn't testing anything

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -212,14 +225,15 @@ def test_create_project_from_workspace(self, mock_file_iter, mock_file_exist, mo
...
+                      self.manager_user, detail='s\nNA19675\nNA19678\nHG00735')
 
-        # Test logged in locally
+            # Test logged in locally
```
hanars, line 236: misaligned indent

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -155,12 +163,18 @@ def create_project_from_workspace(request, namespace, name):
...
+    sample_ids = '\n'.join(['s'] + [individual.individual_id for individual in updated_individuals])
     try:
-        _send_load_data_email(project, updated_individuals, data_path, request_json['sampleType'], request.user)
+        temp_path = save_temp_data(sample_ids)
+        mv_file_to_gs(temp_path, ids_path, user=request.user)
     except Exception as ee:
-        message = 'AnVIL loading request email exception: {}'.format(str(ee))
-        logger.error(message, request.user)
+        logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
+                     detail=sample_ids)
```
hanars, line 174: for the detail its better to send the list of ids rather than the string

hanars (changes requested):

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Thank you. I must be more careful with this kind of mistake.

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: Good to know that `detail` can be a list or any other object.

hanars (approved):

PR #2044, opened 2021-08-13
Performance improvements

hanars:
A variety of changes that speed up the react rendering time, mainly on the project pages. For RGP, reduces the initial JS rendering time from ~23 seconds to ~8 seconds

### DIFF
--- a/ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
+++ b/ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
@@ -5,9 +5,10 @@ import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { connect } from 'react-redux'
 
-import ReduxFormWrapper from 'shared/components/form/ReduxFormWrapper'
+import StateChangeForm from 'shared/components/form/StateChangeForm'
 import { Select } from 'shared/components/form/Inputs'
 import TextFieldView from 'shared/components/panel/view-fields/TextFieldView'
+import { camelcaseToTitlecase, toSnakecase } from 'shared/utils/stringUtils'
 import { updateIndividual } from 'redux/rootReducer'
 
 import {
@@ -19,6 +20,10 @@ const StatusContainer = styled.span`
   display: inline-block;
   whitespace: nowrap;
   min-width: 220px;
+  
+  .ui.form {
+    width: 100%;
+  }
   
   .ui.selection.active.dropdown .menu {
     max-height: none;
@@ -34,12 +39,9 @@ const STATUS_FORM_FIELDS = [{
 
 const CaseReviewStatusDropdown = React.memo(props =>
   <StatusContainer>
-    <ReduxFormWrapper
-      onSubmit={props.updateIndividualStatus}
-      form={`editCaseReviewStatus-${props.individual.individualGuid}`}
+    <StateChangeForm
+      updateField={props.updateIndividualField}
       initialValues={props.individual}
-      closeOnSuccess={false}
-      submitOnChange
       fields={STATUS_FORM_FIELDS}
     />
     {/* edit case review discussion for individual: */}
@@ -66,15 +68,15 @@ export { CaseReviewStatusDropdown as CaseReviewStatusDropdownComponent }
 
 CaseReviewStatusDropdown.propTypes = {
   individual: PropTypes.object.isRequired,
-  updateIndividualStatus: PropTypes.func.isRequired,
+  updateIndividualField: PropTypes.func.isRequired,
   updateIndividualDiscussion: PropTypes.func.isRequired,
 }
 
 const mapDispatchToProps = (dispatch, ownProps) => {
   return {
-    updateIndividualStatus: (updates) => {
+    updateIndividualField: individualField => (value) => {
       dispatch(updateIndividual({
-        individualGuid: ownProps.individual.individualGuid, individualField: 'case_review_status', ...updates }))
+        individualGuid: ownProps.individual.individualGuid, individualField: toSnakecase(camelcaseToTitlecase(individualField)), [individualField]: value }))
     },
     updateIndividualDiscussion: (updates) => {
       dispatch(updateIndividual({
--- a/ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
+++ b/ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
@@ -76,7 +76,10 @@ const mapDispatchToProps = (dispatch, ownProps) => {
   return {
     updateIndividualField: individualField => (value) => {
       dispatch(updateIndividual({
-        individualGuid: ownProps.individual.individualGuid, individualField: toSnakecase(camelcaseToTitlecase(individualField)), [individualField]: value }))
+        individualGuid: ownProps.individual.individualGuid,
+        individualField: toSnakecase(camelcaseToTitlecase(individualField)),
+        [individualField]: value,
+      }))
     },
     updateIndividualDiscussion: (updates) => {
       dispatch(updateIndividual({
--- a/ui/pages/Project/components/ProjectOverview.jsx
+++ b/ui/pages/Project/components/ProjectOverview.jsx
@@ -22,7 +22,7 @@ import {
   getAnalysisStatusCounts,
   getProjectAnalysisGroupFamiliesByGuid,
   getProjectAnalysisGroupIndividualsByGuid,
-  getProjectAnalysisGroupSamplesByGuid,
+  getProjectAnalysisGroupSamplesByTypes,
   getProjectAnalysisGroupMmeSubmissions,
 } from '../selectors'
 import EditFamiliesAndIndividualsButton from './edit-families-and-individuals/EditFamiliesAndIndividualsButton'
@@ -173,25 +173,11 @@ const mapMatchmakerStateToProps = (state, ownProps) => ({
 
 const MatchmakerOverview = connect(mapMatchmakerStateToProps)(Matchmaker)
 
-const Dataset = React.memo(({ project, samplesByGuid, user }) => {
-  const loadedProjectSamples = Object.values(samplesByGuid).reduce((acc, sample) => {
-    const loadedDate = (sample.loadedDate).split('T')[0]
-    if (!acc[sample.sampleType]) {
-      acc[sample.sampleType] = {}
-    }
-    if (!acc[sample.sampleType][sample.datasetType]) {
-      acc[sample.sampleType][sample.datasetType] = {}
-    }
-    acc[sample.sampleType][sample.datasetType] = {
-      ...acc[sample.sampleType][sample.datasetType],
-      [loadedDate]: (acc[sample.sampleType][sample.datasetType][loadedDate] || 0) + 1,
-    }
-    return acc
-  }, {})
+const Dataset = React.memo(({ project, samplesByType, user }) => {
 
-  const datasetSections = Object.keys(loadedProjectSamples).sort().reduce((acc, sampleType) => ([
+  const datasetSections = Object.keys(samplesByType).sort().reduce((acc, sampleType) => ([
     ...acc,
-    ...Object.entries(loadedProjectSamples[sampleType]).map(([datasetType, loadedSampleCounts]) => ({
+    ...Object.entries(samplesByType[sampleType]).map(([datasetType, loadedSampleCounts]) => ({
       key: `${sampleType}-${datasetType}`,
       title: `${SAMPLE_TYPE_LOOKUP[sampleType].text}${DATASET_TITLE_LOOKUP[datasetType] || ''} Datasets`,
       content: Object.keys(loadedSampleCounts).sort().map(loadedDate =>
@@ -236,13 +222,13 @@ const Dataset = React.memo(({ project, samplesByGuid, user }) => {
 
 Dataset.propTypes = {
   project: PropTypes.object.isRequired,
-  samplesByGuid: PropTypes.object.isRequired,
+  samplesByType: PropTypes.object.isRequired,
   user: PropTypes.object.isRequired,
 }
 
 const mapDatasetStateToProps = (state, ownProps) => ({
   user: getUser(state),
-  samplesByGuid: getProjectAnalysisGroupSamplesByGuid(state, ownProps),
+  samplesByType: getProjectAnalysisGroupSamplesByTypes(state, ownProps),
 })
 
 const DatasetOverview = connect(mapDatasetStateToProps)(Dataset)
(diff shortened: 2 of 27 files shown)


ui/pages/Project/components/ProjectPageUI.jsx
```diff
@@ -79,79 +76,64 @@ const NO_DETAIL_FIELDS = [
   { id: FAMILY_FIELD_DESCRIPTION, colWidth: 6 },
 ]
 
-const ProjectPageUI = React.memo((props) => {
-  const exportUrls = [
-    { name: 'Families', data: props.familyExportConfig },
-    { name: 'Individuals', data: props.individualsExportConfig },
-    { name: 'Samples', data: props.samplesExportConfig },
-  ]
-
```
hanars, line 88: this component did not change except `exportUrls` is a passed in props now instead of being computed here

ui/pages/Project/selectors.js
```diff
@@ -248,51 +283,37 @@ export const getVisibleFamilies = createSelector(
...
-    const getSortKey = FAMILY_SORT_LOOKUP[familiesSortOrder](individualsByGuid, samplesByGuid)
-
-    return orderBy(visibleFamilies, [getSortKey], [familiesSortDirection > 0 ? 'asc' : 'desc'])
+    const getSortKey = FAMILY_SORT_LOOKUP[familiesSortOrder](individualsByGuid, samplesByFamily)
+    return visibleFamilies.slice(0).sort((a, b) => getSortKey(a).localeCompare(getSortKey(b)) * familiesSortDirection)
   },
 )
 
-export const getEntityExportConfig = (project, rawData, tableName, fileName, fields) => ({
+export const getEntityExportConfig = ({ project, tableName, fileName, fields }) => ({
```
ShifaSZ, line 299: I have a question about why the object parameter is used rather than separate ones.  Is it just for optional parameters? How about using the default parameters like Python?

ShifaSZ (changes requested):
It is a huge project and looks great. To help me review it, can you add some brief explanations about how the performance is improved?

ui/pages/Project/selectors.js
hanars: because remembering an exact argument order is hard, especially when not all the fields are required and some calls to this function only provide some of them. Plus, using the object you kind of get to name the arguments which makes it more readable. So for instance with this change 2 different calls would look like 
```
getEntityExportConfig({ project, tableName: 'project_families', fileName: 'families', fields: FAMILY_EXPORT_DATA })
...
getEntityExportConfig({ project, fileName: 'template', fields })
```
Whereas using the separate parameters that would read like
```
getEntityExportConfig(project, 'project_families', 'families', FAMILY_EXPORT_DATA)
...
getEntityExportConfig(project, null, 'template', fields)
```
I just find it much more readable/ maintainable the way it is now

ui/pages/Project/components/FamilyTable/CaseReviewStatusDropdown.jsx
```diff
@@ -5,9 +5,10 @@ import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { connect } from 'react-redux'
 
-import ReduxFormWrapper from 'shared/components/form/ReduxFormWrapper'
+import StateChangeForm from 'shared/components/form/StateChangeForm'
```
hanars, line 8: `ReduxFormWrapper` holds  A LOT of state -  its designed to have users edit different fields and then hit submit and do something with all the form data, so the backing form library is keeping all the form field data in the state, including validating inputs and keeping validation state in the state. Its also designed to dispatch an action when the form submits and the render differently depending on whether it is currently submitting or whether submission has succeeded/ failed, which is more state. Also, its designed to be used in a modal, so it maintains the state of if the modal should be ope or closed. This is all really helpful because almost every form we use in seqr has this behavior so having this helper class that manages all that state is great. 

However, there are some places where we are using a "form" that doesn't do ANY of that. Its just a bunch of inline inputs that dispatch an acton for each separate input whenever that specific input changes. Theres no case where you ever need the state of the whole "form" input at once (no submit button at all), no validation, no success/ failure callbacks for submissions, no modal. So setting up and listening to all that state actually adds a bunch of unnecessary overhead. Therefore, I made a lightweight helper `StateChangeForm` that will render inputs that look the same as the other form class, but without setting up any of the state

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -154,91 +213,85 @@ const ProjectOverview = React.memo((
...
 
-  const mmeSubmissionCount = mmeSubmissions.length
-  const deletedSubmissionCount = mmeSubmissions.filter(({ deletedDate }) => deletedDate).length
+Dataset.propTypes = {
+  project: PropTypes.object.isRequired,
+  samplesByType: PropTypes.object.isRequired,
+  user: PropTypes.object.isRequired,
+}
 
-  return (
```
hanars, line 167: I broke this single big component that is connected to the state needed to render everything inside it into smaller helper components that only connect to the state they are actually using. This makes it more modular, and means that only the sub-components that actually need to re-render do so when the state changes, instead of rerendering the whole thing because part changed

ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -115,7 +115,7 @@ const BaseProjectSavedVariants = React.memo(({ project, analysisGroup, loadProje
 
   const getUpdateTagUrl = (newTag) => {
     const isCategory = categoryOptions.includes(newTag)
-    props.updateTable({ categoryFilter: isCategory ? newTag : null })
+    props.updateTableField('categoryFilter')(isCategory ? newTag : null)
```
hanars, line 118: this changed because the new `StateChangeForm` requires a different signature for change functions than before

ui/pages/Project/constants.js
```diff
@@ -155,8 +154,8 @@ export const FAMILY_FILTER_OPTIONS = [
     value: SHOW_DATA_LOADED,
     category: 'Data Status:',
     name: 'Data Loaded',
-    createFilter: (individualsByGuid, samplesByGuid) => family =>
-      familyVariantSamples(family, individualsByGuid, samplesByGuid).filter(sample => sample.isActive).length > 0,
+    createFilter: (individualsByGuid, user, samplesByFamily) => family =>
+      (samplesByFamily[family.familyGuid] || []).filter(sample => sample.isActive).length > 0,
```
hanars, line 158: the `samplesByFamily` selector will only be run once so these functions can reuse its computation, instead of recomputing the sample list for every single family every time the filter is applied

ui/pages/Project/selectors.js
```diff
@@ -55,21 +54,35 @@ export const getCurrentProject = createSelector(
...
 export const getProjectFamiliesByGuid = createSelector(getFamiliesGroupedByProjectGuid, getProjectGuid, selectEntitiesForProjectGuid)
 export const getProjectAnalysisGroupsByGuid = createSelector(getAnalysisGroupsGroupedByProjectGuid, getProjectGuid, selectEntitiesForProjectGuid)
+const getProjectSamplesByGuid = createSelector(getSamplesGroupedByProjectGuid, getProjectGuid, selectEntitiesForProjectGuid)
+
+const getAnalysisGroupGuid = (state, props) => ((props || {}).match ? props.match.params.analysisGroupGuid : (props || {}).analysisGroupGuid)
+
+const getCurrentAnalysisGroup = createSelector(
+  getProjectAnalysisGroupsByGuid,
+  getAnalysisGroupGuid,
+  (projectAnalysisGroupsByGuid, analysisGroupGuid) => analysisGroupGuid && projectAnalysisGroupsByGuid[analysisGroupGuid],
```
hanars, line 64: this just makes it more readable, it was really confusing how the props were getting the analysis group in every seperate function

ui/pages/Project/selectors.js
```diff
@@ -55,21 +54,35 @@ export const getCurrentProject = createSelector(
...
+    return analysisGroup.familyGuids.reduce(
       (acc, familyGuid) => ({ ...acc, [familyGuid]: projectFamiliesByGuid[familyGuid] }), {},
     )
   },
 )
 
+export const getProjectAnalysisGroupIndividualsCount = createSelector(
+  getProjectAnalysisGroupFamiliesByGuid,
+  familiesByGuid =>
+    Object.values(familiesByGuid).reduce((acc, family) => acc + family.individualGuids.length, 0),
```
hanars, line 83: running a count is much faster than keeping track of all the objects and then counting the lengths at the end, especially because it saves a ton of memory usage

ui/pages/Project/selectors.js
```diff
@@ -82,28 +95,34 @@ export const getProjectAnalysisGroupIndividualsByGuid = createSelector(
...
-  getSamplesByGuid,
-  getProjectAnalysisGroupIndividualsByGuid,
-  (samplesByGuid, individualsByGuid) =>
-    Object.values(individualsByGuid).reduce((acc, individual) => ({
-      ...acc,
-      ...individual.sampleGuids.reduce((sampleAcc, sampleGuid) => (
-        { ...sampleAcc, [sampleGuid]: samplesByGuid[sampleGuid] }
-      ), {}),
-    }), {}),
+export const getProjectAnalysisGroupSamplesByTypes = createSelector(
```
hanars, line 98: before, we were formatting data in a selector and then post-processing it in the render method of the component. This just makes the selector process the data so it is exactly the way the component needs it and no computation is redone if a rerender occurs. It also makes use of the already-computed `samplesByFamily`, instead of recomputing that here

ui/pages/Project/selectors.js
```diff
@@ -82,28 +95,34 @@ export const getProjectAnalysisGroupIndividualsByGuid = createSelector(
...
+      acc[typeKey][loadedDate] = (acc[typeKey][loadedDate] || 0) + 1
+      return acc
+    }, {}),
 )
 
 export const getProjectAnalysisGroupMmeSubmissions = createSelector(
   getMmeSubmissionsByGuid,
   getProjectAnalysisGroupFamiliesByGuid,
-  getProjectAnalysisGroupIndividualsByGuid,
+  getIndividualsByFamily,
```
hanars, line 119: `getProjectAnalysisGroupIndividualsByGuid` did not otherwise have to run for the main page render but `getIndividualsByFamily` so refactoring this allows us to skip running that selector and its computation entirely

ui/pages/Project/selectors.js
```diff
@@ -207,33 +226,49 @@ export const getFamiliesSortDirection = createSelector(
...
   getProjectAnalysisGroupFamiliesByGuid,
+  getFamiliesBySearchString,
   getIndividualsByGuid,
-  getSamplesByGuid,
+  getSamplesByFamily,
   getUser,
   getFamiliesFilter,
   getFamiliesSearch,
-  (familiesByGuid, individualsByGuid, samplesByGuid, user, familiesFilter, familiesSearch) => {
-    const searchFilter = familiesSearch ? family =>
```
hanars, line 224: before, whenever anyone would type a new character into the filter bar we would re-compute the search string for each family and check if it matched the new filter. Now, we have one selector that will compute all the search strings if anything has been searched, and then we reuse that computation whenever the filter value changes to check for the new search string in them. This lets us reuse a lot of computations, which is good

ui/redux/selectors.js
```diff
@@ -65,94 +64,90 @@ export const getFamiliesGroupedByProjectGuid = createSelector(getFamiliesByGuid,
...
- *
- * @param state {object} global Redux state
- */
-export const getSortedIndividualsByFamily = createSelector(
-  getIndividualsByFamily,
+const getSortedIndividuals = createSelector(
+  getIndividualsByGuid,
   getMmeSubmissionsByGuid,
-  (familyIndividuals, mmeSubmissionsByGuid) => {
+  (individualsByGuid, mmeSubmissionsByGuid) => {
```
hanars, line 84: sorting all the individuals, and then going through them in order and grouping them by family is WAY faster than grouping them and then sorting each grouped list. Made the same change for the sorted samples as well

ui/redux/selectors.js
```diff
@@ -65,94 +64,90 @@ export const getFamiliesGroupedByProjectGuid = createSelector(getFamiliesByGuid,
...
-            familyAcc[type] = {}
-          }
-          familyAcc[type][individualGuid] = sample
-        })
-        return familyAcc
-      }, {}),
-    }), {})
-  },
+  (individualsByGuid, igvSamplesByGuid) =>
+    Object.values(igvSamplesByGuid).reduce((acc, sample) => {
```
hanars, line 140: doing a single iteration through all the samples is faster than going through the family-individual groups and then iterating thorough all the individuals within that

ui/shared/components/graph/HorizontalStackedBar.jsx
```diff
@@ -21,10 +21,10 @@ const BarContainer = styled.div.attrs(props => ({
   text-align: center;
   border: 1px solid gray;`
 
-const BarSection = styled(({ to, ...props }) => React.createElement(to ? Link : 'div', { to, ...props }))`  
+const BarSection = styled(({ to, ...props }) => React.createElement(to ? Link : 'div', { to, ...props })).attrs(
```
hanars, line 24: if we have a lot of the same bar graph showing different breakdowns, like on the project page where we have a graph for each family's tags, this change means we compute one set of classes used by all of them instead of computing a different set of classes for each graph to make the breakdown render the correct percentages

ui/shared/components/panel/family.jsx
```diff
@@ -58,7 +58,7 @@ BaseFirstSample.propTypes = {
 }
 
 const mapSampleDispatchToProps = (state, ownProps) => ({
-  firstFamilySample: getFirstSampleByFamily(state)[ownProps.familyGuid],
+  firstFamilySample: (getSamplesByFamily(state)[ownProps.familyGuid] || [])[0],
```
hanars, line 61: `getFirstSampleByFamily` didn't do anything other than this one line, so storing essentially a copy of the data in memory wasn't worth it

ui/shared/utils/constants.js
```diff
@@ -389,14 +388,6 @@ export const INDIVIDUAL_EXPORT_DATA = [].concat(
   INDIVIDUAL_ID_EXPORT_DATA, INDIVIDUAL_CORE_EXPORT_DATA, [INDIVIDUAL_HAS_DATA_EXPORT_CONFIG], INDIVIDUAL_HPO_EXPORT_DATA,
 )
 
-export const familyVariantSamples = (family, individualsByGuid, samplesByGuid) => {
```
hanars, line 392: as explained above with individuals, it is way faster to have one selector where everything is sorted, rather than sorting each sub-group individually

hanars:
> It is a huge project and looks great. To help me review it, can you add some brief explanations about how the performance is improved?

I went through and added comments explaining where a lot of the improvements are coming from

ShifaSZ (approved):
It is an amazing optimization. I can't get all the reasoning behind it, but I can see the outcomes are quite good.

PR #2046, opened 2021-08-16
Support alias for elastisearch index

hanars:


seqr/views/apis/dataset_api_tests.py
```diff
@@ -296,6 +202,150 @@ def test_add_variants_dataset(self, mock_open, mock_random, mock_redis):
         self.assertSetEqual({sv_sample_guid, existing_index_sample_guid, new_sample_type_sample_guid}, {sample.guid for sample in sample_models})
         self.assertSetEqual({True}, {sample.is_active for sample in sample_models})
 
+    @urllib3_responses.activate
+    def test_add_variants_dataset_errors(self):
+        url = reverse(add_variants_dataset_handler, args=[PROJECT_GUID])
+        self.check_data_manager_login(url)
```
ShifaSZ, line 208: I assume the tests for the errors have no change.

ShifaSZ (changes requested):
I have some questions which need to be confirmed.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -79,6 +79,15 @@ def __init__(self, families, previous_search_results=None, inheritance_search=No
...
             ))
+        elif len(self.index_metadata) > len(self.samples_by_family_index):
+            # One of the indices is an alias
+            additional_meta_indices = set(self.index_metadata.keys()) - set(self._indices)
+            alias = next(ind for ind in self._indices if ind not in self.index_metadata)
+            self._indices.remove(alias)
+            self._indices += additional_meta_indices
+            alias_samples = self.samples_by_family_index.pop(alias)
+            for alias_index in additional_meta_indices:
+                self.samples_by_family_index[alias_index] = alias_samples
```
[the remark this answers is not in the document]
hanars: oh good point. I'll add handling for multiple indices

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -2403,6 +2404,40 @@ def test_get_es_variants_index_alias(self):
...
+                ],
+                '_name': 'F000011_11'
+            }}]
+        }
+        self.assertExecutedSearches([
+            dict(index=NO_LIFT_38_INDEX_NAME, **expected_search),
+            dict(index=INDEX_NAME, **expected_search),
+        ])
+        _set_cache('index_metadata__{}'.format(SECOND_INDEX_NAME), None)
+
```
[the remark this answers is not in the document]
hanars: no, since the results are mocked and theres no special processing for alias search theres no point in testing the results. This is consistent with many of the tests. in this file

seqr/views/apis/dataset_api_tests.py
hanars: I added a couple, and the order had to change because its now calling validate on the index metadata before it is validating that there are samples in the index, instead of the other way around. But nothing was removed. I'll comment on the new ones so you can find them

seqr/views/apis/dataset_api_tests.py
```diff
@@ -296,6 +202,150 @@ def test_add_variants_dataset(self, mock_open, mock_random, mock_redis):
...
+                'sampleType': 'WES',
+                'genomeVersion': '37',
+                'sourceFilePath': 'test_data.vds',
+                'datasetType': 'SV',
+            }, "properties": MAPPING_PROPS_WITH_SAMPLES}}})
+        response = self.client.post(url, content_type='application/json', data=ADD_DATASET_PAYLOAD)
+        self.assertEqual(response.status_code, 400)
+        self.assertDictEqual(response.json(), {'errors': ['Index "test_index" has dataset type SV but expects VARIANTS']})
+
+        urllib3_responses.replace_json('/{}/_mapping'.format(INDEX_NAME), {
```
hanars, line 271: this and the following test case are new to test the alias validation

ShifaSZ (approved):
Can't be better.

PR #2048, opened 2021-08-17
add new es data disks and nodes

hanars:
Changes required to implement stage 1 for adding new es data nodes as outlined in this ticket: https://github.com/broadinstitute/seqr-private/issues/1024

This would be taken live by running the following
```
# Create the new disks
./servctl deploy init-cluster gcloud-prod
# Create the new ES nodes
./servctl deploy elasticsearch gcloud-prod
```


deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
```diff
@@ -166,6 +166,61 @@ spec:
         resources:
           requests:
             storage: {{ ES_DATA_DISK_SIZE }}
+  - name: data
+    count:  {{ ES_DATA_OLD_NUM_PODS }}
```
hanars, line 170: this is a direct copy-paste of the data configuration above, but changed out `ES_DATA_NUM_PODS` for `ES_DATA_OLD_NUM_PODS ` and `ES_DATA_DISK_SIZE` for `ES_DATA_OLD_DISK_SIZE`

deploy/kubernetes/shared-settings.yaml
```diff
@@ -31,7 +33,8 @@ ELASTICSEARCH_JVM_MEMORY: 22g
 ELASTICSEARCH_CONTAINER_MEMORY: 44Gi
 ELASTICSEARCH_MAX_SHARDS_PER_NODE: 450
```
sjahl, line 34: Should we also have corresponding _OLD values for container memory and max shards, to account for the smaller disk?

sjahl (commented):
This looks OK to me -- just the one question in-line with regards to the container memory and max shard count.

My understanding of these changes are that we're adding a new pool called data-nodes, and the old data pool is staying as-is, just with some added `_OLD_` variables to keep the old values intact.

deploy/kubernetes/shared-settings.yaml
hanars: container memory has to do with the k8 pods which aren't changing so that should be the same. And max shards per node is a function of heap memory, not disk size so that also stays the same

hanars: Unless you think we should stop requesting the full node pool memory for each pod?

deploy/servctl_utils/other_command_utils.py
```diff
@@ -79,9 +79,12 @@ def delete_all(deployment_target):
 
 
 def get_disk_names(disk, settings):
-    num_disks = settings.get('{}_NUM_DISKS'.format(disk.upper().replace('-', '_'))) or 1
+    disk_setting_prefix = disk.upper().replace('-', '_')
+    num_disks = settings.get('{}_NUM_DISKS'.format(disk_setting_prefix)) or 1
+    name_suffix = settings.get('{}_DISK_NAME_SUFFIX'.format(disk_setting_prefix)) or ''
```
ShifaSZ, line 84: Is it more neatly by using `settings.get(ENV_STRING, default_value)`?

ShifaSZ (changes requested):

deploy/servctl_utils/other_command_utils.py
hanars: this handles the empty string better

ShifaSZ (approved):
Looks good to me.

deploy/kubernetes/shared-settings.yaml
sjahl: Is the container memory variable used as the resource limit for each data pod? Or is the memory limit the total shared between all pods in the elasticsearch cluster?

My confusion here is: if that container memory variable is used to set the request/limit for each pod, then I'm surprised that a data pod with 1TB of disk space would require as much memory as a data pod with 3+TBs of disk attached... but this is mostly drawn from experience with other indexing systems. So I'm open to being surprised by elasticsearch 😀

sjahl: Just looking at https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#shard-size-best-practices

My assumption was that smaller disks might impact shard size. And there is a few memory/disk considerations to think about if the shard size is changing.

hanars: So the disk size is unrelated to *shard size* - shard size is determined in the loading pipeline when we decide how many shards a given index should be, which we explicitly set when we load the data. It might be worth investigating our choices in the pipeline to better optimize shard size, but those would be unrelated changes in a different repo.  

The *number of shards per node* is limited by available heap memory - 
```
The number of shards you can hold on a node will be proportional to the amount of heap you have 
available, but there is no fixed limit enforced by Elasticsearch. 
A good rule-of-thumb is to ensure you keep the number of shards per node below 20 per GB heap it has configured.
```
So the heap memory is actually what we are using to determine our shard limit. Heap memory is set via the `ELASTICSEARCH_JVM_MEMORY` variable which is used in the `ES_JAVA_OPTS` env variable for ECK. Elasticsearch documentation states you want your heap to be no more than 50% of available memory, thus that variable is currently half of the `ELASTICSEARCH_CONTAINER_MEMORY`. And then `ELASTICSEARCH_MAX_SHARDS_PER_NODE` is roughly 20x the heap memory

I don't really know how smaller disks will impact node memory usage. We are currently using 76-82% of available heap for our data nodes which seems to me like a good utilization. Presumably, nodes with fewer shards need less memory and since we will be splitting the shards into 11 instead of 3 there will be substantially fewer shards on each node. 

The other question is how many nodes we will need in our data node pool. Currently the node pool uses `n1-highmem-8` which have just under 48Gb of allocatable memory. Therefore, with the elasticsearch data node pods requesting 44Gb they each get their own node. If we dropped the resource requests by half, we could have 2 pods per kubernetes node which would allow us to save a ton of resources. However, if we reduce the memory by less than half we would still only fit one pod (data node) per kubernetes node and in that case we just have a lot of unused resources. So if we don't feel safe decreasing the available memory by that much I'd rather leave it as is and have all our resources be available.

hanars:
Discussed offline and are going with a different approach, see ticket for details

Issue #1891, opened 2021-05-29
Custom feature implementation to integrate with PanelApp


tommyli:
We are planning to implement integration with [PanelApp AU](https://panelapp.agha.umccr.org/), a custom feature for
Seqr.  PanelApp is a service allowing researchers and experts curate and share Gene Lists.  These gene lists are
accessible via a REST API.

We've listed a few things to consider:

* Should implementation be full client side logic since PanelApp already provide an API?  It seems overkill otherwise,
  especially for the simple use case of browsing panels and importing genes from panels.
* However, more sophisticated use cases would require server side integration.  E.g. searching for panels by gene names.
* Server side integration is also more extensible.  E.g. a researcher have already asked for exclusion of incidental
  finding genes (presumably some list like [Incidentalome](https://panelapp.agha.umccr.org/panels/126/) from PanelApp AU
  or from [ACMG Recommendations for Reporting of Incidental Findings in Clinical Exome and Genome
  Sequencing](https://www.ncbi.nlm.nih.gov/clinvar/docs/acmg/)) when doing Location search.
* Server side implementation is also feasible because:
  * PanelApp provides APIs to download all genes per sample
  * Data is not that large, only hundreds of panels (257 as of writing)
  * Data growth rate is not that large
  * Data staleness is not an issue because these lists are **manually curated** and maintained by community in the first
    place
* There should be some Django config setting to feature toggle this (e.g. setting values are `None`, `PanelAppAustralia`,
  `PanelAppGenomicsEngland`).
* The obvious page to show this feature is in the [Gene Lists
  page](https://seqr.broadinstitute.org/summary_data/gene_lists).
* Is your team going to implement something similar?

We would also like to use this as an opportunity for us to collaborate closely with your team.  Implementing this as a
feature toggle that's off by default should be low risk but it is our plan to have this feature eventually submitted for
PR.  Would your team consider accepting a feature like this?

hanars:
Overall, sounds very interesting. I would have to reach out to our team to see if this is something we would be interested in taking live here at the Broad, but I think either way we would be open to taking this in to our main branch as a feature that is off by default, as I agree there is little harm there. And even if the Broad doesn't plan to use it, it could certainly be valuable to other groups with their own installations. 

We currently have no plans to work on anything like this, so you should feel free to go ahead without worrying about duplicating efforts. 

[...]

Issue #1983, opened 2021-07-22
Require descriptions for new gene lists


stephditroia:
**Is your feature request related to a problem? Please describe.**
There are several public gene lists without descriptions. We would like to make descriptions mandatory when creating a new gene list. The hover-over can provide specific guidance on what to enter in the description if needed. 

**Additional context**
![image](https://user-images.githubusercontent.com/44381119/126708651-9c266949-ca68-4a59-b96c-b8d426c2e33d.png)

hanars:
added

PR #2052, opened 2021-08-19
Backend PanelApp implementation for #1891

tommyli:
Backend implementation of PanelApp ready for your team to review (#1891).  In particular:

* Is using Django apps + ENV var a good approach for keeping such features optional?  If your team is happy with it then it may be a good model going forward to collaborate new features on?
* Another approach is to implement it similar to other `reference_data_*` but I feel current impl keeps panelapp even more separate. 
* The UI is not implemented but the changes to the store are what's needed for the UI store to hold new info for PanelApp fields.  These changes should be safe regardless if `panelapp` is installed in the backend.


tommyli:
I'm having trouble understanding why the Codacy docstring issues appear when I'm following the same convention as other docstrings in the codebase.

ui/redux/selectors.js
```diff
@@ -20,6 +20,7 @@ export const getVariantFunctionalDataByGuid = state => state.variantFunctionalDa
 export const getMmeSubmissionsByGuid = state => state.mmeSubmissionsByGuid
 export const getMmeResultsByGuid = state => state.mmeResultsByGuid
 export const getGenesById = state => state.genesById
+export const getPaGenesById = state => state.pagenesById
```
hanars, line 23: you don't seem to use this data anywhere, or the `paLocusList` data on the locus list itself. Is there a separate PR where you show this information? If not, it would be way more performant not to fetch/ return the data to the client

hanars:
> I'm having trouble understanding why the Codacy docstring issues appear when I'm following the same convention as other docstrings in the codebase.

because like 99% of the docstrings in seqr fail codacy, we've never gone back and fixed legacy issues

panelapp/panelapp_utils.py
```diff
@@ -0,0 +1,146 @@
...
+            'version': panel_json.get('version') or None,
+            'version_created': panel_json.get('version_created') or None,
+            'raw_data': panel_json,
+        }, user)
+
+        return existing
+    else:
+        seqr_locus_list = create_model_from_json(SeqrLocusList, {
+            'name': panel_json['name'],
+            'description': None,
```
hanars, line 127: It probably would be clearer for users if the description indicated that this was imported from panel app. Either that or wherever you show the description in the UI, show some additional panel app data to make it clearer

hanars:
> * Is using Django apps + ENV var a good approach for keeping such features optional?  If your team is happy with it then it may be a good model going forward to collaborate new features on?

I think its a pretty clean approach to conditionally add functionality, and for future feature requests I think it would be a good approach

> * The UI is not implemented but the changes to the store are what's needed for the UI store to hold new info for PanelApp fields.  These changes should be safe regardless if `panelapp` is installed in the backend.

The changes to the UI are safe as is, but not very useful. Based on this branch, if there are any UI enhancements that come with the panel app no one using this repo would be able to see them. I think it would be better to include the code changes for the UI display and either 
- set a constant in the metadata that is used by the client code to conditionally show features, similar to `googleLoginEnabled` OR
- make sure the client code gracefully handles the case where the store variables are empty/ null and don't change the display in that case. 

However, after talking to the team we are actually going to take this live for our group! So we are happy to take this code as-is and set the neccessary environment variable to enable it, or if you would prefer to take out the conditionals we are happy to take it as an unconditional part of the seqr code base as well

.github/workflows/unit-tests.yml
```diff
@@ -55,6 +55,12 @@ jobs:
       run: |
         coverage run --source="./matchmaker","./seqr","./reference_data" --omit="*/migrations/*","*/apps.py" manage.py test -p '*_tests.py' -v 2 reference_data seqr matchmaker
         coverage report --fail-under=99
+    - name: Run coverage tests for panelapp
+      env:
+        ENABLE_PANEL_APP: "True"
+      run: |
+        coverage run --source="./panelapp" --omit="*/migrations/*","*/apps.py" manage.py test -p '*_tests.py' -v 2 panelapp
+        coverage report --fail-under=99
```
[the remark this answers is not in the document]
tommyli: Yeah, I initially tried something like `@mock.patch` but it didn't work because the tests required `panelapp` DB migrations.  There are ways to mock in `settings.installed_apps` too but it made the test a bit cumbersome so I thought this was a cleaner approach.

panelapp/views.py
```diff
@@ -0,0 +1,13 @@
+from panelapp.panelapp_utils import import_all_panels
+from seqr.utils.logging_utils import SeqrLogger
+from seqr.views.utils.json_utils import create_json_response
+from seqr.views.utils.permissions_utils import data_manager_required
+
+logger = SeqrLogger(__name__)
+
+
+@data_manager_required
+def import_panelapp_handler(request):
```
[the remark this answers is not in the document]
tommyli: Fair enough.  Our instance can only be accessed via VPN so we're a little spoilt in the security sense.

We were going to implement a UI trigger for this but your idea of using a cron job on deployed host is probably even better and safer.  Will remove this and keep the manage command.

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -722,9 +728,17 @@ def _process_result(result, locus_list):
...
+            pa_locus_list_json = _get_json_for_model(locus_list.palocuslist, user=user)
+            result.update({
+                'paLocusList': pa_locus_list_json,
+            })
+
     prefetch_related_objects(locus_lists, 'created_by')
     prefetch_related_objects(locus_lists, 'locuslistgene_set')
     prefetch_related_objects(locus_lists, 'locuslistinterval_set')
+    if ENABLE_PANEL_APP:
+        prefetch_related_objects(locus_lists, 'palocuslist')
```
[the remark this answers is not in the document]
tommyli: Will fix.

ui/redux/selectors.js
tommyli: See answer below regarding UI.

panelapp/panelapp_utils.py
tommyli: Yes agree, it'll be more consistent with requests like [Require descriptions for new gene lists](https://github.com/broadinstitute/seqr/issues/1983).  There are a few fields from `PaLocusList` I could use to populate the description but we have yet to agree on how PanelApp gene lists should be displayed.  See answer below regarding UI.

tommyli:
> However, after talking to the team we are actually going to take this live for our group! So we are happy to take this code as-is and set the neccessary environment variable to enable it, or if you would prefer to take out the conditionals we are happy to take it as an unconditional part of the seqr code base as well

Great!  I will append to this PR with fixes that addresses some of the above issues and remove the `ENABLE_PANEL_APP` conditionals.  That'll make the codebase even cleaner.  The history of this PR will retain the history on an approach to feature toggle.

tommyli:
> The changes to the UI are safe as is, but not very useful. Based on this branch, if there are any UI enhancements that come with the panel app no one using this repo would be able to see them. I think it would be better to include the code changes for the UI display and either

Indeed, a separate PR will be raised for the UI changes.  Here are some of our thoughts:
* I was so happy to see your team implemented [Gene list search](https://github.com/broadinstitute/seqr/pull/2051) :).  It's great timing so whatever we do, it'll be built on top of that.  From the backend changes of #2051, I may end up raising a new PR that supersedes this current one and raise another PR for the UI.
* Whatever the UI design, we will ensure the UI will handle cases whether PanelApp gene lists exists or not.
* The additional fields from PanelApp included in `PaLocusList` do not really "fit" into the current UI table.  We may end up having a third section (i.e. "My Gene Lists" + "Public Gene Lists" + "PanelApp").  However, the users are happy to simply have those fields crammed into `description`, that'll be good.
* The Gene Lists details page will also be updated to include additional data such as `confidence_level`.

Future changes, most likely separate PRs.
* PanelApp genes also includes data such as "mode of inheritance" and "disease phenotypes".  These will be incorporated in variant search results.
* Add new search/filter in the gene lists filter when doing variant search

Any comments/thoughts/ideas welcome of course.

hanars:
One comment on separate PRs is we don't like to merge incomplete functionality. So if we want to keep the front end code to a separate PR thats fine, but then this PR shouldn't have any front-end changes in it, including changes to the data returned to the client. right now if we released this code we would be slowing down a bunch of API requests to return data we wouldn't use, which isn't great. So I would say wither we want this PR to include some of the basic UI enhancements, maybe populating the "description" column with structured data from panel app if it exists, or we should seperate all that functionality into the future UI PR

tommyli:
Closing, please refer to new PR #2107 which supersedes this.

PR #2057, opened 2021-08-19
fix bug in search results

hanars:


ShifaSZ (approved):
Can't find any problems.

PR #2060, opened 2021-08-20
Send a slack or email message when an AnVIL dataset has been added to the project.

ShifaSZ:
See the title and the connected issue.

### DIFF
--- a/seqr/views/apis/dataset_api.py
+++ b/seqr/views/apis/dataset_api.py
@@ -121,7 +121,7 @@ def add_variants_dataset_handler(request, project_guid):
         Data Manager from seqr
         """.format(
             user=request.user,
-            date=loaded_date.date(),
+            date=project.created_date.date().strftime('%B %d, %Y'),
             namespace=project.workspace_namespace,
             name=project.workspace_name,
             proj_name=project.name,
--- a/seqr/views/apis/dataset_api.py
+++ b/seqr/views/apis/dataset_api.py
@@ -4,12 +4,14 @@ from collections import defaultdict
 from django.db.models import prefetch_related_objects
 from django.utils import timezone
 
+from settings import SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL
+from seqr.utils.communication_utils import send_html_email, safe_post_to_slack
 from seqr.models import Individual, Sample, Family
 from seqr.views.utils.dataset_utils import match_sample_ids_to_sample_records, \
     validate_index_metadata_and_get_elasticsearch_index_samples, load_mapping_file
 from seqr.views.utils.json_utils import create_json_response
-from seqr.views.utils.orm_to_json_utils import get_json_for_samples
-from seqr.views.utils.permissions_utils import get_project_and_check_permissions, data_manager_required
+from seqr.views.utils.orm_to_json_utils import get_json_for_samples, get_project_collaborators_by_username
+from seqr.views.utils.permissions_utils import get_project_and_check_permissions, data_manager_required, has_analyst_access
 
 
 @data_manager_required
@@ -109,6 +111,31 @@ def add_variants_dataset_handler(request, project_guid):
     ]
     Family.bulk_update(
         request.user, {'analysis_status': Family.ANALYSIS_STATUS_ANALYSIS_IN_PROGRESS}, guid__in=family_guids_to_update)
+
+    if project.workspace_name and project.workspace_namespace:
+        message_content = """
+        Hi {user},
+        We are following up on your request to load data from AnVIL on {date}.
+        We have loaded data from the AnVIL workspace “{namespace}/{name}” to the corresponding seqr project {proj_name}. {sample_count} samples are currently loaded. Let us know if you have any questions.
+        Thanks,
+        Data Manager from seqr
+        """.format(
+            user=request.user,
+            date=loaded_date.date(),
+            namespace=project.workspace_namespace,
+            name=project.workspace_name,
+            proj_name=project.name,
+            sample_count=len(matched_sample_id_to_sample_record),
+        )
+        if has_analyst_access(project):
+            safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, message_content)
+        else:
+            send_html_email(
+                message_content,
+                subject='AnVIL data have been loaded into seqr',
+                to=sorted([user['email'] for user in get_project_collaborators_by_username(request.user, project).values()
+                           if user['hasEditPermissions']]),
+            )
 
     response_json = _get_samples_json(matched_sample_id_to_sample_record, inactivate_sample_guids, project_guid)
     response_json['familiesByGuid'] = {family_guid: {'analysisStatus': Family.ANALYSIS_STATUS_ANALYSIS_IN_PROGRESS}
--- a/seqr/views/utils/permissions_utils.py
+++ b/seqr/views/utils/permissions_utils.py
@@ -67,7 +67,7 @@ pm_or_data_manager_required = _user_has_policies_and_passes_test(
 superuser_required = _user_has_policies_and_passes_test(lambda user: user.is_superuser)
 
 
-def _has_analyst_access(project):
+def has_analyst_access(project):
     return project.projectcategory_set.filter(name=ANALYST_PROJECT_CATEGORY).exists()
 
 def get_project_and_check_permissions(project_guid, user, **kwargs):
@@ -155,7 +155,7 @@ def has_project_permissions(project, user, can_edit=False):
         permission_level = CAN_EDIT
 
     return user_is_data_manager(user) or \
-           (user_is_analyst(user) and _has_analyst_access(project)) or \
+           (user_is_analyst(user) and has_analyst_access(project)) or \
            user.has_perm(permission_level, project) or \
            anvil_has_perm(user, permission_level, project)
 


hanars:
This workflow does not match what is specified in the ticket.  If it is an internal project, regardless of if the project is in AnVIL, it needs to send a slack message to the `seqr-data-loading` channel. The slack messages need to match the format of the slack messages that are already being manually sent in that channel, including all the new sample IDS. A sample slack message will look like: 

X new samples are loaded in https://seqr.broadinstitute.org/project/<guid>/project_page
```
sample_1, sample_2, ...
```

ShifaSZ:
> he slack messages need to match the format of the slack messages that are already being manually sent in that channel, including all the new sample IDS

It seems that I don't have access to the `seqr-data-loading` channel. Can you add me to the channel?

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
     Family.bulk_update(
         request.user, {'analysis_status': Family.ANALYSIS_STATUS_ANALYSIS_IN_PROGRESS}, guid__in=family_guids_to_update)
 
+    if project.workspace_name and project.workspace_namespace:
```
hanars, line 115: So as I've said before, this flow does not match the one specified in the ticket. If it is as an internal project, REGARDLESS OF WHETHER IT IS IN ANVIL OR NOT, it needs to send the notification

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
...
+        if has_analyst_access(project):
+            safe_post_to_slack(
+                SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL,
+                """{num_sample} new samples are loaded in https://seqr.broadinstitute.org/project/{guid}/project_page
+                ```{samples}```
+                """.format(
+                    num_sample=len(matched_sample_id_to_sample_record),
```
hanars, line 122: so this is not the number of new samples that were added, this is the number of total samples that were added. We need to let them know which samples were added that never had a record in seqr before

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
...
+                ))
+        else:
+            send_html_email(
+                """Hi {user},
+We are following up on your request to load data from AnVIL on {date}.
+We have loaded data from the AnVIL workspace “{namespace}/{name}” to the corresponding seqr project {proj_name}. {num_sample} samples are currently loaded. Let us know if you have any questions.
+Thanks,
+Data Manager from seqr
+""".format(
+                    user=request.user,
```
hanars, line 134: if we have a display name for them (first/ last) we want to use that, otherwise we should use their email. We never want to use the user id

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
...
+                """.format(
+                    num_sample=len(matched_sample_id_to_sample_record),
+                    guid=project.guid,
+                    samples=[id for id in matched_sample_id_to_sample_record.keys()]
+                ))
+        else:
+            send_html_email(
+                """Hi {user},
+We are following up on your request to load data from AnVIL on {date}.
+We have loaded data from the AnVIL workspace “{namespace}/{name}” to the corresponding seqr project {proj_name}. {num_sample} samples are currently loaded. Let us know if you have any questions.
```
hanars, line 130: we should have a link to the workspace and the project page. Also, lets redo the content of the email to 
```
Hi {user},
We are following up on your request to load data from AnVIL to seqr on {date}.
We have loaded {num_sample} samples from the AnVIL workspace <a>{namespace}/{name}</a> to the corresponding seqr project <a>{proj_name}</a>. 
- The seqr team
```

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
...
+Data Manager from seqr
+""".format(
+                    user=request.user,
+                    date=project.created_date.date().strftime('%B %d, %Y'),
+                    namespace=project.workspace_namespace,
+                    name=project.workspace_name,
+                    proj_name=project.name,
+                    num_sample=len(matched_sample_id_to_sample_record),
+                ),
+                subject='AnVIL data have been loaded into seqr',
```
hanars, line 141: Lets change this to "New data available in seqr"

seqr/views/apis/dataset_api.py
```diff
@@ -110,6 +112,37 @@ def add_variants_dataset_handler(request, project_guid):
...
+                to=sorted([user['email'] for user in get_project_collaborators_by_username(request.user, project).values()
```
hanars, line 142: you should just directly fetch the managers, rather than fetching the managers, fetching the collaborators, merging those lists together, and then separating them out. Theres also really no need to convert the collaborator list to json so you can then pull one field out

seqr/views/apis/dataset_api_tests.py
```diff
@@ -63,6 +67,7 @@ def test_add_variants_dataset(self, mock_random):
 
         mock_random.return_value = 98765432101234567890
 
+        mock_analyst_access.return_value = False
```
hanars, line 70: there are actual test fixture projects that do and do not have this access. Use the correct fixture data, do not mock what are essentially database values

seqr/views/utils/permissions_utils.py
```diff
@@ -67,7 +67,7 @@ def decorator(view_func):
 superuser_required = _user_has_policies_and_passes_test(lambda user: user.is_superuser)
 
 
-def _has_analyst_access(project):
+def has_analyst_access(project):
```
hanars, line 70: if this function is being used outside the file it needs a clearer name. Maybe `project_has_analyst_access`

hanars (changes requested):

seqr/views/apis/dataset_api.py
ShifaSZ: I see. I need to switch around the conditions for internal and AnVIL projects.

seqr/views/apis/dataset_api.py
ShifaSZ: So, I'm going to update `match_sample_ids_to_sample_records()` to return a list of new samples.

seqr/views/apis/dataset_api.py
ShifaSZ: Hmm, I should use a lower-level helper.

seqr/views/utils/permissions_utils.py
ShifaSZ: Can we change the name to something like `is_internal_project`?

hanars: no because "internal" isn't a word we use anywhere in the seqr code base and `is_analyst_access_project` is grammatically worse than `project_has_analyst_access`

seqr/views/apis/dataset_api.py
ShifaSZ: I have just found that the `user` here should be the project manager, not the data manager who edits the datasets. We don't know exactly who created the data loading request at this point.  Maybe we should use the project creator as the data loading requestor.

hanars: The project creator will be one of the project managers, but we want the email top go to all project managers, as I stated in the ticket

ShifaSZ: Okay. How about just greeting the creator and sending it to all the managers' mailboxes?

seqr/views/apis/dataset_api.py
```diff
@@ -112,6 +115,35 @@ def add_variants_dataset_handler(request, project_guid):
...
+            """.format(
+                num_sample=len(new_samples),
+                guid=project.guid,
+                samples=[sample.sample_id for sample in new_samples]
+            ))
+    elif project_has_anvil(project):
+        permission_levels = get_workspace_collaborator_perms(request.user, project.workspace_namespace, project.workspace_name)
+        send_html_email("""Hi {user},
+We are following up on your request to load data from AnVIL on {date}.
+We have loaded {num_sample} sample(s) from the AnVIL workspace <a>https://anvil.terra.bio/#workspaces/{namespace}/{name}</a> to the corresponding seqr project <a>https://seqr.broadinstitute.org/project/{guid}/project_page</a>. Let us know if you have any questions.
```
hanars, line 132: It looks like we already have a constant for `https://anvil.terra.bio/` in communication_utils.py - we should move it to settings.py and re-use the constant here, rather than hardcoding the same URL in multiple places

seqr/views/apis/dataset_api_tests.py
```diff
@@ -69,11 +72,12 @@ def test_add_variants_dataset(self, mock_random):
         }}, method=urllib3_responses.POST)
         MOCK_FILE_ITER.return_value = StringIO('NA19678_1,NA19678\n')
 
-        response = self.client.post(url, content_type='application/json', data=json.dumps({
-            'elasticsearchIndex': INDEX_NAME,
-            'mappingFilePath': 'mapping.csv',
-            'datasetType': 'VARIANTS',
-        }))
+        with mock.patch('seqr.views.utils.permissions_utils.ANALYST_PROJECT_CATEGORY', 'analyst-projects'):
```
hanars, line 75: you should patch the whole function or none of it. We have "Non-Analyst Project" as a test fixture thats not an analyst project that you can use for testing the non analyst project case, or if that really doesn't work for some reason do a second test function to test the analyst notification on its own

seqr/views/apis/dataset_api_tests.py
```diff
@@ -118,6 +122,13 @@ def test_add_variants_dataset(self, mock_random):
         self.assertEqual(len(updated_sample_models), 3)
         self.assertSetEqual({INDEX_NAME}, {sample.elasticsearch_index for sample in updated_sample_models})
 
+        mock_send_email.assert_not_called()
+        mock_send_slack.assert_called_with(
+            'seqr-data-loading',
+            '1 new sample(s) are loaded in https://seqr.broadinstitute.org/project/{guid}/project_page\n            ```[\'NA19678_1\']```\n            '.format(
```
hanars, line 128: its important to test that we are only notifying on the new samples, but it looks like the test input data only includes new samples to begin with. we need to update that

seqr/views/apis/dataset_api_tests.py
```diff
@@ -160,6 +172,18 @@ def test_add_variants_dataset(self, mock_random):
         self.assertSetEqual({sv_sample_guid, existing_index_sample_guid}, {sample.guid for sample in sample_models})
         self.assertSetEqual({True}, {sample.is_active for sample in sample_models})
 
+        if self.test_local:
```
hanars, line 175: this would be clearer if you named this variable `ANVIL_DISABLED`

hanars (changes requested):

seqr/views/apis/dataset_api.py
```diff
@@ -112,6 +115,35 @@ def add_variants_dataset_handler(request, project_guid):
...
+            SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL,
+            """{num_sample} new sample(s) are loaded in https://seqr.broadinstitute.org/project/{guid}/project_page
+            ```{samples}```
+            """.format(
+                num_sample=len(new_samples),
+                guid=project.guid,
+                samples=[sample.sample_id for sample in new_samples]
+            ))
+    elif project_has_anvil(project):
+        permission_levels = get_workspace_collaborator_perms(request.user, project.workspace_namespace, project.workspace_name)
```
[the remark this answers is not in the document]
ShifaSZ: I will change it to `project_user_permission`.

seqr/views/apis/dataset_api.py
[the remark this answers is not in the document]
ShifaSZ: How about `We have loaded {num_sample} samples from the <a href=https://anvil.terra.bio/#workspaces/{namespace}/{name}>AnVIL workspace</a> to the corresponding seqr project <a href=https://seqr.broadinstitute.org/project/{guid}/project_page>{project_name}</a>`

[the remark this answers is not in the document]
ShifaSZ: How about `We have loaded {num_sample} samples from the AnVIL workspace <a href=https://anvil.terra.bio/#workspaces/{namespace}/{name}>{namespace}/{name}</a> to the corresponding seqr project <a href=https://seqr.broadinstitute.org/project/{guid}/project_page>{project_name}</a>`

ShifaSZ: Should we get it from an environment variable?

hanars: no I think hardcoding it is okay for now

seqr/views/apis/dataset_api_tests.py
ShifaSZ: I will add tests for slack messages after adding the two additive datasets. The notice should include the new samples only.

seqr/views/apis/dataset_api_tests.py
ShifaSZ: I found family 14 and individual 18 belong to the "Non-Analyst Project" in the fixture data. I will use these data to test the non-analyst project case.

settings.py
```diff
@@ -185,6 +185,7 @@
 }
 
 TERRA_API_ROOT_URL = os.environ.get('TERRA_API_ROOT_URL')
+ANVIL_BASE_URL = 'https://anvil.terra.bio/'
```
hanars, line 188: you should update communication_utils.py to use this constant instead of having it defined separately there

hanars (changes requested):

hanars:
Unit tests are now failing, but otherwise looks good!

PR #2065, opened 2021-08-23
Mme results validation

hanars:


matchmaker/views/matchmaker_api.py
```diff
@@ -23,7 +23,7 @@
 from settings import BASE_URL, MME_ACCEPT_HEADER, MME_NODES, MME_DEFAULT_CONTACT_EMAIL, \
     MME_SLACK_SEQR_MATCH_NOTIFICATION_CHANNEL, MME_SLACK_ALERT_NOTIFICATION_CHANNEL
 
-logger = logging.getLogger(__name__)
+logger = SeqrLogger(__name__)
```
ShifaSZ, line 26: This logger has no tests. You may want to add them.

matchmaker/views/matchmaker_api_tests.py
```diff
@@ -401,20 +408,27 @@ def test_search_individual_mme_matches(self, mock_nodes, mock_email, mock_slacke
...
+            mock.call('matchmaker_alerts', alert_a_slack_message, **slack_kwargs),
+            mock.call('matchmaker_alerts', alert_b_slack_message, **slack_kwargs),
             mock.call('matchmaker_seqr_match', message, **slack_kwargs),
         ])
         mock_logger.error.assert_has_calls([
             mock.call(
                 'Slack error: Unable to connect to slack: Original message in channel ({}) - {}'.format(
-                    'matchmaker_alerts', alert_slack_message
+                    'matchmaker_alerts', alert_a_slack_message
             )),
```
ShifaSZ, line 427: Mock and test the logger in the `communication_utils` module is a little weird. How about just mock `safe_post_to_slack`? And, maybe add some separate tests for the `communication_utils` module.

ShifaSZ (changes requested):
Looks good mostly. Some suggestions for tests.

matchmaker/views/matchmaker_api_tests.py
hanars: no, this has better coverage of what actually happens and makes sure we adequately test when slack messages fail. I think removing test coverage is a terrible idea!

matchmaker/views/matchmaker_api_tests.py
```diff
@@ -470,6 +503,7 @@ def test_update_mme_submission(self, mock_mme_nodes):
         self.assertEqual(response.reason_phrase, 'Individual is required for a new submission')
 
         # Test successful creation
+        mock_logger.reset_mock()
```
ShifaSZ, line 506: It is unnecessary to reset it since it hasn't been used up to this point.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #2082, opened 2021-08-30
clean up family render field constants

hanars:
No functional changes to this PR, but cleans up constants so their usage is clearer. Will make an upcoming PR I'm working on easier to read


ShifaSZ (approved):
Can't see any issue.

PR #2087, opened 2021-09-01
Project context cleanup

hanars:
Moves  functionality for getting all the project details and child entities from 2 separate sets of functions in the project api and variant search api into a shared utility that is used by both. No changes to functionality, but will make changes in an upcoming PR easier to implement/ read

### DIFF
--- a/seqr/views/utils/project_context_utils.py
+++ b/seqr/views/utils/project_context_utils.py
@@ -14,6 +14,13 @@ def get_projects_child_entities(projects, user, is_analyst=None):
     if is_analyst is None:
         is_analyst = user_is_analyst(user)
 
+    response = _fetch_child_entities(projects, project_guid, user, is_analyst, has_case_review_perm)
+
+    _add_tag_types(response['projectsByGuid'], project_guid)
+
+    return response
+
+def _fetch_child_entities(projects, project_guid, user, is_analyst, has_case_review_perm):
     projects_by_guid = {p['projectGuid']: p for p in get_json_for_projects(projects, user)}
 
     family_models = Family.objects.filter(project__in=projects)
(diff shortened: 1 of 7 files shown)


seqr/views/utils/project_context_utils.py
```diff
@@ -0,0 +1,140 @@
...
+        group['projectGuid'] = project_id_to_guid[group.pop('projectId')]
+
+    for project in response['projectsByGuid'].values():
+        project['locusListGuids'] = []
+    prefetch_related_objects(locus_lists_models, 'projects')
+    for locus_list in locus_lists_models:
+        for project in locus_list.projects.all():
+            response['projectsByGuid'][project.guid]['locusListGuids'].append(locus_list.guid)
+
+def _add_child_ids(response):
```
ShifaSZ, line 102: Would it be more descriptive to rename it to something like `_add_individual_guids`?

seqr/views/utils/project_context_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    if project_guid:
+        response['projectsByGuid'][project_guid]['locusListGuids'] = list(response['locusListsByGuid'].keys())
+    else:
+        _add_parent_ids(response, projects, family_models, individual_models, locus_lists_models)
+
+    _add_child_ids(response)
+
+    return response
+
+def _add_parent_ids(response, projects, family_models, individual_models, locus_lists_models):
```
ShifaSZ, line 68: I'm curious that how does it relate to `parent`?

ShifaSZ (changes requested):

seqr/views/utils/project_context_utils.py
hanars: The parent models. As in, a project is a parent of a family, a family is a parent of an individual, etc

seqr/views/utils/project_context_utils.py
hanars: no because it is adding `sampleGuids`, and `igvSampleGuids` as well. It is adding the guids for all of the child models, so this is a much more descriptive name

seqr/views/utils/project_context_utils.py
ShifaSZ: Okay. I thought the parents in the family pedigree.

seqr/views/utils/project_context_utils.py
ShifaSZ: I see what it means now. All the projects, families, individuals, samples, and igvSamples in the response need both their parent models and child models' guids.

ShifaSZ (approved):

PR #2096, opened 2021-09-02
Fix new sample notification

hanars:
We should only send a slack update for samples if data has never been loaded for that individual before, not just if the sample models itself is new. Also cleans up the `match_sample_ids_to_sample_records` a bit and moves shared behavior into the shared utility function

### DIFF
--- a/seqr/views/apis/dataset_api.py
+++ b/seqr/views/apis/dataset_api.py
@@ -59,7 +59,7 @@ def add_variants_dataset_handler(request, project_guid):
     loaded_date = timezone.now()
     ignore_extra_samples = request_json.get('ignoreExtraSamplesInCallset')
     try:
-        matched_sample_id_to_sample_record, included_families = match_sample_ids_to_sample_records(
+        samples, included_families = match_sample_ids_to_sample_records(
             project=project,
             user=request.user,
             sample_ids=sample_ids,
@@ -75,7 +75,7 @@ def add_variants_dataset_handler(request, project_guid):
         return create_json_response({'errors': [str(e)]}, status=400)
 
     activated_sample_guids, inactivated_sample_guids = update_variant_samples(
-        matched_sample_id_to_sample_record.values(), request.user, elasticsearch_index, loaded_date, dataset_type, sample_type)
+        samples, request.user, elasticsearch_index, loaded_date, dataset_type, sample_type)
 
     updated_samples = Sample.objects.filter(guid__in=activated_sample_guids)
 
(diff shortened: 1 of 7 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -109,34 +110,29 @@ def match_sample_ids_to_sample_records(
         sample_type (string): one of the Sample.SAMPLE_TYPE_* constants
         dataset_type (string): one of the Sample.DATASET_TYPE_* constants
         elasticsearch_index (string): an optional string specifying the index where the dataset is loaded
-        create_sample_records (bool): whether to create new Sample records for sample_ids that
-            don't match existing Sample records, but do match individual_id's of existing
-            Individual records.
         sample_id_to_individual_id_mapping (object): Mapping between sample ids and their corresponding individual ids
         loaded_date (object): datetime object
+        raise_no_match_error (bool): whether to raise an exception if no sample matches are found
+        unmatched_error_template (string): optional template to use to raise an exception if samples are unmatched
```
ShifaSZ, line 116: The `unmatched_error_template` is both a condition and a template. To avoid the intuition of just a template, maybe, we can rename it to `raise_unmatched_template`. And, the comments for it can be changed to something like `whether to raise an exception if there are unmatched samples when the template is not None`.

seqr/views/apis/dataset_api.py
```diff
@@ -116,15 +86,25 @@ def add_variants_dataset_handler(request, project_guid):
...
+            sample.individual_id for sample in Sample.objects.filter(
+                individual__in=updated_individuals, sample_type=sample_type, dataset_type=dataset_type,
+            ).exclude(elasticsearch_index=elasticsearch_index)}
+        previous_loaded_individuals.update(matched_individual_ids)
+        new_sample_ids = [
+            sample.sample_id for sample in updated_samples if sample.individual_id not in previous_loaded_individuals]
         safe_post_to_slack(
             SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL,
-            """{num_sample} new samples are loaded in {base_url}project/{guid}/project_page
+            """{num_sample} new {sample_type}{dataset_type} samples are loaded in {base_url}project/{guid}/project_page
```
ShifaSZ, line 99: Need a space between `{sample_type}` and `{dataset_type}`?

ShifaSZ (changes requested):
Mostly great cleaning up.

seqr/views/apis/dataset_api.py
hanars: no because we do not  show dataset type if its its a a standard variant, if you look at how `dataset_type` is passed to the format you will see that it handles conditionally adding the space

ShifaSZ (approved):

PR #2097, opened 2021-09-02
log error for long running tasks

hanars:
the seqr account authenticated for searching in ES does not (and should not!) have sufficient permission to cancel long running searches. Instead, log an error so we are notified if this is happening

### DIFF
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -1031,8 +1031,11 @@ class EsSearch(object):
         try:
             return search.using(self._client).execute()
         except elasticsearch.exceptions.ConnectionTimeout as e:
-            canceled = self._delete_long_running_tasks()
-            logger.warning('ES Query Timeout. Canceled {} long running searches'.format(canceled), self._user)
+            tasks = self._get_long_running_tasks()
+            if tasks:
+                logger.error('ES Query Timeout: Found {} long running searches'.format(len(tasks)), self._user, detail=tasks)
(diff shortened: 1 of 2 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -1031,8 +1031,11 @@ def _execute_search(self, search):
         try:
             return search.using(self._client).execute()
         except elasticsearch.exceptions.ConnectionTimeout as e:
-            canceled = self._delete_long_running_tasks()
-            logger.warning('ES Query Timeout. Canceled {} long running searches'.format(canceled), self._user)
+            tasks = self._get_long_running_tasks()
+            if tasks:
+                logger.error('ES Query Timeout: Found {} long running searches'.format(len(tasks)), self._user, detail=tasks)
+            else:
+                logger.warning('ES Query Timeout. No long running searches found', self._user)
```
ShifaSZ, line 1038: Should it also be `logger.error()`?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_search.py
hanars: no this is a normal thing that happens sometimes when you try to do a query thats too complex for es. Its a normal, expected application error and we do NOT want to be spammed with alerts when thats happening. If you look in our middlewear, you'll see that 504s are always warnings

ShifaSZ (approved):

PR #2098, opened 2021-09-02
Optimized the IGV track data loading proxy to Google Storage.

ShifaSZ:


seqr/views/apis/igv_api.py
```diff
@@ -126,9 +128,44 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
 
 
+def _stream_gs(request, gs_path):
+    headers = {'Authorization': 'Bearer {}'.format(get_access_token())}
+    range_header = request.META.get('HTTP_RANGE')
+    if range_header:
+        headers['Range'] = range_header
+
+    bucket_object = gs_path[5:].split('/', 1)
+    for retry in range(2):  # retry automatically if the access token expired
```
hanars, line 144: this is a messy way to ensure you have a valid access token, especially because its such a fast command to run and get the token. I would get rid of the global variable and just generate the access token as you need it

seqr/views/apis/igv_api_tests.py
```diff
@@ -18,26 +18,35 @@
...
-        mock_subprocess.assert_called_with(
-            'gsutil -u anvil-datastorage cat gs://fc-secure-project_A/sample_1.cram.gz',
-            stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+        self.assertEqual(next(response.streaming_content), b'\n'.join(STREAMING_READS_CONTENT))
+        mock_subprocess.assert_called_with(['gcloud', 'auth', 'print-access-token'], capture_output=True)
+        mock_file_exist.assert_called_with('gs://project_A/sample_1.bam.bai', user=self.collaborator_user)
+        self.assertEqual(responses.calls[0].request.headers.get('Range'), 'bytes=100-200')
+        self.assertEqual(responses.calls[0].request.headers.get('Authorization'), 'Bearer token')
+
+        responses.replace(responses.GET, 'https://project_a.storage.googleapis.com/sample_1.bai',
```
hanars, line 43: you deleted us updating the url to `url = reverse(fetch_igv_track, args=[PROJECT_GUID, 'gs://fc-secure-project_A/sample_1.cram.gz'])`

I would still use that different URL here, although its okay to have it return an error. We want to test that its not checking if the `bai` file exists

hanars (changes requested):
Awesome work! One thing I want to double check, this will change how all IGV files are handled. Did you check to see that crams, gcnv, and rnaseq all still work as expected?

seqr/views/apis/igv_api.py
ShifaSZ: Getting token with the 'gcloud auth' command is not fast because `subprocess` is slow. Each time when IGV loading a track, it needs tens of loading requests. We can't generate the access token for each of the requests. Another way to avoid retrying is to check the token expiration time.

seqr/views/apis/igv_api.py
```diff
@@ -126,9 +128,44 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
     return _stream_file(request, igv_track_path)
 
 
+def _stream_gs(request, gs_path):
+    headers = {'Authorization': 'Bearer {}'.format(get_access_token())}
+    range_header = request.META.get('HTTP_RANGE')
+    if range_header:
+        headers['Range'] = range_header
+
+    bucket_object = gs_path[5:].split('/', 1)
```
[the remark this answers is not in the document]
ShifaSZ: I found that the parameter for the `lstrip()` is a character set rather than a string. We should use `replace('gs://', '', 1).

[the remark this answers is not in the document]
hanars: replace works too, but just so you know whatever documentation you found for lstrip is just wrong. from my local machine: 
image

seqr/views/apis/igv_api.py
```diff
@@ -126,9 +130,37 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
+        'https://{bucket}.storage.googleapis.com/{object}'.format(bucket=bucket_object[0], object=bucket_object[1]),
+        headers=headers,
+        stream=True)
+
+    return StreamingHttpResponse(response.iter_content(chunk_size=65536), status=response.status_code,
+                                 content_type='application/octet-stream')
+
+
+def _get_access_token(user):
+    if not hasattr(_get_access_token, 'timer') or (time.time() - _get_access_token.timer) > EXPIRATION_TIME_IN_SECONDS:
```
hanars, line 156: I like this so much less than what you had before. If you disagree with my PR feedback, comment and explain why you don't want to go that route, do not just implement something totally different. This makes absolutely no sense in python - is this supposed to be a class? Are you making a function that also secretly can have properties assigned to it? This is incredibly un-pythonic - Functions should never have properties. Nothing should ever hav properties that may or may not exist the way these do

I suggested REDIS to handle caching and timeouts, because that is literally what redis does - it caches something for a fixed period of time, which is exactly what you are doing here. All you have done is reimplemented a cache, but like as a confusing set of object/ function properties. If you have a problem with redis, comment and explain why, don't go ahead and try to implement it on you own

ShifaSZ: As I replied to your previous comment, the token will be used very frequently while IGV is loading a track. There are at least 20 separated loading requests for a gCNV track. Therefore, I don't like to add any extra time for each loading request. I even don't like to use Redis for the token because Redis using HTTP call to get the cached data which can have a significant delay.
If you like, I can change it to a class.

hanars: please time redis and tell me what the percentage change would be. Quite frankly, going from 640ms to 660ms is a totally acceptable slow done

ShifaSZ: Surprisingly to me, the difference is small. It just adds 5% more latency with Redis. Small packages increase much more delay while the large ones can't see any differences.
image

I will use Redis for the token.

hanars: yeah, its good to know for future reference that redis is so quick!

seqr/views/apis/igv_api.py
```diff
@@ -2,16 +2,22 @@
...
 from seqr.views.utils.file_utils import save_uploaded_file
 from seqr.views.utils.json_to_orm_utils import get_or_create_model_from_json
 from seqr.views.utils.json_utils import create_json_response
 from seqr.views.utils.orm_to_json_utils import  get_json_for_sample
 from seqr.views.utils.permissions_utils import get_project_and_check_permissions, check_project_permissions, \
     login_and_policies_required, pm_or_data_manager_required
+from seqr.utils.redis_utils import safe_redis_get_json, safe_redis_set_json
+
+EXPIRATION_TIME_IN_SECONDS = 3600 - 5
+IGV_CACHE_KEY = 'seqr-igv-redis_cache_key'
```
hanars, line 20: Its a bit odd to put the words "cache_key" in a cache key. This is also an incredibly confusing variable name - it sounds like a prefix for caching igv data, it would be much clearer if you specify that this is caching the access token (which isn't even IGV specific). I would name both the variable and the key something like `GS_STORAGE_ACCESS_TOKEN`

seqr/views/apis/igv_api.py
```diff
@@ -126,9 +132,38 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
+                                 content_type='application/octet-stream')
+
+
+def _get_access_token(user):
+    access_token = safe_redis_get_json(IGV_CACHE_KEY)
+    if not access_token or (time.time() - access_token['timer']) > EXPIRATION_TIME_IN_SECONDS:
+        process = run_command('gcloud auth print-access-token', user=user)
+        if process.wait() == 0:
+            access_token = {'timer': time.time(), 'token': next(process.stdout).decode('utf-8').strip()}
+            safe_redis_set_json(IGV_CACHE_KEY, access_token)
```
hanars, line 163: Like I explained in previous reviews, redis allows you to set key/values that have an expiration. You should not be storing expiration numbers in json in setting them and checking them. You should be setting your cache with an explicit expiration so REDIS will handle deleting things that are expire - 
```
access_token = safe_redis_get_json(GS_STORAGE_ACCESS_TOKEN)
if not access_token:
    process = run_command('gcloud auth print-access-token', user=user)
    if process.wait() == 0:
        access_token = next(process.stdout).decode('utf-8').strip()
        safe_redis_set_json(GS_STORAGE_ACCESS_TOKEN, access_token, expire=EXPIRATION_TIME_IN_SECONDS)
return access_token
```

hanars (changes requested):

seqr/views/apis/igv_api.py
ShifaSZ: I did it as you suggest. But the Codacy keeps reporting 'possible hardcoded credential'. I agree that it's not for IGV only. Maybe I can use `GS_STORAGE_SECRET_CACHE_KEY` for the name of the constant and `gs_storage_secrect_cache_entry` for the cache key itself. I hope the Codacy won't report any error.

hanars: using bad variable names to get around codacy is a bad idea. putting the word "secret" in is even worse because a secret is a specific kubernetes thing in our deployment and this is not a secret. We could try adding "temp" in to indicate that this roatates. Does it complain about 
```
GS_STORAGE_TEMP_ACCESS_KEY = 'GS_STORAGE_TEMP_ACCESS_KEY'
```

ShifaSZ: I use `GS_STORAGE_ACCESS_CACHE_KEY` now. Codacy doesn't complain. Do you think it is okay?

seqr/views/apis/igv_api.py
hanars: > Awesome work! One thing I want to double check, this will change how all IGV files are handled. Did you check to see that crams, gcnv, and rnaseq all still work as expected?

@ShifaSZ any answer for this?

seqr/views/apis/igv_api.py
hanars: I think its fine. Also a thing I forgot, if you put the comment `# nosec` at the end fo the line codacy won't complain, so in the future we can also do that

seqr/views/apis/igv_api.py
ShifaSZ: I have just double-checked that the alignment, gcnv, and rnaseq work fine. They fetch track data from the GS.

hanars (changes requested):

seqr/views/apis/igv_api_tests.py
```diff
@@ -18,26 +19,49 @@
...
-        mock_subprocess.assert_called_with(
-            'gsutil cat -r 100-200 gs://project_A/sample_1.bai',
-            stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True)
+        self.assertEqual(next(response.streaming_content), b'\n'.join(STREAMING_READS_CONTENT))
+        self.assertEqual(responses.calls[0].request.headers.get('Range'), 'bytes=100-200')
+        self.assertEqual(responses.calls[0].request.headers.get('Authorization'), 'Bearer token1')
+        mock_get_redis.assert_called_with(GS_STORAGE_ACCESS_CACHE_KEY)
+        mock_set_redis.assert_called_with(GS_STORAGE_ACCESS_CACHE_KEY, 'token1', expire=EXPIRATION_TIME_IN_SECONDS)
+        mock_subprocess.assert_has_calls([
+            mock.call('gsutil -u anvil-datastorage ls gs://fc-secure-project_A/sample_1.bam.bai', stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True),
```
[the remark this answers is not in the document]
ShifaSZ: Good point. I may need the dev environment to do a test for the `fc-secure` buckets.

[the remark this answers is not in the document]
hanars: why?? you should be able to connect locally, the same way you do for non-secure buckets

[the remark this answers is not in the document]
ShifaSZ: I don't have access to the `anvil-datastorage` project?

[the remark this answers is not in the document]
ShifaSZ: It seems that I need to add a `x-goog-user-project` header.
image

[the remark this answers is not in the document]
hanars: You don't need access to the project, you are literally just telling google the name of a billing project you are not authorizing with that project in any way. 

I assumed we want the json API and not the XML api, but if either works then yeah I guess you can use the XML with the header. But also, for both json and xml apis your URL looks wrong - you are doing `'https://{bucket}.storage.googleapis.com/{object}'` but the API's both start their URL with `https://storage.googleapis.com` and then put the bucket in the path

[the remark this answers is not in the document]
ShifaSZ: I found both URL formats work. But I change it as your suggested. It seems a JSON API. But our data is not JSON. So, it is weird. Anyway, it works now.

seqr/views/apis/igv_api.py
```diff
@@ -126,9 +131,39 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
     return _stream_file(request, igv_track_path)
 
 
+def _stream_gs(request, gs_path):
+    headers = {'Authorization': 'Bearer {}'.format(_get_access_token(request.user))}
+    range_header = request.META.get('HTTP_RANGE')
+    if range_header:
+        headers['Range'] = range_header
+    if gs_path.startswith('gs://fc-secure'):
+        headers['x-goog-user-project'] = 'anvil-datastorage'
```
hanars, line 146: Can we make this a helper method in file_utils as well, since we do the same check there? I just feel like theres a chance that someday we would need to change this project ID or the logic to check for it and I would rather only have one place to go

hanars (changes requested):

seqr/views/apis/igv_api.py
hanars: Maybe a helpoer like `get_google_project_id` that takes the gs path and returns either `'anvil-datastorage'` or None, and then here we would use it
```
google_project = get_google_project(gs_pth)
if google_project:
     headers['x-goog-user-project'] = google_project
```

ShifaSZ: It seems that we could split the Google Storage codes into another file. Such as gs_utils.py

ShifaSZ: > Maybe a helpoer like `get_google_project` that takes the gs path and returns either `'anvil-datastorage'` or None, and then here we would use it

Oh... I see what you mean.

seqr/utils/file_utils.py
```diff
@@ -2,40 +2,68 @@
...
-    return _run_command(command, user=user)
+    return run_command(command, user=user)
 
 
-def _is_google_bucket_file_path(file_path):
+def is_google_bucket_file_path(file_path):
     return file_path.startswith("gs://")
 
 
+def get_google_project(gs_path):
```
hanars, line 35: should be a private function since its never used outside the file

seqr/utils/file_utils.py
```diff
@@ -2,40 +2,68 @@
...
+from seqr.utils.redis_utils import safe_redis_get_json, safe_redis_set_json
 
 logger = SeqrLogger(__name__)
 
+EXPIRATION_TIME_IN_SECONDS = 3600 - 5
+GS_STORAGE_ACCESS_CACHE_KEY = 'gs_storage_access_cache_entry'
 
-def _run_command(command, user=None):
+
+def run_command(command, user=None):
```
hanars, line 13: this is no longer used outside this file so it seems okay to make it a private method again

seqr/utils/file_utils.py
```diff
@@ -2,40 +2,68 @@
...
-def _is_google_bucket_file_path(file_path):
+def is_google_bucket_file_path(file_path):
     return file_path.startswith("gs://")
 
 
+def get_google_project(gs_path):
+    return 'anvil-datastorage' if gs_path.startswith('gs://fc-secure') else None
+
+
+def get_gs_rest_api_headers(range_header, gs_path, user=None):
```
hanars, line 39: These do not belong in file utils. I would either keep it where it was in igv_api or make an explicit gs_utils file

hanars (changes requested):

seqr/utils/file_utils.py
ShifaSZ: I will move them back to the igv_api. The gs_utils is too lightweight for now.

seqr/utils/file_utils.py
ShifaSZ: After I move the get_gs_rest_api_headers() back, it is used by igv_api.py now.

seqr/utils/file_utils.py
ShifaSZ: After I move the get_gs_rest_api_headers() back, it is still used by igv_api.py.

seqr/views/apis/igv_api.py
```diff
@@ -126,9 +132,43 @@ def fetch_igv_track(request, project_guid, igv_track_path):
...
+
+    return StreamingHttpResponse(response.iter_content(chunk_size=65536), status=response.status_code,
+                                 content_type='application/octet-stream')
+
+
+def _get_gs_rest_api_headers(range_header, gs_path, user=None):
+    headers = {'Authorization': 'Bearer {}'.format(_get_access_token(user))}
+    if range_header:
+        headers['Range'] = range_header
+    headers['x-goog-user-project'] = get_google_project(gs_path)
```
hanars, line 157: If there is no project we should not send the header, sending None in the header is bad practice

hanars (changes requested):

seqr/views/apis/igv_api.py
ShifaSZ: I thought the tool (package) would remove None headers.

hanars (approved):

PR #2107, opened 2021-09-08
Mcri/feat panelapp upstream v2

tommyli:
This PR replaces previous PR #2052 and retrofits changes from previous PR onto a later code base with [gene lists search feature](https://github.com/broadinstitute/seqr/pull/2051) that your team implemented, it complements well with PanelApp.  This PR is for feature #1891.  Summary of changes from previous PR:

* Same import_all_panels() from [PanelApp Australia](https://panelapp.agha.umccr.org/) feature.  Also tested import works with [PanelApp Genomics England](https://panelapp.genomicsengland.co.uk/)
* Removed API for `import_all_panels()`.  This means `import_all_panels()` can only be triggered using Django management commands
* Additional attributes from `pagene` are excluded when calling all `locus_lists()`, this is for performance reason
* Removed ENABLE_PANEL_APP feature toggle (but kept commit history for future reference)
* Incorporated PanelApp tests as part of main coverage test run
* UI changes to include RED, AMBER, GREEN (reflects PanelApp confidence levels) background colour when viewing genes within a specific gene list for PanelApp gene lists.  These colours are consistent with the colours already used in Seqr variant results background colours.


panelapp/management/tests/import_all_panels_tests.py
```diff
@@ -0,0 +1,17 @@
+import mock
+
+from django.core.management import call_command
+from django.test import TestCase
+
+
+class ImportAllPanelsTest(TestCase):
```
hanars, line 7: I don't think you need this test file at all - in test_import_all_panels in the pa_locus_list_api_tests file you test this in much greater depth so I don't think this file really adds anything

panelapp/models.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    panel_app_id = models.IntegerField(null=False, blank=False, unique=True)
+    disease_group = models.TextField(null=True, blank=True)
+    disease_sub_group = models.TextField(null=True, blank=True)
+    status = models.TextField(null=True, blank=True)
+    version = models.TextField(null=True, blank=True)
+    version_created = models.DateTimeField(null=True, blank=True)
+    url = models.TextField(max_length=2000, null=False, blank=False)
+    raw_data = JSONField(default=dict)
```
hanars, line 23: What is the logic behind storing all of the data as both structured fields on the model and as a raw data json blob?  For fields that we rely on showing in the application it makes total sense to have them as structured columns. But for all the fields we aren't going to return or display, there is a question of whether we should be storing them at all. I can understand wanting to keep the full record, but it seems like a lot of database usage for not a lot of value, especially if its being stored twice. I think either we should
- get rid of any column that isn't going to be used by the client side and just keep all the data in raw_data
- keep data we think might be important as structured columns the way they are now but get rid of the raw_data field entirely
- keep data we think might be important as structured columns the way they are now but remove those values from the raw_data field, and only keep values in raw_data that are not otherwise stored

panelapp/panelapp_utils.py
```diff
@@ -0,0 +1,160 @@
...
+                .get('90', {}) \
+                .get('ensembl_id')
+        else:
+            return None
+
+    panels_url = '{}/panels/?page=1'.format(PANEL_APP_API_URL)
+
+    all_panels = _get_all_panels(panels_url, [])
+
+    for panel in all_panels:
```
hanars, line 30: roughly how long does this manage command take?

panelapp/panelapp_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        with transaction.atomic():
```
hanars, line 31: is there any risk of successfully updating some of the panel app lists but then having subsequent ones fail? Like if that happens will that be okay, or will that be bad?

panelapp/panelapp_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        update_model_from_json(existing.seqr_locus_list, new_seqrlocuslist_json, user)
+        update_model_from_json(existing, new_palocuslist_json, user)
+
+        return existing
+    else:
+        seqr_locus_list = create_model_from_json(SeqrLocusList, new_seqrlocuslist_json, user)
+        pa_locus_list = PaLocusList.objects.create(seqr_locus_list=seqr_locus_list, panel_app_id=panel_app_id)
+        update_model_from_json(pa_locus_list, new_palocuslist_json, user)
+
+        return pa_locus_list
```
hanars, line 147: totally okay to leave it as is, but I think it would be more readable to not duplicate the parts of the logic that are the same whether or not there was an existing list - 
```
pa_locus_list = _safe_get_locus_list(panel_app_id)
...
if pa_locus_list:
    update_model_from_json(existing.seqr_locus_list, new_seqrlocuslist_json, user)
else:
    seqr_locus_list = create_model_from_json(SeqrLocusList, new_seqrlocuslist_json, user)
    pa_locus_list = PaLocusList.objects.create(seqr_locus_list=seqr_locus_list, panel_app_id=panel_app_id)
update_model_from_json(pa_locus_list, new_palocuslist_json, user)
return pa_locus_list
```

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -711,20 +711,41 @@ def _process_result(result, locus_list):
...
+        if hasattr(locus_list, 'palocuslist'):
+            pa_locus_list_json = _get_json_for_model(locus_list.palocuslist, user=user)
+            result.update({
+                'paLocusList': pa_locus_list_json,
+            })
+
     prefetch_related_objects(locus_lists, 'created_by')
     prefetch_related_objects(locus_lists, 'locuslistgene_set')
     prefetch_related_objects(locus_lists, 'locuslistinterval_set')
+    prefetch_related_objects(locus_lists, 'palocuslist')
```
hanars, line 747: the prefetch should probably be conditional on `include_pagenes`

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -696,7 +696,7 @@ def get_json_for_gene_notes_by_gene_id(gene_ids, user):
     return notes_by_gene_id
 
 
-def get_json_for_locus_lists(locus_lists, user, include_genes=False, include_project_count=False, is_analyst=None):
+def get_json_for_locus_lists(locus_lists, user, include_genes=False, include_pagenes=False, include_project_count=False, is_analyst=None):
```
hanars, line 699: Is there a reason to make `include_pagenes` its own kwarg? What would the case be where `include_genes` is True and `include_pagenes` is False?

ui/shared/components/StyledComponents.jsx
```diff
@@ -3,22 +3,24 @@ import styled from 'styled-components'
 import { Button, Header, Icon, Label, Table } from 'semantic-ui-react'
 import { NavLink } from 'react-router-dom'
 
-const BaseButtonLink = styled(({ color, padding, ...props }) => <Button {...props} />).attrs(
+const BaseButtonLink = styled(({ color, padding, background, ...props }) => <Button {...props} />).attrs(
```
hanars, line 6: I'm a little confused why you would want to assign something specifically designed to look like an html link a background. It strikes me that a link with a background color would not look great or be particularly readable. I think maybe an icon next to the link in the color, perhaps with a hover text to explain the color/ more details would be better. Or if you really want a background, it would be better to use an actual button and not a link

panelapp/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+    confidence_level = models.CharField(max_length=1, choices=CONFIDENCE_LEVEL_CHOICES, null=False)
+    biotype = models.TextField(null=True, blank=True)
+    penetrance = models.TextField(null=True, blank=True)
+    mode_of_pathogenicity = models.TextField(null=True, blank=True)
+    mode_of_inheritance = models.TextField(null=True, blank=True)
+    raw_data = JSONField(default=dict)
+
+    class Meta:
+        json_fields = ['confidence_level', 'penetrance', 'mode_of_pathogenicity', 'mode_of_inheritance']
```
hanars, line 53: you appear to only be using `confidence_level` in the frontend. I would say you should either update the front end to use the other fields, or stop returning them

hanars (changes requested):

panelapp/models.py
tommyli: Accepting option 2 - The structured columns already anticipate known future needs so removing `raw_data` entirely makes sense.

panelapp/panelapp_utils.py
tommyli: Takes around 5-10 minutes for PanelApp Australia.  PanelApp England has just under double the number of panels/genelists and it took around 25 minutes.  On our end, we intend to have a daily cron schedule for this.

panelapp/panelapp_utils.py
tommyli: Agree your logic is more readable, changes applied.

seqr/views/utils/orm_to_json_utils.py
tommyli: Yes, when first querying for all genelists (`/summary_data/gene_lists`), `include_genes` is set to true so UI filter can include filtering by gene symbol.  However, the filter in the UI do not require any of the fields from `pagene` so including all of them upfront increases the returned json payload size a fair bit and slows things down.

panelapp/models.py
tommyli: Agree, only include `confidence_level` for now and will include others when needed later.

ui/shared/components/StyledComponents.jsx
tommyli: Additional padding was added to make the background look more like a button.  The intent is to have this look like a link button but still somewhat consistent with what normal non-pagenes look like.  Attached are screenshots for both.

Screen Shot 2021-09-11 at 13 19 19
Screen Shot 2021-09-11 at 13 19 06

Note in second screenshot, it was taken with a mouse hover over gene "CACNB4" so it's more like a link button.

In a future PR, we may well include additional details with some kind of hover as you suggested.  However, those details still needs some finalising.  Is it OK to include confidence_level/colour for now?  Users of PanelApp are well aware of what the colours mean already and the colours add most value for now.

seqr/views/utils/orm_to_json_utils.py
tommyli: Fixed.

panelapp/panelapp_utils.py
tommyli: That's a good point, have fixed the code to continue with importing remaining panels even if one fails unexpectedly.

tommyli:
Thanks @hanars for the great feedback, they were useful!  See above replies, for the simple and obvious ones, I've marked the conversation as resolved.  The rest I've left open for further tracking.

ui/shared/components/StyledComponents.jsx
hanars: this really seems like a misuse of the component, I'm sorry but I'm not willing to accept this change. It looks weird to have the blue link text on a colored background! I would say either we need to go with a different component if there is a pagene, either adding an icon next to it or using a regular basic button, or we could go with changing the link color itself, which is already supported and will look fine

panelapp/models.py
```diff
@@ -0,0 +1,51 @@
+from django.db import models
+from django.db.models import JSONField
```
hanars, line 2: this import is unused

panelapp/models.py
```diff
@@ -0,0 +1,51 @@
...
+
+from seqr.models import LocusList as SeqrLocusList, LocusListGene as SeqrLocusListGene
+
+
+class PaLocusList(models.Model):
+    """PanelApp extension of seqr.models.LocusList."""
```
hanars, line 8: add a blank line before the docstring

panelapp/models.py
```diff
@@ -0,0 +1,51 @@
...
+    version = models.TextField(null=True, blank=True)
+    version_created = models.DateTimeField(null=True, blank=True)
+    url = models.TextField(max_length=2000, null=False, blank=False)
+
+    class Meta:
+        json_fields = ['panel_app_id']
+
+
+class PaLocusListGene(models.Model):
+    """PanelApp extension of seqr.models.LocusListGene."""
```
hanars, line 29: add a blank line before the docstring

hanars (changes requested):

david-ma:
I tried link color, but it's a bit too hard to read without increasing font weight or size. It gets flagged by Google Chrome for not having enough contrast.
Screen Shot 2021-09-14 at 1 57 39 pm
Screen Shot 2021-09-14 at 1 56 51 pm Screen Shot 2021-09-14 at 1 55 41 pm Screen Shot 2021-09-14 at 1 55 05 pm

Using circle icons looks great though!
Screen Shot 2021-09-14 at 2 48 43 pm
And I've included descriptions of the colours from the Panel App handbook as hover-over title text: https://panelapp.genomicsengland.co.uk/media/files/PanelApp_Handbook_V18.pdf

panelapp/models.py
tommyli: Sorry, I must be missing something.  Are there more details on how to pass these style guides?  I've tried many combinations of formatting styles and can't see to get Codacy happy.  Indeed, are there conflicts with some of these pydocstyle rules?  https://github.com/PyCQA/pydocstyle/issues/141

ui/shared/components/StyledComponents.jsx
tommyli: Please see reply from @david-ma below.

hanars:
> Using circle icons looks great though! So I'll prep this UI for this pull request.
 
Awesome! Agreed that this looks great. Thanks for being willing to iterate on this

panelapp/models.py
hanars: I'm sorry, our codacy rules for docstrings are a nightmare. If that round of adjustments didn't fix the error, we can merge it with the codacy issues and I'll get them cleaned up later

hanars (approved):

hanars:
Thanks all so much for this great contribution!! Can't wait to take this live :)

hanars:
As an update, created this Pr to fix a couple small bugs and add support for multiple sources for panel app lists : https://github.com/broadinstitute/seqr/pull/2115

panelapp/models.py
tommyli: Just to satisfy my curiosity, what were the Codacy changes required to make it happy?  Not urgent of course, it's more for my future reference.

PR #2110, opened 2021-09-09
IGV coordinates for SVs

hanars:


ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -319,6 +319,7 @@ const IgvPanel = React.memo(({ variant, igvSampleIndividuals, individualsByGuid,
     variant.chrom,
     (variant.genomeVersion !== project.genomeVersion && variant.liftedOverPos) ? variant.liftedOverPos : variant.pos,
     100,
+    variant.end && variant.end - variant.pos,
```
ShifaSZ, line 322: It seems that we need to increase the range and not just add an end offset.

ShifaSZ (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
hanars: no, we want the start position of the locus to be relative to the start of the SV and the end position to be relative to the end of the SV, and if you pass the endOffset to `getLocus` that is exactly what happens

PR #2116, opened 2021-09-14
Igv track options

ShifaSZ:
It is still draft because of styling. Most of the functionalities are ready.


ShifaSZ:
The latest RNA-Seq IGV UI.
image

hanars (changes requested):
At a high level, this adds an INCREDIBLE level of complexity, for what is ultimately not so complex. You do not need to split out the entire state of component by type in order to show groupings, you can keep the state simple and have the checkbox groups handle the filtering/ merging of values on their own.

The final implementation for the main `FamilyReads` component needs to have only 3 state fields: 
`openFamily` and `sampleTypes` as it currently has, and `referenceTracks`. 
As part of `updateSampleTypes` you should add logic to to also filter `referenceTracks` to remove anything that is no longer relevant, so `IgvPanel` can simply add all the `referenceTracks` to the reference with minimal parsing. You should define clear mappings/logic of which reference  should be shown for a given sample type, and you should not need the `parseTrackType` helper method - create your constants in a way that they don't need so much mapping! 

For your panel:
```
const dnaSampleOptions = DNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
const rnaSampleOptions = RNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
      <Segment>
          { dnaSampleOptions &&
            <CheckboxGroup
              groupLabel="DNA Tracks"
              value={this.state.sampleTypes} // it should just ignore the values not in the checkbox group
              options={dnaTrackOptions}
              onChange={this.updateSampleTypes}
            />
        }
          }
          { rnaTrackOptions &&
            <div>
              <CheckboxGroup
                groupLabel="RNA Tracks"
                value={this.state.sampleTypes}
                options={rnaTrackOptions}
                onChange={this.updateSampleTypes}
              />
              <b>Reference Tracks</b>
              <CheckboxGroup
                groupLabel="GTEx Tracks"
                value={this.state.referenceTracks}
                options={GTEX_TRACK_OPTIONS}
                onChange={this.updateReferenceTracks}
              />
              <CheckboxGroup
                groupLabel="Mappability Tracks"
                value={this.state.referenceTracks}
                options={MAPPABILITY_TRACK_OPTIONS}
                onChange={this.updateReferenceTracks}
              />
            </div>
          }
        </Segment>
```

In the future, I would also ask that you do different PRs for different features. If one depends on the other, do the changes for the next feature branched off of one of the other feature branches you are working on. I think ultimately you would have less work to completely redo if I had been able to review this at an earlier stage

ShifaSZ:
> As part of `updateSampleTypes` you should add logic to also filter `referenceTracks` to remove anything that is no longer relevant, so `IgvPanel` can simply add all the `referenceTracks` to the reference with minimal parsing.

These are very good suggestions. I have a question for "add all the `referenceTracks` to the reference", do you mean adding those tracks to the `reference` parameter of the `IGV` component? The `reference` parameter will be eventually the `reference` parameter for igv.createBrowser. And the tracks in the `reference` won't be removed and loaded dynamically. So I think the new `referenceTracks` that we are adding need to be removable. What do you think?

hanars:
> These are very good suggestions. I have a question for "add all the `referenceTracks` to the reference", do you mean adding those tracks to the `reference` parameter of the `IGV` component? The `reference` parameter will be eventually the `reference` parameter for igv.createBrowser. And the tracks in the `reference` won't be removed and loaded dynamically. So I think the new `referenceTracks` that we are adding need to be removable. What do you think?

Oh I see the issue. Yes, you are correct, it seems like the best way to have the component properly manage adding/ removing tracks is to pass these reference tracks to the `IGV` component at the end of the `tracks` array prop instead of in the `reference` props like I was suggesting. I still think it makes sense to have `referenceTracks` in the `FamilyReads` state, and pass that and `sampleTypes` to the `IgvPanel`. Then you can have `IgvPanel` handle constructing the `tracks` list the way it does now, and add the passed in referenceTracks to that list

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -66,11 +67,17 @@ const GCNV_TRACK_OPTIONS = {
...
+  format: 'gtf',
+  height: 100,
+}
+
 const TRACK_OPTIONS = {
   [ALIGNMENT_TYPE]: ALIGNMENT_TRACK_OPTIONS,
   [COVERAGE_TYPE]: COVERAGE_TRACK_OPTIONS,
   [JUNCTION_TYPE]: JUNCTION_TRACK_OPTIONS,
   [GCNV_TYPE]: GCNV_TRACK_OPTIONS,
+  [ANNOTATION_TYPE]: ANNOTATION_TRACK_OPTIONS,
```
hanars, line 80: we do not support adding these tracks in seqr and that was not part of any of the tickets for this so I don't know why you added this, but please take it out as we do not have any intention of supporting this

ShifaSZ: We have a SegDups track for mappability track types which is an "annotation" type.

ShifaSZ:
> I still think it makes sense to have `referenceTracks` in the `FamilyReads` state, and pass that and `sampleTypes` to the `IgvPanel`.

I'm coming up with a solution that merges `referenceTracks` into `sampleTypes`. It looks neater for both FamilyReads and IgvPanel. I'll push the code soon and you will see.

ui/shared/components/panel/family/FamilyReads.jsx
hanars: thats a reference track so you should hard code the constant with the appropriate settings, not  change the track options for actual tracks

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -395,25 +663,52 @@ class FamilyReads extends React.PureComponent {
...
+        </Segment>
         <Segment>
           <ButtonLink onClick={this.hideReads} icon={<Icon name="remove" color="grey" />} floated="right" size="large" />
           <VerticalSpacer height={20} />
           <IgvPanel
             variant={variant}
             igvSampleIndividuals={igvSampleIndividuals}
             sampleTypes={this.state.sampleTypes}
+            gtexRefs={this.state.gtexReferences}
+            mappabilityRefs={this.state.mappabilityRefs}
```
ShifaSZ, line 711: These two lines need to be deleted.

ui/shared/components/panel/family/FamilyReads.jsx
ShifaSZ: I could hard code the track options for this reference track. But how about the other reference tracks which have types like COVERAGE_TYPE and JUCTION_TYPE, should I use the same track options as the actual tracks with the same types?

ShifaSZ:
The UI looks like the picture below now.
image

ui/shared/components/panel/family/FamilyReads.jsx
hanars: I think the reference tracks need to be as clear as possible, which with all the current mapping they really aren't. Maybe do something like this to remove some duplication but keep things explicit
```
const GTEX_TRACKS = [...].map(track => ({ ...GTEX_PROPS, ...track })
const MAPPING_TRACKS = [...].map(track => ({ ...MAPPING_PROPS, ...track })
```

ShifaSZ: Good idea. I updated the code. Let me know if any further issues.

ShifaSZ:
> > The UI looks like the picture below now.
> 
> Well, when you don't have any RNA tracks selected you should not be able to select RNA reference tracks or show them, so this is not great

The options for RNA tracks won't be displayed if there is no IGV RNA-seq data. But if the RNA-seq data is configured, it seems that we should allow users to see and use the related options even when the RNA tracks are unselected.

hanars:
> The options for RNA tracks won't be displayed if there is no IGV RNA-seq data. But if the RNA-seq data is configured, it seems that we should allow users to see and use the related options even when the RNA tracks are unselected.

No, like I said, if no RNA tracks are selected, the RNA reference tracks should not be selectable or displayed

hanars:
The back and forth in this review has gotten to difficult to handle for me and I can't figure out what is and is not ready for review at this point. I am going to decline this PR. If you have implementation questions about how something is supposed to work, you can ask them on the github issues themselves. When this is ready for my review using the simplified state that we have discussed you should put up a new PR and I will review that. Also, if it is at all possible to put https://github.com/broadinstitute/seqr-private/issues/1032 into its own PR, I think that would make the next round of reviews much easier.  If you feel you still don't understand what is needed for this implementation, fee free to schedule a meeting with me.

ShifaSZ:
Okay. I'd like to have a meeting with you about what is happening here.

hanars:
> Okay. I'd like to have a meeting with you about what is happening here.

then like I said, schedule it with me

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -395,18 +663,43 @@ class FamilyReads extends React.PureComponent {
...
+                value={this.getIncludedSampleTypes(GTEX_TRACK_OPTIONS)}
+                options={GTEX_TRACK_OPTIONS}
+                onChange={this.updateGtexReferences}
+              />
+              <CheckboxGroup
+                groupLabel="Mappability Tracks"
+                value={this.getIncludedSampleTypes(MAPPABILITY_TRACK_OPTIONS)}
+                options={MAPPABILITY_TRACK_OPTIONS}
+                onChange={this.updateMappabilityRefs}
+              />
```
hanars, line 699: Should be 
```
this.state.sampleTypes.some(sampleType => isRnaType(sampleType)) &&
              <CheckboxGroup
                groupLabel="GTEx Tracks"
                value={this.state.referenceTracks}
                options={GTEX_TRACK_OPTIONS}
                onChange={this.updateReferenceTracks}
              />
              <CheckboxGroup
                groupLabel="Mappability Tracks"
                value={this.state.referenceTracks}
                options={MAPPABILITY_TRACK_OPTIONS}
                onChange={this.updateReferenceTracks}
              />
```

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -395,18 +663,43 @@ class FamilyReads extends React.PureComponent {
...
-        }
+          }
+          { rnaTrackOptions &&
+            <div>
+              <CheckboxGroup
+                groupLabel="RNA Tracks"
+                value={this.getIncludedSampleTypes(RNA_TRACK_TYPE_OPTIONS)}
+                options={rnaTrackOptions}
+                onChange={this.updateRnaSampleTypes}
+              />
```
hanars, line 686:
```
{ dnaTrackOptions &&
<CheckboxGroup
groupLabel="DNA Tracks"
              value={this.state.sampleTypes}
              options={dnaTrackOptions}
              onChange={this. updateSampleTypes}
/>
}
{ rnaTrackOptions &&
<CheckboxGroup
groupLabel="RNA Tracks"
              value={this.state.sampleTypes}
              options={rnaTrackOptions}
              onChange={this. updateSampleTypes}
/>
}
```

PR #2120, opened 2021-09-15
fix UI build error

hanars:


ui/shared/components/panel/variants/Variants.jsx
```diff
@@ -26,13 +26,13 @@ const StyledVariantRow = styled(({ isCompoundHet, isSV, severity, ...props }) =>
...
+      return '#eaa8a857'
     } else if (severity === 0) {
-      return VARIANT_BACKGROUND_COLORS.amber
+      return '#f5d55c57'
     } else if (severity < 0) {
-      return VARIANT_BACKGROUND_COLORS.green
+      return '#21a92624'
     } else if (isSV) {
-      return VARIANT_BACKGROUND_COLORS.lightBlue
+      return '#f3f8fa'
```
ShifaSZ, line 35: How about using named constants for these colors?

ShifaSZ (changes requested):

ui/shared/components/panel/variants/Variants.jsx
hanars: this PR is reverting a change that accidentally got committed that is completely breaking out UI build. The purpose is to revert that change back, not to make optimizations, so that is totally out of scope

ShifaSZ (approved):

PR #2128, opened 2021-09-20
Project-specific HGMD

hanars:


ui/pages/Search/selectors.js
```diff
@@ -154,6 +155,14 @@ export const getDatasetTypes = createSelector(
   },
 )
 
+export const getHasHgmdPermission = createSelector(
+  getUser,
+  getProjectsInput,
+  getProjectsByGuid,
+  (user, projectGuids, projectsByGuid) => user.isAnalyst || projectGuids.some(
+    projectGuid => (projectsByGuid[projectGuid] || {}).enableHgmd),
```
ShifaSZ, line 163: Just some of the projects enabling HGMD access are enough to allow the user to access HGMD of all projects. Is it too loose in permission control?

ShifaSZ (changes requested):

hanars:
> I have a general security question: is it secure enough to enforce the HGMD access control with the frontend? Maybe, we should use the backend to filter out the data to which the user doesn't have access.

that is out of scope of the PR, as thats what we are already doing and have been doing for years

ui/pages/Search/selectors.js
hanars: no, because if a PERSON has access to any project that we have agreed to grant HGMD access to, that means that PERSON has access to a HGMD license, which is what actually matters

ShifaSZ (approved):

PR #2129, opened 2021-09-20
Split the DNA/RNA tracks.

ShifaSZ:
It implements part of issue # 1032. The RNA-seq reference tracks come in later PR.

### DIFF
--- a/ui/shared/components/form/Inputs.jsx
+++ b/ui/shared/components/form/Inputs.jsx
@@ -266,8 +266,7 @@ const InlineFormGroup = styled(Form.Group).attrs({ inline: true })`
 
 export const CheckboxGroup = React.memo((props) => {
   const { value, options, label, groupLabel, onChange, ...baseProps } = props
-  const otherValue = value.filter(val => !options.find(opt => opt.value === val))
-  const realLength = value.length - otherValue.length
+  const numSelected = options.filter(opt => value.includes(opt.value)).length
   return (
     <List>
       <List.Item>
@@ -275,14 +274,15 @@ export const CheckboxGroup = React.memo((props) => {
           <BaseSemanticInput
             {...baseProps}
             inputType="Checkbox"
-            checked={realLength === options.length}
-            indeterminate={realLength > 0 && realLength < options.length}
+            checked={numSelected === options.length}
+            indeterminate={numSelected > 0 && numSelected < options.length}
             label={groupLabel || label}
             onChange={({ checked }) => {
+              const remainValue = value.filter(val => !options.find(opt => opt.value === val))
               if (checked) {
-                onChange(options.map(option => option.value).concat(otherValue))
+                onChange(options.map(option => option.value).concat(remainValue))
               } else {
-                onChange(otherValue)
+                onChange(remainValue)
               }
             }}
           />
--- a/ui/shared/components/form/Inputs.jsx
+++ b/ui/shared/components/form/Inputs.jsx
@@ -266,6 +266,8 @@ const InlineFormGroup = styled(Form.Group).attrs({ inline: true })`
 
 export const CheckboxGroup = React.memo((props) => {
   const { value, options, label, groupLabel, onChange, ...baseProps } = props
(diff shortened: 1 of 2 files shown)


ui/shared/components/form/Inputs.jsx
```diff
@@ -266,21 +266,23 @@ const InlineFormGroup = styled(Form.Group).attrs({ inline: true })`
 
 export const CheckboxGroup = React.memo((props) => {
   const { value, options, label, groupLabel, onChange, ...baseProps } = props
+  const otherValue = value.filter(val => !options.find(opt => opt.value === val))
+  const realLength = value.length - otherValue.length
```
hanars, line 270: I don't think you need to compute the list of values that are and are not in the list of options, its much cleaner to compute the selected options, and then only compute the updated lists in the onChange method itself, as theres no need to precompute values that may never be used
```
const numSelected = options.filter(opt => value.includes(opt)).length
...
checked={numSelected === options.length}
indeterminate={numSelected > 0 && numSelected < options.length}
...
// select all
onChange([...Set([...value, ...options.map(option => option.value)])])
//deselect all
onChange(value.filter(val => !options.find(opt => opt.value === val)))
```

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -395,18 +398,28 @@ class FamilyReads extends React.PureComponent {
     />
 
     const igvSampleIndividuals = this.state.openFamily && (igvSamplesByFamilySampleIndividual || {})[this.state.openFamily]
+    const dnaTrackOptions = DNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals && igvSampleIndividuals[value])
```
hanars, line 401: it would be neater to add ` || {}` to the assignment for `igvSampleIndividuals` so you don't have to keep checking for null

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -394,19 +397,29 @@ class FamilyReads extends React.PureComponent {
...
-    const reads = igvSampleIndividuals ?
+    const igvSampleIndividuals = (this.state.openFamily && (igvSamplesByFamilySampleIndividual || {})[this.state.openFamily]) || {}
+    const dnaTrackOptions = DNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
+    const rnaTrackOptions = RNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
+    const reads = Object.keys(igvSampleIndividuals).length ?
       <Segment.Group horizontal>
-        {Object.keys(igvSampleIndividuals).length > 1 &&
-          <Segment>
+        <Segment>
+          { dnaTrackOptions &&
```
hanars, line 406: 1) empty lists are "truthy" in javascript so I don't think this is the behavior you want, it will render the checkbox group without any options instead of rendering nothing
2) While the entire `reads` component is shown if there are any sample types, before this change the Segment showing the control panel used to be shown only if there was > 1 sample type. But now, the segment will always be shown, which is not the behavior we want

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
ShifaSZ: Thanks for #1. I will correct it. We always want the control panel for the RNA tracks. Does the RNA always have two sample types (both coverage/wig and junction/spliceJunctions)? If yes, I may keep the condition `Object.keys(igvSampleIndividuals).length > 1` for displaying the control panel.

hanars: The behavior for whether or not to show the control panel is either if there are more than one sample type OR if there are any RNA sample types. So for example, for the many samples in seqr that have just crams and nothing else there should be no control panel at all. The behavior does need to change here, because even if there is only one sample type, if its an RNA type we need to show the reference track options

hanars (approved):

PR #2133, opened 2021-09-21
Created filtering for in silico filters

MladenZeljic:
Created the custom annotation filters on Variant Search page, under "In Silico Filters" section:
image

The searching part is currently handled via backend, and the constructed search query is sent to elasticsearch

Based on the type of the option, the specific comparison will be done on the backend:
-  >= comparison for numeric values as agreed via emails
- matching for non numeric values


ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -140,17 +140,17 @@ class Predictions extends React.PureComponent {
         }
         {
           predictorFields.length > NUM_TO_SHOW_ABOVE_THE_FOLD &&
-            <Transition.Group animation="fade down" duration="500">
```
hanars, line 143: why would you change this?

ui/shared/components/panel/search/constants.js
```diff
@@ -375,6 +375,114 @@ export const LOCATION_FIELDS = [
   },
 ]
 
+export const IN_SILICO_FIELDS = [
+  {
+    name: 'cadd',
+    label: 'CADD',
+    labelHelp: 'Enter a numeric value for Cadd prediction. During filtering, some variants might be left out, since their annotation value might be rounded up or down from their original value',
```
hanars, line 382: The `During filtering, some variants might be left out, since their annotation value might be rounded up or down from their original value` should be removed from all of these. This language was never discussed or approved by me

ui/shared/components/panel/search/constants.js
```diff
@@ -375,6 +375,114 @@ export const LOCATION_FIELDS = [
...
+    component: StyledSemanticInput,
+    inputType: 'Input',
```
hanars, line 384: Numeric inputs should all have `type: 'number'` as a property

ui/shared/components/panel/search/constants.js
```diff
@@ -375,6 +375,114 @@ export const LOCATION_FIELDS = [
...
+  },
+  {
+    name: 'revel',
+    label: 'Revel',
+    labelHelp: 'Enter a numeric value for Revel prediction. During filtering, some variants might be left out, since their annotation value might be rounded up or down from their original value',
+    component: StyledSemanticInput,
```
hanars, line 390: put them in a Grid so they get padding, don't style the component

ui/shared/components/panel/search/constants.js
```diff
@@ -375,6 +375,114 @@ export const LOCATION_FIELDS = [
...
+    labelHelp: 'Enter a numeric value for STRVCTVRE prediction. During filtering, some variants might be left out, since their annotation value might be rounded up or down from their original value',
+    component: StyledSemanticInput,
+    inputType: 'Input',
+  },
+  {
+    name: 'polyphen',
+    label: 'Polyphen',
+    labelHelp: 'Enter a value for Polyphen prediction',
+    component: StyledSemanticInput,
+    inputType: 'Input',
```
hanars, line 440: All predictor fields that are string values need to be dropdowns where users select one of the allowed options. Free text is completely unacceptable

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -170,6 +194,13 @@ export const LOCATION_PANEL = {
   helpText: 'Filter by variant location. Entries can be either gene symbols (e.g. CFTR) or intervals in the form <chrom>:<start>-<end> (e.g. 4:6935002-87141054) or separated by tab. Variant entries can be either rsIDs (e.g. rs61753695) or variants in the form <chrom>-<pos>-<ref>-<alt> (e.g. 4-88047328-C-T). Entries can be separated by commas or whitespace.',
 }
 
+export const IN_SILICO_PANEL = {
+  name: 'in_silico',
+  headerProps: { title: 'In Silico Filters' },
+  fields: IN_SILICO_FIELDS,
+  fieldLayout: inSilicoFieldLayout,
```
hanars, line 201: this field needs the following `helpText`:

Filter by in-silico predictors. For numeric filters, any variant with a score greater than or equal to the provided filter value will be returned. For text filters, variants with exactly matching classifications will be returned.

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -103,6 +104,29 @@ export const ANALYST_PATHOGENICITY_PANEL = pathogenicityPanel(true)
 export const PATHOGENICITY_PANEL = pathogenicityPanel(false)
 
 const ANNOTATION_GROUP_INDEX_MAP = ANNOTATION_GROUPS.reduce((acc, { name }, i) => ({ ...acc, [name]: i }), {})
+const IN_SILICO_FILTER_ROW_CHUNK_SIZE = 3
+
+export const inSilicoFieldLayout = (fieldComponents) => {
```
hanars, line 109: You can fit way more than 3 filters per row. Additionally, I think you need to create a visual distinction between numeric and text filters. Try putting all the numeric filter in one row and all the text filters in another row. If those rows look crowded, you can split them into up to 2 rows for each type

hanars (changes requested):

PR #2136, opened 2021-09-23
All user demo project

hanars:


seqr/views/utils/permissions_utils.py
```diff
@@ -209,7 +212,7 @@ def get_project_guids_user_can_view(user):
             workspace=Concat('workspace_namespace', Value('/'), 'workspace_name')).filter(
             workspace__in=workspaces).only('guid')]
 
-    safe_redis_set_json(cache_key, project_guids, expire=TERRA_WORKSPACE_CACHE_EXPIRE_SECONDS)
+    safe_redis_set_json(cache_key, sorted(project_guids), expire=TERRA_WORKSPACE_CACHE_EXPIRE_SECONDS)
```
ShifaSZ, line 215: Why is `sorted` required? If it is merely for test churn reasons, I would like to update the tests.

ShifaSZ (changes requested):
The `all_user_demo==True` and `ProjectCategory__name=='Demo'` seem redundant. Should them be unified?

seqr/views/utils/permissions_utils.py
hanars: it is for test churn but anyways its better to have consistent order in the cache anyways, rather than have the order change every time

hanars:
> The `all_user_demo==True` and `ProjectCategory__name=='Demo'` seem redundant. Should them be unified?

No, I specifically built it so it needs 2 unrelated conditions to be met so no one ever accidentally makes a project public to all users

ShifaSZ (approved):

PR #2156, opened 2021-10-05
Upgrade ingress-nginx to handle Ingress API deprecations

### LINKED ISSUE #2150: Upgrade ingress-nginx to latest release for new k8s v1 apiVersion
Due to changes in k8s v1.22: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

The v1beta1 version of the Ingress API that we're using is going away soon in our GKE release channel. Google says that they won't force upgrade GKE clusters that are using this API until March 2022, but we should upgrade this as soon as we can in order to keep getting kubernetes updates.

I think all we need to do is upgrade ingress-nginx, and then upgrade any of our own Ingress objects to the new API.

https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.0.3

sjahl:
Some background info:

- The Ingress API version that we, and our version of ingress-nginx were using has been deprecated, and will be removed from kubernetes v1.22. This means when google upgrades our GKE cluster to that version, our ingress could stop working. https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes
- ingress-nginx released a new backwards incompatible version (1.0.0, and subsequently 1.0.3) to handle this api deprecation: https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.0.3

This PR upgrades our deploy_nginx function to run ingress-nginx 1.0.3, updates our modified versions of the ingress-nginx, Deployment/Service/ConfigMap, and updates the seqr Ingress object to the new Ingress v1 API spec.

closes #2150


hanars:
This looks fine to me, but our policy is to not merge to master until immediately before we deploy. So you should feel free to merge whenever that happens

sjahl:
Ah, yeah, I meant to set the merge target as dev

sjahl:
Going to merge this into dev, as it's deployed there. The upgrade unfortunately results in a few minutes of downtime, so I'll need to schedule around ~10 minutes of downtime when it's time to merge the dev branch to master. Or, we can deploy nginx independently whenever is convenient.

PR #2159, opened 2021-10-05
Bump webpack version

hanars:


ui/.eslintrc
```diff
@@ -22,35 +22,63 @@
       "babel-module": {}
     }
   },
+  //"reportUnusedDisableDirectives": true, // TODO re-enable
   "rules": {
     "semi": [2, "never"],
     "no-shadow": ["error", { "allow": ["error"] }],
     "arrow-body-style": 0,
+    "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
     "func-names": 0,
```
hanars, line 31: we bumped the version for eslint and for the eslint plugins we extend, and the new recommended eslint styles do not completely match up with our code style. We should actually be using the latest recommendations for code style, but that involves changing many of the files in the repo. To keep this PR readable and focused on the webpack changes I disabled many of the failing new rules, and will do a follow-up PR to re-enable them and update the code

ShifaSZ (changes requested):
Mostly looks good to me. One minor comment.

PR #2160, opened 2021-10-07
Added getting token expiry before caching it.

ShifaSZ:
Changed the fixed token expiration time to one fetched from Google.

### DIFF
--- a/seqr/views/apis/igv_api.py
+++ b/seqr/views/apis/igv_api.py
@@ -11,11 +11,10 @@ from seqr.utils.redis_utils import safe_redis_get_json, safe_redis_set_json
 from seqr.views.utils.file_utils import save_uploaded_file
 from seqr.views.utils.json_to_orm_utils import get_or_create_model_from_json
 from seqr.views.utils.json_utils import create_json_response
-from seqr.views.utils.orm_to_json_utils import  get_json_for_sample
+from seqr.views.utils.orm_to_json_utils import get_json_for_sample
 from seqr.views.utils.permissions_utils import get_project_and_check_permissions, check_project_permissions, \
     login_and_policies_required, pm_or_data_manager_required
 
-EXPIRATION_TIME_IN_SECONDS = 3600 - 5
 GS_STORAGE_ACCESS_CACHE_KEY = 'gs_storage_access_cache_entry'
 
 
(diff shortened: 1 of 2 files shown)


seqr/views/apis/igv_api.py
```diff
@@ -161,13 +160,25 @@ def _get_gs_rest_api_headers(range_header, gs_path, user=None):
...
+
 def _get_access_token(user):
     access_token = safe_redis_get_json(GS_STORAGE_ACCESS_CACHE_KEY)
     if not access_token:
         process = run_command('gcloud auth print-access-token', user=user)
         if process.wait() == 0:
             access_token = next(process.stdout).decode('utf-8').strip()
-            safe_redis_set_json(GS_STORAGE_ACCESS_CACHE_KEY, access_token, expire=EXPIRATION_TIME_IN_SECONDS)
+            expires_in = _get_token_expiry(access_token)
+            safe_redis_set_json(GS_STORAGE_ACCESS_CACHE_KEY, access_token, expire=expires_in-1)
```
hanars, line 181: given that there is probably still some lag, I would do `expires_in-5`

seqr/views/apis/igv_api_tests.py
```diff
@@ -59,11 +61,11 @@ def test_proxy_google_to_igv(self, mock_set_redis, mock_get_redis, mock_subproce
...
         self.assertEqual(response.status_code, 200)
-        self.assertIsNone(responses.calls[1].request.headers.get('Range'))
-        self.assertEqual(responses.calls[1].request.headers.get('Authorization'), 'Bearer token2')
-        self.assertIsNone(responses.calls[1].request.headers.get('x-goog-user-project'))
+        self.assertIsNone(responses.calls[3].request.headers.get('Range'))
+        self.assertEqual(responses.calls[3].request.headers.get('Authorization'), 'Bearer token2')
+        self.assertIsNone(responses.calls[3].request.headers.get('x-goog-user-project'))
         mock_get_redis.assert_called_with(GS_STORAGE_ACCESS_CACHE_KEY)
-        mock_set_redis.assert_called_with(GS_STORAGE_ACCESS_CACHE_KEY, 'token2', expire=EXPIRATION_TIME_IN_SECONDS)
+        mock_set_redis.assert_called_with(GS_STORAGE_ACCESS_CACHE_KEY, 'token2', expire=3598)
```
hanars, line 68: we should probably test the case where redis does have a valid token as well

PR #2177, opened 2021-10-13
Eslint - fix import cycle

hanars:
This updates our code to conform with the updated eslint rule prohibiting cyclical dependencies, which is best practice that we want to be following anyways. This PR moves utility functions into utility files rather than having them exported directly from their current files to remove the cyclical dependencies we currently have. All of the code is copy-pasted into new files, nothing actually changed

### DIFF
--- a/ui/shared/components/page/PageHeaderLayout.jsx
+++ b/ui/shared/components/page/PageHeaderLayout.jsx
@@ -30,7 +30,7 @@ const BreadcrumbContainer = styled.div`
   }
 `
 
-export const PageHeaderLayout = React.memo(({
+const PageHeaderLayout = React.memo(({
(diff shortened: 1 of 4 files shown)


ui/shared/components/page/PageHeaderLayout.jsx
```diff
@@ -0,0 +1,146 @@
...
+    margin-bottom: 10px !important;
+  }
+  
+  a.active {
+    color: #111 !important;
+    cursor: auto !important;
+  }
+`
+
+export const PageHeaderLayout = React.memo(({
```
ShifaSZ, line 33: How about export it as default?

ShifaSZ (approved):

PR #2180, opened 2021-10-14
Mme update performance

hanars:
Searching after updating/ creating a matchmaker submission can be slow, causing the request to time out. This makes users incorrectly think that the update to the submission failed, when in reality it succeeded and only searching for updated results failed. This changes the behavior to not perform a search as part of the update request, and instead trigger a separate search request after the submission update succeeds


seqr/views/utils/orm_to_json_utils.py
```diff
@@ -854,5 +854,7 @@ def get_json_for_matchmaker_submissions(models, individual_guid=None, additional
     )
 
 
-def get_json_for_matchmaker_submission(submission, **kwargs):
-    return _get_json_for_model(submission, get_json_for_models=get_json_for_matchmaker_submissions, **kwargs)
\ No newline at end of file
+def get_json_for_matchmaker_submission(submission):
+    return _get_json_for_model(
+        submission, get_json_for_models=get_json_for_matchmaker_submissions, individual_guid=submission.individual.guid,
+        additional_model_fields=['contact_name', 'contact_href', 'submission_id'])
```
ShifaSZ, line 860: Needs a newline at the end.

ShifaSZ (changes requested):
Looks good to me.

ShifaSZ (approved):

PR #2182, opened 2021-10-14
update IGV range size for SVs

hanars:


ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -179,11 +181,12 @@ ReadButtons.propTypes = {
 const IgvPanel = React.memo((
   { variant, igvSampleIndividuals, individualsByGuid, project, sampleTypes, rnaReferences },
 ) => {
+  const size = variant.end && variant.end - variant.pos
   const locus = variant && getLocus(
     variant.chrom,
     (variant.genomeVersion !== project.genomeVersion && variant.liftedOverPos) ? variant.liftedOverPos : variant.pos,
-    100,
-    variant.end && variant.end - variant.pos,
+    size ? Math.max(parseInt(size / 3, 10), MIN_LOCUS_RANGE_SIZE) : MIN_LOCUS_RANGE_SIZE,
```
ShifaSZ, line 188: My guess is that `parseInt` here is for rounding. Would `Math.round` be more straightforward?

ShifaSZ (approved):

PR #2187, opened 2021-10-18
Adding IGV junction visibility options

ShifaSZ:


ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -82,9 +84,11 @@ const getIgvTracks = (igvSampleIndividuals, individualsByGuid, sampleTypes) => {
               type: 'merged',
               name: track.name,
               height: track.height,
+              updated: minJunctionEndsVisible,
               tracks: [coverageTrack, track],
             }
           }
+          track.updated = minJunctionEndsVisible
```
hanars, line 91: I'm confused why you are setting a property called `updated` on the track - from the TGG viewer code it looks like `minJunctionEndsVisible` is an option that should be passed directly to the track itself - https://github.com/broadinstitute/tgg-viewer/blob/master/src/redux/selectors.js#L182

ShifaSZ: The IGV.js doesn't update the display even when the track options are changed. The IGV component in seqr has to update (or refresh) the display after the options of the tracks are changed. Currently, only deleted and added tracks are refreshed by the IGV component. I added a new option `updated`. The IGV component will refresh the track if the value of the `updated` option is changed. The IGV.js doesn't care about any unknown options.
The TGG-Viewer's junction visibility options aren't working at least on my computer. I don't have a chance to check with Ben yet.

hanars: I think it would be better to instead define a constant list of track properties to check to see if they changed and check those for reloading, rather than sending a meaningless extra property which can only support one property for update, especially since you never actually set the `minJunctionEndsVisible` property on the track itself. Something like
```
const TRACK_UPDATE_PROPERTIES = ['minJunctionEndsVisible']
...
const prevTracksById = prevProps.tracks.reduce((acc, track) => ({ ...acc, [getTrackId(track)]: track }), {})
const prevTrackIds = Object.keys(prevTracksById)
...
tracks.forEach((track) => {
  const prevTrack = track.name && prevTracksById[track.name]
  if (prevTrack) {
    const shouldUpdate = TRACK_UPDATE_PROPERTIES.some(property => track[property] !== prevTrack[property])
    if (shouldUpdate) {
      this.browser.removeTrackByName(track.name)
      this.browser.loadTrack(track)
    }
  }
})
```


> The TGG-Viewer's junction visibility options aren't working at least on my computer.

We can't add a control panel to seqr that does not work! If this is not working then this absolutely can not be deployed, and in fact you should not have PR'd a change until that issues is resolved.
That said, since you never actually add the property to the junction track, that might be why its not working, so I would try updating as I suggested and seeing if that fixes the issue

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -189,7 +199,9 @@ const IgvPanel = React.memo((
     size,
   )
 
-  const tracks = rnaReferences.concat(getIgvTracks(igvSampleIndividuals, individualsByGuid, sampleTypes))
+  const rnaRefs = updateJunctionRefs(rnaReferences, minJunctionEndsVisible)
```
hanars, line 202: this seems odd to me. Do we actually want to update the reference track display as well?

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
hanars: So in that case, I think it is a bit cleaner to post-process all the track together, the way TGG viewer does, rather than separately updating the reference tracks and regular tracks. Also, l I think the name `updateJunctionRefs` is not as descriptive as it could be and currently seems to only handle merged tracks and not pure junction tracks. I would name it something more generic in case we add any other settings that require post processing in the future, maybe something like `applyUserTrackSettings`. So the change here would just be
```
const tracks = applyUserTrackSettings(rnaReferences.concat(getIgvTracks(igvSampleIndividuals, individualsByGuid, sampleTypes)))
```

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -178,8 +179,22 @@ ReadButtons.propTypes = {
   showReads: PropTypes.func,
 }
 
+const optionReducer = trackType => (acc, { type, field, value }) => ({
+  ...acc,
+  ...(trackType === type) ? { [field]: value } : {},
```
hanars, line 184: I think this would be more readable as 
```
trackType === type ? { ...acc, [field]: value } : acc
```
and I also think it needs a more descriptive name, maybe `addOptionsToTrackType`

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -329,6 +358,7 @@ class FamilyReads extends React.PureComponent {
             igvSampleIndividuals={igvSampleIndividuals}
             sampleTypes={sampleTypes}
             rnaReferences={rnaReferences}
+            options={[{ type: JUNCTION_TYPE, field: 'minJunctionEndsVisible', value: minJunctionEndsVisible }]}
```
hanars, line 361: This is a premature optimization, and it will immediately conflict with react-perf linting rules I reenable in my open PR. Revert this back to `minJunctionEndsVisible` being a direct prop

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -189,7 +204,9 @@ const IgvPanel = React.memo((
     size,
   )
 
-  const tracks = rnaReferences.concat(getIgvTracks(igvSampleIndividuals, individualsByGuid, sampleTypes))
+  const tracks = applyUserTrackSettings(
+    rnaReferences.concat(getIgvTracks(igvSampleIndividuals, individualsByGuid, sampleTypes)), options,
+  )
```
hanars, line 209: As I say below, we should be passing `minJunctionEndsVisible`, and I think we don't wan to get too bogged down in making this generic given that we don't exactly know what  future enhancements would be. That said, I think we would want to apply all options for a track type at once, not iterate through multiple times if we end up needing to add multiple properties to the same track type. I think we should replace `options` here with 
`{ [JUNCTION_TYPE]: { minJunctionEndsVisible } }` and then update `applyUserTrackSettings` accordingly

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -178,8 +179,15 @@ ReadButtons.propTypes = {
   showReads: PropTypes.func,
 }
 
+const applyUserTrackSettings = (tracks, options) => tracks.map(track => ({
+  ...Object.keys(options).includes(track.type) ? { ...track, ...options[track.type] } : track,
```
hanars, line 183: its more performant to just try and access the value directly, rather than convert the object to a key list and check if the list contains a string. So just `options[track.type] ? { ...track, ...options[track.type] } : track` and the same for the merged track below

hanars (approved):

PR #2189, opened 2021-10-18
Bump IGV version to show data range and remove the autoscale option.

ShifaSZ:
Bump the IGV version because the IGV v2.8.6 can't display the data range of tracks properly.  Remove the `autoscale option for gCNV tracks to fix the scale from 0 to 5.


hanars:
okay, but a change in every single one of our minor package versions for no clear reason is an unacceptable risk that I can't approve. Also, your JS unit test are failing, which indicates to me that the versions you have installed do not work properly for us. 
When you checked out dev and ran npm install, did that generate a change in your package.lock? Because a clean npm install should install everything in package.lock without changing it

PR #2193, opened 2021-10-20
Eslint react perf

hanars:
re-enable eslint rules for react performance enhancements. Specifically, this prevents the creation of objects, arrays, and functions as props in the same scope as the component they are passed to. This improved react performance, as otherwise React will always see these props as changed and will re-render the component. 

For many of these change, it made sense to convert a functional component to a react class, so the diff is quite large despite not much actually changing


ui/shared/components/panel/family/FamilyLayout.jsx
```diff
@@ -26,11 +26,11 @@ const FamilyLayout = React.memo((
     <FamilyGrid annotation={annotation} offset={offset}>
       <Grid.Row>
         {(leftContent || !useFullWidth) && <Grid.Column width={3}>{leftContent}</Grid.Column>}
-        {compact ? fields.map(
+        {compact ? (fields || []).map(
           field => <Grid.Column width={field.colWidth || 1} key={field.id}>{fieldDisplay(field)}</Grid.Column>,
         ) : (
           <Grid.Column width={getContentWidth(useFullWidth, leftContent, rightContent)}>
-            {fields.map(field => fieldDisplay(field))}
+            {(fields || []).map(field => fieldDisplay(field))}
```
ShifaSZ, line 33: Are there any real use-cases that `fields` can be null?

ui/shared/components/panel/genes/GeneDetail.jsx
```diff
@@ -35,11 +37,11 @@ const GeneSection = React.memo(({ details }) => (
...
   </CompactGrid>
-))
+)
 
-GeneSection.propTypes = {
-  details: PropTypes.arrayOf(PropTypes.object),
-}
+// GeneSection.propTypes = {
+//   details: PropTypes.arrayOf(PropTypes.object),
+// }
```
ShifaSZ, line 44: Do we still need these lines?

ShifaSZ (changes requested):
I just ran through quickly.

ui/pages/SummaryData/components/SavedVariants.jsx
```diff
@@ -64,27 +65,14 @@ TAG_OPTIONS.push({
...
 
-const BaseSavedVariants = React.memo(({ loadAllProjectSavedVariants, geneDetail, ...props }) => {
-  const { params } = props.match
-  const { tag, gene } = params
-
-  const getUpdateTagUrl = selectedTag => `/summary_data/saved_variants/${selectedTag}${gene ? `/${gene}` : ''}`
-
-  const getGeneHref = selectedGene => `/summary_data/saved_variants/${tag || SHOW_ALL}/${selectedGene.key}`
+const getUpdateTagUrl =
+  (selectedTag, match) => `/summary_data/saved_variants/${selectedTag}${match.params.gene ? `/${match.params.gene}` : ''}`
```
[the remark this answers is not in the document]
hanars: no, literally the entire purpose of this PR is to stop defining function props in the same scope that they are used

ui/shared/components/panel/genes/GeneDetail.jsx
hanars: no, good catch

PR #2194, opened 2021-10-20
better mme notification

hanars:


matchmaker/views/external_api.py
```diff
@@ -82,7 +82,11 @@ def mme_match_proxy(request, originating_node_name):
     try:
         _generate_notification_for_incoming_match(results, incoming_query, originating_node_name, query_patient_data)
     except Exception as e:
```
ShifaSZ, line 84: The `Exception` can be more specific to `ValueError`.

ShifaSZ (changes requested):

matchmaker/views/external_api.py
hanars: no, we need to handle every error no matter what kind of exception happens, even if its a key error from parsing a malformed result or something else we have no way of expecting

PR #2198, opened 2021-10-20
Panel app display updates

hanars:


ui/pages/Search/components/filters/LocusListSelector.jsx
```diff
@@ -40,7 +40,7 @@ class BaseLocusListDropdown extends React.Component {
           label="Gene List"
           value={locusList.locusListGuid}
           onChange={locusListGuid => onChange({ locusListGuid })}
-          options={[{ text: 'None', value: null }].concat(projectLocusListOptions)}
+          options={projectLocusListOptions}
```
ShifaSZ, line 43: Why should the `null` option be removed?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -20,9 +20,15 @@ const INLINE_STYLE = {
   display: 'inline-block',
 }
 
-const BaseGeneLabelContent = styled(
-  ({ color, label, maxWidth, containerStyle, dispatch, ...props }) => <Label {...props} size="mini" color={color || 'grey'} content={label} />,
-)`
+const BaseGeneLabelContent = styled(({ color, customColor, label, maxWidth, containerStyle, dispatch, ...props }) => {
```
ShifaSZ, line 23: It looks like that `GeneLabelContent` is the same as `BaseGeneLabelContent`. Why don't we remove one of them?

PR #2201, opened 2021-10-21
fix gene detail note

hanars:


hanars:
fixed in https://github.com/broadinstitute/seqr/pull/2193

PR #2233, opened 2021-10-26
Docker Cleanup Part 1

sjahl:
I think this is ready for an initial review. More changes to come eventually, but I think this is the smallest unit of work that improves our docker build to the point where we can start automating it. The resulting image here should be a little bit smaller in size than our current one, and should maintain equivalent functionality so that the `docker_build()` method could still be used for generating release images.

Major changes include:
- Rebasing our image on `python:3.7-slim-bullseye`. This is the official debian-based python image that provides python 3.7.12 presently. We also get upgraded from debian buster (which was EOL), but I don't think there's much we depend on there.
- Removing all the steps that download and compile python, since that is functionally provided by the python base image
- Moving all of the apt package installations into a single `RUN` instruction
- Removes duplicate gcloud/gsutil SDK installations, kubectl, and the full postgres database package installation (since I think we only need the psql client).
- Configures a [hadolint](https://github.com/hadolint/hadolint) github action to lint the seqr Dockerfile. I've also applied fixes to the issues raised by the linter
- A cloudbuild configuration that can be used to trigger automatic builds of the seqr image (currently in a testing repository) when the seqr `dev` branch gets new commits.

Closes https://github.com/broadinstitute/seqr-private/issues/1061, https://github.com/broadinstitute/seqr-private/issues/1063

Touches https://github.com/broadinstitute/seqr-private/issues/1062, https://github.com/broadinstitute/seqr-private/issues/1065

### DIFF
--- a/.cloudbuild/seqr-docker.cloudbuild.yaml
+++ b/.cloudbuild/seqr-docker.cloudbuild.yaml
@@ -1,0 +1,12 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  dir: 'deploy/docker/seqr'
+  args: [ 'build', '.', '-t', 'seqr-docker-cleanup-testing:$COMMIT_SHA']
+- name: 'gcr.io/cloud-builders/docker'
+  dir: 'deploy/docker/seqr'
+  args: [ 'tag', 'seqr-docker-cleanup-testing:$COMMIT_SHA', 'us-central1-docker.pkg.dev/tgg-sre-testing/seqr-testing/seqr-docker-cleanup-testing:$COMMIT_SHA']
+
+images:
+  - 'us-central1-docker.pkg.dev/tgg-sre-testing/seqr-testing/seqr-docker-cleanup-testing:$COMMIT_SHA'
+
+timeout: 900s
--- a/.cloudbuild/seqr-docker.cloudbuild.yaml
+++ b/.cloudbuild/seqr-docker.cloudbuild.yaml
@@ -1,12 +1,12 @@
 steps:
 - name: 'gcr.io/cloud-builders/docker'
   dir: 'deploy/docker/seqr'
-  args: [ 'build', '.', '-t', 'seqr-docker-cleanup-testing:$COMMIT_SHA']
+  args: [ 'build', '.', '-t', 'seqr:$COMMIT_SHA' ]
 - name: 'gcr.io/cloud-builders/docker'
   dir: 'deploy/docker/seqr'
-  args: [ 'tag', 'seqr-docker-cleanup-testing:$COMMIT_SHA', 'us-central1-docker.pkg.dev/tgg-sre-testing/seqr-testing/seqr-docker-cleanup-testing:$COMMIT_SHA']
+  args: [ 'tag', 'seqr:$COMMIT_SHA', 'gcr.io/seqr-project/seqr:$COMMIT_SHA']
 
 images:
-  - 'us-central1-docker.pkg.dev/tgg-sre-testing/seqr-testing/seqr-docker-cleanup-testing:$COMMIT_SHA'
+  - 'gcr.io/seqr-project/seqr:$COMMIT_SHA'
 
 timeout: 900s
--- a/.github/workflows/docker-lint.yaml
+++ b/.github/workflows/docker-lint.yaml
@@ -9,6 +9,7 @@ on:
     paths:
       - deploy/docker/seqr/Dockerfile
       - .hadolint.yaml
+      - .github/workflows/docker-lint.yaml
   # pull_request:
   #   types: [opened, synchronize, reopened]
 
--- a/.github/workflows/docker-lint.yaml
+++ b/.github/workflows/docker-lint.yaml
@@ -1,0 +1,22 @@
+name: Dockerfile Linting
+
+# Run the test suite on pushes (incl. merges) to master and dev
+# Run the test suite when a PR is opened, pushed to, or reopened
+on:
+  push:
+    branches:
+      - docker-rework
+    paths:
+      - deploy/docker/seqr/Dockerfile
+      - .hadolint.yaml
+  # pull_request:
+  #   types: [opened, synchronize, reopened]
+
+jobs:
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2 
+      - uses: hadolint/hadolint-action@v1.5.0
+        with:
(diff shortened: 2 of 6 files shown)


ShifaSZ (dismissed):

deploy/docker/seqr/Dockerfile
```diff
@@ -102,7 +43,7 @@ RUN git clone -q --branch $SEQR_GIT_BRANCH $SEQR_REPO
 WORKDIR /seqr
 
 # install seqr dependencies
-RUN pip install wheel && pip install -r requirements.txt
+RUN pip install --no-cache-dir wheel && pip install --no-cache-dir -r requirements.txt
 
 ARG SEQR_SERVICE_PORT
 ENV SEQR_SERVICE_PORT=$SEQR_SERVICE_PORT
```
sjahl, line 49: @hanars @ShifaSZ Do either of you see any value in making the seqr service port a variable like this? I'm not sure I see a case where this would need to change, and propose getting rid of this and setting the expose port to 8000 statically.

If someone's running the container in an infrastructure where using port 8000 is problematic, the "right" way to deal with that is for them to change the port on the kubernetes Service, or publish a different port with the docker `-P` flag.

sjahl: In other words, the `EXPOSE` directive should always be 8000 as long as gunicorn is configured to listen on that port in https://github.com/broadinstitute/seqr/blob/master/deploy/docker/seqr/config/gunicorn_config.py#L2 -- setting it to anything else won't work.

hanars: The only reason I can think of to make this configurable is if someone has a local installation where they need to change the port, but as long as thats something they can change via how they run docker compose, I'm fine to make this static

sjahl:
So, with the seqr port as a docker build arg, someone would have to rebuild
the container and update the gunicorn config if they wanted to change the
port that seqr listens on.

If the container always exposes port 8000, they can reuse the same image
without changing anything. Instead, they would change which port their
kubernetes Service or docker-compose configuration was mapping back to the
container. The expose directive doesn’t need to change for that to be
configurable.

Does that make sense?

.cloudbuild/seqr-docker.cloudbuild.yaml
```diff
@@ -0,0 +1,12 @@
+steps:
```
hanars, line 1: It looks like this no longer has an automated trigger hooked up to it - is that correct?

.github/workflows/docker-lint.yaml
```diff
@@ -0,0 +1,23 @@
+name: Dockerfile Linting
+
+# Run the test suite on pushes (incl. merges) to master and dev
+# Run the test suite when a PR is opened, pushed to, or reopened
+on:
+  push:
+    branches:
+      - docker-cleanup-part-1
```
hanars, line 8: I assume we want to update this before we merge

hanars (commented):
This all looks great to me. I'm at an approval but just want to come up with a deployment plan before actually hitting that final approve button. It strikes me that this change should have no user-facing impact, but we still probably want to consider like 15 minutes or so of scheduled downtime for this. Would we want to build and push up the docker image in advance, and at our deploy time just redeploy seqr with the new image? Or would we want to wait to build the image until we are ready to release it? Also, do you think its better to merg ths branch to master before or after we build the image?

.github/workflows/docker-lint.yaml
sjahl: Yeah, we'll uncomment the `pull_request` section here, and also have the job run on changes to the dev and master branches. Goal is to have the linter run on pull requests, and then again post-merge.

sjahl:
> This all looks great to me. I'm at an approval but just want to come up with a deployment plan before actually hitting that final approve button. It strikes me that this change should have no user-facing impact, but we still probably want to consider like 15 minutes or so of scheduled downtime for this. Would we want to build and push up the docker image in advance, and at our deploy time just redeploy seqr with the new image? Or would we want to wait to build the image until we are ready to release it? Also, do you think its better to merg ths branch to master before or after we build the image?

@hanars Yep, there's a few little cleanup branch-naming things that you pointed out, and then I think it's ready to go. Let me think about a release plan for a bit, and I'll post back here. Generally speaking, I think it makes sense to merge this into dev, build a new image, and then redeploy seqr-dev -- If we can, we'll want to isolate that deployment to only the docker image updating, so that we don't conflate problems with a seqr app-code update. With what we learn there, we can plan for prod/master.

.cloudbuild/seqr-docker.cloudbuild.yaml
hanars: agreed, with the way we currently handle static assets we can't have an automated build from the dev branch

sjahl:
Ok, I think the ship plan here is, starting with dev:

- merge this PR into dev
- build and push a new docker image (and tag gcloud-dev) using our current servctl facility
- `servctl deploy -d seqr` to pick up the new container image.
- Test that the application is running as expected

If all goes well there, then I think this can go out as normal with the next seqr prod release (with the additional step of rebuilding and tagging the gcloud-prod seqr image). That'll require a few minutes of downtime, as we'll need to cycle the seqr container to pick up the new image. If we don't have a prod release planned imminently, I'm ok with shipping just this, depending on what else has creeped into the dev..master diff at that point.

hanars:
So we never have a planned seqr release that involves building a docker image and deleting and restarting the pod, our planned deployments all involve running the `redeploy_seqr.sh` helper that just updates the code running in the existing image. Once we switch to automated docker builds that will obviously change, but for now this will need a special deployment. Because of that, I think it might be easier if we PR this directly to master instead of dev. I would still say you do all the testing on dev that you propose here but you can do that off of your branch. Then we could change the base of this PR to master, and I will plan a small amount of downtime to do this deployment. Does that make sense to you?

sjahl:
Yep, that makes sense. I was able to test a local install today with the new image, and I'll try and get that image pushed and tested in dev by tomorrow. I'll switch the merge base on this once that seems ready to go.

sjahl:
@hanars I think the checklist is good to go -- I'm unsure of whether image changes like the ones in this PR should go in the changelog (particularly since these shouldn't have much of a user impact). I left that one un-checked, and am happy to update the changelog if you think I should.

hanars:
I think these changes do belong in the changelog, since it is mainly maintained for users who are running seqr locally via the docker image. While this should have no impact on them, I still think it is worth noting

hanars (approved):

PR #2250, opened 2021-10-29
Lazy load assets

hanars:
Using webpack and react, we can choose to lazily load imports so they are only included when a user actually needs them. These lazily loaded components are nto added to the main entrypoint bundle, and are instead loaded as their own asset files as needed. See https://reactjs.org/docs/code-splitting.html#named-exports

Theres no perfect way to do this, but in general for larger libraries that are not used through most of the views in seqr I opted to have them load lazily. Overall, reducers the main assets as follows:
```
app.js: 842.58 KB ->  356.33 KB 
app.css: 100.37->  98.9 KB 
```


ui/pages/Dashboard/components/CategoryIndicator.jsx
```diff
@@ -1,19 +1,11 @@
...
 import PropTypes from 'prop-types'
-import randomMC from 'random-material-color'
-import styled from 'styled-components'
-import { Icon } from 'semantic-ui-react'
+import { Loader } from 'semantic-ui-react'
 
 import { ButtonLink } from 'shared/components/StyledComponents'
 import EditProjectCategoriesModal from './EditProjectCategoriesModal'
 
-const getColor = categoryNames => (
```
hanars, line 10: thsi was all copied into another file so it could be lazily imported here

ui/pages/Dashboard/components/ComputedColoredIcon.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import randomMC from 'random-material-color'
+import styled from 'styled-components'
```
hanars, line 3: this is copied form the previous file so does not need review

ui/shared/components/form/XHRUploaderWithEvents.jsx
```diff
@@ -0,0 +1,148 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+
+import React from 'react'
```
hanars, line 4: copied from another file, does not need review

ui/shared/components/panel/search/LabeledSlider.jsx
```diff
@@ -0,0 +1,36 @@
+import styled from 'styled-components'
+import Slider from 'react-rangeslider'
+import 'react-rangeslider/lib/index.css'
+
+export default styled(Slider).attrs(props => ({
```
hanars, line 5: copied from another file, does not need review

ui/shared/components/panel/view-pedigree-image/LazyPedigreeImagePanel.jsx
```diff
@@ -0,0 +1,379 @@
+import React from 'react'
```
hanars, line 1: this is just a renamed file, I don't know why git is not picking that up, but it doesn't need review

ui/shared/utils/stringUtils.js
```diff
@@ -14,7 +11,4 @@ export const toCamelcase = s => (s || '').split(' ').map(
   (word, i) => word && (i > 0 ? word[0].toUpperCase() : word[0].toLowerCase()) + word.slice(1),
 ).join('')
 
-const TEXT_RENDERERS = Object.keys(markdownRenderers).reduce((acc, k) => ({ ...acc, [k]: markdownRenderers.text }), {})
-
-export const stripMarkdown =
-  s => ReactDOMServer.renderToStaticMarkup(<ReactMarkdown renderers={TEXT_RENDERERS}>{s || ''}</ReactMarkdown>).trim()
+export const stripMarkdown = s => removeMd(s || '').replaceAll('\n', ' ')
```
hanars, line 14: use a lightweight library for this instead of the very heavy libraries, which are changed to be lazily loaded in the one other place that needs them

ui/pages/Dashboard/components/ComputedColoredIcon.jsx
ShifaSZ: I'm curious that the module `ComputedColoredIcon` doesn't look big. Why should it be a separate lazy-loaded module?

ShifaSZ (changes requested):

ui/pages/Dashboard/components/ComputedColoredIcon.jsx
hanars: because it uses the `randomMC` module which is used nowhere else in the app. Separating out this component removes that node module from the main bundle

ShifaSZ (approved):

PR #2253, opened 2021-11-02
In silico updates

hanars:
Updates the layout and search behavior of the in silico filters, including moving the splice AI filter to be part of the "annotation" filter instead of keeping it with the other in silico filters. Also, cleans up some in-silico related server side and client side constants for better code reuse


ShifaSZ (changes requested):
Looks great. There is one thing that maybe can be improved.

ShifaSZ (approved):

PR #2260, opened 2021-11-02
hide long lists of datasets

hanars:


ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -181,17 +181,48 @@ const mapMatchmakerStateToProps = (state, ownProps) => ({
...
+
+    const total = allLoads.length
+    if (total < 6 || showAll) {
+      return allLoads
+    }
+
+    return [
+      ...allLoads.slice(0, 2),
+      <ButtonLink key="show" padding="5px 0" onClick={this.show}>{`Show ${total - 5} additional datasets`}</ButtonLink>,
+      ...allLoads.slice(total - 3),
```
ShifaSZ, line 213: A nitpick issue is to show the "show more" button only when it needs more lines to show all. So, if the length is 6, then it is better to show them all. So the condition for showing all should be `total <= 6`.

ShifaSZ (changes requested):

ui/pages/Project/components/ProjectOverview.jsx
hanars: no I intentionally did strict less because if the length is 6 we do not want to show them all, we only want to show the first 5. thats why we use `total - 5` when calculating the number that are hidden

ShifaSZ (approved):

PR #2277, opened 2021-11-08
Mme variant link

hanars:


ui/shared/components/panel/variants/FamilyVariantTags.jsx
```diff
@@ -224,6 +224,8 @@ const FamilyVariantTags = React.memo(({
               variantId={variantId}
               tagOptions={projectTagTypes}
               displayMetadata
+              linkTagType="seqr MME"
+              tagLinkUrl={`/project/${family.projectGuid}/family_page/${family.familyGuid}/matchmaker_exchange`}
```
ShifaSZ, line 228: It is weird to me that you add the MME link to the `VariantTagField`. The requirement is to add the link to the `saved variant`. It seems more reasonable to add the link beside the family ID.

ShifaSZ (changes requested):

ui/shared/components/panel/variants/FamilyVariantTags.jsx
hanars: From the ticket:
```
 Would be cool if the Matchbox tag linked directly to the MME page when you clicked on it. 
```

I agreed , and thought that was a better user interface than adding yet another link to an already crowded layout

ShifaSZ (approved):

PR #2278, opened 2021-11-09
Filter by submitter

hanars:


ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -15,34 +15,49 @@ import {
...
   VARIANT_PER_PAGE_FIELD,
+  {
+    name: 'savedBy',
+    component: connect(mapSavedByInputStateToProps)(Dropdown),
+    inline: true,
+    selection: false,
+    fluid: false,
+    label: 'Saved By:',
+    placeholder: 'Select a user',
+  },
```
ShifaSZ, line 57: I would be more consistent if also defining a constant for the `savedBy` fields.

ShifaSZ (changes requested):

ui/pages/Project/components/SavedVariants.jsx
hanars: not really, we only define constants once they are used in multiple places

ShifaSZ (approved):

PR #2288, opened 2021-11-10
Clinvar override frequency filter

hanars:


seqr/utils/elasticsearch/es_search.py
```diff
@@ -202,21 +203,35 @@ def filter_by_in_silico(self, in_silico_filters):
         if in_silico_filters:
             self.filter(_in_silico_filter(in_silico_filters))
 
-    def filter_by_frequency(self, frequencies):
+    def filter_by_frequency(self, frequencies, pathogenicity=None):
```
ShifaSZ, line 206: You may also want to change the function name to reflex the pathogenic overwriting.

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_search.py
hanars: not really, I think that makes it more complicated. The actually pathogenicity filtering is still happening elsewhere

ShifaSZ (approved):

PR #2291, opened 2021-11-15
error handling for invalid family id update

hanars:


seqr/views/apis/family_api.py
```diff
@@ -41,12 +41,44 @@ def edit_families_handler(request, project_guid):
...
+    if family_guids:
+        family_models.update({f.guid: f for f in Family.objects.filter(project=project, guid__in=family_guids)})
+        if len(family_models) != len(family_guids):
+            missing_guids = set(family_guids) - set(family_models.keys())
+            return create_json_response({'error': 'Invalid family guids: {}'.format(', '.join(missing_guids))}, status=400)
+
+        updated_family_ids = {
+            fields[FAMILY_ID_FIELD]: family_models[fields['familyGuid']].family_id for fields in modified_families
+            if fields.get('familyGuid') and fields.get(FAMILY_ID_FIELD) and \
+                fields[FAMILY_ID_FIELD] != family_models[fields['familyGuid']].family_id}
```
ShifaSZ, line 55: The `family_models[fields['familyGuid']]` could be undefined when the family with a GUID of `fields['familyGuid']` couldn't be found. Can this case happen?

seqr/views/apis/family_api.py
```diff
@@ -41,12 +41,44 @@ def edit_families_handler(request, project_guid):
         return create_json_response(
             {}, status=400, reason="'families' not specified")
 
+    family_guids = [f['familyGuid'] for f in modified_families if f.get('familyGuid')]
```
ShifaSZ, line 44: Would it be cleaner by using the code below?
```
family_guids = [x for x in [f.get('familyGuid') for f in modified_families] if x is not None]
```

ShifaSZ (changes requested):

seqr/views/apis/family_api.py
hanars: no, a single list comprehension is much faster and easier to read the a nested double list comprehension

seqr/views/apis/family_api.py
ShifaSZ: I mean if `family_models[fields['familyGuid']]` is undefined or None, then `family_models[fields['familyGuid']].family_id` will be an error.

hanars: there is error handling 5 lines above this that returns an error to the user if there isn't a model for any of the guids, so thats completely impossible

ShifaSZ: Okay, I see what you mean.

ShifaSZ (approved):

PR #2292, opened 2021-11-15
Add IGV variant/gene range options

ShifaSZ:


ui/shared/components/graph/IGV.jsx
```diff
@@ -42,7 +43,15 @@ class IGV extends React.PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    const { tracks } = this.props
+    const { tracks, locus } = this.props
+    if (this.browser && locus !== prevProps.locus) {
+      if (this.container) {
+        igv.removeBrowser(this.browser)
+        igv.createBrowser(this.container, { ...this.props }).then((browser) => {
```
ShifaSZ, line 50: Maybe there is a better way to refresh the browser when the locus changed.

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -197,10 +199,40 @@ const getVariantLocus = (variant, project) => {
   )
 }
 
-const IgvPanel = React.memo((
-  { variant, igvSampleIndividuals, individualsByGuid, project, sampleTypes, rnaReferences, minJunctionEndsVisible },
+const getGeneLocus = (variant, genesById, project) => {
+  if (variant.transcripts) {
+    const geneIds = Object.keys(variant.transcripts)
+    const geneId = geneIds.length > 0 ? geneIds[0] : ''
```
ShifaSZ, line 205: Which gene will be displayed in the case of multiple genes for a variant?

hanars: the main gene, there is a `getVariantMainGeneId` helper function you can use

ui/shared/components/graph/IGV.jsx
hanars: probably, I suggest you read the IGV JS documentation to see

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -348,6 +393,12 @@ class FamilyReads extends React.PureComponent {
                 />
                 {sampleTypes.some(sampleType => RNA_TRACK_TYPE_LOOKUP.has(sampleType)) && (
                   <div>
+                    <Divider horizontal>Range</Divider>
```
hanars, line 396: Per the ticket: This option will be available for all data types.
This means its always shown, not just if an RNA seq type is shown

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -197,10 +199,40 @@ const getVariantLocus = (variant, project) => {
...
 
-const IgvPanel = React.memo((
-  { variant, igvSampleIndividuals, individualsByGuid, project, sampleTypes, rnaReferences, minJunctionEndsVisible },
+const getGeneLocus = (variant, genesById, project) => {
+  if (variant.transcripts) {
+    const geneIds = Object.keys(variant.transcripts)
+    const geneId = geneIds.length > 0 ? geneIds[0] : ''
+    const gene = genesById[geneId]
+    if (gene) {
+      const genomeVersion = (project.genomeVersion === GENOME_VERSION_38) ? 'Grch38' : 'Grch37'
```
hanars, line 208: Use GENOME_VERSION_LOOKUP to get the display value

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -197,10 +199,40 @@ const getVariantLocus = (variant, project) => {
...
+      const size = gene[`codingRegionSize${genomeVersion}`]
+      return getLocus(
+        gene[`chrom${genomeVersion}`],
+        gene[`start${genomeVersion}`],
+        size ? Math.max(Math.round(size / 3), MIN_LOCUS_RANGE_SIZE) : MIN_LOCUS_RANGE_SIZE,
+        size,
+      )
+    }
+  }
+  // If gene locus is not available, return the variant locus
```
hanars, line 218: if gene locus is not available we should not show the option at all, selecting "gene" and having nothing change is a really bad user experience

hanars: thinking about it, it might be better to dynamically generate the range dropdown options the way we do for `dnaTrackOptions`. Then you make the value the actual locus itself and pass that - something like 
```
const geneLocus = variant && getGeneLocus(variant, genesById, project)
const rangeOptions = geneLocus ? [{ value: getVariantLocus(variant, project), text: 'Variant' },
  { value: geneLocus, text: 'Gene' },] : null
...
{rangeOptions && ... <RadioGroup ...}
```

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -367,19 +390,29 @@ class FamilyReads extends React.PureComponent {
                 )}
               </div>
             )}
+            { locusOptions && (
```
hanars, line 393: this should display above the rna seq tracks/filters

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -326,6 +345,10 @@ class FamilyReads extends React.PureComponent {
       openFamily && (igvSamplesByFamilySampleIndividual || {})[openFamily]) || {}
     const dnaTrackOptions = DNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
     const rnaTrackOptions = RNA_TRACK_TYPE_OPTIONS.filter(({ value }) => igvSampleIndividuals[value])
+    const project = openFamily && projectsByGuid[familiesByGuid[openFamily].projectGuid]
+    const geneLocus = project && variant && getGeneLocus(variant, genesById, project)
+    const locusOptions = geneLocus ? [{ value: variant && getVariantLocus(variant, project), text: 'Variant' },
```
hanars, line 350: `geneLocus` is already conditional on `variant`, so theres no need to do `variant && getVariantLocus` here

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -235,19 +246,23 @@ class FamilyReads extends React.PureComponent {
...
     minJunctionEndsVisible: 0,
+    locus: null,
   }
 
   showReads = familyGuid => sampleTypes => () => {
+    const { variant, projectsByGuid, familiesByGuid } = this.props
     this.setState({
       openFamily: familyGuid,
       sampleTypes,
+      locus: variant && getVariantLocus(variant, projectsByGuid[familiesByGuid[familyGuid].projectGuid]),
```
hanars, line 265: not crucial, but it might be helpful/ more readable to write a helper function like `this.getProjectForFamily(familyGuid)` that is used  both here and in the main render method

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
ShifaSZ: Don't you think the rna seq tracks should stay together with the DNA tracks?

hanars: no, like I said, I think the order should be: select track type, select range, select any rna-specific stuff

hanars (approved):

PR #2124, opened 2021-09-16
Add acmg classification to saved variants

athegaul:
ACMG Classification is added to saved variants. It includes migrations, the API change, and UI change. Below are screenshots that show changes on UI.

image

image

image

image

image

image

image

image

image

image

PR #2293, opened 2021-11-15
Acmg classification updates

hanars:
Cleans up the ACMG code submitted through an external contributor in https://github.com/broadinstitute/seqr/pull/2124

This fixes some small bugs and console errors from missing keys and malformed HTML tables and some linting issues. I also cleaned up the way they were using state in one particular file, as that was really anti-pattern to our usual approach. Aside form that one file, however, I am not cleaning up style differences aws I don't have the bandwidth to do that for all external contributions


ui/shared/components/panel/acmg/AcmgCriteria.jsx
```diff
@@ -17,6 +19,19 @@ const CATEGORY_CONFIGS = [
   { classification: 'Likely Benign', isThisClassification: value => value >= -6 && value <= -1 },
 ]
 
+const formatCriteria = (criteria) => {
```
hanars, line 22: this was copy-pasted from the now-deleted AcmgScoreCriteria.jsx

ui/shared/components/panel/acmg/AcmgCriteria.jsx
```diff
@@ -164,23 +192,29 @@ class AcmgCriteria extends React.PureComponent {
...
+    const { score, criteria, classify } = this.state
     const { formWarning } = this.state
 
-    if (criteria.length > 0) {
-      setAcmgClassification(this.getNewScore(criteria))
-    }
-
     return (
       <div>
+        <Table celled padded>
```
hanars, line 201: this was copy-pasted from the now-deleted AcmgScoreCriteria.jsx

ui/shared/components/panel/acmg/AcmgCriteria.jsx
```diff
@@ -29,16 +44,31 @@ export const getNewScoreValue = (criteria) => {
...
+    const { classify = 'Unknown', criteria = [] } = variant.acmgClassification || {}
+    // eslint-disable-next-line react/state-in-constructor
+    this.state = { formWarning: '', criteria, classify, score: getNewScoreValue(criteria) }
+  }
+
+  setCriteria(criteria) {
+    const newState = { criteria }
+    if (criteria.length > 0) {
+      newState.score = getNewScoreValue(criteria)
+      const newCategory = CATEGORY_CONFIGS.find(
```
hanars, line 65: copied from the now-deleted `getNewScore`, below

ui/shared/components/panel/acmg/AcmgCriteria.jsx
```diff
@@ -29,16 +44,31 @@ export const getNewScoreValue = (criteria) => {
...
   }
 
-  state = { formWarning: '' }
+  constructor(props) {
+    super(props)
+
+    const { variant } = this.props
+    const { classify = 'Unknown', criteria = [] } = variant.acmgClassification || {}
+    // eslint-disable-next-line react/state-in-constructor
+    this.state = { formWarning: '', criteria, classify, score: getNewScoreValue(criteria) }
```
ShifaSZ, line 58: I'm curious about how to meet the eslint requirement of no 'react/state-in-constructor' in this case. Can we define the states outside of the constructor? And assign initial values to them in the `componentDidMount` function?

ui/shared/components/panel/acmg/AcmgCriteria.jsx
```diff
@@ -118,7 +144,8 @@ class AcmgCriteria extends React.PureComponent {
       const elements = rows.map((row, rowIdx) => {
         const criteriaUsed = this.getCriteriaUsed()
         return (
-          <Table.Row>
+          // eslint-disable-next-line react/no-array-index-key
+          <Table.Row key={rowIdx}>
```
ShifaSZ, line 148: It seems that the options defined in `ACMG_DROP_DOWN_OPTIONS` have a `key` field. Can we use it for the key?

ui/shared/components/panel/acmg/AcmgRuleSpecification.jsx
```diff
@@ -89,16 +94,17 @@ const tableListItemsForCriteria = (items, criteria) => {
 
 const tableRowWithTableCells = (items, description, cellsWithList = false, color = 'blue') => {
   const row = (
-    <Table.Row>
+    <Table.Row key={description}>
       <Table.Cell textAlign="center">{description}</Table.Cell>
-      {items.map((item) => {
+      {items.map((item, key) => {
```
ShifaSZ, line 99: It would be better to add a key field in each item.

ui/shared/components/panel/acmg/AcmgCriteria.jsx
hanars: no thats terrible practice in React and super prone to bugs. The linter rule encourages us not to use props to initialize state because thats better practice, but for use cases like this where its substantially better code if we do, its fine to disable the linter

ui/shared/components/panel/acmg/AcmgCriteria.jsx
hanars: ideally yes, but they don't. And adding it that would involve me reading through about 100 different constants that I didn't write and am not familiar with, reconfiguring lists of lists to be lists of objects with list keys and then updating the rendering code accordingly, and then figuring out how to add meaningful keys. If you or I had implemented this feature I absolutely would have wanted to do that. But given that this is an external contribution the point was just to get this working without spamming the console with errors, and maybe some day in the future (probably when it needs to be updated) I'll clean it up

ui/shared/components/panel/acmg/AcmgRuleSpecification.jsx
hanars: agreed, but see my comment above

ShifaSZ (approved):

PR #2308, opened 2021-11-18
Bump elasticsearch version version

hanars:
Our current version of elasticsearch is reaching EOL in December, so this bumps us to the latest version. This also removes our maintained copies of docker images, and instead uses the elasticsearch default distributed image. Discussed offline with Steve and we agree that this is a better approach for us going forward

This has been tested on dev and confirmed that it works. Updating the version triggers a rolling update of elasticsearch with no data loss. I would anticipate some search unavailability during the rolling update, but it shouldn't be for very long. 

** Release Checklist**

- [x] Infrastructure changes:
  
  OR 
  
   - no new docker image required
  - [x] Any changes to kubernetes configurations will be deployed through a full seqr kubernetes deployment after merging. All these changes have been tested on dev


sjahl (approved):
This looks good to me!

PR #2309, opened 2021-11-18
fix dismiss warning message

hanars:


ui/shared/components/page/WarningMessages.jsx
```diff
@@ -15,7 +15,7 @@ class WarningMessages extends React.PureComponent {
   state = { hiddenMessages: {} }
 
   hide = message => () => {
-    this.setState(prevState => ({ ...prevState.hiddenMessages, [message]: true }))
+    this.setState(prevState => ({ hiddenMessages: { ...prevState.hiddenMessages, [message]: true } }))
```
ShifaSZ, line 18: Is there any reason to use an updater function rather than a simple state change object like below?
```
const { hiddenMessages } = this.state
this.setState({ hiddenMessages: { hiddenMessages, [message]: true } })
```

hanars: yes, because it violates our linting rules to reference the current state during a state update because it can lead to messy behavior and unnecessary updates

ShifaSZ (approved):

PR #2311, opened 2021-11-19
Add gCNV sample colored line legend.

ShifaSZ:


ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -328,6 +328,35 @@ class FamilyReads extends React.PureComponent {
...
+    const { individualsByGuid } = this.props
+    const sampleColors = this.getSampleColors()
+    return Object.entries(sampleColors).map(
+      ([iGuid, sampleColor]) => (
+        <div key={iGuid}>
+          <ColoredSpan color="black">
+            {individualsByGuid[iGuid].displayName}
+            &nbsp;
+          </ColoredSpan>
+          <ColoredSpan backgroundColor={sampleColor}>&nbsp;</ColoredSpan>
```
hanars, line 354: The name text should be the same color as everything else in this side bar. If thats black, fine, if thats grey, fine, but you should not be adding special color handling for it. 

For the icon itself, use an actual icon, not a span with a background color. Its always best practice for the HTML to be as semantically close to what it is doing as possible- a span is usually an inline container for content while an icon is actually an icon. Given that the only colors we support are red and blue which are standard colors handled by semantic ui this doesn't even need to be a styled component, you can just do `<Icon name="square full" color={sampleColor} />`

hanars (changes requested):

ui/shared/components/panel/family/FamilyReads.jsx
ShifaSZ: The name text's default color is grey. But the color is black for the other text (e.g. the track names) in the sidebar. Should I keep the sample name text in grey?

ShifaSZ: Without the special styling, the sidebar looks as below figure.
image

hanars: probably you should  put the text it inside a `label`, or just in general use the same way of styling this as everything else on the sidebar which probably involves wrapping it in something. Since nothing else here is using ColoredSpan this shouldn't have to either

Also, I know its not in the original screenshot, but can we put the colors to the left of the labels instead of the right so the alignment is nicer? Also, we should add a divider for this section as well, with the title "gCNV Samples"

ShifaSZ: I can add the divider.

For the colored icon, maybe you want it to be consistent with the TGG-Viewer. It is on the right for TGG-viewer.

hanars: Given that I explicitly said "I know its not in the original screenshot" I clearly DON'T want it to look like the TGG viewer, which is a totally different and unrelated application to seqr that we have literally no need or desire to conform to. I want it to actually look good, which is why I suggested what I did

ShifaSZ:
The page looks as follows now.
image

hanars:
As I stated in my comment earlier "we should add a divider for this section as well, with the title "gCNV Samples"" so you need to update the divider title to "gCNV Samples"

ShifaSZ:
> As I stated in my comment earlier "we should add a divider for this section as well, with the title "gCNV Samples"" so you need to update the divider title to "gCNV Samples"

The `Divider` converts its content to capital letters. So "gCNV Samples" becomes "GCNV SAMPLES". "GCNV" looks weird. That's why I try not to include "gCNV".

hanars:
> The `Divider` converts its content to capital letters. So "gCNV Samples" becomes "GCNV SAMPLES". "GCNV" looks weird. That's why I try not to include "gCNV".

GCNV all caps is fine

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -328,6 +328,35 @@ class FamilyReads extends React.PureComponent {
...
+  getSampleColorPanel = () => {
+    const { individualsByGuid } = this.props
+    const sampleColors = this.getSampleColors()
+    return Object.entries(sampleColors).map(
+      ([iGuid, sampleColor]) => (
+        <div key={iGuid}>
+          <Icon name="square full" color={sampleColor} />
+          <label>
+            {individualsByGuid[iGuid].displayName}
+            &nbsp;
```
hanars, line 353: I don't think this space is needed

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -328,6 +328,35 @@ class FamilyReads extends React.PureComponent {
...
 
+  getSampleColors = () => {
+    const { openFamily } = this.state
+    const { igvSamplesByFamilySampleIndividual, individualsByGuid } = this.props
+    const igvSampleIndividuals = (
+      openFamily && (igvSamplesByFamilySampleIndividual || {})[openFamily]) || {}
+    return Object.keys(igvSampleIndividuals[GCNV_TYPE] || {}).reduce(
+      (acc, iGuid) => ({
+        ...acc,
+        [iGuid]: individualsByGuid[iGuid].affected === AFFECTED ? 'red' : 'blue',
```
hanars, line 339: this needs to share code with the function we use when we compute the track types, so if we ever want to change the colors we only need to change one place. Maybe a helper function like `getSampleColor(individual)`

ui/shared/components/panel/family/FamilyReads.jsx
```diff
@@ -328,6 +328,35 @@ class FamilyReads extends React.PureComponent {
...
+      }), {},
+    )
+  }
+
+  getSampleColorPanel = () => {
+    const { individualsByGuid } = this.props
+    const sampleColors = this.getSampleColors()
```
hanars, line 346: This code works, but the control flow is a little confusing. I think it might be easier to understand if you simply get a list of the gcnv individuals and then loop through those and display them. Something like
```
const gcnvIndividuals = Object.keys(igvSampleIndividuals[GCNV_TYPE] || {}).map(iGuid => individualsByGuid[iGuid])
return gcnvIndividuals.map(individual => (
 <div>
 <Icon name="square full" color={getSampleColor(individual)} />
<label>{individual.displayName}</label>
...
))
```

ui/shared/components/StyledComponents.jsx
```diff
@@ -38,6 +38,11 @@ export const ColoredDiv = styled.div`
   color: ${props => props.color} !important;
 `
 
+export const ColoredSpan = styled.span`
```
hanars, line 41: this is unused

hanars (changes requested):

hanars (approved):

PR #2337, opened 2021-12-06
Anvil loading away message

hanars:


ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -118,6 +134,7 @@ const LoadWorkspaceDataForm = React.memo(({ namespace, name }) => (
         see &nbsp;
         <b><a href={VCF_DOCUMENTATION_URL} target="_blank" rel="noreferrer">this documentation</a></b>
       </Message>
+      {WARNING_BANNER ? <Message error compact header={WARNING_HEADER} content={WARNING_BANNER} /> : null}
```
ShifaSZ, line 137: You may want to add a date condition (before Jan 3, 2022) to show the warning because the seqr might not be updated right after the holidays.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -31,6 +32,12 @@
 
 BLOCK_SIZE = 65536
 
+ANVIL_LOADING_EMAIL_DATE = '2021-12-13'
+ANVIL_LOADING_DELAY_EMAIL = """
+We have received your request to load data to seqr from AnVIL. Currently, the Broad Institute is holding an internal 
+retreat or closed for the winter break so we are unable to load data until the 2nd week of January 2022. We appreciate 
+your understanding and support of our research team taking some well-deserved time off and hope you also have a nice break.
+"""
```
ShifaSZ, line 40: It would be ideal to set with environment variables. Then, we can add and remove the warning without updating the software.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -31,6 +32,12 @@
 
 BLOCK_SIZE = 65536
 
+ANVIL_LOADING_EMAIL_DATE = '2021-12-13'
```
ShifaSZ, line 35: You may also want to add an ending date.

ShifaSZ (changes requested):

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
hanars: no I want this to show until whenever we explicitly decide to turn it off in January, we don't yet know what the end date should be

seqr/views/apis/anvil_workspace_api.py
hanars: it is substantially easier to do a code redeploy than an env variable redeploy in seqr. This was designed on purpose to be as easy as possible for. me

seqr/views/apis/anvil_workspace_api.py
hanars: as explained above, no because we don't know the end date yet

ShifaSZ (approved):

PR #2344, opened 2021-12-08
Docker rework part 2

sjahl:
Before this is mergeable, I still need to do some cleanup work (static assets removal), documentation, and plugging in the build triggers via terraform, but I think these changes get us in a good spot for arbitrary branch builds of the seqr docker image, with the goal being to automatically build seqr after merging PRs into dev or master.

This moves us to a multi-stage docker build, which has the following advantages:
- doesn't require us to keep build-time deps in the final image
- reduced image size
- makes it easier to skip unnecessary rebuilds of the python and npm environments. We now only rebuild the python installation if requirements.txt changes, and only rebuild the static assets if the `ui/` folder's contents change.

I also updated the cloudbuild configuration to build images with Google's [kaniko cache](https://cloud.google.com/build/docs/kaniko-cache), which should dramatically reduce build times of the image if you trigger a build within the cache TTL.

This image has some potentially end-user breaking changes:

1. The docker image build now includes npm build steps. This allows us to build an image that contains changes to the ui source code, and the statically built assets aren't tightly coupled with the prod release/master branch anymore. We'll no longer track the statically built assets in git.
2. This image no longer performs any git clone/pulls of a particular github repo branch. The build now copies in the current seqr checkout as is, rather than cloning anything from github. For testing builds, this can contain uncommitted/unpushed changes, arbitrary branches, etc if desired. For our automated release builds, the seqr version in the docker image will be the HEAD of the master or dev branch directly following a PR merge.
    - Downstream users will upgrade seqr by pulling a new docker image and restarting their seqr containers. We should probably discuss the effects of this on local installations.
    - At some point, someone added a `SEQR_GIT_BRANCH` docker `ARG` that allowed for building a docker image based off of a specified git branch: This will no longer be supported, as it's not necessary anymore. To build an image off of an arbitrary branch or commit, simply check out the revision you want to build, and then run the docker build.

Closes https://github.com/broadinstitute/seqr-private/issues/1067
Touches https://github.com/broadinstitute/seqr-private/issues/1062, https://github.com/broadinstitute/seqr-private/issues/1065


deploy/kubernetes/seqr/seqr.gcloud.yaml
```diff
@@ -167,8 +160,6 @@ spec:
           value: "{{ DATABASE_BACKUP_BUCKET }}"
         - name: RUN_CRON_JOBS
           value: "{{ RUN_CRON_JOBS }}"
-        - name: STATIC_MEDIA_DIR
-          value: "{{ STATIC_MEDIA_DIR }}"
```
hanars, line 171: we should remove this variable from `shared-settings.yaml` since its no longer used

deploy/servctl_utils/deploy_command_utils.py
```diff
@@ -200,22 +200,6 @@ def deploy_redis(settings):
 def deploy_seqr(settings):
     print_separator("seqr")
 
-    if settings["BUILD_DOCKER_IMAGES"]:
```
hanars, line 203: `settings["BUILD_DOCKER_IMAGES"]` is False by default so maybe leave the `if` check here and throw an explicit error if someone tries to run this command to build the image?

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -204,7 +204,7 @@ def _get_pedigree_image_url(pedigree_image):
                 pedigree_image = pedigree_image.url
             except Exception:
                 pedigree_image = None
-        return os.path.join("/media/", pedigree_image) if pedigree_image else None
+        return pedigree_image
```
hanars, line 207: 🤦

hanars: are we sure this works if you have local image files?

seqr/urls.py
```diff
@@ -335,11 +335,16 @@
 urlpatterns += [
     url(r'^admin/login/$', RedirectView.as_view(url=LOGIN_URL, permanent=True, query_string=True)),
     url(r'^admin/', admin.site.urls),
-    url(r'^media/(?P<path>.*)$', django.views.static.serve, {
-        'document_root': MEDIA_ROOT,
-    }),
 ]
 
+# If serving /media from the local filesystem
```
hanars, line 340: this change really confused me the first time I read it. Maybe add to the comment something like "not needed if using GCS as the media backend"

hanars (commented):
we will probably want to update our changelog for this

.dockerignore
```diff
@@ -0,0 +1,10 @@
+ui/dist/*
+ui/node_modules/*
+static/*
+!static/fonts/*
+!static/images/*
```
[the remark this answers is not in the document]
sjahl: Yup, as far as I know, docker does't look at the gitignore file when doing `COPY` operations, so it needs to be set here too.

.github/workflows/unit-tests.yml
```diff
@@ -41,35 +41,30 @@ jobs:
...
         java-package: jdk # (jre, jdk, or jdk+fx) - defaults to jdk
         architecture: x64 # (x64 or x86) - defaults to x64
+    - name: Install JS Dependencies
+      run: npm install || true
+      working-directory: ./ui
+    - name: Javascript/NPM Tests
+      run: npm test
+      working-directory: ./ui
+    - name: build production app for python tests
+      run: CI= npm run build
```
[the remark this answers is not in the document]
sjahl: In our last build, that took 1m29s, and the entire end to end pipeline took 6m52s -- I ended up combining the pipelines here since it felt it might save time overall to only run the `npm install` once. It might be worth looking at the python tests that require this build, and seeing if we can refactor those like you mentioned in standup. I think it would be a plus overall to be able to run the python tests without needing the JS stack.

Another option is I can maybe add a prerequisite step that runs the npm stuff first, and then re-uses the resulting node_modules and build output in parallel for the node and python tests afterward. Though, I don't think will actually save that much time.

deploy/docker/seqr/Dockerfile
```diff
@@ -17,13 +17,46 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+
+WORKDIR /build
+
+COPY ./ui ./ui
+
+WORKDIR /build/ui
+RUN npm ci && \
+    npm run build
+
+FROM python:3.7-slim-bullseye as runtime
```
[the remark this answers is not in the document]
sjahl: Yeah, we'll go over this in our meeting tomorrow.

for reference, this is a multi-stage docker build instruction, which is a technique used for optimizing image size and keeping build-time-only dependencies out of the final image: https://docs.docker.com/develop/develop-images/multistage-build/

deploy/docker/seqr/Dockerfile
```diff
@@ -32,29 +65,26 @@ RUN echo "deb http://packages.cloud.google.com/apt cloud-sdk-bullseye main" > /e
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-# DISABLE_CACHE work-around to force git pull on every docker build, based on https://github.com/docker/docker/issues/1996
-ARG DISABLE_CACHE=1
-ARG SEQR_REPO=https://github.com/broadinstitute/seqr
-ARG SEQR_GIT_BRANCH=master
-
-# update seqr repo
-RUN git clone -q --branch $SEQR_GIT_BRANCH $SEQR_REPO
```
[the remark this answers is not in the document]
sjahl: Yup, happy to go over this a little more. The tooling to build the images automatically will be via the cloudbuild step in the .cloudbuild/seqr-docker.cloudbuild.yaml file on this PR, and the build triggers in https://github.com/broadinstitute/seqr-terraform/pull/9/files#diff-1eb08bd6eee28c854f459b9a3661be73eafd153e5f62884ef0493f0b5cc6562f

The idea will eventually be that we'll want a policy that dictates we don't build images locally. The only published images should be ones that go through the trusted build chain from PR merge -> cloudbuild -> GCR

I can definitely see a use case for wanting to run a testing/ad-hoc image in a k8s dev environment somewhere. I think the way forward there is to have a seqr-testing GCR image that we allow local pushes to; using an alternative image for dev/one-off deployments should be an easy thing to accommodate when we work on the k8s manifests later.

In any case, we'll chat about this more 👍

deploy/docker/seqr/Dockerfile
```diff
@@ -32,29 +65,26 @@ RUN echo "deb http://packages.cloud.google.com/apt cloud-sdk-bullseye main" > /e
...
 ENV TERM=xterm
 
-COPY readiness_probe /
-COPY bin/*.sh /usr/local/bin/
-COPY gitconfig /root/.gitconfig
-COPY config/*.py ./
-COPY bashrc /root/.bashrc
+COPY deploy/docker/seqr/readiness_probe /
+COPY deploy/docker/seqr/bin/*.sh /usr/local/bin/
+COPY deploy/docker/seqr/gitconfig /root/.gitconfig
```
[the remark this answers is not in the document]
sjahl: Nope, good catch!

deploy/kubernetes/seqr/seqr.gcloud.yaml
sjahl: will do 👍

deploy/servctl_utils/deploy_command_utils.py
sjahl: ok 👍

seqr/urls.py
sjahl: Ok, I see, I can update the comment. The logic here is that we only want to serve requests to /media/* if we're configured to serve static content from a local media_root directory.

seqr/views/utils/orm_to_json_utils.py
sjahl: I'll double check this, but I tested the code on dev with the GCS stuff turned on and off, and we were getting correct URL's in both cases. My understanding is that the `.url` attribute will prepend the media path onto the image name when returning the value. My belief in this case was that `pedigree_image.url` was returning something like `/media/foo.png`, and then we were doing a no-op operation with `os.path.join`, like this:

```
>>> test_path = "/media/foo.png"
>>> import os
>>> os.path.join("/media/", test_path)
'/media/foo.png'
```

deploy/kubernetes/gcloud-prod-settings.yaml
```diff
@@ -23,6 +23,8 @@ TERRA_API_ROOT_URL: 'https://api.firecloud.org'
 
 CLOUDSQL_AVAILABILITY_TYPE: 'regional'
 
+GCS_MEDIA_ROOT_BUCKET: 'seqr-media-storage-public-prod'
```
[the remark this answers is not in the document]
sjahl: Yes, the bucket will need to be created via https://github.com/broadinstitute/seqr-terraform

Note, that the GCS portion of this PR has moved to #2397 , so we'll cover this over there.

deploy/kubernetes/seqr/seqr.gcloud.yaml
sjahl: will handle this over in #2397

seqr/urls.py
sjahl: resolved in #2397

seqr/views/utils/orm_to_json_utils.py
sjahl: resolving this, as it was handled in the GCR media repo PR

.gitignore
```diff
@@ -5,6 +5,11 @@
 
 ui/node_modules
 ui/jest*
+ui/dist
+
+static/*
```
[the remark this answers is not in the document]
sjahl: I've removed these gitignores for now -- we'll add them back when we actually remove the static assets.

.github/workflows/unit-tests.yml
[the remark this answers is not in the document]
hanars: I think 6+ minutes to get test results back every time we put up a PR is a bit slower than I really want. I know we don't want to block this PR indefinitely, but I think what I would prefer is if you could take out the tests that rely on the assets for now, and take out this extra step here, and then I can work in parallel/ shortly after this goes live on getting those tests re-enabled (or determine if we really need them at all)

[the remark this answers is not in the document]
sjahl: Actually, since we are leaving the assets in on this PR, then we can probably just take this github actions change out for now. Does it make more sense to do that, and then update the test pipeline in the PR where we deprecate assets and/or refactor the tests?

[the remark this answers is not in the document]
hanars: yes that makes sense to me

sjahl:
@hanars OK, just rebased this, I think it's ready for another review pass. I think our order of operations here is:

1. merge this PR
2. create cloudbuild triggers to run the docker builds: https://github.com/broadinstitute/seqr-terraform/pull/9
3. manually trigger cloudbuild to compile a gcloud-dev image and test it
4. do a prod release, and generate a gcloud-prod image
5. create a github discussion post, and email known users

Anything I'm missing?

hanars:
> 4. do a prod release, and generate a gcloud-prod image

To clarify, the cloudbuild trigger would generate this image for us, it would not need a manual trigger like the dev one, right?

deploy/docker/seqr/Dockerfile
```diff
@@ -7,21 +7,53 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update && apt-get install -y --no-install-recommends \
     apt-utils \
     bzip2 \
+    curl \
+    gcc \
+    libc6-dev \
+    libpq-dev \
+    nodejs \
+    npm \
+    redis-tools \
```
hanars, line 16: not particularly important, but I doubt we need `redis-tools` for this stage of the build

hanars (approved):
looks good to me with the deployment plan you outlined!

sjahl:
> > 4. do a prod release, and generate a gcloud-prod image
> 
> To clarify, the cloudbuild trigger would generate this image for us, it would not need a manual trigger like the dev one, right?

yup -- Theoretically, if we wanted to try an automated one in dev, I could merge and apply that terraform prior to this PR.

hanars:
> > > 4. do a prod release, and generate a gcloud-prod image
> > 
> > 
> > To clarify, the cloudbuild trigger would generate this image for us, it would not need a manual trigger like the dev one, right?
> 
> yup -- Theoretically, if we wanted to try an automated one in dev, I could merge and apply that terraform prior to this PR.

Your call which you prefer. If the manual build works on dev it does mean that if the automated build fails on prod the manual one should still work so I'm not too worried about not having tested the automation

PR #2346, opened 2021-12-09
Rna-seq outlier search result gene flag

hanars:
Builds off of https://github.com/broadinstitute/seqr/pull/2345 to add a flag on genes in variants that have outlier RNA seq

### DIFF
--- a/ui/pages/Project/reducers.js
+++ b/ui/pages/Project/reducers.js
@@ -275,7 +275,6 @@ export const reducers = {
   matchmakerMatchesLoading: loadingReducer(REQUEST_MME_MATCHES, RECEIVE_DATA),
   mmeContactNotes: createObjectsByIdReducer(RECEIVE_DATA, 'mmeContactNotes'),
   rnaSeqDataLoading: loadingReducer(REQUEST_RNA_SEQ_DATA, RECEIVE_DATA),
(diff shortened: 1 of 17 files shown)


ui/pages/Project/reducers.js
```diff
@@ -221,7 +221,7 @@ export const loadMmeMatches = (submissionGuid, search) => (dispatch, getState) =
 
 export const loadRnaSeqData = individualGuid => (dispatch, getState) => {
   const data = getState().rnaSeqDataByIndividual[individualGuid]
-  if (!data) {
+  if (!data || Object.values(data).every(({ isSignificant }) => isSignificant)) {
```
ShifaSZ, line 224: It is a little confusing why it need to load data when every individual is significant. Add a comment to the code?

ui/shared/components/panel/variants/VariantGene.test.js
```diff
@@ -4,12 +4,12 @@ import Adapter from 'enzyme-adapter-react-16'
 import configureStore from 'redux-mock-store'
 import VariantGene from './VariantGene'
```
ShifaSZ, line 5: A little confusing for the names of VariantGene and VariantGenes. This one must be VariantGenes class, right?

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #2347, opened 2021-12-10
Bulk edit mee contact

hanars:


matchmaker/views/matchmaker_api.py
```diff
@@ -350,6 +350,26 @@ def send_mme_contact_email(request, matchmaker_result_guid):
     })
 
 
+@login_and_policies_required
+def update_mme_project_contact(request, project_guid):
```
ShifaSZ, line 354: It is adding a new contact. Is it also required to change the existing contacts?

ShifaSZ (changes requested):

matchmaker/views/matchmaker_api.py
hanars: no, the feature request was to add contacts to all submissions in a project. you can already more granularly edit the contacts on a per-submission basis, this is just the bulk edit functionality

ShifaSZ (approved):

PR #2351, opened 2021-12-14
switch to curator for snapshot management

sjahl:
Couple changes here:

- Removes our self-vendored curl image, and just uses the upstream one for our one-off configure-snapshot-repo task
- Adds an image for running elastic.co's Curator (for some reason Elastic doesn't supply a docker image for this tool….)
- Switches our snapshot routine from the curl-based one, to curator. I think this approach is much cleaner, and the cronjob now handles both the snapshot task and the cleanup task directly after making the snapshot.

The retention policy that I put on here was 45 days -- the effect here should be that when we run the cronjob, it should keep the current snapshot, previous month's snapshot, and then delete anything older.

I've got this deployed in dev, and just ran a test run to verify that things are working. PR is against master, since I think this can really be deployed without disturbing the rest of the fish tank.

Closes https://github.com/broadinstitute/seqr-private/issues/1083


deploy/kubernetes/elasticsearch/elasticsearch.gcloud.yaml
```diff
@@ -237,4 +237,4 @@ spec:
       targetPort: 9200
   selector:
     elasticsearch.k8s.elastic.co/cluster-name: elasticsearch
-    elasticsearch.k8s.elastic.co/node-data: "true"
\ No newline at end of file
+    elasticsearch.k8s.elastic.co/node-data: "true"
```
sjahl, line 240: I think my text editor added the newline here… Happy to remove this if we don't want a final newline

hanars (approved):

PR #2352, opened 2021-12-16
feat: Gene location filter using PA confidence level

tommyli:
Hi @hanars,

Sorry it took a while before I got round to look at this again.  Below are screenshots on what the location filter by confidence level looks like.  The list of genes used for filtering is using the union of all three text areas in the case of PanelApp gene lists.

Apart from your usual review, I also need some help in fixing a bug in this implementation.  I spent nearly a whole day trying to figure out why "Exclude locations" checkbox is not working, i.e. the checkbox is not getting checked when I click on it.  Perhaps you could shed some light?  It's probably something silly that I can't see.  In general, `LocationFilter.jsx` is done following the implementation of `CustomInheritanceFilter.jsx` as you previously suggested.

Screen Shot 2021-12-16 at 19 24 10

Screen Shot 2021-12-16 at 19 24 25

On another note, I requested PR with dev instead of master, is that your preference?  Or should we raise PRs against master in the future?


hanars:
So I think overriding the entire `fields` list for the location panel with a single custom field is not the way you want to go, and is probably why the exclude checkbox is not behaving the way you want. I think you should instead replace the first element of the list with a custom `component`, which maintains the `rawItems` name. The you add some special `normalize` and `format` functions such that the form value is always normalized as a string (which is what the server side will want), but add conditional logic so that if a panel app is list is selected it is formatted in the form as a dictionary with the keys `red`, `amber`, and `green`. Then add conditional logic to the component such that if the value passed to the component is a dictionary it renders the three panels correctly, and if the value is a string it renders as it currently is. 

Alternatively, you could add a second field to the `LOCATION_FIELDS` list which also has the name `rawItems` which is just the panel app functionality described above and leave the existing `rawItems` field as is, and then add a custom `fieldLayout` to determine which of the 2 fields should actually be rendered.

I do prefer PRs against dev, so this works for me. As an FYI, the Broad has a holiday closing starting next week so I will not be merging new PRs/ deploying new code until January

tommyli:
Sorry Hana, turns out it was a styling bug, i.e. adding the colored icon broke all checkboxes.  It's fixed now and this PR is ready for your review.

hanars:
So even with fixing the bug, I stand by my comment. It really is more in keeping with how every other filter in seqr works to keep multiple subfields for the locus field for the different components (genes, variants, and exclude) rather than bundling it into one massive complicated field. I'll be happy to review if this gets refactored or I may take a crack it myself after the break, but I can't accept this as-is

tommyli:
That's fair enough and thanks for the pointers.  I think I understand what you mean by only `LOCATION_FIELDS` first element and replacing that with a component that knows how to render normal or PanelApp locus lists.

I'll be away next week and back after new year so won't be able to look at this either until after new year.  Of course, feel free to take over the implementation because I understand sometimes the effort taken reviewing is not much less than re-implementing it.

tommyli:
Superseded by #2376

PR #2359, opened 2021-12-17
create manage command for loading rna seq

hanars:
The file upload is too big to upload via the UI in production, so creating a manage command.

### DIFF
--- a/seqr/management/tests/load_rna_seq_tests.py
+++ b/seqr/management/tests/load_rna_seq_tests.py
@@ -1,0 +1,59 @@
+# -*- coding: utf-8 -*-
+import mock
+
+from django.core.management import call_command
+from django.test import TestCase
+
+from seqr.models import Sample, RnaSeqOutlier
+from seqr.views.utils.orm_to_json_utils import get_json_for_rna_seq_outliers
+
+RNA_FILE_ID = 'tmp_-_2021-03-01T00:00:00_-_test_data_manager_-_new_muscle_samples.tsv.gz'
+EXISTING_SAMPLE_GUID = 'S000150_na19675_d2'
+
+class LoadRnaSeqTest(TestCase):
+    fixtures = ['users', '1kg_project', 'reference_data']
+
+    @mock.patch('seqr.views.apis.data_manager_api.ANALYST_PROJECT_CATEGORY', 'analyst-projects')
+    @mock.patch('seqr.management.commands.load_rna_seq.logger.info')
+    @mock.patch('seqr.management.commands.load_rna_seq.open')
+    @mock.patch('seqr.views.apis.data_manager_api.gzip.open')
+    def test_command(self, mock_gzip_open, mock_open, mock_logger):
+        mock_gzip_file = mock_gzip_open.return_value.__enter__.return_value
+        mock_gzip_file.__next__.return_value = 'sampleID\tgeneID\tdetail\tpValue\tpadjust\tzScore\n'
+        mock_gzip_file.__iter__.return_value = [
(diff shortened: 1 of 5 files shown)


seqr/management/tests/load_rna_seq_tests.py
```diff
@@ -0,0 +1,59 @@
...
+EXISTING_SAMPLE_GUID = 'S000150_na19675_d2'
+
+class LoadRnaSeqTest(TestCase):
+    fixtures = ['users', '1kg_project', 'reference_data']
+
+    @mock.patch('seqr.views.apis.data_manager_api.ANALYST_PROJECT_CATEGORY', 'analyst-projects')
+    @mock.patch('seqr.management.commands.load_rna_seq.logger.info')
+    @mock.patch('seqr.management.commands.load_rna_seq.open')
+    @mock.patch('seqr.views.apis.data_manager_api.gzip.open')
+    def test_command(self, mock_gzip_open, mock_open, mock_logger):
```
ShifaSZ, line 20: It could be cleaner to mock 'seqr.views.apis.data_manager_api.load_rna_seq'.

seqr/management/tests/load_rna_seq_tests.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+        mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19675_D2\tNA19675_1']
+
+        with self.assertRaises(ValueError) as e:
+            call_command('load_rna_seq', RNA_FILE_ID)
+        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19675_D2, NA19675_D3')
+
+        with self.assertRaises(ValueError) as e:
+            call_command('load_rna_seq', RNA_FILE_ID, '--mapping-file', 'map.tsv')
+        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19675_D3')
```
ShifaSZ, line 37: The exception is not handled by the command codes but those of the function `load_rna_seq`. Therefore, maybe the exception shouldn't be test here.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -573,12 +573,10 @@ def test_update_rna_seq(self, mock_logger, mock_open, mock_load_uploaded_file, m
             'Parsed 1 RNA-seq samples',
             'Attempted data loading for 0 RNA-seq samples in the following 0 projects: ',
         ]
```
ShifaSZ, line 575: It could be better to separate the tests for the function `load_rna_seq` and `update_rna_seq`.

ShifaSZ (changes requested):

seqr/management/tests/load_rna_seq_tests.py
hanars: I would rather actually test that the function is being called properly. If you mock the entire function and the function signature changes your tests pass because you mocked the functions but the real world behavior fails because its not getting the expected arguments

seqr/management/tests/load_rna_seq_tests.py
hanars: I wanted some level of sanity testing that the called function is behaving as expected. I think given that these tests are already written I'm not going to remove them so we can have lower test coverage

seqr/views/apis/data_manager_api_tests.py
hanars: This file has one command. This test tests what happens when you call that command. The command does 2 things in a row so it tests that both things are done. Nowhere in our repo do we try to split the test for a single successful call into different tests

ShifaSZ (approved):

PR #2376, opened 2022-01-10
feat: Gene location filter using PA confidence level, v2

tommyli:
Replaces #2352.

@hanars, not sure if this attempt is closer to what you meant in #2352.  I certainly would like your feedback on the need for the API change because the form maintains a new `locusListItems.rawItems` field (when PanelApp list selected) instead of just `rawItems`.

### DIFF
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -29,7 +29,7 @@ class BaseLocusListDropdown extends React.Component {
       let { rawItems } = locusList
 
       if (locusList.paLocusList) {
-        const CONFIDENCE_COLORS = { 1: 'red', 2: 'amber', 3: 'green', 4: 'green' }
+        const CONFIDENCE_COLORS = { 0: 'none', 1: 'red', 2: 'amber', 3: 'green', 4: 'green' }
         rawItems = locusList.items?.reduce((acc, item) => {
           const color = CONFIDENCE_COLORS[item.pagene?.confidenceLevel || 0]
           if (color in acc) {
@@ -37,7 +37,7 @@ class BaseLocusListDropdown extends React.Component {
           }
 
           return acc
-        }, { green: '', amber: '', red: '' })
+        }, { green: '', amber: '', red: '', none: '' })
       }
 
       onChange({ locusListGuid, rawItems })
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -25,8 +25,8 @@ class BaseLocusListDropdown extends React.Component {
   componentDidUpdate(prevProps) {
     const { locusList, onChange } = this.props
     if (prevProps.locusList.rawItems !== locusList.rawItems) {
-      const { locusListGuid, rawItems } = locusList
-      onChange({ locusListGuid, rawItems })
+      const { locusListGuid, isPanelAppList, rawItems, rawItemsGreen, rawItemsAmber, rawItemsRed } = locusList
+      onChange({ locusListGuid, isPanelAppList, rawItems, rawItemsGreen, rawItemsAmber, rawItemsRed })
     }
   }
 
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -3,7 +3,6 @@ import React from 'react'
 import { connect } from 'react-redux'
 import { Dropdown } from 'shared/components/form/Inputs'
 import { LocusListItemsLoader } from 'shared/components/LocusListLoader'
-import { toUniqueCsvString } from 'shared/utils/stringUtils'
 import { getSearchedProjectsLocusListOptions } from '../../selectors'
 
 class BaseLocusListDropdown extends React.Component {
@@ -24,19 +23,35 @@ class BaseLocusListDropdown extends React.Component {
 
   componentDidUpdate(prevProps) {
     const { locusList, onChange } = this.props
+
     if (prevProps.locusList.rawItems !== locusList.rawItems) {
-      const { locusListGuid, isPanelAppList, rawItemsGreen, rawItemsAmber, rawItemsRed } = locusList
+      const { locusListGuid } = locusList
       let { rawItems } = locusList
-      if (isPanelAppList) {
-        rawItems = {
-          rawItems: toUniqueCsvString([rawItemsGreen, rawItemsAmber, rawItemsRed]),
-          green: rawItemsGreen,
-          amber: rawItemsAmber,
-          red: rawItemsRed,
+
+      if (locusList.paLocusList) {
+        const grouped = locusList.items?.reduce((acc, item) => {
+          const confidence = item.pagene?.confidenceLevel || 0
+          const group = acc[confidence] || []
+          group.push(item.display)
+          acc[confidence] = group
+          return acc
+        }, {})
+        if (grouped) {
+          rawItems = {
+            green: grouped['3']?.concat(grouped['4'])?.join(', ') || '',
+            amber: grouped['2']?.join(', ') || '',
+            red: grouped['1']?.join(', ') || '',
+          }
+        } else {
+          rawItems = {
+            green: '',
+            amber: '',
+            red: '',
+          }
         }
       }
 
-      onChange({ locusListGuid, isPanelAppList, rawItems, rawItemsGreen, rawItemsAmber, rawItemsRed })
+      onChange({ locusListGuid, rawItems })
     }
   }
 
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -29,26 +29,15 @@ class BaseLocusListDropdown extends React.Component {
       let { rawItems } = locusList
 
       if (locusList.paLocusList) {
-        const grouped = locusList.items?.reduce((acc, item) => {
-          const confidence = item.pagene?.confidenceLevel || 0
-          const group = acc[confidence] || []
-          group.push(item.display)
-          acc[confidence] = group
+        const CONFIDENCE_COLORS = { 1: 'red', 2: 'amber', 3: 'green', 4: 'green' }
+        rawItems = locusList.items?.reduce((acc, item) => {
+          const color = CONFIDENCE_COLORS[item.pagene?.confidenceLevel || 0]
+          if (color in acc) {
+            acc[color] = [acc[color], item.display].filter(val => val).join(', ')
+          }
+
           return acc
-        }, {})
-        if (grouped) {
-          rawItems = {
-            green: grouped['3']?.concat(grouped['4'])?.join(', ') || '',
-            amber: grouped['2']?.join(', ') || '',
-            red: grouped['1']?.join(', ') || '',
-          }
-        } else {
-          rawItems = {
-            green: '',
-            amber: '',
-            red: '',
-          }
-        }
+        }, { green: '', amber: '', red: '' })
       }
 
       onChange({ locusListGuid, rawItems })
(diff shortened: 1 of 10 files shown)


hanars:
So at a high level this is much closer. However, I think it would be better if you kept the field limited to a single sub-value - `rawItems`. Adding the `locusListItems` as the field to work off of adds unnecessary complexity and this is a bit hard to read. I think if a panel app gene list is selected, the value of `rawItems` should be set to an object with keys `green`, `amber`, and `red`. You will want to make this change directly in the `LocusListSelector` component, not to the `normalize` method. Then the custom `component` for locations box (the `LocusListItemsFilter` component you created) will be substantially simpler and more consistent with how other form fields in seqr works. It will check the type of the `value` passed to it, and if its an object it renders the 3 panels with the correct sub-values and `onChanges` and if `value` is not an object it just renders the vanilla text box. Something like this:

```
const PanelAppItemsFilter = ({ color, value, onChange, ...props }) =>
  <BaseSemanticInput 
    {...props} 
    width={3}
    label={<span><ColoredIcon color={color} {...ICON_PROPS[color]} /> {color} Genes</span>} 
    value={value[color]} 
    onChange={colorVal => onChange({ ...value, [color]: colorVal })}
  />

const LocusListItemsFilter = props => {
  if (props.value $$ typeof props.value === 'object') {
    return [
      <PanelAppItemsFilter {..props} color="red" />,
      <PanelAppItemsFilter {..props} color="amber" />,
      <PanelAppItemsFilter {..props} color="green" />,
    ]
  }
  return <BaseSemanticInput {..props} />
}
```
Let me know if this makes sense/ if you see nay issues with it

tommyli:
Thanks @hanars for your help.  I've made changes as per your suggestions, indeed a lot cleaner.  The need for the API change is still there though in my implementation because I don't know how to cleanly merge the `rawItems.rawItemsGreen`, `rawItems.rawItemsAmber` and `rawItems.rawItemsRed` back into just `rawItems` in a clean way.  Right now, when panel app selected, the values are in `rawItems.rawItems`.

tommyli:
Thanks @hanars for your help. I've made changes as per your suggestions, indeed a lot cleaner. The need for the API change is still there though in my implementation because I don't know how to cleanly merge the `rawItems.rawItemsGreen`, `rawItems.rawItemsAmber` and `rawItems.rawItemsRed` back into just rawItems in a clean way. Right now, when panel app selected, the values are in `rawItems.rawItems` and is handled in the backend.

hanars:
yeah, so ideally we could use the `normalize` and `format` functions to properly format a field so how it is displayed and how it is submitted have different types, but I couldn't figure out how to do that in this case either. I think though that rather than changing the server side behavior, you should update the `onSubmit` function for the `VariantSearchFormContainer` to transform the value there. And also, instead of keeping a nested `rawItems` key that constantly has to be kept in sync with the other 3 values, just create the merged value on submit and keep the form value simple
```
onSubmit: (({ rawItems, ...searchParams }) => {
    const formattedRawItems = (rawItems && typeof rawItems === 'object') ? Object.values(rawItems).joint(', ') : rawItems
    dispatch(navigateSavedHashedSearch({ rawItems: formattedRawItems, ...searchParams }, ownProps.history.push, ownProps.resultsPath))
},
```

ui/redux/selectors.js
```diff
@@ -262,6 +262,25 @@ export const getParsedLocusList = createSelector(
       })
       locusList.items.sort(compareObjects('display'))
       locusList.rawItems = locusList.items.map(({ display }) => display).join(', ')
+
+      if (locusList.paLocusList) {
```
hanars, line 266: this selector is used in a lof of other places where this computation would not be helpful, so I think this isn't a great place to add it. Maybe add it to the `onChange` of the `LocusListSelector` instead

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const { value } = props
+
+  return value && typeof value === 'object' ?
+    [
+      <PanelAppItemsFilter {...props} key="green" color="green" name="rawItemsGreen" />,
+      <PanelAppItemsFilter {...props} key="amber" color="amber" name="rawItemsAmber" />,
+      <PanelAppItemsFilter {...props} key="red" color="red" name="rawItemsRed" />,
+    ] :
+    (
+      configuredField(LOCUS_LIST_RAW_ITEMS_FIELD)
```
hanars, line 81: you shouldn't be calling `configuredField` here, it's already being called one level up. Just return the actual component(s) we want to return with all the props passed through

tommyli: Can you please elaborate on what you mean?  If I remove configuredField, I get error below.

```text
app.js:15195 Uncaught Error: Objects are not valid as a React child (found: object with keys {name, label, labelHelp, component, rows, width}). If you meant to render a collection of children, use an array instead.
    in LocusListItemsFilter (created by renderField)
    in renderField (created by ConnectedField)
    in ConnectedField (created by Connect(ConnectedField))
    in Connect(ConnectedField) (created by Field)
    in Field
```

tommyli: Thanks, with your help, got it working and I understand more on what you mean by letting the higher level form handle the form configurations.

I've also fixed your other reviewed items.  Let me know of other issues after your review, thanks!

ui/pages/Search/components/filters/LocusListSelector.jsx
```diff
@@ -24,8 +23,34 @@ class BaseLocusListDropdown extends React.Component {
...
+          acc[confidence] = group
+          return acc
+        }, {})
+        if (grouped) {
+          rawItems = {
+            green: grouped['3']?.concat(grouped['4'])?.join(', ') || '',
+            amber: grouped['2']?.join(', ') || '',
+            red: grouped['1']?.join(', ') || '',
+          }
+        } else {
```
hanars, line 45: what would the case be that there is no `grouped` but we would still want to display the 3 panels?

ui/pages/Search/components/filters/LocusListSelector.jsx
```diff
@@ -24,8 +23,34 @@ class BaseLocusListDropdown extends React.Component {
...
   componentDidUpdate(prevProps) {
     const { locusList, onChange } = this.props
+
     if (prevProps.locusList.rawItems !== locusList.rawItems) {
-      const { locusListGuid, rawItems } = locusList
+      const { locusListGuid } = locusList
+      let { rawItems } = locusList
+
+      if (locusList.paLocusList) {
+        const grouped = locusList.items?.reduce((acc, item) => {
```
hanars, line 32: I think it would be neater to just reduce once into the object you want: 
```
const CONFIDENCE_COLORS = { '1': 'red', '2': 'amber', '3': 'green', '4': 'green'}
rawItems = locusList.items?.reduce((acc, item) => {
  const color = CONFIDENCE_COLORS[item.pagene?.confidenceLevel || 0]
  if (color in acc) {
    acc[color] = [acc[color], item.display].filter(val => val).join(', ')
  }
  return acc
}, { red: '', amber: '', green: ''})
```

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+import { Icon, Popup } from 'semantic-ui-react'
+import { BaseSemanticInput } from 'shared/components/form/Inputs'
+import { ColoredIcon } from 'shared/components/StyledComponents'
+import { PANEL_APP_CONFIDENCE_DESCRIPTION, PANEL_APP_CONFIDENCE_LEVEL_COLORS } from 'shared/utils/constants'
+import { camelcaseToTitlecase } from 'shared/utils/stringUtils'
+
+const PA_LABEL_HELP = 'A list of genes, can be separated by commas or whitespace.'
+const PA_ICON_PROPS = {
+  green: {
+    name: 'circle',
```
hanars, line 12: since this is always the same it would be cleaner to make it a prop directly instead of a configuration

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+    title: PANEL_APP_CONFIDENCE_DESCRIPTION[2],
+    color: PANEL_APP_CONFIDENCE_LEVEL_COLORS[2],
+  },
+  red: {
+    name: 'circle',
+    title: PANEL_APP_CONFIDENCE_DESCRIPTION[1],
+    color: PANEL_APP_CONFIDENCE_LEVEL_COLORS[1],
+  },
+}
```
hanars, line 26: I think this would be cleaner as something like
```
Object.entries({ '1': 'red', '2': 'amber', '3': 'green'}).reduce((acc, [confidence, key]) => ({ ...acc, [key]: { 
  title: PANEL_APP_CONFIDENCE_DESCRIPTION[confidence], 
  color: PANEL_APP_CONFIDENCE_LEVEL_COLORS[confidence],
} }), {})
```

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+    </label>
+  )
+
+  return (
+    <BaseSemanticInput
+      {...props}
+      inputType="TextArea"
+      width={3}
+      label={iconLabel}
+      labelHelp={PA_LABEL_HELP}
```
hanars, line 49: it seems you set `labelHelp={PA_LABEL_HELP}` here but also in your definition for `iconLabel` - shouldn't it only be needed once?

ui/shared/components/panel/search/VariantSearchFormContainer.jsx
```diff
@@ -19,8 +19,14 @@ VariantSearchFormContainer.propTypes = {
 }
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  onSubmit: (searchParams) => {
-    dispatch(navigateSavedHashedSearch(searchParams, ownProps.history.push, ownProps.resultsPath))
+  onSubmit: ({ search, ...searchParams }) => {
+    const { rawItems } = search.locus
```
hanars, line 23: I don't think `search.locus` is always defined. Maybe better to do
```
let restructuredSearchParams = searchParams
if (search.locus) {
  ...
}
dispatch(navigateSavedHashedSearch(restructuredSearchParams, ...
```

hanars (changes requested):
looks really good! just some minor feedback

ui/pages/Search/components/filters/LocusListSelector.jsx
tommyli: None, so the else clause is redundant.

tommyli:
You're such a great reviewer, thanks for the feedback.

Have applied suggested changes.  As I was going through, I noticed genes with confidence level = 0 also needed handling so added a fourth text area.

ui/shared/components/panel/search/VariantSearchFormContainer.jsx
```diff
@@ -19,8 +19,17 @@ VariantSearchFormContainer.propTypes = {
 }
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  onSubmit: (searchParams) => {
-    dispatch(navigateSavedHashedSearch(searchParams, ownProps.history.push, ownProps.resultsPath))
+  onSubmit: ({ search, ...searchParams }) => {
+    const { rawItems } = search?.locus || ''
+    const formattedRawItems = (rawItems && typeof rawItems === 'object') ? toUniqueCsvString(Object.values(rawItems)) : rawItems
```
hanars, line 24: probably makes more sense to move these lines inside the `if` block

hanars (approved):
Once chane I noted I'd prefer to go in, but once thats made you are good to merge this

ui/shared/components/panel/search/VariantSearchFormContainer.jsx
tommyli: Done.

PR #2381, opened 2022-01-11
Gene search AF cutoff

hanars:


ui/shared/components/panel/search/GeneBreakdown.jsx
```diff
@@ -39,13 +39,6 @@ const COLUMNS = [
     format: row => (
       <SearchResultsLink
         location={row.geneId}
-        projectFamilies={Object.entries(row.families.reduce((acc, { family }) => {
-          if (!acc[family.projectGuid]) {
-            acc[family.projectGuid] = []
-          }
-          acc[family.projectGuid].push(family.familyGuid)
-          return acc
-        }, {})).map(([projectGuid, familyGuids]) => ({ projectGuid, familyGuids }))}
```
ShifaSZ, line 48: Is there any reason that this parameter can be eliminated?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -350,6 +350,16 @@ GeneDetails.propTypes = {
   rnaSeqData: PropTypes.object,
 }
 
+const GeneSearchLinkWithPopup = props => (
+  <Popup
+    trigger={
+      <GeneSearchLink {...props} />
+    }
+    content="Search for all variants with AF < 10% in this gene present in any affected individual"
```
ShifaSZ, line 358: `AF < 10%` is just the default value. It is not true after user changes it. Add `by default` to the sentence?

ShifaSZ (changes requested):

ui/shared/components/panel/search/GeneBreakdown.jsx
hanars: yes it was not being used at all, and when I was trying to figure out how the component was working its presence was super confusing

ui/shared/components/panel/variants/VariantGene.jsx
hanars: it is going to open up a search page with search results with those exact filters applied, its not pre-populating a form you can adjust. Also this was explicitly asked for in the ticket

ShifaSZ: I see.

ShifaSZ (approved):

PR #2383, opened 2022-01-12
add code contribution guidelines

hanars:
Add guidance for external code contributions to our README

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,3 +29,32 @@ On-prem installs can be created using docker-compose:
 For notes on how to update an older instance, see  	
 
 [Update/Migration Instructions](deploy/MIGRATE.md)
+
+
+## Contributing to seqr
+
+(Note: section inspired by, and some text copied from, [GATK](https://github.com/broadinstitute/gatk#contribute))
+
+We welcome all contributions to seqr. 
+Code should be contributed via GitHub pull requests against the main seqr repository.
+
+If you’d like to report a bug but don’t have time to fix it, you can submit a
+[GitHub issue](https://github.com/broadinstitute/seqr/issues/new?assignees=&labels=bug&template=bug_report.md&title=)
+
+For larger features, feel free to discuss your ideas or approach in our 
+[discussion forum](https://github.com/broadinstitute/seqr/discussions)
+
+To contribute code:
+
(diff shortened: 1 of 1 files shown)


sjahl (approved):
I think this looks great!

A suggestion for future thought (I don't think we need to block merging on this): A lot of the contributor guide(line)s that I like tend to include instructions on how to set up a development environment that resembles the way we work. For us, that might include instructions on how to set up a virtualenv, and how to run tests locally, etc. This kind of guide is also useful for our own internal hires.

Some good examples are:

- https://hail.is/docs/0.2/getting_started_developing.html
- https://matrix-org.github.io/synapse/latest/development/contributing_guide.html (something like this is way overkill for us, but kind of shows the length at which one could go :) )

hanars:
> A suggestion for future thought (I don't think we need to block merging on this): A lot of the contributor guide(line)s that I like tend to include instructions on how to set up a development environment that resembles the way we work.

Yeah, so we actually don't have any documentation anywhere for how to set up seqr for development 😬 , which is definitely a huge gap (I just spent a very painful few hours trying to help Avery do it from my memory). I made a ticket to write this documentation so I don't keep forgetting

README.md
```diff
@@ -29,3 +29,32 @@ On-prem installs can be created using docker-compose:
...
+[discussion forum](https://github.com/broadinstitute/seqr/discussions)
+
+To contribute code:
+
+* Submit a GitHub pull request against the master branch.
+* Break your work into small, single-purpose patches whenever possible. 
+However, do not break apart features to the point that they are not functional 
+(i.e. updates that require changes to both front end and backend code should be submitted as a single change)
+* For larger features, add a detailed description to the pull request to explain the changes and your approach
+* Make sure that your code passes all our tests and style linting
```
ShifaSZ, line 54: Do the tests refer to those automated by GitHub? More specific descriptions are preferred.

ShifaSZ (changes requested):

README.md
hanars: this is intentionally vague, it is up to the contributor if they want to set up tests to run locally or rely on our automated github tests, depending on what kind of change they are contributing

ShifaSZ (approved):

PR #2384, opened 2022-01-13
Select gene list on enter

hanars:
The feature request is to be able to hit "enter" to add a gene list. Forms automatically submit when you hit enter, so refactoring this to actually be a form


ShifaSZ (approved):
Looks good to me.

PR #2386, opened 2022-01-13
gCNV consequences and new calls

hanars:


seqr/utils/elasticsearch/constants.py
```diff
@@ -334,8 +335,11 @@ def get_prediction_response_key(key):
     'end': {},
     'start': {},
     'num_exon': {},
-    'geneIds': {'response_key': 'geneIds'},
+    'geneIds': {'response_key': 'geneIds', 'format_value': list},
     'defragged': {'format_value': bool},
+    'prev_call': {'format_value': bool},
```
ShifaSZ, line 340: Would it be better to use `prev_identical`?

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1644,7 +1650,7 @@ def test_filtered_get_es_variants(self):
     def test_sv_get_es_variants(self):
         setup_responses()
         search_model = VariantSearch.objects.create(search={
-            'annotations': {'structural': ['DUP']},
+            'annotations': {'new_structural_variants': ['NEW']},
```
ShifaSZ, line 1653: Use `NEW_SV_FIELD` instead of `'new_structural_variants'`?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/constants.py
hanars: no because thats not whats in the index?

seqr/utils/elasticsearch/es_utils_tests.py
hanars: why? we only hardcode constant in tests when we use them in many places and in this test file we barely use constants for the search dictionaries and this wasn't a constant before?

ShifaSZ: That must be fine for a test file. I didn't notice it.

ShifaSZ (approved):

PR #2387, opened 2022-01-13
Fix for reference data MGI parsing

tommyli:
Running `python -u manage.py update_all_reference_data --use-cached-omim` in our local instance resulted in errors.  One was a data issue in DBNSFP and the second was a change in format in the MGI report.  Raising this PR for reference.

I suspect this is not regularly run at Broad?  What's the process used at Broad to maintain these gene associated reference datasets?

### DIFF
--- a/reference_data/management/commands/update_dbnsfp_gene.py
+++ b/reference_data/management/commands/update_dbnsfp_gene.py
@@ -26,6 +26,7 @@ EXCLUDE_FIELDS = (
     'SORVA_LOF', 'Essential_gene', 'chr', 'MIM', 'OMIM', 'RVIS_percentile_EVS', 'RVIS_EVS', 'HIPred',
 )
 
+REFSEQ_OVERRIDE = {'NM_021006': 'ENSG00000256515'}
 
 class DbNSFPReferenceDataHandler(ReferenceDataHandler):
 
@@ -38,6 +39,8 @@ class DbNSFPReferenceDataHandler(ReferenceDataHandler):
                          for k, v in record.items() if not k.startswith(EXCLUDE_FIELDS)}
         parsed_record["function_desc"] = parsed_record["function_desc"].replace("FUNCTION: ", "")
         parsed_record['gene_id'] = parsed_record['gene_id'].split(';')[0]
+        if record['Refseq_id'] and REFSEQ_OVERRIDE.get(record['Refseq_id']):
+            parsed_record['gene_id'] = REFSEQ_OVERRIDE.get(record['Refseq_id'])
 
         gene_names = [record['Gene_name']]
         for gene_name_key in ['Gene_old_names', 'Gene_other_names']:
--- a/reference_data/management/commands/update_dbnsfp_gene.py
+++ b/reference_data/management/commands/update_dbnsfp_gene.py
@@ -26,8 +26,6 @@ EXCLUDE_FIELDS = (
     'SORVA_LOF', 'Essential_gene', 'chr', 'MIM', 'OMIM', 'RVIS_percentile_EVS', 'RVIS_EVS', 'HIPred',
 )
 
-REFSEQ_OVERRIDE = {'NM_021006': 'ENSG00000256515'}
-
 class DbNSFPReferenceDataHandler(ReferenceDataHandler):
 
     model_cls = dbNSFPGene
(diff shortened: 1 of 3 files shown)


hanars:
I don't really understand the dbnsfp issue you are seeing - the file it uses a Broad-hosted copy of dbnsfp that we have successfully loaded before so  regardless of how infrequently we have used it, if it has ever worked it should continue to work. Just want to understand whats going on here before I approve these changes. 

For MGI yeah they unfortunately do not version their data so we are vulnerable to them changing their formatting. I'll make a ticket for us to create our own Broad hosted copy so that in the future we can be guaranteed stability. 

In terms of updates, we update OMIM and HPO weekly in our seqr instance, everything else we wait for a specific release that our analysts ask us to update to. For instance, with dbnsfp we are using a pinned 4.0 version so running the update command should do nothing, but at some point we will want to go to v5 and will make a PR when that is wanted and manually run the command once

tommyli:
Sorry, I should've been more clear that there are two separate issues in this PR.  The `REFSEQ_OVERRIDE` in `update_dbnsfp_gene.py` is a hack to work around a data issue I was having when running `update_mgi()`.  Without it, I get:

```text
django.db.utils.IntegrityError: duplicate key value violates unique constraint \"reference_data_mgi_gene_id_marker_id_220a3b30_uniq\"
DETAIL:  Key (gene_id, marker_id)=(636609, MGI:98260) already exists.
", "@type": "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent"}
reference_data_mgi_gene_id_marker_id_220a3b30_uniq
```

After more digging at the code and data, I noticed the root cause was in `update_dbnsfp_gene.py`, it created a duplicate entry for the same gencode `gene_id` for different `entrez_gene_id`.  Below table should make it clear (after running `update_dbnsfp_gene.py`, version `dbNSFP4.0_gene`):

```sql
SELECT rdg.id, rdg.gene_id, rdg.gene_symbol, rdd.refseq_id, rdd.entrez_gene_id, rdd.gene_names
FROM reference_data_geneinfo rdg
  JOIN reference_data_dbnsfpgene rdd ON rdg.id = rdd.gene_id
WHERE rdd.entrez_gene_id IN ('6349', '414062');
```

| id | gene\_id | gene\_symbol | refseq\_id | entrez\_gene\_id | gene\_names |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 636609 | ENSG00000276085 | CCL3L1 | NM\_021006 | 6349 | CCL3L1;D17S1718;SCYA3L;SCYA3L1;G0S19-2;LD78BETA |
| 636609 | ENSG00000276085 | CCL3L1 | NM\_001001437 | 414062 | CCL3L3;MGC12815 |

The `REFSEQ_OVERRIDE` code hack is simply to override

```sql
SELECT rdg.gene_id, rdg.gene_symbol
FROM reference_data_geneinfo rdg
WHERE rdg.gene_id IN ('ENSG00000256515', 'ENSG00000276085');
```

| gene\_id | gene\_symbol |
| :--- | :--- |
| ENSG00000256515 | CCL3L3 |
| ENSG00000276085 | CCL3L1 |

What do you propose for handling these data issues?

hanars:
So the ensembl ID for dbnsfp is supposed to be unique - I think its an oversight that we didn't enforce that at the database level. One idea could be to change `entrez_gene_id` to `entrez_gene_ids` and allow it to have multiple values, and then make the gene names list a deduplicated combination of all the gene names form both entrez IDs. But I would recommend separating out the dbnsfp and MGI changes into 2 different PRs

tommyli:
The `REFSEQ_OVERRIDE` hack is now reverted, this branch should only contain the changes for MGI parsing, please review.

tommyli:
> One idea could be to change entrez_gene_id to entrez_gene_ids and allow it to have multiple values, and then make the gene names list a deduplicated combination of all the gene names form both entrez IDs

So did you mean updating `update_dbnsfp_gene.py` and rename column `entrez_gene_id` to `entrez_gene_ids` and store the entrez gene ids as a `;` delimited string like `gene_names`?  If so, would there be a performance issue since the parsing in `DbNSFPReferenceDataHandler` is currently streamed.  Or have I completely misunderstood what you meant?

hanars:
> So did you mean updating `update_dbnsfp_gene.py` and rename column `entrez_gene_id` to `entrez_gene_ids` and store the entrez gene ids as a `;` delimited string like `gene_names`? If so, would there be a performance issue since the parsing in `DbNSFPReferenceDataHandler` is currently streamed. Or have I completely misunderstood what you meant?

That is what I meant, and it wouldn't be a performance issue necessarily because you could still stream, but it would be a pretty hefty refactor to be able to check what has already been parsed. seqr currently simply does not support loading multiple entries for the same ensembl id- we just throw out multiples. So the change you are proposing is actually very large

tommyli:
I will stick with `REFSEQ_OVERRIDE` for now in our codebase since it's only one data hack.

Given the large refactoring effort, is it worth progressing with your suggestion?  Especially given this is debatably an upstream data issue?  Or do you see this as something that eventually needs to be implemented for more robust handling of upstream datasets?

Either way, can this PR be accepted since the `REFSEQ_OVERRIDE` hack is reverted?

hanars:
I don't feel strongly about the refactoring effort, we've been living fine for a while just missing a couple enrez ids. We won;t accept your `REFSEQ_OVERRIDE` hack to our main branch and I do want you to be aware that having multiple entries for a single ensembl ID might result in some bugs

hanars (approved):

hanars:
approved and rebased to our dev branch as that is where we like to merge things

PR #2397, opened 2022-01-18
Use gcs media backend

sjahl:
This PR adds a configuration option to use GCS to host the `MEDIA_ROOT` directory, with the goal of removing our dependence on a persistent GCP volume to store the data. This will enable us to redeploy seqr using more traditional kubernetes rollout semantics (which will be the way forward after #2344 )

If a `GCS_MEDIA_ROOT_BUCKET` isn't defined, the app will fall back to using a local directory for the `MEDIA_ROOT`, and will serve requests to `/media` with the builtin `django.views.static.serve` server. The intention here is for the behavior to remain unchanged for seqr local/docker-compose installs.

### DIFF
--- a/seqr/views/utils/orm_to_json_utils.py
+++ b/seqr/views/utils/orm_to_json_utils.py
@@ -205,7 +205,7 @@ def _get_json_for_families(families, user=None, add_individual_guids_field=False
                 pedigree_image = pedigree_image.url
             except Exception:
                 pedigree_image = None
-        return os.path.join("/media/", pedigree_image) if pedigree_image else None
+        return pedigree_image
 
     analyst_users = set(User.objects.filter(groups__name=ANALYST_USER_GROUP) if ANALYST_USER_GROUP else [])
     def _process_result(result, family):
--- a/seqr/views/utils/orm_to_json_utils.py
+++ b/seqr/views/utils/orm_to_json_utils.py
@@ -3,7 +3,6 @@ Utility functions for converting Django ORM object to JSON
 """
 
 import json
-import os
 from collections import defaultdict
 from copy import deepcopy
 from django.db.models import prefetch_related_objects, Prefetch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # _seqr_ Changes
 
+## 1/25/22
+* Add support for serving the django media root out of a google cloud storage bucket
+
 ## dev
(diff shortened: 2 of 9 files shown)


seqr/views/utils/orm_to_json_utils.py
```diff
@@ -205,7 +204,7 @@ def _get_pedigree_image_url(pedigree_image):
                 pedigree_image = pedigree_image.url
             except Exception:
                 pedigree_image = None
-        return os.path.join("/media/", pedigree_image) if pedigree_image else None
+        return pedigree_image
```
sjahl, line 207: @hanars I confirmed that this works in dev -- 

When the GCS bucket is enabled, the URL returned is a google object URL to the image in the bucket, and when it's disabled we're getting a `/media/foobar/baz.jpg` image as expected. Prepending the `/media` was redundant.

sjahl:
@hanars I could do a deploy at 7:30PM on either Wednesday or Friday this week. It takes a few minutes (took around 10mins in dev) to sync the local /seqr_static_files volume up to GCS. I'd probably schedule an hour of downtime to give some buffer time in case of any problems. I'll have to restart the seqr container to apply the seqr deployment manifest.

Plan is to merge this into dev, and then do a dev -> master PR for the prod release. Let me know if we've got anything in the pipeline before wed/fri that shouldn't be shipped.

hanars:
Wednesday is better for me than Friday. Since I don't think there's anything on this branch not already in master, it will be easier to change this PR to directly off master and release it directly on Wednesday. That way we don't need to worry about building assets etc on Wednesday if we do end up merging anything else to dev before then.

sjahl:
- [ ] All unit test are passing and coverage has not substantively dropped
- [ ] No new dependency vulnerabilities have been introduced
- [ ] Static assets have been built and committed if needed
- [ ] If any python requirements are updated, they are noted and here and will be updated before deploying: 
- [ ] Any database migrations are noted here, and will be run before deploying: 
- [ ] All major changes are recorded in the changelog, and the changelog has been updated to reflect the latest release
- [ ] Any new endpoints are explicitly tested to ensure they are only accessible to correctly permissioned users
- [ ] No secrets have been committed
- [ ] Infrastructure changes:
  
  OR 
  
  - [ ] Any changes to seqr's infrastructure are already been deployed to a new docker image, and these changes will be released via that image. All these changes have been tested before release on the docker image
  - [ ] All these changes have bee vetted for potential vulnerabilities

PR #2409, opened 2022-01-21
Upgrade ECK to 1.9.1 and ES to 7.16.3

sjahl:
This bumps elasticsearch to the latest ECK operator and 7.x elasticsearch release. I PR'd this against master, since I think this will likely get deployed without anything else; let me know if that's not the case.

There was also still a `docker_build()` call in the `deploy_elasticsearch()` method, which I removed since I believe we're no longer using our custom elasticsearch image.

I think the deploy steps here are just to run the servctl deploy command for elasticsearch. But, I haven't done this particular upgrade before, so if you can confirm that, please do 👍 

Closes https://github.com/broadinstitute/seqr-private/issues/1087

### DIFF
--- a/deploy/kubernetes/shared-settings.yaml
+++ b/deploy/kubernetes/shared-settings.yaml
@@ -1,6 +1,6 @@
 #####################################################
 ####### SETTINGS THAT MAY NEED TO BE MODIFIED #######
-ELASTICSEARCH_VERSION: 7.15.2
+ELASTICSEARCH_VERSION: 7.16.3
 LINKERD_VERSION: 'stable-2.10.0'
 
 ES_DATA_DISK_TYPE: pd-ssd
--- a/deploy/servctl_utils/deploy_command_utils.py
+++ b/deploy/servctl_utils/deploy_command_utils.py
@@ -107,8 +107,6 @@ def deploy_secrets(settings, components=None):
 def deploy_elasticsearch(settings):
     print_separator("elasticsearch")
 
-    docker_build("elasticsearch", settings, ["--build-arg ELASTICSEARCH_SERVICE_PORT=%s" % settings["ELASTICSEARCH_SERVICE_PORT"]])
-
     if settings["ONLY_PUSH_TO_REGISTRY"]:
         return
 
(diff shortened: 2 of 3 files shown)


hanars:
Since its going straight to master adding the release checklist

- [x] Infrastructure changes:
  
  OR 
  
    -  N/A 
    - @sjahl have we tested in dev?

sjahl:
I tested this in dev this morning. Unfortunately, I don't think we can upgrade both ES and the operator version with a single rolling restart of elasticsearch (both the image update and the operator update trigger restarts, and I don't think there's a way we can really apply both upgrades simultaneously). So, we'll need to schedule downtime for two rolling restarts.

Given that, the ship process should like like this:

1. Run `servctl deploy elasticsearch gcloud-prod`: updates components to the new elasticsearch images, triggers rolling restart.

During this upgrade, it looks like elasticsearch creates an index that contains deprecation warnings. The index was configured with a replica, which couldn't schedule all of its shards on the single node dev cluster. Result being, the cluster went into a yellow state, and the operator upgrade will refused to restart the pods. I don't think this is a problem for prod, but if this is encountered, we just set the number of replicas on the index to 0:

Find the shard/index that won't allocate: `curl http://elasticsearch-es-http:9200/_cluster/allocation/explain?pretty=true`

Set the replicas to 0 on it:
```
curl -X PUT "elasticsearch-es-http:9200/.ds-.logs-deprecation.elasticsearch-default-2022.01.24-000001/_settings?pretty" -H 'Content-Type: application/json' -d'
{
  "index" : {
    "number_of_replicas" : 0
  }
}
'
```

Once that's resolved, you can replace the CRDs with the new version, and apply the operator manifests as follows

2. Run `kubectl replace -f https://download.elastic.co/downloads/eck/1.9.1/crds.yaml`
3. Run kubectl apply -f `https://download.elastic.co/downloads/eck/1.9.1/operator.yaml`: This triggers a rolling restart.

hanars:
Thanks for looking into this. Do you want to do this deploy Wednesday night along with the gcs media release so wet just have one big down time, or would you rather do it separately? How long do you think we should communicate for this down time?

sjahl:
Yeah, I can tack this on to the wednesday deployment. I would say let's schedule two hours total (my suspicion is that the GCS stuff won't take a full hour). But, it's been a while since I did something that triggered a rolling restart of the prod ES cluster, so I don't have a good frame of reference of how how it takes to get it back to green state. Happy to extend the window if you've got some thoughts there.

hanars:
Last time I did a rolling restart of es it took like 30-40 minutes, so communicating 2 hours overall sounds right to me

PR #2420, opened 2022-01-25
clean up dependency vulnerabilities

hanars:


ShifaSZ:
Do you want to add a changelog for this since it relates to security/regulatory?

hanars:
> Do you want to add a changelog for this since it relates to security/regulatory?

Thanks for checking, but no.  The changelog is for users to know about changes that will effect them. So either changes that will require additional work for people running local seqr (like database migrations or configuration changes) or substantial changes to the application that a user would notice and care about. Since local installations use pre-built static assets bumps to ui patch versions don't effect anyone

Issue #2085, opened 2021-09-01
Deletions and duplications called by gCNV on exome data 


Glemiret:
As discussed on the meeting on August 31, deletions and duplications called by the gCNV pipeline on exome data should be named “exome deletion” and “exome duplication” in seqr. The options of “exome deletion” and “exome duplication” should also appear in the annotations section (under structural) in seqr searches. The words “deletion” and “duplication” should be kept and applied to CNVs identified from genome data.

hanars:
we will wait to work on this until the next exome gCNV callset is available for loading

hanars:
Rethinking this - we will do the backend work that will allow us to support data loaded either the way it currently is or the way it will need to be for this functionality to work. Then, we will reload all the data, and after that we will implement the front end changes to take advantage of the backend changes, and deprecate the bridge functionality

hanars:
Bridge functionality has been added to seqr, and the loading pipeline had been updated accordingly. we are now blocked on reloading all gcnv data using the updated pipeline

Glemiret:
Hi Hana, is it now possible to apply the changes in this ticket?

hanars:
this has been implemented. There are now separate annotations for "gCNV Deletion" and "gCNV Duplication" that will work separately from the genome SV "Deletion" and "duplication"

PR #2421, opened 2022-01-25
explicitly search for gcnv sv annotations

hanars:


ui/shared/utils/constants.js
```diff
@@ -483,12 +483,22 @@ export const VEP_GROUP_SV_NEW = 'new_structural_variants'
...
-    description: 'A large deletion',
+    description: 'A large deletion called from exome data',
+    text: 'gCNV Deletion',
+    value: 'gCNV_DEL',
+  },
+  {
+    description: 'A large duplication called from exome data',
+    text: 'gCNVDuplication',
+    value: 'gCNV_DUP',
+  },
```
ShifaSZ, line 494: The text fields are not so unified. Should the `gCNVDuplication` need a blank in it? According to issue #2085, should we use `exome` instead of `gCNV` for the text fields?

ShifaSZ (changes requested):

ui/shared/utils/constants.js
hanars: we talked offline and agreed "gCNV" was clearer than "exome". But good catch on the typo!

ShifaSZ (approved):

PR #2426, opened 2022-01-27
remove settings backup job

hanars:
Now that the seqr static files disk is gone we don't need to be running a cron job to back it up

- [x] Infrastructure changes:
  
  OR 
  
  - [x] Any changes to kubernetes configurations will be deployed through a full seqr kubernetes deployment after merging. All these changes have been tested on dev
  - can't *really* test in dev because cron jobs are disabled there, but should be fine


deploy/docker/seqr/entrypoint.sh
```diff
@@ -24,9 +24,6 @@ if [ -e "/.config/service-account-key.json" ]; then
     cp /.config/boto /root/.boto
 fi
 
-# link to persistent disk dir with static files
-mkdir -p /seqr_static_files/generated_files
```
sjahl, line 28: On this, I'm wondering if we do in fact want to leave something here for local install users that don't use a GCS bucket?

Do you think doing this conditionally if the `STATIC_MEDIA_DIR` env var is set is appropriate? Maybe we should have a convo about what to do about this in the case of local installs -- we probably want to avoid having people accumulate local state in a non-persistent location in the container.

sjahl (commented):
Interesting, with a name like "settings_backup" I don't know that I would have guessed that it had much to do with the seqr_static_files disk 😄 

Just the one question about the media dir on local installs. I think the way things are now, if local install users are doing anything that puts pedigree images or other uploads in the media dir, we're probably wiping out state if their container dies for some reason (or, in the future if they are pulling new docker images for updates)

deploy/docker/seqr/entrypoint.sh
sjahl: I guess if we use something like a volume mount in the docker-compose file, we won't need to do anything in the entrypoint to create this directory for local install users

hanars: I think using a volume for local installations is probably the way to go here. django will create the nested `generated_files` directory on its own if its needed so with the volume mounted local users wont need any additional set up here

sjahl (approved):

PR #2443, opened 2022-02-03
show node stats for es status page

hanars:


seqr/views/apis/data_manager_api.py
```diff
@@ -53,7 +56,7 @@ def elasticsearch_status(request):
     return create_json_response({
         'indices': indices,
         'diskStats': list(disk_status.values()),
-        'elasticsearchHost': ELASTICSEARCH_SERVER,
```
hanars, line 56: this was added back when we used to actually update the static IP address of the ES cluster fairly frequently and it was important to know what that was. Now that we use a kuberenetes service, this is just always the service name and doesn't tell us anything helpful

PR #2444, opened 2022-02-03
allow deletion of analyzed families

hanars:


seqr/fixtures/1kg_project.json
```diff
@@ -382,6 +382,17 @@
         "success_story": "Differential treatement",
         "success_story_types": ["A", "D"]
     }
+},
+    {
+    "model": "seqr.familyanalysedby",
+    "pk": 1,
+    "fields": {
+        "guid": "FAB000001_1",
```
ShifaSZ, line 390: Should we explicitly test the entry deleted while deleting the family?

ShifaSZ (changes requested):

seqr/fixtures/1kg_project.json
hanars: no

ShifaSZ (approved):

PR #2447, opened 2022-02-03
use stable copy of mgi data

hanars:


reference_data/management/commands/update_mgi.py
```diff
@@ -22,20 +22,21 @@ def __init__(self, **kwargs):
...
-        return super(MGIReferenceDataHandler, self).get_gene_for_record(record)
-
+        entrez_gene = self.entrez_id_to_gene.get(record.pop('entrez_gene_id'))
+
+        try:
+            return super(MGIReferenceDataHandler, self).get_gene_for_record(record)
+        except ValueError as e:
+            if entrez_gene:
+                return entrez_gene
+            raise e
```
ShifaSZ, line 39: When returns `entrez_gene`, the `record['gene_symbol']` won't be deleted. It is different from the code before. Is that what you want?

ShifaSZ (changes requested):

reference_data/management/commands/update_mgi.py
hanars: it will be because `super(MGIReferenceDataHandler, self).get_gene_for_record(record)` does that

ShifaSZ: okay

ShifaSZ (approved):

PR #2457, opened 2022-02-07
remove unused freq filters from ui

hanars:


ui/shared/components/panel/search/constants.js
```diff
@@ -357,8 +345,7 @@ export const SV_FREQUENCIES = [
   },
 ]
 
-export const FREQUENCIES = [...SNP_FREQUENCIES]
-FREQUENCIES.splice(5, 0, ...SV_FREQUENCIES)
+export const FREQUENCIES = [...SNP_FREQUENCIES, ...SV_FREQUENCIES]
```
ShifaSZ, line 348: Besides the order, the ticket also asks to change the name of `SV Callset` to `This SV Callset`.

ShifaSZ (changes requested):

ui/shared/components/panel/search/constants.js
hanars: oh good catch thank you!

ShifaSZ (approved):

PR #2466, opened 2022-02-10
move child entity code to search file

hanars:
This is another " no changed functionality" PR. Takes a utility function that has a conditional option to do lots of extra stuff, and moves the extra stuff into the one file that actually needs it


seqr/views/apis/variant_search_api.py
```diff
@@ -366,6 +380,33 @@ def _get_projects_details(projects, user, project_category_guid=None):
     return response
 
 
+def _add_parent_ids(response, projects, family_models, individual_models):
```
ShifaSZ, line 383: Although the two `_add_parent_ids` are in two different files,  it is still a little confusing. Maybe you want to eliminate this function and inline the codes.

ShifaSZ (approved):

PR #2470, opened 2022-02-11
Fix bugs with comp het search on structural consequences

hanars:


ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -364,12 +364,16 @@ const GeneSearchLinkWithPopup = props => (
   />
 )
 
+export const getGeneConsequence = (geneId, variant) => {
+  const geneTranscripts = variant.transcripts[geneId]
+  return geneTranscripts && geneTranscripts.length > 0 &&
+    (geneTranscripts[0].majorConsequence || '').replace(/_/g, ' ')
+}
+
```
ShifaSZ, line 372: It is ideal if this can be defined in a util file. But it still works.

ShifaSZ (approved):

Issue #2437, opened 2022-01-31
Possible race condition in gcloud auth configuration


sjahl:
From the crash this weekend. Two errors of note here. I've included the Seqr stacktrace below, but I don't think it's actually what's important. The problems that we had should probably be caught prior to arriving at the seqr codepath. This occurred when the container was starting up: we failed to authenticate with `gcloud` and then (expectedly) seqr died on startup because it had no auth identity and then just stopped. We have a very long `sleep` in the seqr entrypoint, which is why we end up sitting here

1) On container startup, we failed to connect to the google authentication api . From what I can tell from other errors, we even had trouble talking to github and PyPI at startup. So, I think either the node was having network issues, or it just takes a while for the network interface to become available when the pod starts up. I don't have a great answer for why this seems to be happening more, but my suspicion is that without the GCP volume attachment for static media, the container is actually starting up faster than it used to.

```
+ gcloud auth activate-service-account --key-file /.config/service-account-key.json
ERROR: gcloud crashed (TransportError): HTTPSConnectionPool(host='oauth2.googleapis.com', port=443): Max retries exceeded with url: /token (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f6dcbc1cdf0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

2) This is the seqr stacktrace, included for completeness. I don't really think there's much to be done on the seqr side of things here since the gcloud authentication is happening outside of the app.

```
[2022-01-30 03:23:45 -0500] [161] [ERROR] Exception in worker process
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 586, in spawn_worker
    worker.init_process()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base.py", line 135, in init_process
    self.load_wsgi()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base.py", line 144, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 52, in load
    return self.load_wsgiapp()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 41, in load_wsgiapp
[...]

PR #2472, opened 2022-02-14
Add retries to gcloud cli auth configuration

sjahl:


sjahl:
Closes #2437

settings.py
```diff
@@ -395,7 +396,18 @@
...
                                                capture_output=True, text=True, shell=True).stdout.split('\n')[0]
     if not SERVICE_ACCOUNT_FOR_ANVIL:
-        raise Exception('Error starting seqr - gcloud auth is not properly configured')
+        # attempt to acquire a service account token
+        if os.path.exists('/.config/service-account-key.json'):
+            SERVICE_ACCOUNT_FOR_ANVIL = \
+                re.findall('\[(.*)\]',
+                           subprocess.run(['gcloud', 'auth', 'activate-service-account', '--key-file',
+                                           '/.config/service-account-key.json'],
+                                           capture_output=True, text=True).stderr)[0] # nosec
```
sjahl, line 405: i had to add the `# nosec` here to pass codacy. It initially complained about the use of `shell=True`, which i thought was reasonable. But, after doing what I thought it wanted, and using a `shell=False`, it complained about unsanitized command input... but the input here is a static list. Let me know if you have any thoughts on this (or thoughts on how to improve the logic in this section in general, @hanars )

hanars (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## dev
 
+* Update seqr dockerfile to improve immutability and build automation (REQUIRES IMAGE UPDATE)
+* Adds NPM asset build to the seqr dockerfile
```
hanars, line 6: I think this was an error and meant as part of a different PR?

hanars (changes requested):

sjahl:
Whoops, yeah, forgot to switch branches!

hanars (approved):

PR #2479, opened 2022-02-22
ensure that we also push the latest docker tag on auto builds

sjahl:
The cloudbuild file was only pushing the commit SHA tag and the `gcloud-{env}` tag (I had incorrectly assumed that cloudbuild would bump the `latest` tag on its own).

Let me know if there's other tags that we'd like to push. It looks like previously we were pushing a timestamp tag, but I'm unsure of what the utility of that was. Happy to add it if we'd like to keep it.

### DIFF
--- a/.cloudbuild/seqr-docker.cloudbuild.yaml
+++ b/.cloudbuild/seqr-docker.cloudbuild.yaml
@@ -3,6 +3,7 @@ steps:
   args:
   - --destination=gcr.io/seqr-project/seqr:${COMMIT_SHA}
   - --destination=gcr.io/seqr-project/seqr:${_CUSTOM_BRANCH_TAG}
+  - --destination=gcr.io/seqr-project/seqr:latest
   - --dockerfile=deploy/docker/seqr/Dockerfile
   - --cache=true
   - --cache-ttl=168h


sjahl:
The build failure on this has exposed kind of a nasty pip problem. An upstream dependency (markupsafe, which I believe gets pulled in via django -> jinja -> markupsafe) introduced a breaking change in a minor release, and our pip install has automatically pulled in that new package, since we don't resolve any versions deeper than the ones in our requirements.txt.

The long term fix is to utilize something like [pip-tools](https://pypi.org/project/pip-tools/), which will generate a requirements file that will keep our whole dependency tree stable. Will chat with @hanars when she gets back.

hanars:
our kubernetes configs and the local installs use the `gcloud-prod` (or `gcloud-dev`) tags, not the latest tag, so thats probably the most important to move. but seems like the pip issue may be blocking

sjahl:
> our kubernetes configs and the local installs use the `gcloud-prod` (or `gcloud-dev`) tags, not the latest tag, so thats probably the most important to move. but seems like the pip issue may be blocking

Yup -- the `_CUSTOM_BRANCH_TAG` variable in the cloudbuild is the one that applies gcloud-dev or -prod, based on which branch triggered the build. With this PR, the tags that get pushed are:

- git commit sha
- gcloud-dev (when building the dev branch), or gcloud-prod (when building master)
- latest

hanars (approved):

PR #2482, opened 2022-02-23
Enable seqr to automatically trigger airflow AnVIL dags

averywpx:


settings.py
```diff
@@ -376,6 +376,10 @@
 
 SERVICE_ACCOUNT_FOR_ANVIL = None
 
+AIRFLOW_CLIENT_ID = os.environ.get('AIRFLOW_CLIENT_ID')
+AIRFLOW_WEBSERVER_ID = os.environ.get('AIRFLOW_WEBSERVER_ID')
+DAG_VERSION = '0.0.1'
```
hanars, line 381: `DAG_VERSION` should be in `anvil_workspace_api.py`, it doesn't need to be a shared setting

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -100,6 +104,7 @@ def create_project_from_workspace(request, namespace, name):
     """
     # Validate that the current user has logged in through google and has sufficient permissions
     workspace_meta = check_workspace_perm(request.user, CAN_EDIT, namespace, name, can_share=True, meta_fields=['workspace.bucketName'])
+    workspace_meta = {'workspace' : {'bucketName' : 'my_bucket'}}
```
hanars, line 107: make sure not to commit mocked out testing code in the PR

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -120,6 +125,7 @@ def create_project_from_workspace(request, namespace, name):
 
     # Add the seqr service account to the corresponding AnVIL workspace
     added_account_to_workspace = add_service_account(request.user, namespace, name)
+    added_account_to_workspace = False
```
hanars, line 128: make sure not to commit mocked out testing code in the PR

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -139,6 +145,7 @@ def create_project_from_workspace(request, namespace, name):
 
     # Validate the VCF to see if it contains all the required samples
     samples = get_vcf_samples(data_path)
+    samples = ['sample1', 'sample2']
```
hanars, line 148: ditto

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
                      detail=sample_ids)
 
+    # trigger dags
```
hanars, line 184: I think it would be most readable to have all of this logic inside a single helper function, something like `_trigger_data_loading`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
     )
 
     safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, message_content)
+
+
+def generate_AnVIL_variables(project, data_path, sample_type):
```
hanars, line 273: for all of these helper functions, since they are only used in this file its best to mark them as private by starting all of the function names with an underscore (`_generate_AnVIL_variables` instead of ` generate_AnVIL_variables`)

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    pipeline_dag = {
```
hanars, line 274: in general, helper functions should either encapsulate functionality that is reused in many places, or functionality that is complicated. One-liners that are only used once are better as inline code than helper functions

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
                      detail=sample_ids)
 
+    # trigger dags
+    anvil_type = 'AnVIL_{sample_type}'.format(sample_type=request_json['sampleType'])
+
+    anvil_variables = get_variables(anvil_type)
+    updated_anvil_variables = generate_AnVIL_variables(project, data_path, request_json['sampleType'])
+    projects = anvil_variables["active_projects"]
```
hanars, line 189: so for the dag, we no longer care about `active_projects` we care about `projects_to_run`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
...
+    updated_projects = updated_anvil_variables["active_projects"]
+
+    if anvil_variables != updated_anvil_variables: 
+        update_variables(anvil_type, updated_anvil_variables)
```
hanars, line 193: I think we should always run `update_variables`, theres no real-world case where you would be in this endpoint and the project would already be in the DAG. This is a case of premature optimization

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
                      detail=sample_ids)
 
+    # trigger dags
+    anvil_type = 'AnVIL_{sample_type}'.format(sample_type=request_json['sampleType'])
```
hanars, line 185: so this isn't really an anvil "type", this is something specific to the dag although its not clear to me what. It would probably be better to rename this to something more descriptive

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
...
+    projects = anvil_variables["active_projects"]
+    updated_projects = updated_anvil_variables["active_projects"]
+
+    if anvil_variables != updated_anvil_variables: 
+        update_variables(anvil_type, updated_anvil_variables)
+
+    dag_id = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=anvil_type, version=DAG_VERSION)
+
+    if set(projects) != set(updated_projects):
+        dag_task_ids = get_task_ids(dag_id)
```
hanars, line 198: this code block makes sense as a helper function, as its a but confusing what its doing. Maybe name it something like `_wait_for_dag_variable_update` to be clearer

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    pipeline_dag = {
+        "active_projects": [project.guid],
+        "es_enable_export": 'true',
+        "vcf_path": data_path,
+        "project_path": '{}v1'.format(_get_loading_project_path(project, sample_type)),
+        "projects_to_run": [project.guid],
+    }
+    return pipeline_dag
+
+def get_variables(variable_key):
```
hanars, line 283: the word "variables" usually refers to variables in the code, so the name of this function is a little confusing. Maybe something like `_get_dag_variables`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,6 +181,27 @@ def create_project_from_workspace(request, namespace, name):
...
+
+    if anvil_variables != updated_anvil_variables: 
+        update_variables(anvil_type, updated_anvil_variables)
+
+    dag_id = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=anvil_type, version=DAG_VERSION)
+
+    if set(projects) != set(updated_projects):
+        dag_task_ids = get_task_ids(dag_id)
+        updated_task_ids = get_task_ids(dag_id)
+        while(set(dag_task_ids) == set(updated_task_ids)):
```
hanars, line 200: so this definitely works, but I do wonder if theres a more explicit way to do this then just seeing if anything changed. What do the task ids looks like, in the end?

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+
+def get_task_ids(dag_id):
+
+    endpoint = 'api/v1/dags/{}/tasks'.format(dag_id)  
+    airflow_response = make_request(endpoint, method='GET')
+
+    tasks = airflow_response['tasks']
+    task_lst = []
+    for task_dict in tasks:
+        task_lst += [task_dict['task_id']]
```
hanars, line 313: this is fine, but its neater to do things like this as a list comprehension:
```
task_ids = [task_dict['task_id'] for task_dict in tasks]
```

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    return task_lst
+    
+
+def trigger_dag(dag_id):
+    json_data = {
+        'conf' : {}
```
hanars, line 319: what does this empty "conf" do?

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        }
+    endpoint = 'api/v1/dags/{}/dagRuns'.format(dag_id) 
+    make_request(endpoint, method='POST', json=json_data)
+
+
+def make_request(endpoint, method, **kwargs):
```
hanars, line 325: so making really generic helpers like this is nice in that they are completely extensible, but frustrating in that its hard to know how you can use them, and its hard to know how they are being used which makes it hard to ever update behavior. I think the following function signature would be much clearer:
```
def _make_airflow_api_request(endpoint, method='GET', json=None, timeout=90):
```

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    # Set the default timeout, if missing
+    if 'timeout' not in kwargs:
+        kwargs['timeout'] = 90
+
+    # Obtain an OpenID Connect (OIDC) token from metadata server or using service
+    # account.
+    google_open_id_connect_token = id_token.fetch_id_token(Request(), AIRFLOW_CLIENT_ID)
+    
+    webserver_url = (
```
hanars, line 334: adding strings together works, but is not very pythonic or easy to maintain. You want to use either `.format` or use an f string as follows:
```
webserver_url = f'https://{AIRFLOW_WEBSERVER_ID}.appspot.com/{endpoint}'
```
You might also want to consider hardcoding the api path here, as you currently duplicate it in every call to this helper:
```
webserver_url = f'https://{AIRFLOW_WEBSERVER_ID}.appspot.com/api/v1/{endpoint}'
```

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -240,3 +268,89 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        + '.appspot.com/'
+        + endpoint
+    )
+    resp = requests.request(
+        method, webserver_url,
+        headers={'Authorization': 'Bearer {}'.format(
+            google_open_id_connect_token)}, **kwargs)
+
+    if resp.status_code == 403:
+        raise Exception('Service account does not have permission to '
```
hanars, line 346: so we definitely need to add better error handling to the dag triggering. In general, we want the user to see success no matter what and notify ourselves if something goes wrong. This means that for whatever top-level helper function you wrap the whole triggering code in, you want to put it in a try/except block:

```
try:
    _trigger_data_loading(...)
except Exception as e:
    logger.error(e)
    safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, str(e))
```

let me know if you have any questions about this

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -3,6 +3,11 @@
 import time
 import tempfile
 from datetime import datetime
+import requests
+import base64
```
hanars, line 7: I don't think this is used

hanars (changes requested):
This is overall really great! There are a lot of small change requests here to make the code a bit more readable and maintainable in the long run, but this is really awesome work!

seqr/views/apis/anvil_workspace_api.py
averywpx: `requests` is used in `_make_airflow_api_request`, and base64 is not used

seqr/views/apis/anvil_workspace_api.py
hanars: So the try/except handling for this is now good, but in general raising base class `Exception` is not good python style, you really want to always try and raise a more specific exception class, even though in this case your error handling will catch exceptions of all classes. I think the best thing to do here is rather than raising your own custom exceptions on bad responses, use the built in `resp.raise_for_status()`. This will raise an exception on any non-200 response with a useful exception message, and won't require any additional handling on your part

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -223,7 +235,7 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
   
         The sample IDs to load have been uploaded to {ids_path}.  
   
-        DAG for the loading pipeline:
+        {dag_name} is triggered with following:
```
hanars, line 238: I would leave the word "DAG" in here to make it clear what was triggered, and have this be `DAG {dag_name} was triggered with following`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -236,7 +248,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
         project_name=project.name,
         sample_type=sample_type,
         genome_version=GENOME_VERSION_LOOKUP.get(project.genome_version),
+        dag_name = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=sample_type, version=DAG_VERSION),
         dag=json.dumps(pipeline_dag, indent=4),
```
hanars, line 252: so the `pipeline_dag` variable is being constructed in this function, and is not the actual post body that the dag was triggered with. I think it would be better if we actually used the real post body in this message instead

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -236,7 +248,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
 
     safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, message_content)
+
+def _trigger_anvil_dags(project, data_path, user_input):
+    genome_test_type = 'AnVIL_{sample_type}'.format(sample_type=user_input['sampleType'])
+
+    anvil_variables = _get_dag_variables(genome_test_type)
+    updated_anvil_variables = {
+        "active_projects": [project.guid],
+        "es_enable_export": 'true',
```
hanars, line 263: I though there was away to set this for a given dag and not need to set it on every dag run. Lets do that instead

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -236,7 +248,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+ 
+    _update_variables(genome_test_type, updated_anvil_variables)
+
+    dag_id = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=genome_test_type, version=DAG_VERSION)    
+    _wait_for_dag_variable_update(projects, updated_projects, dag_id)
+    
+    _trigger_dag(dag_id)
+    
+def _wait_for_dag_variable_update(projects, updated_projects, dag_id):
+    if set(projects) != set(updated_projects):
```
hanars, line 279: as discussed in our one-on-one, we want to update this behavior to check that at least one task_id contains the project guid in it, instead of waiting for the list of tasks to change

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -236,7 +248,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
         project_name=project.name,
         sample_type=sample_type,
         genome_version=GENOME_VERSION_LOOKUP.get(project.genome_version),
+        dag_name = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=sample_type, version=DAG_VERSION),
         dag=json.dumps(pipeline_dag, indent=4),
     )
 
     safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, message_content)
+
+def _trigger_anvil_dags(project, data_path, user_input):
```
hanars, line 257: this is very nitpicky, but we only will ever trigger one dag at a time so using the plural "dags" in this fuction name is confusing. I think also its better to use a more generic name explaining what is happening, something like `_trigger_data_loading` since DAG is not exactly a common-knowledge thing in coding. For the very dag-specific helper functions, like `_wait_for_dag_variable_update` it is okay to keep the word dag in the function name, but for this parent function I think its better to be less specific and more descriptive

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,8 +178,16 @@ def create_project_from_workspace(request, namespace, name):
...
 
-    # Send a slack message to the slack channel
-    _send_load_data_slack_msg(project, ids_path, data_path, request_json['sampleType'], request.user)
+    # use airflow api to trigger AnVIL dags
+    try:
+        _trigger_anvil_dags(project, data_path, request_json)
+        # Send a slack message to the slack channel
+        _send_load_data_slack_msg(project, ids_path, data_path, request_json['sampleType'], request.user)
+    except Exception as e:
+        logger.error(e)
```
hanars, line 187: the `logger` class here requires a secondary argument with the user - this needs to be `logger.error(e, request.user)`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -174,8 +178,16 @@ def create_project_from_workspace(request, namespace, name):
...
+        safe_post_to_slack(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, str(e))
```
hanars, line 188: So I discussed this with the loading team, and we've determined the following is the notification behavior we want if the DAG fails to trigger for any reason:
1) We should always send the "success" message (`_send_load_data_slack_msg`) regardless of whether or not the dag actually was triggered. This allows that channel to have a full record of all loading requests we receive, regardless of how the DAG ends up being triggered. You do not need to change the text of the message if the trigger does fail, it is okay to leave it as is with "DAG was triggered"
2) We want to send notifications of failure to the `seqr_loading_notifications` channel (you will need to add a new variable for that). This message needs to include the error that occurred, the dag name, and the dag variables. I think this will be a good template:
```
"""
ERROR triggering AnVIL loading for project {project_guid}: {e} 
  
DAG {dag_name} should be triggered with following: 
```{dag}```
"""
```
3) We should leave the `logger.error(e, request.user)` as it is

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +244,92 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        
+        DAG {dag_name} should be triggered with following: 
+        ```{dag}```
+        """.format(
+            project_guid = project.guid,
+            e = e,
+            dag_name = "seqr_vcf_to_es_AnVIL_{anvil_type}_v{version}".format(anvil_type=sample_type, version=DAG_VERSION),
+            dag = json.dumps(pipeline_dag, indent=4)
+        )
+    safe_post_to_slack(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, message_content)
```
hanars, line 266: these actually need to go to a different channel - `seqr_loading_notifications` . You may need to add a new variable for this

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
         # Test saving ID file exception
         url = reverse(create_project_from_workspace, args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME2])
         mock_mv_file.side_effect = Exception('Something wrong while moving the ID file.')
+        # Test triggering dag exception
+        responses.replace(responses.PATCH, 
+                      '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL), 
+                      json={'test': 2},
```
hanars, line 366: what does this body do? if its a generic error response it is more readable to do something like `{'error': 'not found'}` or anything thats more clearly an error

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+    ]
+}}```
+        """.format(guid=project2.guid)
+        slack_calls = [
+            mock.call(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message_on_failure),
+            mock.call(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, slack_message_as_normal),
+        ]
+        mock_slack.assert_has_calls(slack_calls)
+        mock_send_email.assert_not_called()
+
```
hanars, line 426: we should also test that the `responses` for this call look the way we want them to. You'll want to add `responses.calls.reset()` before this call and then add the responses tests down here

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+    "project_path": "gs://seqr-datasets/v02/GRCh38/AnVIL_WES/{guid}/v1",
+    "projects_to_run": [
+        "{guid}"
+    ]
+}}```
+        """.format(
+            guid=project2.guid,
+            airflow_url = MOCK_AIRFLOW_URL
+        )
+        #mock_slack.assert_called_with(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message_on_failure)
```
hanars, line 400: this should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+    "projects_to_run": [
+        "{guid}"
+    ]
+}}```
+        """.format(guid=project2.guid)
+        slack_calls = [
+            mock.call(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message_on_failure),
+            mock.call(SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, slack_message_as_normal),
+        ]
+        mock_slack.assert_has_calls(slack_calls)
```
hanars, line 424: So this works, but testing the success message twice is kind of overkill, and in fact it will make it hard for us in the future to ever update it since its being tested an extra time here. The better way to do this would be to do `self.assert_any_call(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, slack_message_on_failure)` which will test that the failure slack was sent without checking any other slack calls

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -100,7 +104,7 @@ def create_project_from_workspace(request, namespace, name):
     """
     # Validate that the current user has logged in through google and has sufficient permissions
     workspace_meta = check_workspace_perm(request.user, CAN_EDIT, namespace, name, can_share=True, meta_fields=['workspace.bucketName'])
-
+
```
hanars, line 107: adding this space violates our style rules

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -176,7 +180,14 @@ def create_project_from_workspace(request, namespace, name):
...
 
-    # Send a slack message to the slack channel
+    # use airflow api to trigger AnVIL dags
+    try:
+        _trigger_data_loading(project, data_path, request_json['sampleType'])
+    except Exception as e:
+        logger.error(str(e), request.user)
+        _send_slack_msg_on_failure_trigger(e, project, data_path, request_json['sampleType'])
+    
+    # Send a slack message to the slack channel
```
hanars, line 190: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +244,92 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        "vcf_path": data_path,
+        "project_path": '{}v1'.format(_get_loading_project_path(project, sample_type)),
+        "projects_to_run": [project.guid],
+    }
+    return dag_variables
+    
+def _wait_for_dag_variable_update(dag_id, project):
+    updated_project = project.guid
+    dag_projects = _get_task_ids(dag_id)
+    while updated_project not in ''.join(dag_projects):
```
hanars, line 294: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +244,92 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    endpoint = 'variables/{}'.format(key)
+    val_str = json.dumps(val)
+    json_data= {
+        "key": key,
+        "value": val_str
+        }
+    _make_airflow_api_request(endpoint, method='PATCH', json=json_data)
+
+def _get_task_ids(dag_id):
+    endpoint = 'dags/{}/tasks'.format(dag_id)
```
hanars, line 307: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +244,92 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    endpoint = 'dags/{}/tasks'.format(dag_id)  
+    airflow_response = _make_airflow_api_request(endpoint, method='GET')
+
+    tasks = airflow_response['tasks']
+    task_ids = [task_dict['task_id'] for task_dict in tasks]
+    return task_ids
+    
+
+def _trigger_dag(dag_id):
+    endpoint = 'dags/{}/dagRuns'.format(dag_id)
```
hanars, line 316: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +170,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
                                            mock_send_email, mock_add_service_account,
                                            mock_has_service_account, mock_load_file, mock_api_logger, mock_time,
                                            mock_utils_logger):
+        # Set up api responses
+        # update variables
+        responses.add(responses.PATCH, 
+                      '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL), 
+                      json={'test': 1},
+                      status=200)
+        # get task id
```
hanars, line 184: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +170,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
                                            mock_has_service_account, mock_load_file, mock_api_logger, mock_time,
                                            mock_utils_logger):
+        # Set up api responses
+        # update variables
+        responses.add(responses.PATCH, 
+                      '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL), 
+                      json={'test': 1},
+                      status=200)
+        # get task id 
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL),
```
hanars, line 185: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +170,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
+                      status=200)
+        # get task id 
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL), 
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=DAG_TASKS_RESP,
+                      status=200)
+        
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL), 
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=UPDATE_DAG_TASKS_RESP,
```
hanars, line 192: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +170,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=DAG_TASKS_RESP,
+                      status=200)
+        
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL), 
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=UPDATE_DAG_TASKS_RESP, 
+                      status=200)
+        # trigger dag
+        responses.add(responses.POST, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/dagRuns'.format(MOCK_AIRFLOW_URL),
```
hanars, line 195: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
         # Test saving ID file exception
         url = reverse(create_project_from_workspace, args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME2])
         mock_mv_file.side_effect = Exception('Something wrong while moving the ID file.')
+        # Test triggering dag exception
+        responses.replace(responses.PATCH,
```
hanars, line 364: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +360,70 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
         self.assertEqual(response.status_code, 200)
-        mock_api_logger.error.assert_called_with('Uploading sample IDs to Google Storage failed. Errors: Something wrong while moving the ID file.',
-                      self.manager_user, detail=['HG00735', 'NA19675', 'NA19678'])
+        project2 = Project.objects.get(workspace_namespace=TEST_WORKSPACE_NAMESPACE, workspace_name=TEST_NO_PROJECT_WORKSPACE_NAME2)
+        
+        logger_calls = [
+            mock.call('Uploading sample IDs to Google Storage failed. Errors: Something wrong while moving the ID file.',
+                      self.manager_user, detail=['HG00735', 'NA19675', 'NA19678']),
+            mock.call(f'404 Client Error: Not Found for url: {MOCK_AIRFLOW_URL}/api/v1/variables/AnVIL_WES',
```
hanars, line 376: there is extra space at the end of this line that should be removed

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -176,7 +180,14 @@ def create_project_from_workspace(request, namespace, name):
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
                      detail=sample_ids)
 
-    # Send a slack message to the slack channel
+    # use airflow api to trigger AnVIL dags
+    try:
```
hanars, line 184: it would be cleaner if you moved all the error handling code, including the slack message, into the `_trigger_data_loading` function so the control flow here is cleaner

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -176,6 +179,9 @@ def create_project_from_workspace(request, namespace, name):
         logger.error('Uploading sample IDs to Google Storage failed. Errors: {}'.format(str(ee)), request.user,
                      detail=sample_ids)
 
+    # use airflow api to trigger AnVIL dags
+    _trigger_data_loading(project, data_path, request_json['sampleType'], request)
+
```
hanars, line 184: theres extra whitespace here that needs to be removed

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +239,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        """.format(
+            project_guid = project.guid,
+            e = e,
+            dag_name = "seqr_vcf_to_es_AnVIL_{anvil_type}_v{version}".format(anvil_type=sample_type, version=DAG_VERSION),
+            dag = json.dumps(pipeline_dag, indent=4)
+        )
+    safe_post_to_slack(SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, message_content)
+
+def _trigger_data_loading(project, data_path, sample_type, request):
+    try:
```
hanars, line 264: extra whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -45,6 +46,66 @@
 
 TEMP_PATH = '/temp_path/temp_filename'
 
+TEST_GUID=f'P_{TEST_NO_PROJECT_WORKSPACE_NAME}'
+MOCK_TOKEN = 'mock_openid_bearer'
+MOCK_AIRFLOW_URL = 'http://testairflowserver'
+UPDATED_ANVIL_VARIABLES = {
+    "key": "AnVIL_WES",
+    "value": json.dumps({
+        "active_projects": [TEST_GUID],
```
hanars, line 55: extra whitespace at the end of the line for all the values in this dictionary

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -45,6 +46,66 @@
...
+                    "task_id": "scale_dataproc_cluster",
+                },
+                {
+                    "task_id": f"skip_compute_project_subset_{TEST_GUID}",
+                }
+                ],
+            "total_entries": 6
+        }
+
+
```
hanars, line 108: there are way too many extra lines at the end of this test now

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -91,7 +152,10 @@ def test_anvil_workspace_page(self, mock_logger):
         self.assertEqual(response.status_code, 302)
         self.assertEqual(response.url, '/login/google-oauth2?next=/workspace/my-seqr-billing/anvil-1kg%2520project%2520n%25C3%25A5me%2520with%2520uni%25C3%25A7%25C3%25B8de')
         self.mock_get_ws_access_level.assert_not_called()
-
+
```
hanars, line 155: extra whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +169,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
     @mock.patch('seqr.views.apis.anvil_workspace_api.tempfile.NamedTemporaryFile')
+    @responses.activate
     def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_file_iter, mock_file_exist, mock_slack,
                                            mock_send_email, mock_add_service_account,
                                            mock_has_service_account, mock_load_file, mock_api_logger, mock_time,
                                            mock_utils_logger):
+        # Set up api responses
+        # update variables
+        responses.add(responses.PATCH,
+                      '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL),
```
hanars, line 180: extra whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -105,10 +169,33 @@ def test_anvil_workspace_page(self, mock_logger):
...
+                      status=200)
+        # get task id
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL),
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=DAG_TASKS_RESP,
+                      status=200)
+        # get task id again if the response of the previous requset didn't include the updated guid
+        responses.add(responses.GET, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL),
+                      headers={'Authorization': 'Bearer {}'.format(MOCK_TOKEN)},
+                      json=UPDATE_DAG_TASKS_RESP,
```
hanars, line 191: extra whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +359,67 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
         self.assertEqual(response.status_code, 200)
-        mock_api_logger.error.assert_called_with('Uploading sample IDs to Google Storage failed. Errors: Something wrong while moving the ID file.',
-                      self.manager_user, detail=['HG00735', 'NA19675', 'NA19678'])
+        project2 = Project.objects.get(workspace_namespace=TEST_WORKSPACE_NAMESPACE, workspace_name=TEST_NO_PROJECT_WORKSPACE_NAME2)
+        
+        logger_calls = [
+            mock.call('Uploading sample IDs to Google Storage failed. Errors: Something wrong while moving the ID file.',
+                      self.manager_user, detail=['HG00735', 'NA19675', 'NA19678']),
+            mock.call(f'404 Client Error: Not Found for url: {MOCK_AIRFLOW_URL}/api/v1/variables/AnVIL_WES',
```
hanars, line 375: extra whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +359,67 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+    "projects_to_run": [
+        "{guid}"
+    ]
+}}```
+        """.format(
+            guid=project2.guid,
+            airflow_url = MOCK_AIRFLOW_URL
+        )
+                        
+        slack_message_as_normal = """
```
hanars, line 400: this is no longer used

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -45,6 +46,66 @@
 
 TEMP_PATH = '/temp_path/temp_filename'
 
+TEST_GUID=f'P_{TEST_NO_PROJECT_WORKSPACE_NAME}'
+MOCK_TOKEN = 'mock_openid_bearer'
```
hanars, line 50: our security linter is being annoying at this one. Just add the comment `# nosec` to the end oof this line

hanars (changes requested):

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -241,10 +357,48 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+    ]
+}}```
+        """.format(
+            guid=project2.guid,
+            airflow_url = MOCK_AIRFLOW_URL
+        )
+        mock_slack.assert_any_call(SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, slack_message_on_failure)
+        responses.calls.reset()
+        mock_send_email.assert_not_called()
+        self.assertEqual(len(responses.calls), 0)
```
hanars, line 400: we expect this to be 1, move the reset before to post call

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -239,12 +355,53 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+            guid=project2.guid,
+            airflow_url = MOCK_AIRFLOW_URL
+        )
+        mock_slack.assert_any_call(SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, slack_message_on_failure)
+        mock_send_email.assert_not_called()
+        self.assertEqual(len(responses.calls), 1)
+        self.assertEqual(responses.calls[0].request.url, '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL))
+        self.assertEquals(responses.calls[0].request.method, "PATCH")
+        self.assertEqual(responses.calls[0].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+        self.assertEquals(responses.calls[0].response.json(), {'Error': '404 Client Error'})
```
hanars, line 404: this line isn't really necessary - since you configure the fake response in the test and we test for it in the failure message I think its better not to explicitly test the fake response here and only test the the request was sent as expected

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -210,13 +299,40 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
             TEMP_PATH, 'gs://seqr-datasets/v02/GRCh38/AnVIL_WES/{guid}/base/{guid}_ids.txt'.format(guid=project.guid),
             user=self.manager_user
         )
+
+        # Test triggering anvil dags
+        self.assertEqual(len(responses.calls), 4)
+
```
hanars, line 305: remove the extra whitespace

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +238,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        genome_test_type = 'AnVIL_{sample_type}'.format(sample_type=sample_type)
+
+        updated_anvil_variables = _construct_dag_variables(project, data_path, sample_type)
+
```
hanars, line 267: remove the extra whitespace

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +238,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+        _update_variables(genome_test_type, updated_anvil_variables)
+
+        dag_id = "seqr_vcf_to_es_{anvil_type}_v{version}".format(anvil_type=genome_test_type, version=DAG_VERSION)
+        _wait_for_dag_variable_update(dag_id, project)
+
```
hanars, line 272: remove trailing whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -210,13 +299,40 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+        # update variables
+        self.assertEqual(responses.calls[0].request.url, '{}/api/v1/variables/AnVIL_WES'.format(MOCK_AIRFLOW_URL))
+        self.assertEquals(responses.calls[0].request.method, "PATCH")
+        self.assertDictEqual(json.loads(responses.calls[0].request.body), UPDATED_ANVIL_VARIABLES)
+        self.assertEqual(responses.calls[0].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+
```
hanars, line 311: remove trailing whitespace

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -238,7 +238,95 @@ def _send_load_data_slack_msg(project, ids_path, data_path, sample_type, user):
...
+    _make_airflow_api_request(endpoint, method='PATCH', json=json_data)
+
+def _get_task_ids(dag_id):
+    endpoint = 'dags/{}/tasks'.format(dag_id)
+    airflow_response = _make_airflow_api_request(endpoint, method='GET')
+
+    tasks = airflow_response['tasks']
+    task_ids = [task_dict['task_id'] for task_dict in tasks]
+    return task_ids
+
```
hanars, line 310: trailing whitespace

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -210,13 +299,40 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
...
+        self.assertEqual(responses.calls[1].request.url, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL))
+        self.assertEqual(responses.calls[1].request.method, 'GET')
+        self.assertEqual(responses.calls[1].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+        self.assertEqual(responses.calls[1].response.json(), DAG_TASKS_RESP)
+
+        self.assertEqual(responses.calls[2].request.url, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/tasks'.format(MOCK_AIRFLOW_URL))
+        self.assertEqual(responses.calls[2].request.method, 'GET')
+        self.assertEqual(responses.calls[2].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+        self.assertEqual(responses.calls[2].response.json(), UPDATE_DAG_TASKS_RESP)
+
```
hanars, line 322: trailing whitespce

hanars (changes requested):

hanars (approved):

PR #2484, opened 2022-02-24
React 17

hanars:
bumps react and redux to the latest major versions, and updates other related packages as needed. Swaps out the library for setting the page title as the previously used one is mo longer supported


ui/pages/Dashboard/Dashboard.jsx
```diff
@@ -1,11 +1,5 @@
...
 
 import ProjectsTable from './components/ProjectsTable'
 
-export default () => (
-  <div>
-    <DocumentTitle title="seqr: home" />
-    <ProjectsTable />
-  </div>
-)
+export default () => <ProjectsTable />
```
ShifaSZ, line 5: The component doesn't add any new attributes. Can we get rid of it and use `ProjectsTable` directly?

ui/shared/components/page/PageHeader.jsx
```diff
@@ -35,18 +35,28 @@ const NO_HEADER_PAGES = [
...
-  path: `/${page}`,
-  component: () => <SimplePageHeader page={page} {...props} />,
+  path: `/${page}/:subPage?`,
+  component: ({ match }) => <SimplePageHeader page={page} subPage={match.params.subPage} {...props} />,
 }))
 
-const nullComponent = () => null
+const EmptyHeader = ({ match }) => {
+  const page = match.path.split('/').pop()
+  useSeqrTitle(NO_HEADER_PAGE_TITLES[[page]] || page)
```
ShifaSZ, line 54: I don't see why to add square bracket `[]` to `page` when using it as a key.

ShifaSZ (changes requested):

ui/pages/Dashboard/Dashboard.jsx
```diff
@@ -1,11 +1,243 @@
 import React from 'react'
```
hanars, line 1: this is just the moved/renamed ProjectsTable.jsx , nothing actually changed in this file

ShifaSZ (approved):

PR #2492, opened 2022-02-25
Fix excel export for many families

hanars:
Excel has a column limit that we run into when we try to export too many families per variant

### DIFF
--- a/seqr/views/apis/variant_search_api.py
+++ b/seqr/views/apis/variant_search_api.py
@@ -7,6 +7,7 @@ from django.contrib.postgres.aggregates import ArrayAgg
 from django.core.exceptions import MultipleObjectsReturned
 from django.db.utils import IntegrityError
 from django.db.models import Q, prefetch_related_objects
+from math import ceil
 
 from reference_data.models import GENOME_VERSION_GRCh37
 from seqr.models import Project, Family, Individual, SavedVariant, VariantSearch, VariantSearchResults, ProjectCategory, \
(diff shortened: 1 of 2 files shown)


seqr/views/apis/variant_search_api_tests.py
```diff
@@ -312,6 +312,28 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts, mock_erro
              '', '', '', '', '']]
         self.assertEqual(response.content, ('\n'.join(['\t'.join(line) for line in expected_content])+'\n').encode('utf-8'))
 
+        # test export with max families
+        with mock.patch('seqr.views.apis.variant_search_api.MAX_FAMILIES_PER_ROW', 1):
```
ShifaSZ, line 316: It seems too extreme to set `MAX_FAMILIES_PER_ROW` to 1. Can it be 2 or 3?

ShifaSZ (changes requested):

seqr/views/apis/variant_search_api.py
```diff
@@ -265,26 +269,47 @@ def export_variants_handler(request, search_hash):
...
+                continue
+
+            num_split = ceil(len(variant['familyGuids']) / MAX_FAMILIES_PER_ROW)
+            gens_per_row = ceil(len(variant['genotypes']) / num_split)
+            gen_keys = list(variant['genotypes'].keys())
+            for i in range(num_split):
+                split_var = deepcopy(variant)
+                split_var['familyGuids'] = variant['familyGuids'][i*MAX_FAMILIES_PER_ROW:(i+1)*MAX_FAMILIES_PER_ROW]
+                split_gen = set(gen_keys[i*gens_per_row:(i+1)*gens_per_row])
+                split_var['genotypes'] = {k: v for k, v in variant['genotypes'].items() if k in split_gen}
```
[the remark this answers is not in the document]
hanars: No, dict comprehension like this are the better way to do it

seqr/views/apis/variant_search_api_tests.py
hanars: No, there are only max 2 families in the fixture days so setting it to anything over 1 doesn't do anything. Adding new fixture data would be difficult and would break many other tests, and there's no real benefit anyways. Using 1 here does just as good a job testing the behavior as using a bigger number

ShifaSZ (approved):

PR #2509, opened 2022-03-04
Clean up search dataset types

hanars:
This refactors how we handle setting dataset-type specific search filters in seqr. In anticipation of adding the new Mito dataset type, I thought our current approach would be hard to scale. This approach sets the default search panels to be the panels for all data set types- currently that is SNP and SV, but in the future that will also include mito. Then there is a `DATASET_TYPE_PANEL_PROPS` that allows props for certain panels to be overridden based on dataset type. When we add new dataset types in the future, this means there is one clear place to set these props, instead of needing to check eery different panel. 

This also removes the option for SV-only search panels. This is not a case we have, and in the (unlikely) event we ever do end up with a family or 2 that don't have SNP data, I think it is totally fine to show all the SNP filters as well. In fact, I think we should always show the main SNP filter. This means we can greatly reduce the number of panel options when we add mito - instead of adding every possible combination of the 3 available dataset types, we will only add those that include SNPs - so in total "VARIANTS" (current), "SV,VARIANTS" (current default), "MITO,VARIANTS" (new), and "MITO,SV,VARIANTS (new, will be the new default).


ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -169,115 +147,57 @@ const secondaryPanel = panel => ({
...
-    />
-  </Form.Group>
-)
-
-SVFrequecyHeaderFilter.propTypes = {
-  value: PropTypes.object,
-  onChange: PropTypes.func,
-}
-
-const SV_QS_FILTER_FIELD = {
```
hanars, line 209: these were moved to the shared constants, as we want the default panels to have all dataset types

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -169,115 +147,57 @@ const secondaryPanel = panel => ({
...
-    ...secondaryPanel(ANNOTATION_PANEL_MAP[DATASET_TYPE_VARIANT_CALLS]),
-    fieldLayout: annotationFieldLayout([HIGH_IMPACT_GROUPS, MODERATE_IMPACT_GROUPS, CODING_IMPACT_GROUPS]),
-  },
-}
-
-const svCallsetChange = (onChange, initialValues) => val => onChange(
-  { ...initialValues, [SV_CALLSET_FREQUENCY]: val },
-)
-
-const SVFrequecyHeaderFilter = ({ value, onChange }) => (
```
hanars, line 192: this si no longer needed as we no longer support sv-only panels

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -169,115 +147,57 @@ const secondaryPanel = panel => ({
...
+  const typePanels = PANELS.map(panel => ({ ...panel, ...(typePanelProps[panel.name] || {}) }))
   return {
     ...typeAcc,
-    [type]: [true, false].reduce((analystAcc, hasHgmdBool) => {
-      const analystPanels = typePanels.map(
-        ({ hasHgmdPermission, ...panel }) => (hasHgmdPermission === undefined ? panel : hasHgmdPermission[hasHgmdBool]),
-      )
+    [type]: [true, false].reduce((hgmdAcc, hasHgmdBool) => {
+      const hgmdPanels = typePanels.map(panel => (
+        (!hasHgmdBool && panel.name === HGMD_PATHOGENICITY_PANEL) ? { ...panel, ...NO_HGMD_PANEL_PROPS } : panel
```
hanars, line 193: changes how we set the hgmd/non-hgmd panel props so the main `PANELS` list can be a simple list of panels instead of having all these confusing dictionaries in it. Since HGMD is so specific to pathogenicity theres no need to support this as a generalizable configuration for all panels

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -95,23 +93,20 @@ const LeftAligned = styled.div`
 
 const LazyLabeledSlider = props => <React.Suspense fallback={<Loader />}><LabeledSlider {...props} /></React.Suspense>
 
-const JsonSelectPropsWithAll = (options, all) => ({
+export const JsonSelectPropsWithAll = (options, all) => ({
```
ShifaSZ, line 96: The name `JsonSelectPropsWithAll` doesn't make a lot of sense. How about changing to another name like `getPathPanelProps`?

ShifaSZ (changes requested):

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
hanars: this is copy pasted and they are the properties for a select component with json values and an "all" option which is used by several non-pathogenic related components, so I will leave it as is

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -169,115 +147,57 @@ const secondaryPanel = panel => ({
...
+const PANEL_MAP = [ALL_DATASET_TYPE, DATASET_TYPE_VARIANT_CALLS].reduce((typeAcc, type) => {
+  const typePanelProps = DATASET_TYPE_PANEL_PROPS[type] || {}
+  const typePanels = PANELS.map(panel => ({ ...panel, ...(typePanelProps[panel.name] || {}) }))
   return {
     ...typeAcc,
-    [type]: [true, false].reduce((analystAcc, hasHgmdBool) => {
-      const analystPanels = typePanels.map(
-        ({ hasHgmdPermission, ...panel }) => (hasHgmdPermission === undefined ? panel : hasHgmdPermission[hasHgmdBool]),
-      )
+    [type]: [true, false].reduce((hgmdAcc, hasHgmdBool) => {
```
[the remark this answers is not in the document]
hanars: yes working on it now

PR #2511, opened 2022-03-04
Show family analysis group

hanars:


seqr/views/apis/project_api.py
```diff
@@ -191,6 +191,15 @@ def project_individuals(request, project_guid):
         'individualsByGuid': {i['individualGuid']: i for i in individuals},
     })
 
+@login_and_policies_required
+def project_analysis_groups(request, project_guid):
+    project = get_project_and_check_permissions(project_guid, request.user)
+
+    return create_json_response({
+        'projectsByGuid': {project_guid: {'analysisGroupsLoaded': True}},
+        'analysisGroupsByGuid': get_project_analysis_groups([project], project_guid)
```
ShifaSZ, line 200: I'm curious about why a new API is required. Since `analysisGroupsByGuid` should already be loaded with `project_overview` API, why not just add a per-family selector enough?

hanars: because on the family page the project_overview endpoint isn't called, but we still need to display this data. The reducer function works so it checks what has already been loaded, so on the project page, where the overview loaded this data, the endpoint is not called but on the family page, where this data is not loaded, it calls the endpoint

ShifaSZ (approved):

PR #2517, opened 2022-03-07
properly clear search form loading on success

hanars:
small bug introduced in the FormWrapper change, for the search form we want the loading spinner to be explicitly tied to the variants loading and not the internal submitting promise. The only place we explicitly pass a loading prop to the form is for the search form, so this gets the desired behavior


ShifaSZ (approved):
Looks good to me.

PR #2519, opened 2022-03-08
Allow hom alt SNPs + DEL in comp het search

hanars:


ShifaSZ (changes requested):
Mostly looks good to me.

ShifaSZ (approved):

PR #2523, opened 2022-03-08
seqr stat breakup

hanars:
Breaks up seqr stats based on internal projects vs external projects, and breaks apart sample counts by dataset type. No specific ticket, I just got frustrated by not being able to easily find these stats


ShifaSZ (approved):
Looks good to me.

PR #2530, opened 2022-03-09
reset from submission state on external change

hanars:
This fixes the bug where you can't change the search form filters after submitting


ShifaSZ (approved):
Looks good to me.

PR #2532, opened 2022-03-09
add locus list validation

hanars:
This does not necessarily solve the underlying issue of the linked ticket (I am still trying to replicate the bug) but adds a better user experience for adding gene lists and showing if somehow no lists have been selected


ShifaSZ (approved):
Looks good to me.

PR #2537, opened 2022-03-10
Various nodejs fixes for the docker image

sjahl:
Two goals on this one:

1. I wanted to separate the `npm ci` (install) step from the `npm build` step. This should make it so that the docker build won't have to reinstall all node_modules every time something in `ui/` changes. It should only have to do that when either the docker cache has expired, or if the package/package-lock change.
2. Upgrades node to 14.x, and npm to v7 to match our developer laptop env. This appears to fix the issues i mentioned at standup with regards to the lockfile that is being generated with npm v7.

    What's odd, is that the version of NPM we were getting from Debian was already v7, so it was able to handle the new lockfile, even with the older nodejs. The nodejs 14.x that comes from nodesource installs npm v6, which apparently couldn't handle the new lockfile, and is where the bug with trying to clone things over ssh cropped up. The previous workaround for that problem was to configure git to rewrite github URLs to https, but upgrading NPM seems to make that not required anymore (and it's probably the right move if we're using npm 7 on our developer laptops).

    So, long story short, if we want to continue using npm 7 and nodejs 14.x, this should do it.

### DIFF
--- a/deploy/docker/seqr/Dockerfile
+++ b/deploy/docker/seqr/Dockerfile
@@ -9,16 +9,24 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     bzip2 \
     curl \
     gcc \
+    # git needed for the nodejs install
+    git \
     libc6-dev \
     libpq-dev \
-    nodejs \
-    npm \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
 ENV VIRTUAL_ENV=/opt/venv
 RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# install nodejs14 repo since debian is still on node 12.x
+RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash -
+# installs the nodejs and npm packages
+RUN apt-get install -y --no-install-recommends \
+    nodejs \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt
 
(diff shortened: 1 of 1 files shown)


hanars (approved):
looks good. Maybe add a comment explaining the purpose the `COPY ./ui/package.json ./ui/package.json`, as naively it looks like that doesn't do anything at all

sjahl:
> looks good. Maybe add a comment explaining the purpose the `COPY ./ui/package.json ./ui/package.json`, as naively it looks like that doesn't do anything at all

I can actually just be more intelligent about the workdir's -- I'll push a change that switches directories less often, and the copys will be a little less confusing. I'll throw a comment in too.

deploy/docker/seqr/Dockerfile
```diff
@@ -9,33 +9,46 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+# install seqr python dependencies
 RUN pip install --no-cache-dir wheel && \
     pip install --no-cache-dir -r requirements.txt && \
     ln -s python3 /opt/venv/bin/python3.7
 
-WORKDIR /build
+# install seqr nodejs dependencies
+WORKDIR /build/ui
+# Copy the package configs from the build context -> the temporary build container, for generating node_modules
+COPY ./ui/package.json .
```
hanars, line 45: If we are in `/build/ui` wouldn't the file be `.package.json`? Or am I missing something?

sjahl: The two arguments to the `COPY` command are `<source> <destination>`, where `<source>` is the path to the file in the build context (the seqr repo) and `<destination>` is the place inside the container that you want to copy the file to. So we're taking ui/package.json and ui/package-lock.json from [here](https://github.com/broadinstitute/seqr/tree/master/ui), and copying them into `/build/ui` inside the container.

PR #2550, opened 2022-03-15
fix cron

hanars:
Changes to the docker file changed which python is correctly configured, crontab has not been working for a couple weeks. Also, our cron did not like the notation I used for months in one of the commands so I updated it


sjahl (approved):
Whoops! Thanks for catching that. New crontab spec looks good to me.

If it's helpful, https://crontab.guru/ is my go-to bookmark for making cron do the thing

PR #2554, opened 2022-03-15
do not allow null individual id for mme submissions

hanars:


matchmaker/migrations/0004_alter_matchmakersubmission_individual.py
```diff
@@ -0,0 +1,19 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('matchmaker', '0003_auto_20201123_2111'),
+    ]
```
ShifaSZ, line 11: You may want to add a changelog for the need for migration.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #2555, opened 2022-03-15
Add mitochondrial data display (draft, tests haven't been updated).

ShifaSZ:


seqr/utils/elasticsearch/constants.py
```diff
@@ -338,6 +361,12 @@ def get_prediction_response_key(key):
 }
 GENOTYPE_FIELDS_CONFIG.update(BASE_GENOTYPE_FIELDS_CONFIG)
 GENOTYPE_FIELDS_CONFIG.update({field: {} for field in SNP_QUALITY_FIELDS.keys()})
+MITO_GENOTYPE_FIELDS_CONFIG = {
+    'hl': {},
+    'mito_cn': {},
+    'contamination': {},
+}
+GENOTYPE_FIELDS_CONFIG.update(MITO_GENOTYPE_FIELDS_CONFIG)
```
hanars, line 369: we should not be updating the genotype fields used for variants, we should be updating the logic that selects which genotype config to use to choose between the 3 different available data topes: variants, sv, and mito. The gentoypes for mito should be different than the gentoypes for variants, just as the gentoypes for svs is already different

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -451,6 +453,189 @@
...
+      "samples_gq_30_to_35" : [ ],
+      "samples_gq_35_to_40" : [ ],
+      "samples_gq_40_to_45" : [ ],
+      "samples_gq_45_to_50" : [ ],
+      "samples_gq_50_to_55" : [ ],
+      "samples_gq_55_to_60" : [ ],
+      "samples_gq_60_to_65" : [
+        "MCAI043"
+      ],
+      "samples_gq_65_to_70" : [ ],
```
hanars, line 484: empty fields like this should not be exporting to elasticsearch. If they are thats a bug you should fix, and either way you shouldn't include them in the fixture data

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1718,6 +1920,20 @@ def test_sv_wgs_get_es_variants(self):
...
+        setup_responses()
+        search_model = VariantSearch.objects.create(search={})
+        results_model = VariantSearchResults.objects.create(variant_search=search_model)
+        results_model.families.set(self.families)
+
+        variants, _ = get_es_variants(results_model, num_results=2)
+        self.maxDiff = None
+        self.assertListEqual([v for v in variants if v['chrom'] == 'M'], [PARSED_MITO_VARIANT])
+
+        self.assertExecutedSearch(size=4, index=','.join([SV_WGS_INDEX_NAME, MITO_WGS_INDEX_NAME]))
```
hanars, line 1935: this should also be searching in the variants index, not just SV and mito

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -61,6 +61,14 @@ const addDividedLink = (links, name, href) => links.push((
   </span>
 ), <a key={name} href={href} target="_blank" rel="noreferrer">{name}</a>)
 
+const isTrnaOrRrna = (geneNames) => {
```
hanars, line 64: you should be using the `gencodeGeneType` information from the gene to determine this, not the gene name

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -342,6 +357,15 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
       ).map(e => <div key={e}>{e}</div>)]}
       <VerticalSpacer height={5} />
       <VariantLocusListLabels variant={variant} familyGuids={variant.familyGuids} />
+      {variant.highConstraintRegion && (
+        <span>
+          <HorizontalSpacer width={12} />
+          <Popup
+            trigger={<Label color="red" horizontal size="tiny">High Constraint Region</Label>}
+            content="The variant locate in one of the high constraint region"
```
hanars, line 365: there doesn't seem to be any additional information provided in the popup, so I would just make this a plain label

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,22 +182,34 @@ const POPULATIONS = [
     queryParams: { [GENOME_VERSION_37]: 'dataset=gnomad_sv_r2_1' },
     helpMessage: GNOMAD_SV_CRITERIA_MESSAGE,
   },
+  {
+    field: 'gnomad_mito',
+    fieldTitle: 'gnomAD mitochondria',
```
hanars, line 187: use mito instead of the full mitochondria,for this and helix

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,22 +182,34 @@ const POPULATIONS = [
...
 
-  const hasAcPops = POPULATIONS.filter(pop => populations[pop.field] && populations[pop.field].ac)
+  const hasAcPops = POPULATIONS.filter(pop => populations[pop.field] && populations[pop.field].ac !== null)
   const hasGlobalAfPops = POPULATIONS.filter(pop => (
     populations[pop.field] && populations[pop.field].filter_af &&
     (populations[pop.field].filter_af !== populations[pop.field].af)))
   const hasHelpMessagePops = POPULATIONS.filter(
     pop => pop.helpMessage && populations[pop.field] && populations[pop.field].af !== null,
   )
+  const hasAcHetPops = POPULATIONS.filter(pop => populations[pop.field] && populations[pop.field].ac_het !== null)
```
hanars, line 208: also need to check for undefined

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -208,13 +220,26 @@ const Frequencies = React.memo(({ variant }) => {
...
         {hasGlobalAfPops.length > 0 && hasAcPops.length > 0 && <VerticalSpacer height={5} />}
-        {hasAcPops.length > 0 && <Popup.Header content="Allele Counts" />}
+        {(hasAcPops.length > 0 || hasAcHetPops.length > 0 || hasMaxHlPops.length > 0) && <Popup.Header content="Allele Counts" />}
         <Popup.Content>
           {hasAcPops.map(pop => (
             <div key={pop.field}>
               {`${pop.fieldTitle}: ${populations[pop.field].ac} out of ${populations[pop.field].an}`}
             </div>
           ))}
+          {hasAcHetPops.map(pop => (
```
hanars, line 230: these aren't allele counts, you need to add a different header and section for these

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -152,7 +152,13 @@ export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) => (
         content={
           <div>
             <b>Warning: </b>
-            {warning}
+            {warning.split('\n').reduce((acc, message) => (acc ? (
```
hanars, line 155: This is pretty confusing behavior. It looks like you are trying to get the first warning message to display inline with the "Warning:" header and then everything else to display on its own row. I think it would be better to pass in an array and check the type here, and then if its an array show each entry on its own row

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -166,6 +172,12 @@ export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) => (
         <Allele isAlt={genotype.numAlt > (isHemiX ? 0 : 1)} variant={variant} textAlign="right" />
         /
         {isHemiX ? '-' : <Allele isAlt={genotype.numAlt > 0} variant={variant} textAlign="left" />}
+        {genotype.mitoCn && (
+          <div>
+            &nbsp;CN:&nbsp;
+            {genotype.mitoCn.toFixed(0)}
```
hanars, line 178: why would `mitoCn` ever be anything other than an integer? This should be returning as an integer from the pipeline/ elasticsearch, not getting post-processed here

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -232,6 +247,8 @@ const genotypeDetails = (genotype, variant, genesById) => {
   ]
 }
 
+const addWarning = (warning, moreMessage) => (warning ? `${warning}.\n ${moreMessage}` : moreMessage)
```
hanars, line 250: as above, making this into a string so you can split it later is confusing. Just keep it as an array

ui/shared/utils/constants.js
```diff
@@ -1119,6 +1126,10 @@ export const PREDICTOR_FIELDS = [
   { field: 'metasvm', group: MISSENSE_IN_SILICO_GROUP, indicatorMap: INDICATOR_MAP },
   { field: 'gerp_rs', group: MISSENSE_IN_SILICO_GROUP, noSeverity: true, min: -13, max: 7 },
   { field: 'phastcons_100_vert', group: MISSENSE_IN_SILICO_GROUP, noSeverity: true },
+  { field: 'apogeeScore', warningThreshold: 0.5, dangerThreshold: 0.5, fieldTitle: 'APOGEE' },
+  { field: 'hapDefiningVariant', fieldTitle: 'Haplogroup Defining', indicatorMap: { Y: { color: 'green', value: '' } } },
```
hanars, line 1130: you should set the field in the elasticsearch config to return this with the desired display, `fieldTitle` is a last resort for things that need special casing. The other fields do have special casing so will need an explicit field title, but it would still be better for the field to better match the display (i.e. apogee instead of apogeeScore, mitotip instead of mitotipTrnaPrediction, htmvar instead of HmtVarScore)

hanars (changes requested):
One high level concern I have is that you don't seem to be adding a new Sample dataset type for mito variants. This should involve changes to the Sample models, and you should make sure that the index metadata is properly set to  MITO on the index so when we validate the index metadata it is checking it (which also means updating the UI for adding datasets). Its okay if you want to save that work for the search PR, but I just want to make sure it doesn't get missed, and also want to make sure the index metadata is set properly as I'm not sure how you were able to test this without adding the index failing

ShifaSZ:
> and also want to make sure the index metadata is set properly as I'm not sure how you were able to test this without adding the index failing

Good point. The `dataset_type` of the samples is `VARIANT` in my debugging environment. It is not failing when displaying them as SNPs. I can test what happens if I change the field to `MITO`.

hanars:
> The `dataset_type` of the samples is `VARIANT` in my debugging environment. It is not failing when displaying them as SNPs. 

So my concern here is the elastcsearch index metadata should have a MITO dataset type, and seqr should be checking that they match. So I am worried that there is an issue with the pipeline and it is not properly setting the index metadata

> I can test what happens if I change the field to `MITO`.

 Do this as part of the search PR. It is lower priority than working on the SV display changes, and I don't want to block the rest of the PR on it

seqr/utils/elasticsearch/es_utils_tests.py
ShifaSZ: Agree. The empty fields do be exported to ES. I can create a PR to fix the pipeline. But the SNP pipeline will be changed as well.

seqr/utils/elasticsearch/es_utils_tests.py
ShifaSZ: The fixture data for family 14 have only SV and MITO samples.

seqr/utils/elasticsearch/es_utils_tests.py
hanars: The SNP pipeline already doesn't export these fields. Are you sure the mito pipeline does?

seqr/utils/elasticsearch/es_utils_tests.py
hanars: so we should use a different family. Its crucial to test we can search in both

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: How about adding a header "Heteroplasmic statistics" for displaying ac_het/af_het/max_hl?

hanars: I think just "Heteroplasmy" should be fine

ui/shared/utils/constants.js
ShifaSZ: Okay, I can change the backend config.  If we use `apogee` as the response key, is it okay to display it as `Apogee`? How about MitoTIP and HmtVar, can they be displayed as `Mitotip` and `Hmtvar`?

ShifaSZ: It seems that I can use 'mitoTIP' and 'hmtVar' for the response keys.

seqr/utils/elasticsearch/constants.py
ShifaSZ: Updated

ui/shared/components/panel/variants/Annotations.jsx
ShifaSZ: I found that no mainTranscriptId was returned from the backend and the ES index looks normal. It should be a backend issue. But I have to stop here.

ShifaSZ: I change to use the `gencodeGeneType` now. I couldn't see the field because my local reference data DB was bad. I updated the DB. It works now.

ShifaSZ:
The latest screenshot for the mito variant display.
image

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -303,7 +331,9 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
           trigger={<Label horizontal size="mini" content={previousCall.content} color={previousCall.color} />}
         />
       )}
-      {`${genotype.gq || genotype.qs || '-'}${variant.svType ? '' : genotype.numAlt >= 0 && `, ${genotype.ab ? genotype.ab.toPrecision(2) : '-'}`}`}
+      {`${genotype.gq || genotype.qs || '-'}${
+        variant.svType ? '' : genotype.numAlt >= 0 && `, ${genotype.ab ? genotype.ab.toPrecision(2) :
+          `${genotype.hl ? genotype.hl.toPrecision(2) : '-'}`}`}`}
```
hanars, line 336: this is really hard to read now. I think this might be cleaner to do the followig:
```
let secondaryQuality
const showSecondaryQuality = !variant.svType && genotype.numAlt >= 0
if (showSecondaryQuality) {
  secondaryQuality = genotype.ab || genotype.hl
}
...
const content = (
...
  {genotype.gq || genotype.qs || '-'}
  {showSecondaryQuality && `, ${secondaryQuality ? secondaryQuality.toPrecision(2) : '-'}`}
...
)
```

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
```diff
@@ -313,6 +329,13 @@
     'dbnsfp_SIFT_pred': {},
     'StrVCTVRE_score': {'response_key': 'strvctvre'},
 }
+MITO_PREDICTION_FIELDS_CONFIG = {
+    'APOGEE_score': {'response_key': 'apogee'},
+    'hap_defining_variant': {'response_key': 'haplogroup_defining', 'format_value': lambda k: 'Y' if k else None},
```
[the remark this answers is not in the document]
ShifaSZ: It seems that there is a naming rule for the predictions like `PREDICT-SOURCE`_`PREDICT-NAME`. When we don't use `response_key` to rename the field, it will use `PREDICT-NAME` as the default response key. If we want to get rid of `response_key`, we should name the mito prediction fields in the ES index as `mitimpact_apogee`, `hap_haplogroup_defining`, `mitotip_mitoTIP`, and `hmtvar_hmtVar`. The name `mitimpact_apogee` looks reasonable but the rest names are a little weird. What do you think about these field names? @hanars

seqr/utils/elasticsearch/es_search.py
```diff
@@ -780,7 +781,7 @@ def _parse_hit(self, raw_hit):
 
         sorted_transcripts = [
             {_to_camel_case(k): v for k, v in transcript.to_dict().items()}
-            for transcript in hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
+            for transcript in hit.get(SORTED_TRANSCRIPTS_FIELD_KEY, [])
```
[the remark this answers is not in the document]
ShifaSZ: The field still can be undefined if we set the `write_null_values` option to False while writing to the ES index. For the SNP pipeline, the option is set to True. That's why `sortedTranscriptConsequences` is always defined. It seems that we shouldn't write the null fields in the feature.

seqr/views/apis/family_api_tests.py
```diff
@@ -483,6 +483,6 @@ def test_get_family_rna_seq_data(self):
         self.assertDictEqual(response.json(), {
             'M': {
                 'individualData': {'NA19675_1': 8.38},
-                'rdgData': [8.38, 1.01],
+                'rdgData': [1.01, 8.38],
```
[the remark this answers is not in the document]
ShifaSZ: I believe it is an order churn of `RnaSeqTpm.objects.filter()`

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -734,13 +875,22 @@
     'gnomad_svs_ID',
     'bothsides_support',
 ]
-
+POPULATIONS = ['callset', 'exac', 'g1k', 'gnomad_exomes', 'gnomad_genomes', 'gnomad_svs', 'sv_callset', 'topmed']
```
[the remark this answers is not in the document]
ShifaSZ: Okay, I can flatten it.

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -166,6 +170,12 @@ export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) => (
         <Allele isAlt={genotype.numAlt > (isHemiX ? 0 : 1)} variant={variant} textAlign="right" />
         /
         {isHemiX ? '-' : <Allele isAlt={genotype.numAlt > 0} variant={variant} textAlign="left" />}
+        {genotype.mitoCn && (
+          <div>
+            &nbsp;CN:&nbsp;
+            {genotype.mitoCn}
```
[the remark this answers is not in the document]
ShifaSZ: Good point. Basically, the mitochondria fields are updated from SNP variants. So I didn't compare it with the SVs. Also, the mito_cn is the number of the copy of the mitochondrial DNA, not the copy of a variant. Maybe we shouldn't display it similarly to the CN of SVs.

seqr/utils/elasticsearch/constants.py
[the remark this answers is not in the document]
hanars: I think they are fine given that many of our predictor names are weird like that (we have `eigen_Eigen_phred` and `mpc_MPC`). But if you really prefer not to give some of them weird names its okay to use  `response_key` here, its up to you

seqr/utils/elasticsearch/es_search.py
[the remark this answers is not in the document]
hanars: we opted not to make the change about writing null values and declined that PR and will not be making that change in the future, so this should still be defined

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -248,22 +259,32 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
   } else if (isCompoundHet && [individual.maternalGuid, individual.paternalGuid].every(missingParentVariant(variant))) {
-    warning = 'Variant absent in parents'
+    warnings = ['Variant absent in parents']
   }
 
   if (hasCnCall) {
     const cnWarning = getGentoypeCnWarning(genotype, variant.svType, isHemiX)
     if (cnWarning) {
-      warning = warning ? `${warning}. ${cnWarning}` : cnWarning
+      warnings = warnings.concat(cnWarning)
```
hanars, line 274: don't reassign the list. Just use `warnings.push` to update in place and make `warnings` a `const`

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -409,6 +419,12 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
       ).map(e => <div key={e}>{e}</div>)]}
       <VerticalSpacer height={5} />
       <VariantLocusListLabels variant={variant} familyGuids={variant.familyGuids} />
+      {variant.highConstraintRegion && (
```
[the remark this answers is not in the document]
ShifaSZ: Okay. Move it next to `lofDetails`.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -225,7 +256,7 @@ const Frequencies = React.memo(({ variant }) => {
           </Popup.Content>
         ))}
       </Popup>
-    ) : freqContent
+      ) : freqContent
```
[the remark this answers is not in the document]
ShifaSZ: Spaces are added because the position of the left parenthesis is changed for additional conditions.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -216,6 +229,24 @@ const Frequencies = React.memo(({ variant }) => {
             </div>
           ))}
         </Popup.Content>
+        {((hasGlobalAfPops.length > 0 || hasAcPops.length > 0) &&
+            (hasAcHetPops.length > 0 || hasMaxHlPops.length > 0)) && <VerticalSpacer height={5} />}
+        {(hasAcHetPops.length > 0 || hasMaxHlPops.length > 0) && <Popup.Header content="Heteroplasmy" />}
+        <Popup.Content>
+          {hasAcHetPops.map(pop => (
+            <div key={pop.field}>
+              {populations[pop.field].ac_het && `${pop.fieldTitle} heteroplasmic AF:
```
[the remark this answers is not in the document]
ShifaSZ: I want to display both `af_het` and `ac_het` in a format of "<population> heteroplasmic AF: {af_het} AC_het={ac_het} out of {an}" if anyone of these two fields is available. Should it be checking `af_het` or `ac_het`?

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
hanars: WGS SVs show both a copy number and a genotype and they look good. Reuse the code. If reusing the code does not look good, it means you are not displaying it the same way and you need to change the mito display layout to match the SV display layout

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
hanars: if you are checking `ac_het` thats how you define `hasAcHetPops` sos theres no need to check anything. Are there cases where you have ac_het but not af_het? If so you need to handle them. If not, you don't need a conditional

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -143,6 +143,13 @@ const svGenotype = (genotype, isHemiX) => {
   )
 }
 
+const mitoCopyNumber = cn => (
+  <div>
+    &nbsp;
+    { copyNumberGenotype(cn, false)}
```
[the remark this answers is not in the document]
hanars: the ui server is constantly running linting. If you look in the terminal where you ran `npm run start` and you see errors or warnings, linting is failing

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -143,6 +143,13 @@ const svGenotype = (genotype, isHemiX) => {
   )
 }
 
+const mitoCopyNumber = cn => (
```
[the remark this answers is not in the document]
ShifaSZ: I do use the function `copyNumberGenotype` as it in your previous comment. The problem is that the space isn't enough to display both genotype and copy number. As the picture below, the number is out of its parent component and not displayed. I inlined the code for mito copy number and replaced `div` tag with `br` tag for displaying the `cn` in a new line. Do you think it is okay?
image

ui/shared/components/panel/variants/VariantIndividuals.jsx
ShifaSZ: Okay. Also, I found the way I used the function `.concat()` is not correct. I should use a message array for its parameter.

ui/shared/components/panel/variants/VariantIndividuals.jsx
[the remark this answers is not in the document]
hanars: for genome SVs CN displays on another line. Per my previous several comments, you need to match the layout to what we use for SVs

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
ShifaSZ: According to the dataset we have, `ac_het` and `af_het` are always defined and non-zero at the same time. Checking `ac_het` only looks okay.

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
ShifaSZ: Assuming you are talking about displaying `mito_cn`, agree, if I add a `<br />` before `copyNumberGenotype()` both SV and MITO will display in the same way.

ShifaSZ:
Latest mito variant display examples.
image

seqr/utils/elasticsearch/es_search.py
```diff
@@ -807,7 +807,8 @@ def _parse_hit(self, raw_hit):
         })
         return result
 
-    def _parse_genotypes(self, raw_hit, hit, index_family_samples, is_sv):
+    def _parse_genotypes(self, raw_hit, hit, index_family_samples, data_type):
+        is_sv = data_type == Sample.DATASET_TYPE_SV_CALLS
```
hanars, line 811: this doesn't appear to be used

seqr/utils/elasticsearch/es_search.py
```diff
@@ -832,7 +833,7 @@ def _is_matched_sample(family_guid, sample):
                        for sample_id, sample in samples_by_id.items())]
 
         genotypes = {}
-        genotype_fields_config = SV_GENOTYPE_FIELDS_CONFIG if is_sv else GENOTYPE_FIELDS_CONFIG
+        genotype_fields_config = genotype_fields_config = GENOTYPE_FIELDS[data_type]
```
hanars, line 836: this looks like a typo, assigning to the variable twice

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,6 +182,16 @@ const POPULATIONS = [
     queryParams: { [GENOME_VERSION_37]: 'dataset=gnomad_sv_r2_1' },
     helpMessage: GNOMAD_SV_CRITERIA_MESSAGE,
   },
+  {
+    field: 'gnomad_mito',
+    fieldTitle: 'gnomAD mito',
```
hanars, line 187: this should link to gnomad, the same as the regular gnomad v3 variants

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
hanars: Okay, but `hasAcHetPops` already did this check so you shouldn't check it again here

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -76,39 +76,38 @@ const clinvarUrl = (clinvar) => {
...
-            significance={variant.hgmd.class}
-            href={`https://my.qiagendigitalinsights.com/bbp/view/hgmd/pro/mut.php?acc=${variant.hgmd.accession}`}
-            formatName={hgmdName}
-          />
-        </span>
-      )}
+  return pathogenicity.map(([title, linkProps]) => (
+    <span key={title}>
+      <HorizontalSpacer width={5} />
+      {/* eslint-disable-next-line react/jsx-one-expression-per-line */}
```
hanars, line 105: we shouldn't disable linting unless we really need to. You can change the following line to pass linting by making it 
```
<b>{`${title}:`}</b>
```

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -76,39 +76,38 @@ const clinvarUrl = (clinvar) => {
...
-          <PathogenicityLink
-            significance={variant.hgmd.class}
-            href={`https://my.qiagendigitalinsights.com/bbp/view/hgmd/pro/mut.php?acc=${variant.hgmd.accession}`}
-            formatName={hgmdName}
-          />
-        </span>
-      )}
+  return pathogenicity.map(([title, linkProps]) => (
+    <span key={title}>
+      <HorizontalSpacer width={5} />
```
hanars, line 104: for the first link, does this add extra spacing that isn't currently there?

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -166,6 +166,12 @@ export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) => (
         <Allele isAlt={genotype.numAlt > (isHemiX ? 0 : 1)} variant={variant} textAlign="right" />
         /
         {isHemiX ? '-' : <Allele isAlt={genotype.numAlt > 0} variant={variant} textAlign="left" />}
+        {genotype.mitoCn && (
+          <span>
```
hanars, line 170: can we change the `copyNumberGenotype` helper to include the `br` so there is less copy-paste happening?

hanars (changes requested):
Can you please add a changelog entry for this

seqr/utils/elasticsearch/es_search.py
ShifaSZ: We used it twice. But only one is left after we changed `GENOTYPE_FIELDS_CONFIG` to `GENOTYPE_FIELDS[data_type]`. I will remove `is_sv` by using its value directly.

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/Pathogenicity.jsx
ShifaSZ: You are right. There is an extra spacing for Clinvar compared to the existing code. I can remove it by using the mapping index as a condition.

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -121,24 +121,23 @@ Allele.propTypes = {
   variant: PropTypes.object,
 }
 
-const copyNumberGenotype = (cn, isHemiX) => (
+const copyNumberGenotype = (cn, isHemiX, newline) => (isCalled(cn) && (
```
hanars, line 124: In general, if there is a parameter that is not always set it is best for it to be the last. So since we always explicitly set `newline` but not `isHemiX ` its better to swap the order of these so fo rthe mito cn case it would just be `copyNumberGenotype(genotype.mitoCn, true)`

hanars (changes requested):
Adding a changelog entry is blocking approval

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -76,39 +76,37 @@ const clinvarUrl = (clinvar) => {
...
-          <PathogenicityLink
-            significance={variant.hgmd.class}
-            href={`https://my.qiagendigitalinsights.com/bbp/view/hgmd/pro/mut.php?acc=${variant.hgmd.accession}`}
-            formatName={hgmdName}
-          />
-        </span>
-      )}
+  return pathogenicity.map(([title, linkProps], index) => (
+    <span key={title}>
+      {index && <HorizontalSpacer width={5} />}
```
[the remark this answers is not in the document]
ShifaSZ: You are right. Add `!!` before `index`.

hanars (approved):

PR #2566, opened 2022-03-17
Display translocation destination location

ShifaSZ:


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -332,13 +372,14 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
         variant.liftedOverPos ? (
           <div>
             hg19:
-            <UcscBrowserLink variant={variant} useLiftover includeEnd={!!variant.svType || !variant.ref} />
+            <UcscBrowserLink variant={variant} useLiftover includeEnd={!!svType || !variant.ref} />
+            {translocLink(liftedTargetLocus, true)}
```
hanars, line 376: similar to above, but we want the second link to  be shown conditonially on `endChrom` and use the `isEnd` parameter alone with the `useLiftover` parameter. We should also update `UcscBrowserLink` to always take the lifted end coordinates from `rg37LocusEnd` if thats in the variant object

hanars (changes requested):

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -311,7 +350,8 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
       {mainGeneId && <VariantGenes mainGeneId={mainGeneId} showMainGene={showMainGene} variant={variant} />}
       {(mainGeneId && Object.keys(variant.transcripts || {}).length > 1) && <VerticalSpacer height={10} />}
       <LargeText>
-        <b><UcscBrowserLink variant={variant} includeEnd={!!variant.svType} /></b>
+        <b><UcscBrowserLink variant={variant} includeEnd={!!svType} /></b>
+        {translocLink(targetLocus)}
```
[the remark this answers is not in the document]
ShifaSZ: Sounds good.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -824,6 +825,20 @@ def _parse_genome_versions(self, result, index_name, hit):
...
 
+    def _parse_transloc_xstop(self, result, index_name):
+        xstop = result.pop(XSTOP_FIELD, None)
+        if self.index_metadata[index_name]['genomeVersion'] != GENOME_VERSION_GRCh38 or\
+                self.index_metadata[index_name]['datasetType'] != Sample.DATASET_TYPE_SV_CALLS or\
+                self.index_metadata[index_name]['sampleType'] != Sample.SAMPLE_TYPE_WGS:
+            return
+        if xstop:
+            endChrom, end = get_chrom_pos(xstop)
+            if endChrom != result['chrom'] and result['svType'] == TRANSLOCATION_TYPE:
```
ShifaSZ, line 836: Maybe we should remove the condition of `endChrom != result['chrom']` because the translocation can happen in the same chromosome. We may still need to display the start and the target coordination separately when the chromosomes are the same.

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -19,9 +19,15 @@ const LargeText = styled.div`
...
-  const pos = parseInt(useLiftover ? variant.liftedOverPos : variant.pos, 10)
+const TRANSLOCATION_SV = 'BND'
+
+const UcscBrowserLink = ({ variant, useLiftover, includeEnd, isEnd }) => {
+  const variantChrom = isEnd ? variant.endChrom : variant.chrom
+  const variantPos = isEnd ? variant.end : variant.pos
+  const liftedVariantChrom = isEnd ? variant.rg37LocusEnd?.contig : variant.liftedOverChrom
+  const liftedVariantPos = isEnd ? variant.rg37LocusEnd?.position : variant.liftedOverPos
+  const chrom = useLiftover ? liftedVariantChrom : variantChrom
+  const pos = parseInt(useLiftover ? liftedVariantPos : variantPos, 10)
```
hanars, line 30: you end up assigning a lot of variables here, many of which don't get used. I think it would be cleaner to use `let` variables and if/else blocks here. while the code is much longer, its much more explicit to the reader about how the final variablew are getting set
```
let chrom = variant.chrom
let rawPos = variant.pos
if (useLiftover) {
  if (isEnd && rg37LocusEnd) {
    chrom = variant.rg37LocusEnd.contig
    pos = variant.rg37LocusEnd.position
  } else {
    chrom = variant.liftedOverChrom
    rawPos = variant.liftedOverPos
  }
} else if (isEnd) {
  chrom = variant.endChrom || chrom
  rawPos = variant.end
}
const pos = parseInt(rawPos, 10)
```

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -259,7 +269,7 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
           <Transcripts variant={variant} />
         </Modal>
       )}
-      {svType && end && (
+      {svType && end && svType !== TRANSLOCATION_SV && (
```
hanars, line 272: we really do not want to special case translocation where it isn't neccesssary. In this case, what we actually care about is whether `end !== pos`, as we never want to show a size of 0

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -320,7 +330,13 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
       {mainGeneId && <VariantGenes mainGeneId={mainGeneId} showMainGene={showMainGene} variant={variant} />}
       {(mainGeneId && Object.keys(variant.transcripts || {}).length > 1) && <VerticalSpacer height={10} />}
       <LargeText>
-        <b><UcscBrowserLink variant={variant} includeEnd={!!variant.svType} /></b>
+        <b><UcscBrowserLink variant={variant} includeEnd={!!svType && !endChrom} /></b>
+        {endChrom && (
+          <span>
+            ;&nbsp
+            <b><UcscBrowserLink variant={variant} isEnd /></b>
```
hanars, line 337: put this entire section inside the `b` tag above, it is better for the semi colon to be bolded too

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -320,7 +330,13 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
       {mainGeneId && <VariantGenes mainGeneId={mainGeneId} showMainGene={showMainGene} variant={variant} />}
       {(mainGeneId && Object.keys(variant.transcripts || {}).length > 1) && <VerticalSpacer height={10} />}
       <LargeText>
-        <b><UcscBrowserLink variant={variant} includeEnd={!!variant.svType} /></b>
+        <b><UcscBrowserLink variant={variant} includeEnd={!!svType && !endChrom} /></b>
```
hanars, line 333: this is the one place where I think it is appropriate to check if `svType` is a translocation, as you are correct that we want to show translocations differently even if they are within the same chromosome

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -341,13 +357,19 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
         variant.liftedOverPos ? (
           <div>
             hg19:
-            <UcscBrowserLink variant={variant} useLiftover includeEnd={!!variant.svType || !variant.ref} />
+            <UcscBrowserLink variant={variant} useLiftover includeEnd={(!!svType || !variant.ref) && !endChrom} />
+            {endChrom && (
+              <span>
+                ;&nbsp
+                <UcscBrowserLink variant={variant} isEnd />
```
hanars, line 364: this doesn't seem to actually set `useLiftover`

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
```diff
@@ -224,7 +225,7 @@
         '_script': {
             'type': 'number',
             'script': {
-               'source': "(doc.containsKey('svType') && doc['svType'].value == 'BND') ? -50 : doc['start'].value - doc['end'].value"
+               'source': "(doc.containsKey('svType') && doc['svType'].value == TRANSLOCATION_TYPE) ? -50 : doc['start'].value - doc['end'].value"
```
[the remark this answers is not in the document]
ShifaSZ: My bad

seqr/utils/elasticsearch/es_search.py
ShifaSZ: Some of the CPX and INS variants have a `CHR2` which is different from the `locus.contig`. I don't think we should display the second coordination for these variants. Changing the `end` fields with the `END2` values could be an error for these variants. The followings are some examples of these CPX and INS variants. 
image

seqr/utils/elasticsearch/es_search.py
```diff
@@ -704,7 +705,7 @@ def _parse_hit(self, raw_hit):
 
         sorted_transcripts = [
             {_to_camel_case(k): v for k, v in transcript.to_dict().items()}
-            for transcript in hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
+            for transcript in hit.get(SORTED_TRANSCRIPTS_FIELD_KEY, [])
```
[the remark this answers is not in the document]
ShifaSZ: The current SNP pipeline sets `write_null_values` to `True`. So it writes all the null fields such as the empty `samples_gq_X_to_Y` fields to the ES. As we discussed a few days ago, we shouldn't write those fields. Some mito variants have an empty `SORTED_TRANSCRIPTS_FIELD_KEY`. I'm not sure it is normal or not. When I set `write_null_values` to `False`, we get some variants that don't have a `SORTED_TRANSCRIPTS_FIELD_KEY` key. That's why the change is necessary.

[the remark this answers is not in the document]
hanars: As I said when I declined that PR, thats out of scope for this work. As I declined the PR, we should not be generating indices with that code. As such, we should not make this change

seqr/utils/elasticsearch/es_search.py
hanars: I'm happy to have the UI code check if its a translocation, and not display endChrom otherwise. I'm not okay dong this server side. My initial comment here still stands

hanars: for all of these examples it looks like end is the same as start, and end2 is different so I think we absolutely want these to display the same way as translocations

ui/shared/components/panel/variants/Annotations.jsx
ShifaSZ: okay

seqr/utils/elasticsearch/es_search.py
ShifaSZ: My concern for these examples is that the CHR2/END2 don't conform to the definition. The END2 is for the CHR2 per the definition. But If you check the CPX_INTERVALS in the above examples, you can see the END2 values don't look like for CHR2. In the contrast, they look more like they are for the locus.contig. I'm also not sure about the CHR2/END2 for the INS variants. That's why I wanted to exclude the non-translocation types when generating the `endChrom`.

seqr/utils/elasticsearch/es_search.py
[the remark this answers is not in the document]
ShifaSZ: Okay. I have to remove the mito samples from my local env or I can use a different project for mito.

seqr/utils/elasticsearch/es_search.py
hanars: Change the logic so "end" is always coming from xstop, as I asked above. Change the UI logic to appropriately display translocations and complex variants. For these complex variants it looks like we don't want to show either END or END2, given that END is the same value as start. Therefore, the following UI will work and the server side code here will be clean and maintainable as I suggested. 
```
<b>
  <UcscBrowserLink variant={variant} includeEnd={!!svType && !endChrom} />
  {endCrom && svType === TRANSLOCATION_SV && 
    <span>
      ;&nbsp;
      <UcscBrowserLink variant={variant} isEnd />
    </span>
  }
</b>
```

seqr/utils/elasticsearch/es_search.py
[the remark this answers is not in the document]
hanars: the only thing you should be working on between now and when you get back form China is this PR and SV loading. You should not be working on mito. I therefore really don't care what you do to your local environment with regards to mito, but you need to figu it out so it stops blocking this unrelated code

seqr/utils/elasticsearch/es_search.py
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/Annotations.jsx
hanars: actually, you may want to change this to `svType && end && !endChrom` so it displays properly for those complex SVs we were discussing as well. In any case where there is an `endChrom` we probably can't accurately compute a size

seqr/utils/elasticsearch/es_search.py
ShifaSZ: We always display the second coordination for the translocation SVs even if the CHR2 is the same as the locus.contig. Is that right?

hanars: Yes that is correct. So instead of `endCrom && svType === TRANSLOCATION_SV` the conditional for showing the end coordinate as a separate link should just be svType === TRANSLOCATION_SV` . Good catch

seqr/utils/elasticsearch/es_search.py
```diff
@@ -824,6 +825,16 @@ def _parse_genome_versions(self, result, index_name, hit):
             'liftedOverPos': lifted_over_pos,
         })
 
+    def _parse_transloc_xstop(self, result, index_name):
```
hanars, line 828: the name of this function is a bit confusing, I think just `_parse_xstop` is fine

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -19,15 +19,28 @@ const LargeText = styled.div`
...
+    pos = end
+  }
+  pos = parseInt(pos, 10)
   let genomeVersion = useLiftover ? variant.liftedOverGenomeVersion : variant.genomeVersion
   genomeVersion = genomeVersion === GENOME_VERSION_37 ? '19' : genomeVersion
 
   const highlight = variant.ref && `hg${genomeVersion}.chr${chrom}:${pos}-${pos + (variant.ref.length - 1)}`
   const highlightQ = highlight ? `highlight=${highlight}&` : ''
-  const endOffset = variant.end && variant.end - variant.pos
+  const endOffset = !endChrom && end && end - variant.pos
```
hanars, line 43: I'm a little confused about this check. Maybe it would be better to check `includeEnd` here instead

seqr/utils/elasticsearch/es_search.py
```diff
@@ -824,6 +825,16 @@ def _parse_genome_versions(self, result, index_name, hit):
...
+        xstop = result.pop(XSTOP_FIELD, None)
+        if xstop:
+            endChrom, end = get_chrom_pos(xstop)
+            if endChrom != result['chrom'] or end != result['end']:
```
hanars, line 832: endChrom should only be set if it is different from chrom, so this logic does not work. This should be 2 seperate if statements:
```
if end_chrom != result['chrom']:
  result['endChrom'] = end_chrom
if end != result['end']:
  result['end'] = end
```

seqr/utils/elasticsearch/es_search.py
```diff
@@ -824,6 +825,16 @@ def _parse_genome_versions(self, result, index_name, hit):
             'liftedOverPos': lifted_over_pos,
         })
 
+    def _parse_transloc_xstop(self, result, index_name):
+        xstop = result.pop(XSTOP_FIELD, None)
+        if xstop:
+            endChrom, end = get_chrom_pos(xstop)
```
hanars, line 831: python variables should use underscores not snake case, so this should be end_chrom

seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -433,6 +433,8 @@
       'algorithms': ['wham', 'manta'],
       'sc': 7,
       'contig': '2',
+      'rg37_locus': {'contig': '2', 'position': 49272526},
+      'rg37_locus_end': {'contig': '2', 'position': 49273037},
```
hanars, line 437: you should also update `xstop` so you can test the behavior for endChrom

hanars (changes requested):

seqr/utils/elasticsearch/es_search.py
hanars, line 828: `self` is also not used, which means this should be declared as a staticmethod -
 
```
@staticmethod
def _parse_xstop(result, index_name):
```

ui/shared/components/panel/variants/Annotations.jsx
ShifaSZ: We can only compare the `end` and the `pos` when the `endChrom` is undefined. Otherwise, the `end` is for another chromosome. It doesn't make sense to compare them.
The `includeEnd` is checked when using the `endOffset` below. It can also be checked here but I think it also works without changing it.

seqr/utils/elasticsearch/es_search.py
ShifaSZ: The logic `endChrom != result['chrom'] or end != result['end']` doesn't work in the tests. To be honest, I don't know why. I think the chrom:end in `xstop` is different from that in the result only when there is CHR2 and END2 in the pipeline input. So we can use the condition to detect CHR2 and END2.
But, I'll go with your original suggestion and assign the `endChrom` and `end` when the `end_chrom` is different from `result['chrom']`. The only issue is that we can't display the second coordination if the translocation target is in the same chromosome.
I'm hesitating to check and assign `end` separately. It could be wrong for some CPX variants.

ShifaSZ:
There are some CPX variants with valid CHR2 and END2. They are displayed as below.
image
It doesn't show the length. It can be an issue.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -82,6 +82,20 @@ def __init__(self, families, previous_search_results=None, inheritance_search=No
         self._no_sample_filters = False
         self._any_affected_sample_filters = False
 
+    @staticmethod
+    def _parse_xstop(result):
+        xstop = result.pop(XSTOP_FIELD, None)
+        if xstop:
+            end_chrom, end = get_chrom_pos(xstop)
+            if end_chrom != result['chrom'] or end != result['end']:
+                if result['svType'] == 'INS':
```
hanars, line 91: I talked to Alba about this one. It turns out for insertions they use the `CHR2` field to tell you where the sv came from, but we don't want to display the source as the position the way we do with translocations/ complex SVs. I think its out of scope to change this in the pipeline so this seemed like a reasonable workaround for now, but I am open to discussion

seqr/utils/elasticsearch/es_search.py
```diff
@@ -82,6 +82,20 @@ def __init__(self, families, previous_search_results=None, inheritance_search=No
         self._no_sample_filters = False
         self._any_affected_sample_filters = False
 
+    @staticmethod
+    def _parse_xstop(result):
+        xstop = result.pop(XSTOP_FIELD, None)
+        if xstop:
+            end_chrom, end = get_chrom_pos(xstop)
+            if end_chrom != result['chrom'] or end != result['end']:
```
hanars, line 90: checking if either is different means that it is easy to have endChrom defined for all translocations even if they are within the same chromosome. I think it is okay because for translocations we always want to show endChrom, for complex variants END2 is only defined in cases we would want to show it, and for insertions I added the special case handling below anyways

seqr/utils/elasticsearch/es_search.py
```diff
@@ -690,6 +700,11 @@ def _parse_hit(self, raw_hit):
             result['_sort'] = [_parse_es_sort(sort, self._sort[i]) for i, sort in enumerate(raw_hit.meta.sort)]
 
         self._parse_genome_versions(result, index_name, hit)
+        self._parse_xstop(result)
+
+        # If an SV has genotype-specific coordinates that differ from the main coordinates, use those
+        if is_sv and genotypes:
+            self._set_sv_genotype_coords(genotypes, result)
```
hanars, line 707: I moved this to after `_parse_xstop` so we set the variant's end before tying to update it for specific genotypes. I did not see any translocations or complex variants with genotype-specific start/end positions but figured this was the better way to go and it resolved a bunch of test issues

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -19,28 +19,73 @@ const LargeText = styled.div`
   font-size: 1.2em;
 `
 
-const UcscBrowserLink = ({ variant, useLiftover, includeEnd }) => {
-  const chrom = useLiftover ? variant.liftedOverChrom : variant.chrom
-  const pos = parseInt(useLiftover ? variant.liftedOverPos : variant.pos, 10)
-  let genomeVersion = useLiftover ? variant.liftedOverGenomeVersion : variant.genomeVersion
-  genomeVersion = genomeVersion === GENOME_VERSION_37 ? '19' : genomeVersion
+const UcscBrowserLink = ({ genomeVersion, chrom, pos, refLength, endOffset }) => {
```
hanars, line 22: I made the logic for the link a bit simpler so it is passed the specific coordinates it needs, and moved the logic into the new `VariantPosition` component that parses out all the coordinates including the end and then renders the correct link(s)

hanars:
@ShifaSZ I can't add you as a reviewer because you made the PR, but please take a look and let me know what you think. I tried to add helpful comments

seqr/utils/elasticsearch/es_search.py
ShifaSZ: For genome SV, no genotype-specific start/end fields according to the VCF header.

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -117,11 +162,15 @@ const BaseSearchLinks = React.memo(({ variant, mainTranscript, genesById }) => {
 
   const seqrLinkProps = { genomeVersion: variant.genomeVersion, svType: variant.svType }
   if (variant.svType) {
-    seqrLinkProps.location = `${variant.chrom}:${variant.pos}-${variant.end}%20`
+    if (variant.endChrom && variant.endChrom !== variant.chrom) {
+      seqrLinkProps.location = `${variant.chrom}:${variant.pos - 50}-${variant.pos + 50} ${variant.endChrom}:${variant.end - 50}-${variant.end + 50}`
+    } else {
+      seqrLinkProps.location = `${variant.chrom}:${variant.pos}-${variant.end}%20`
```
ShifaSZ, line 168: I'm not very familiar with the location property. Does it require a '%20' at the end? If yes, then a '%20' needs to be added to the end of the first assignment.

seqr/utils/elasticsearch/es_search.py
hanars: okay, but for exome there is and I didn't change the logic I just moved it. Changing/skipping this step for genomes is out of scope

seqr/utils/elasticsearch/es_search.py
hanars: I agree that it is a little complex to add it here, but it needs to display in the UI hover where I have added it and not as a complex interval

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -19,28 +19,73 @@ const LargeText = styled.div`
...
+  refLength: PropTypes.number,
+  endOffset: PropTypes.number,
+}
+
+const VariantPosition = ({ variant, svType, useLiftover }) => {
+  let { chrom, pos, end, endChrom, genomeVersion } = variant
+  const { ref, rg37LocusEnd } = variant
+
+  const showEndLocation = svType && endChrom
+  let endOffset = svType && !showEndLocation && end && end - pos
```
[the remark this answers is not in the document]
hanars: yes but using a shared variable means if we ever want to tweak the logic we only need to do so in one olace. Since there is never a situation where we want to show the full end location and the end offset, making `endOffset` conditional on `showEndLocation` is much cleaner

ui/shared/components/panel/variants/Annotations.jsx
hanars: this is setting the location parameter passed to search if you are searching for all of the "same" variants as this one. The search already only includes variants of the same SV type, and if there is no endChrom it maintains current behavior of returning variants where both their start and end are within 20% of the overall SV size of the respective start and end of this variant. In the case of endChrom, this will now instead search for variants (of the same SV type) where their start position is either within 50bp of the start or end of the variant. For endChrom variants, the %20 is not relevant, as the breakpoints are not within the same chromosome and the size is functionally 0 so using a size percentage is a bad idea. However, thinking about this, for translocations/ cpx variants in multiple chromosomes, I think we actually want to return anything where the start is nearby and not worry about filtering against the end. I will change the logic to just be within a 50bp window of the start position for now, and can work with the analysts later if it is not working well for them

ShifaSZ: That's helpful. I thought the '%20' was a blank char.

ShifaSZ:
I don't see more issues.

hanars (approved):
Approved via discussion with @ShifaSZ

PR #2567, opened 2022-03-17
Add GenCC

hanars:


reference_data/management/commands/utils/update_utils.py
```diff
@@ -85,8 +89,8 @@ def update_records(reference_data_handler, file_path=None):
         with open_file(file_path, open_mode) as f:
             header_fields = reference_data_handler.get_file_header(f)
 
-            for line in tqdm(f, unit=" records"):
-                record = dict(zip(header_fields, line.rstrip('\r\n').split('\t')))
+            for line in reference_data_handler.get_file_iterator(f):
+                record = dict(zip(header_fields, line if isinstance(line, list) else line.rstrip('\r\n').split('\t')))
```
ShifaSZ, line 93: Would it be better to do splitting only when the line is an instance of `str`?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -330,18 +356,8 @@ export const GeneDetails = React.memo((
...
-              <Label size="mini" color={OMIM_SECTION.color} content="OMIM" />
-            </Segment>
-            <Segment color={OMIM_SECTION.color}>
-              {omimDetails}
-            </Segment>
-          </OmimSegments>
-        )
-      )}
+      <br />
+      {getDetailSections(GENE_DISEASE_DETAIL_SECTIONS, gene, compact, labelProps)}
```
ShifaSZ, line 360: I could be an extra blank line if there is neither GenCC nor OMIM.

ShifaSZ (changes requested):
Mostly looks awsome.

reference_data/management/commands/utils/update_utils.py
ShifaSZ: How about this comment?

hanars: no its better this way

ShifaSZ (approved):

PR #2568, opened 2022-03-18
Display both sides support.

ShifaSZ:


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -265,12 +265,21 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene }) => {
...
           <HorizontalSpacer width={5} />
           <Popup
             trigger={<Icon name="help circle" />}
-            content={`Algorithms: ${algorithms}`}
+            content={
+              <div>
+                {algorithms && `Algorithms: ${algorithms}.`}
+                {bothsidesSupport && (
+                  <div>
+                    Variant has read-level support for both sides of breakpoint
```
hanars, line 278: per my comment on the ticket, this should just be "Bothsides Support"

hanars (changes requested):

ui/shared/components/panel/variants/Annotations.jsx
ShifaSZ: Update pushed.

hanars (approved):

PR #2582, opened 2022-03-24
match ui to server permissions for project deletion

hanars:
On the server side, projects can only be deleted by the person who created them, which is the desired behavior. This updates the front end code to show the "delete project" button to the correct users


ShifaSZ (commented):
Looks good.

ShifaSZ (approved):

PR #2599, opened 2022-03-30
Update analysis groups and case review UI

hanars:


ui/redux/rootReducer.js
```diff
@@ -88,7 +88,8 @@ export const updateFamily = (values) => {
   const familyField = values.familyField ? `_${values.familyField}` : ''
   return dispatch => new HttpRequestHelper(`${urlBase}/update${familyField}`,
     (responseJson) => {
-      dispatch({ type: RECEIVE_DATA, updatesById: { familiesByGuid: responseJson } })
+      const updatesById = values.rawResponse ? responseJson : { familiesByGuid: responseJson }
+      dispatch({ type: RECEIVE_DATA, updatesById })
```
ShifaSZ, line 92: It is not so straightforward. How about removing the key `analysisGroupsByGuid` from the response and adding it here? So that the responses for these APIs are more similar.

ShifaSZ (changes requested):
Mostly looks good. Two possible improvements.

ui/redux/rootReducer.js
hanars: no thats much worse, it makes it basically impossible to ever tweak these sorts of responses. The only way to clean this up completely is to change about 7 existing endpoints so they return nested in `familiesByGuid` instead of adding the response key here, but thats way out of scope

ShifaSZ (approved):

PR #2617, opened 2022-04-04
Remove the statically generated assets

sjahl:
This PR removes the statically built assets from our git tree. The intention going forward is that the assets are generated for each commit in the docker build, and that users installing seqr without our docker image will have to run `npm build` in the same fashion that they'd have to run `pip install`. This is the deprecation mentioned in https://github.com/broadinstitute/seqr/discussions/2501

The following is some differences I noticed between the master branch's current assets, and the files generated in a new docker build (these may or may not be the result of differences between running the npm build/etc in the container vs on a mac -- removing the assets shouldn't have any effect itself, since those are ignored from our docker builds anyway):

```
sjahl@wm4a6-69e[2]:~/Desktop
 $ diff master-ui-dist.txt dev-ui-dist.txt
31c31
< 854-3cd6690c.js
---
> 854-2bd18fe3.js
34c34
< app-d2374cc2.js
---
> app-7738cea8.js
sjahl@wm4a6-69e[1]:~/Desktop
 $ diff master-static.txt dev-static.txt
31c31
< static/854-3cd6690c.js
---
> static/854-2bd18fe3.js
34c34
< static/app-d2374cc2.js
---
> static/app-7738cea8.js
66c66,68
< hijack-styles.css
---
> hijack.min.css
> hijack.js
> hijack.scss
```


sjahl:
OK, the tests pass here, so I think this is good to merge if it looks good. Do you think it makes sense to merge this into dev first, or should the merge base be kept at master?

hanars (approved):

PR #2619, opened 2022-04-04
Search util cleanup

hanars:
This cleans up the utility functions/ classes for search. This will help with eventually swapping out the backend, and in the mean time cleans up the code to make behavior/ data flow cleaner. The biggest change is moving the logic for filtering variants to a single `filter_variants` helper in the search class itself, rather than keeping that logic in the utility function. As the logic complicated over time, that logic became more complex/ interconnected and it makes more sense to abstract it out

### DIFF
--- a/seqr/utils/elasticsearch/es_gene_agg_search.py
+++ b/seqr/utils/elasticsearch/es_gene_agg_search.py
@@ -8,9 +8,9 @@ class EsGeneAggSearch(EsSearch):
     AGGREGATION_NAME = 'gene aggregation'
     CACHED_COUNTS_KEY = None
 
-    def search(self, **kwargs):
+    def search(self, *args, **kwargs):
         self._aggregate_by_gene()
-        return super(EsGeneAggSearch, self).search(**kwargs)
+        return super(EsGeneAggSearch, self).search(*args, **kwargs)
 
     def _aggregate_by_gene(self):
         searches = [self._search]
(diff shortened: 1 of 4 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -213,16 +211,87 @@ def sort(self, sort):
...
+        if quality_filter and quality_filter.get('vcf_filter') is not None:
+            self._filter(~Q('exists', field='filters'))
+
+        annotations = {k: v for k, v in (annotations or {}).items() if v}
+        new_svs = bool(annotations.pop(NEW_SV_FIELD, False))
+        splice_ai = annotations.pop(SPLICE_AI_FIELD, None)
+        self._allowed_consequences = sorted({ann for anns in annotations.values() for ann in anns})
+
+        inheritance_mode = (inheritance or {}).get('mode')
+        inheritance_filter = (inheritance or {}).get('filter') or {}
```
ShifaSZ, line 242: Would it better to use `.get('filter', {})`?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_gene_agg_search.py
```diff
@@ -8,7 +8,11 @@ class EsGeneAggSearch(EsSearch):
     AGGREGATION_NAME = 'gene aggregation'
     CACHED_COUNTS_KEY = None
 
-    def aggregate_by_gene(self):
+    def search(self, *args, **kwargs):
+        self._aggregate_by_gene()
+        return super(EsGeneAggSearch, self).search(*args, **kwargs)
```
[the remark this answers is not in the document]
hanars: this is how we use `super` literally everywhere else in the repo so I'm going to leave it as is

seqr/utils/elasticsearch/es_search.py
hanars: no, if filter is included and set to null (which can happen from the ui) then everything would break

ShifaSZ (approved):

Issue #2615, opened 2022-04-04
Long project category names can't be shown in dropdown


david-ma:
Long category names can't be shown in the dropdown on the main dashboard:
<img width="330" alt="Screen Shot 2022-04-04 at 4 44 45 pm" src="https://user-images.githubusercontent.com/3320713/161487906-a7ecfacc-c565-4849-b59b-ba71c0ab9d4b.png">

This can be helped by changing the text to "All project categories", and setting `min-width: 14em;`
<img width="327" alt="Screen Shot 2022-04-04 at 4 47 46 pm" src="https://user-images.githubusercontent.com/3320713/161488267-83f6485d-1bee-41d7-9f98-459c71f3dabb.png">

However, the width of the dropdown is still not dynamic and does not grow with longer names.

A better, but slightly hacky solution is to use a display:table, and set:
```
div.visible.menu.transition {
  display: table !important;
}
```

<img width="403" alt="Screen Shot 2022-04-04 at 4 39 19 pm" src="https://user-images.githubusercontent.com/3320713/161488596-3daae9c2-4461-43be-b904-4f1005310e85.png">

Adding `border-top: solid 0.5px #96c8da !important;` makes it look nicer, but I'm hard-coding the color here. Is there a better/preferred way of doing this?

<img width="389" alt="Screen Shot 2022-04-04 at 4 39 40 pm" src="https://user-images.githubusercontent.com/3320713/161488770-f1767525-49d4-4319-9b4a-5c24255de85e.png">

Note that it will still try to make the dropdown menu as narrow as possible, and break on spaces.

hanars:
This issue is that we are rendering the `Select` component with the `fluid` prop and then trying to set the container around it. The better approach is to change `fluid` to `compact`

david-ma:
Ah yes, I see, the [Fluid Semantic UI Dropdown](https://react.semantic-ui.com/modules/dropdown/) is:
> A dropdown can take the full width of its parent.

The FilterContainer has `min-width: 8em` and doesn't know about the options inside Form Select.

I tried using compact, but that just removes the minimum width & it looks like this:
<img width="368" alt="Screen Shot 2022-04-05 at 12 51 55 pm" src="https://user-images.githubusercontent.com/3320713/161670101-121abd96-d861-40ef-828c-28a61f5723b9.png">

I can't think of a good way to make the dropdown dynamic, besides my `display: table;` hack. And yeah, this is low priority, so I'll forget about that idea.

[...]

PR #2621, opened 2022-04-05
Use longer placeholder text, to make the Project Category dropdown wider

david-ma:
This PR is a simpler version which just uses longer placeholder text, no weird CSS.

Changing the placeholder text to "All project categories" is a natural way to make the select field wider, while also giving users more context on what the field is for.

The increase to 14em vs 8em should be enough to accomodate any long word used in a category title.

For issue #2615


hanars:
Created https://github.com/broadinstitute/seqr/pull/2622 instead, which is a cleaner fix

PR #2628, opened 2022-04-07
Fix static assets for local dvelopment

hanars:
We can no longer uses the compiled `app.html` for local development since it doesn't exist so this instead updates seqr to use use the non-compiled `app.html` template for local development and updates tests accordingly


sjahl (approved):
This looks good to me, I'll merge this in to my branch when I get to it this afternoon, but feel free to merge before that if you'd like the free dopamine hit :)

PR #2635, opened 2022-04-08
fix sv type mapping

hanars:
It turns out wires got crossed at some point and we got the mapping of SV types in seqr wrong. See this slack conversation withe Harrison in which he explains that BNDs are not, in fact, transloactions: https://the-tgg.slack.com/archives/CT1J8AGES/p1649436053460989


ShifaSZ (approved):
Looks good to me.

PR #2638, opened 2022-04-11
fix kubectl copy helper

hanars:
No ticket, but the way the quoting works currently does not actually allow this command to run

### DIFF
--- a/deploy/kubectl_helpers/utils/copy.sh
+++ b/deploy/kubectl_helpers/utils/copy.sh
@@ -11,9 +11,11 @@ TO_POD=$4 # boolean indicating whether to copy a file to the pod or from the pod
 
 POD_NAME=$("${DIR}"/get_pod_name.sh "${DEPLOYMENT_TARGET}" "${COMPONENT}")
 if ${TO_POD}; then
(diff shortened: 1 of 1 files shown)


deploy/kubectl_helpers/utils/copy.sh
```diff
@@ -11,9 +11,11 @@ TO_POD=$4 # boolean indicating whether to copy a file to the pod or from the pod
...
+    DST="${POD_NAME}:."
 else
-    COMMAND="${POD_NAME}:${FILE} ."
+    SRC="${POD_NAME}:${FILE}"
+    DST="."
 fi
 
-kubectl cp "${COMMAND}"
\ No newline at end of file
+kubectl cp "${SRC}" "${DST}"
```
ShifaSZ: I don't see any difference. Can you explain?

hanars: It changes the quoting. The generated command before was `kubectl cp ' source dest'` which does not work and now the generated command is `kubectl cp 'source' 'dest'`

ShifaSZ (approved):

Issue #2637, opened 2022-04-11
Invalid seqr search 


lynnpais:
**Describe the bug**
I am getting the following error when trying to use the 'seqr' link to search for other alleles in the callset: 
Invalid search hash: 6048c9aa99dcd1853d49959855c0f46e

**Link to page(s) where bug is occurring**
Link to variant - https://seqr.broadinstitute.org/project/R0384_rare_genomes_project_gen/saved_variants/variant/SV0109950_5139322029_f019668_r
Page with error - https://seqr.broadinstitute.org/variant_search/results/6048c9aa99dcd1853d49959855c0f46e?page=1&sort=in_omim

**Scope of the bug**
For several variants. I tried deleting cache and cookies but that didn't fix things.

**Screenshots**
<img width="1293" alt="Screen Shot 2022-04-11 at 10 08 02 AM" src="https://user-images.githubusercontent.com/37872186/162757527-e961a956-e63f-48bb-8503-5b48e007db29.png">

lynnpais:
Emily tried too and is getting the same error. 

Having the issue for both the 'Gene Search' and 'seqr' links for ALL variants. 

An example for this is MATR3 - https://seqr.broadinstitute.org/project/R0384_rare_genomes_project_gen/saved_variants/variant/SV0109950_5139322029_f019668_r

<img width="1161" alt="Screen Shot 2022-04-11 at 10 11 18 AM" src="https://user-images.githubusercontent.com/37872186/162758440-c7111826-0046-44c5-a15c-de0f70bdf462.png">

hanars:
My best bet is this is an issue with the local storage in chrome being exceeded, as clearing this on user's browsers fixed the issue. It is hard to understand why all the analysts would have this same error at the exact same time, but its the best theory I have. I will update our local storage handler to a) clear out storage when it gets full and b) log descriptive errors when this happens so its easier to debug issues like that in the future (or this issue, if it keeps happening)

PR #2641, opened 2022-04-11
clear local storage on save error

hanars:
[seqr #2637](https://github.com/broadinstitute/seqr/issues/2637)

### DIFF
--- a/ui/shared/utils/localStorage.js
+++ b/ui/shared/utils/localStorage.js
@@ -8,7 +8,9 @@ export const saveState = (label, state) => {
     const serializedState = JSON.stringify(state)
     localStorage.setItem(label, serializedState)
   } catch (err) {
-    // Ignore write errors.
+    console.warn(err) // eslint-disable-line no-console
+    // clear existing state, as most common reason for failure is out of memory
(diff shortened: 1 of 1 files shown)


ui/shared/utils/localStorage.js
```diff
@@ -8,7 +8,9 @@ export const saveState = (label, state) => {
     const serializedState = JSON.stringify(state)
     localStorage.setItem(label, serializedState)
   } catch (err) {
-    // Ignore write errors.
+    console.warn(err) // eslint-disable-line no-console
```
ShifaSZ, line 11: Since we need the error message from our users and they don't usually use the debugging tool to check the message, why don't we use the `window.alert()` to show the error message to the end users?

ShifaSZ (changes requested):

ui/shared/utils/localStorage.js
hanars: because I do NOT want end users to see this, I just want the ability for me to see it if I need to dive into debugging an issue that this causes

ShifaSZ (approved):
Sounds good.

PR #2643, opened 2022-04-11
Fix all project quality filter search

hanars:


seqr/utils/elasticsearch/es_search.py
```diff
@@ -153,8 +153,12 @@ def _filter_families_for_inheritance(self, inheritance_filter, skipped_sample_co
 
             if not self.samples_by_family_index[index]:
                 del self.samples_by_family_index[index]
+                dataset_type = self._get_index_dataset_type(index)
+                self.indices_by_dataset_type[dataset_type].remove(index)
 
-        if len(self.samples_by_family_index) < 1:
+        self._indices = sorted(list(self.samples_by_family_index.keys()))
```
bw2, line 159: self._indices is set to a sorted list in some places (like here) but not others (like line 171).
Not sure if some places in the code assume it's sorted.

seqr/utils/elasticsearch/es_search.py
```diff
@@ -153,8 +153,12 @@ def _filter_families_for_inheritance(self, inheritance_filter, skipped_sample_co
...
+
+        if len( self._indices) < 1:
```
bw2, line 161:
```suggestion
        if len(self._indices) < 1:
```

bw2 (approved):

seqr/utils/elasticsearch/es_search.py
hanars: I think that was just for testing stability so I think its okay, but good catch

Issue #2630, opened 2022-04-08
Load latest RNA-seq data


bw2:
Please load the latest recent RNA-seq files:

<img width="436" alt="image" src="https://user-images.githubusercontent.com/6240170/162474775-4c282e14-cc3c-4c0d-8a69-68bef0ab57de.png">

hanars:
outlier data is loaded. tpm data is a new format so still working on that

PR #2645, opened 2022-04-11
support loading RNA TPM with metadata

hanars:
The file format for the delivered RNA-seq TPM data has changed substantively enough that it made sense to add a new manage command for handling it, rather than trying to adapt the existing command. While we could probably ask to receive the data with the old format, this new format is better as the tissue type is being provided more reliably so we want to support this new format

[seqr #2630](https://github.com/broadinstitute/seqr/issues/2630)


seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -62,51 +107,62 @@ def test_command(self, mock_gzip_open, mock_open, mock_logger):
...
+
+        self._assert_expected_command_results(mock_logger)
 
         # Test fails on mismatched tissue
-        mock_open.return_value.__enter__.return_value.__iter__.return_value[2] = 'NA19678_D1\tNA19678\tfibroblasts'
-        call_command('load_rna_seq_tpm', 'new_file.tsv.gz', MAPPING_FILE_ID, '--ignore-extra-samples')
-        mock_logger.warning.assert_called_with('Skipped data loading for the following 1 samples due to mismatched tissue type: NA19678_D1 (fibroblasts to whole_blood)')
+        mock_gzip_file.__iter__.return_value[5] = 'NA19678_D1\tENSG00000233750\t6.04\tfibroblasts\n'
+        call_command('load_rna_seq_tpm_with_metadata', 'new_file.tsv.gz', '--ignore-extra-samples')
+        mock_logger.warning.assert_called_with(EXPECTED_MISMATCHED_TISSUE_WARNING)
```
ShifaSZ, line 165: It should need a `mock.logger.reset_mock()` before calling the command.

ShifaSZ (changes requested):

seqr/management/commands/load_rna_seq_tpm_with_metadata.py
```diff
@@ -0,0 +1,79 @@
...
+                invalid_tissues[sample] = tissue_type
+                continue
+
+            models = RnaSeqTpm.objects.bulk_create(
+                [RnaSeqTpm(sample=sample, **data) for data in data_by_gene.values()], batch_size=1000)
+            logger.info(f'create {len(models)} RnaSeqTpm for {sample.sample_id}')
+
+        if invalid_tissues:
+            message = ', '.join([
+                f'{sample.sample_id} ({REVERSE_TISSUE_TYPE[expected_tissue]} to {REVERSE_TISSUE_TYPE[sample.tissue_type]})'
```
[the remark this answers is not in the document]
hanars: this is a superuser run manage command. In the unlikely event that we get malformed data, its okay to have unfriendly error messages

seqr/management/commands/load_rna_seq_tpm_with_metadata.py
```diff
@@ -0,0 +1,79 @@
...
+    def _parse_row(self, row):
+        sample_id = row[SAMPLE_ID_COL]
+        if row[TPM_COL] != '0.0' and not sample_id.startswith('GTEX'):
+            prev_tissue = self.sample_id_to_tissue_type.get(sample_id)
+            tissue = row[TISSUE_COL]
+            if not tissue:
+                raise ValueError(f'Sample {sample_id} has no tissue type')
+            if prev_tissue and prev_tissue != tissue:
+                raise ValueError(f'Mismatched tissue types for sample {sample_id}: {prev_tissue}, {tissue}')
+            self.sample_id_to_tissue_type[sample_id] = tissue
```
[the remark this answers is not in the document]
hanars: not really, I think thats less readable. Having things error out earlier is generally better than later, and having less nesting is more readable than more nesting

seqr/management/tests/load_rna_seq_tpm_tests.py
hanars: we never had that for the existing test in this file, but I'll add a test that warning is explicitly not called during a successful run

hanars: No it doesn't. The test has an explicit `mock_logger.warning.assert_not_called()` so we know that on the success run this was not called and therefore this is only testing the failure run

ShifaSZ (approved):

PR #2647, opened 2022-04-12
Update download_reference_data.sh

Abuhaliqa:
Fixed Space typo that would prevent the download of Clinvar data at line 30


hanars (approved):
thank you for fixing this

PR #2654, opened 2022-04-14
Add Clingen info

hanars:


ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -180,6 +193,20 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+    color: 'purple',
+    description: 'ClinGen Dosage Sensitivity',
+    label: 'ClinGen',
+    showDetails: gene => gene.clinGen,
+    detailsDisplay: gene => (
+      <NoBorderTable basic="very" compact="very">
+        {gene.clinGen.haploinsufficiency &&
+          <ClinGenRow value={gene.clinGen.haploinsufficiency} href={gene.clinGen.href} label="Haploinsufficiency" />}
+        {gene.clinGen.triplosensitivity &&
+          <ClinGenRow value={gene.clinGen.triplosensitivity} href={gene.clinGen.href} label="Triplosensitivity" />}
```
ShifaSZ, line 206: Does it need protection when `gene.clinGen` is undefined?

ShifaSZ (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
hanars: No the `showDetails` condition makes sure this only renders when it is defined

ShifaSZ (approved):

Issue #2656, opened 2022-04-19
Clean up gene/variant location search logic


hanars:
Currently, searching by both gene/interval and variants is supported in the seqr location filter, behaving as an OR filter.  This is harder behavior to replicate on the hail backend, and is also not particularly useful (and according to log data is never really used). We will update the behavior of these filters as follows:
- User can only input either genes/interval or variants, but not both. If a user enters genes or gene lists the variant search box becomes greyed out, and similarly if they enter anything into the variant search box the gene search becomes greyed out.
- On the backend, variant filtration is only performed when no genes/ intervals are specified
- The "Exclude Locations" checkbox is only available for genes and not variants search. It is appropriately greyed out in the UI and only applied for gene filters on the backend
- Variants must be provided in a consistant format - a user can specify either Variant IDs or RSIDs but not both. This is enforced on the backend

PR #2659, opened 2022-04-20
Panel app gene search cleanup

hanars:
This is a precursor Pr for https://github.com/broadinstitute/seqr/issues/2656 . It cleans up the way the panel app gene lists are handled in the location panel in preparation for other changes to that panel. It has no changes to display/ behavior/ performance, it just makes the code easier to work with going forward

Change summary:
Currently, panel app gene lists are added to the same form field as regular lists, but they are formatted a a dictionary instead of a string, and the `LocusListItemsFilter` has conditional logic to render it based on type, and then before the form is submitted it post processes the object into a string. This changes the form to use a separate field for panel app list data, uses a decorator to keep the string and object values in sync to remove the need for post processing before submission, and can therefore make the panel app fields themselves top-level fields which cleans up the logic for `LocusListItemsFilter` so it is a simple field type. `LocusListItemsFilter` also now has a `FormSpy` so the filters can be shown or hidden based on other form field values

### DIFF
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -27,11 +27,10 @@ class BaseLocusListDropdown extends React.Component {
 
     if (prevProps.locusList.rawItems !== locusList.rawItems) {
       const { locusListGuid } = locusList
-      let panelAppItems
 
       if (locusList.paLocusList) {
         const CONFIDENCE_COLORS = { 0: 'none', 1: 'red', 2: 'amber', 3: 'green', 4: 'green' } // TODO constant
-        panelAppItems = locusList.items?.reduce((acc, item) => {
+        const panelAppItems = locusList.items?.reduce((acc, item) => {
           const color = CONFIDENCE_COLORS[item.pagene?.confidenceLevel || 0]
           if (color in acc) {
             acc[color] = [acc[color], item.display].filter(val => val).join(', ')
--- a/ui/pages/Search/components/filters/LocusListSelector.jsx
+++ b/ui/pages/Search/components/filters/LocusListSelector.jsx
@@ -27,11 +27,11 @@ class BaseLocusListDropdown extends React.Component {
 
     if (prevProps.locusList.rawItems !== locusList.rawItems) {
       const { locusListGuid } = locusList
-      let { rawItems } = locusList
+      let panelAppItems
 
       if (locusList.paLocusList) {
-        const CONFIDENCE_COLORS = { 0: 'none', 1: 'red', 2: 'amber', 3: 'green', 4: 'green' }
-        rawItems = locusList.items?.reduce((acc, item) => {
+        const CONFIDENCE_COLORS = { 0: 'none', 1: 'red', 2: 'amber', 3: 'green', 4: 'green' } // TODO constant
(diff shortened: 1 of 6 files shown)


ui/shared/components/panel/search/constants.js
```diff
@@ -349,15 +350,30 @@ const SV_FREQUENCIES = [
 
 export const FREQUENCIES = [...SNP_FREQUENCIES, ...SV_FREQUENCIES]
 
+export const LOCUS_FIELD_NAME = 'locus'
+export const PANEL_APP_FIELD_NAME = 'panelAppItems'
+const PANEL_APP_COLORS = [...new Set(
+  Object.entries(PANEL_APP_CONFIDENCE_LEVELS).sort((a, b) => b[0] - a[0]).map(config => config[1]),
+)]
```
ShifaSZ, line 357: Will be the sorting useful when creating a set?

ui/pages/Search/components/filters/LocusListSelector.jsx
```diff
@@ -27,21 +28,17 @@ class BaseLocusListDropdown extends React.Component {
...
-          return acc
-        }, { green: '', amber: '', red: '', none: '' })
+        const panelAppItems = locusList.items?.reduce((acc, item) => {
+          const color = PANEL_APP_CONFIDENCE_LEVELS[item.pagene?.confidenceLevel] || PANEL_APP_CONFIDENCE_LEVELS[0]
+          return { ...acc, [color]: [acc[color], item.display].filter(val => val).join(', ') }
+        }, {})
+        onChange({ locusListGuid, panelAppItems })
+      } else {
+        const { rawItems } = locusList
+        onChange({ locusListGuid, rawItems })
```
ShifaSZ, line 40: Would it be better by changing to `onChange({ locusListGuid, locusList.rawItems })` ?

ShifaSZ (changes requested):
Mostly looks good to me except for some nitpicking issues.

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -1,87 +1,41 @@
...
-    size="small"
-    position="top center"
-  />
+import { PANEL_APP_CONFIDENCE_LEVELS, PANEL_APP_CONFIDENCE_DESCRIPTION, PANEL_APP_CONFIDENCE_LEVEL_COLORS } from 'shared/utils/constants'
+
+const PA_ICON_PROPS = Object.entries(PANEL_APP_CONFIDENCE_LEVELS).reduce((acc, [confidence, colorKey]) => {
+  const color = PANEL_APP_CONFIDENCE_LEVEL_COLORS[confidence]
+  return color ?
+    { ...acc, [colorKey]: { name: 'circle', color, title: PANEL_APP_CONFIDENCE_DESCRIPTION[confidence] } } : acc
+}, {})
```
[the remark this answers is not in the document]
hanars: when `confidence` is 0 `PANEL_APP_CONFIDENCE_LEVEL_COLORS[confidence]` is null, which is the desired behavior as there is no icon for that panel

ui/pages/Search/components/filters/LocusListSelector.jsx
hanars: no

ShifaSZ (approved):

PR #2660, opened 2022-04-21
Location search updates

hanars:
Updates interaction between gene and variant location filters to be easier to support going forward per https://github.com/broadinstitute/seqr/issues/2656


seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1716,6 +1717,7 @@ def test_sv_wgs_get_es_variants(self):
         self.assertListEqual(variants, [PARSED_SV_WGS_VARIANT])
 
         self.assertExecutedSearch(filters=[
+            {'terms': {'rsid': ['rs9876']}},
```
ShifaSZ, line 1720: Does it also need a separated test for the `variant_ids`?

ui/shared/components/panel/search/LocusListItemsFilter.jsx
```diff
@@ -14,28 +14,34 @@ const PA_ICON_PROPS = Object.entries(PANEL_APP_CONFIDENCE_LEVELS).reduce((acc, [
...
+        disabled: shouldDisable(values?.search?.locus || {}),
+        ...props,
+      },
     )}
   </FormSpy>
 )
 
 LocusListItemsFilter.propTypes = {
-  label: PropTypes.string,
+  label: PropTypes.node,
```
ShifaSZ, line 40: Does the `label` need to be a `node`? What I can find in this PR is that all the `labels` are strings.

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_search.py
```diff
@@ -222,13 +222,16 @@ def _filter(self, new_filter):
...
         self._filter_custom(custom_query)
 
         if has_location_filter:
-            self.filter_by_location(
-                genes=genes, intervals=intervals, rs_ids=rs_ids, variant_ids=variant_ids, locus=locus)
+            self._filter(_location_filter(genes, intervals, locus))
+        elif variant_ids:
+            self.filter_by_variant_ids(variant_ids, locus=locus)
+        elif rs_ids:
+            self._filter(Q('terms', rsid=rs_ids))
```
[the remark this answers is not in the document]
hanars: the front end completely prevents users from entering variants if genes are entered and vice versa, so its m actually completely enforced not just a hint

seqr/utils/elasticsearch/es_utils_tests.py
hanars: no I checked and we are already testing that behavior  in other existing tests, but good catch

ui/shared/components/panel/search/LocusListItemsFilter.jsx
hanars: `node` is a prop type that includes any valid react child node, including plain strings and jsx. Before, all the labels were plain strings so this prop type made sense, but the label for variant search box which is now using this component has jsx in it so we need to change the prop type to support that

ShifaSZ (approved):

PR #2687, opened 2022-04-27
Variant rna seq performance

hanars:
Fetching the RNA TPM data for all the saved variants for large projects takes about a full minute. Rather than fetching the data with the variants, we should instead just indicate if a family has data or not, and then fetch it dynamically when the user actually wants to view the data

### DIFF
--- a/ui/shared/components/panel/variants/RnaSeqTpm.jsx
+++ b/ui/shared/components/panel/variants/RnaSeqTpm.jsx
@@ -38,39 +38,40 @@ const PLOT_OPTIONS = {
   yAxisUnit: 'TPM',
 }
 
-const launchGtex = (geneId, tpms) => (gencodeId) => {
-  const tissues = [...new Set(Object.values(tpms).map(({ sampleTissueType }) => sampleTissueType))]
-  const boxplotData = Object.entries(tpms).map(([individual, { tpm }]) => ({ label: individual, data: [tpm] }))
+const launchGtex = (geneId, familyGuid) => (gencodeId) => {
+  new HttpRequestHelper(`/api/family/${familyGuid}/rna_seq_data/${geneId}`,
+    (responseJson) => {
+      const boxplotData = Object.entries(responseJson).reduce((acc, [tissue, { rdgData, individualData }]) => ([
+        ...acc,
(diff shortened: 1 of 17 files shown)


ui/shared/components/panel/variants/RnaSeqTpm.jsx
```diff
@@ -38,39 +38,40 @@ const PLOT_OPTIONS = {
   yAxisUnit: 'TPM',
 }
 
-const launchGtex = (geneId, tpms) => (gencodeId) => {
-  const tissues = [...new Set(Object.values(tpms).map(({ sampleTissueType }) => sampleTissueType))]
-  const boxplotData = Object.entries(tpms).map(([individual, { tpm }]) => ({ label: individual, data: [tpm] }))
+const launchGtex = (geneId, familyGuid) => (gencodeId) => {
```
hanars, line 41: the control flow changes here - before we already had the sample specific data so we formatted that data and fetched the gtex and rdg aggregate data. Now, we first do one request to fetch the sample and rdg aggregate data, and then do a followup request to gtex for the tissues that are returned from the API

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -508,12 +509,14 @@ BaseVariantGene.propTypes = {
...
   compoundHetToggle: PropTypes.func,
+  hasRnaTpmData: PropTypes.bool,
   rnaSeqData: PropTypes.object,
 }
 
 const mapStateToProps = (state, ownProps) => ({
   gene: getGenesById(state)[ownProps.geneId],
-  rnaSeqData: getRnaSeqDataByFamilyGene(state)[ownProps.variant.familyGuids[0]],
+  hasRnaTpmData: getFamiliesByGuid(state)[ownProps.variant.familyGuids[0]]?.hasRnaTpmData,
+  rnaSeqData: getRnaSeqOutilerDataByFamilyGene(state)[ownProps.variant.familyGuids[0]],
```
ShifaSZ, line 519: Do you also want to change the `rnaSeqData` to `rnaSeqOutilerData`?

ShifaSZ (changes requested):
Good PR. I only have two naming comments.

ShifaSZ (approved):

PR #2701, opened 2022-05-04
Fix es pv creation

hanars:
When we moved the configuration for disks from servctl to terraform, it ended up breaking the code for creating persistant volumes for the disks in kubernetes, which in turn breaks our elasticsearch deployment. This updates the pv code to instead create one pv for every data pod. It also fixes a bug in the command error handling that caused commands which raise multiple errors to be ignored if one of the errors is ignorable, instead of checking each error, which caused the deployment to proceed without a pv when that failed

### DIFF
--- a/deploy/servctl_utils/shell_utils.py
+++ b/deploy/servctl_utils/shell_utils.py
@@ -74,8 +74,18 @@ def run(command,
     p.wait()
 
     output = log_buffer.getvalue()
+
     if p.returncode not in ok_return_codes:
-        if ignore_all_errors or (errors_to_ignore and any([error_to_ignore in str(output) for error_to_ignore in errors_to_ignore])):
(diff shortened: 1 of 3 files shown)


deploy/servctl_utils/shell_utils.py
```diff
@@ -74,8 +74,18 @@ def run(command,
     p.wait()
 
     output = log_buffer.getvalue()
+
     if p.returncode not in ok_return_codes:
-        if ignore_all_errors or (errors_to_ignore and any([error_to_ignore in str(output) for error_to_ignore in errors_to_ignore])):
+        should_ignore = False
+        if ignore_all_errors:
```
sjahl, line 80: love it 😄

sjahl (approved):
Looks good to me 👍

Issue #2710, opened 2022-05-10
ClinVar LP/P filter overriding other filters 


Glemiret:
**Describe the bug**
Here is the question: should the ClinVar LP/P filter override any other filters when doing a seqr search? Will a non coding transcript exon variant be returned by a seqr search if it has a ClinVar LP entry even if "non coding transcript exon" is not selected in the annotation section? Anne and I thought so. 
In RGP_248, 2 non coding variants (1 is ClinVar LP) in trans in RNU4ATAC are not returned by seqr recessive searches. Is this a bug in seqr or is this normal behavior? The variant that has a LP entry in ClinVar is not returned if I do a search with just ClinVar LP/P selected as a filter (and nothing else). 

These 2 variants are the answer for this case (RNA seq profile gave us the answer) but these 2 variants were missed by 3 different analysts who analyzed this case in the past. 

I look forward to discuss this case with you and see what you think, thanks!

**Link to page(s) where bug is occurring**
https://seqr.broadinstitute.org/project/R0384_rare_genomes_project_gen/saved_variants/family/F020329_rgp_248/Analyst%20high%20priority

**Scope of the bug**
Just this project

**Screenshots**

![Screen Shot 2022-05-10 at 9 18 32 AM](https://user-images.githubusercontent.com/84086816/167637715-3732cb76-638a-4d09-b599-84f789e02df5.png)

Caustint:
Chiming in as I'm thinking through this... In the example above both variants are in non coding transcripts, but only one of them is  pathogenic in ClinVar. So you'd want the override behavior to be as follows: seqr should return all variant pairs containing _at least one_ P/LP variant, regardless of the annotations for the second hit (ie the presence of one ClinVar P/LP should override the annotation filters for both variants in the pair). Is that right?

Glemiret:
@Caustint  That would be next step, to discuss as a group. My question for this ticket was: is this a bug or is this current seqr behavior?
Before discussing recessive searches, the variant that has a LP entry in ClinVar is **not** returned if I do a search with just ClinVar LP/P selected as a filter (and nothing else). Is this variant not recognized as having a ClinVar entry? Is this a bug?

hanars:
[...]

Issue #229, opened 2017-04-04
For each variant in search results, show all the gene names whose VEP-annotated consequence matches one of the consequences selected by the user in variant filters.


bw2:
Currently, all coding genes are shown. Instead, select gene names to show based on whether the variant effect for that gene is among those selected by the user.

hanars:
Showing all the genes will be resolved by https://github.com/macarthur-lab/seqr/issues/646 

Leaving this open because we would want to highlight all the genes whose consequence was filtered on, and currently we only highlight the gene for the worst annotation

hanars:
Should consider this while working on https://github.com/macarthur-lab/seqr-private/issues/421

hanars:
also, within a gene should show a consequence that was sorted on and not neccessarily the main transcript

hanars:
RNU4ATAC is a good example of this - its always the secondary gene because its non-coding even though the main gene is always intronic

PR #2713, opened 2022-05-10
select main transcript that has filtered annotation

hanars:
Implements the final outstanding feature from https://github.com/broadinstitute/seqr/issues/229 to prevent the confusion as caused in https://github.com/broadinstitute/seqr/issues/2710

Sets a selected main transcript for variants where the worst transcript was not actually one of the filtered transcripts from search. The UI for this already exists for saved variants, so no UI code change is needed to properly display the selection

### DIFF
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -776,11 +776,19 @@ class EsSearch(object):
 
         main_transcript_id = sorted_transcripts[0]['transcriptId'] \
             if len(sorted_transcripts) and 'transcriptRank' in sorted_transcripts[0] else None
+        selected_main_transcript_id = None
+        if self._allowed_consequences and sorted_transcripts[0].get('majorConsequence') not in self._allowed_consequences:
+            selected_main_transcript_id = next((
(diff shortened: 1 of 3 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -776,11 +776,19 @@ def _parse_hit(self, raw_hit):
 
         main_transcript_id = sorted_transcripts[0]['transcriptId'] \
             if len(sorted_transcripts) and 'transcriptRank' in sorted_transcripts[0] else None
+        selected_main_transcript_id = None
+        if self._allowed_consequences and sorted_transcripts[0].get('majorConsequence') not in self._allowed_consequences:
```
bw2, line 780: Is sorted_transcripts always guaranteed to have length > 0?
Looking at line 766 it's hard to understand the order of operations with for/in/or 
```
sorted_transcripts = [
            for transcript in hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
        ]
```

Will this sorted_transcripts = [[]] sometimes? leading to sorted_transcripts[0] == [] instead of {}?

bw2 (approved):

seqr/utils/elasticsearch/es_search.py
hanars: good catch

PR #2723, opened 2022-05-17
PanelApp Mode of Inheritance

david-ma:
For discussion #2700

Show PanelApp Mode of Inheritance in the Genelist Label on the variant_search page, with design as discussed in #2700:

Screen Shot 2022-05-18 at 12 33 30 am

* The freetext MOI from PanelApp is translated into an array of enums, which can be shown as initials on the label if they are of the 4 most common types: AD (Autosomal Dominant/Monoallelic), AR (Autosomal Recessive/Biallelic), XD (X-Linked Dominant), XR (X-Linked Recessive).
* The freetext MOI has been added to the pop-over. And if initials are used, the initials are shown next to the MOI.
* The max-width of the labels has been widened from 7em to 12em.
* The PanelApp link has been added to the pop-over.
* The URL API origin of the PanelApp genelist and the panelApp ID are provided with the locusList, so this link can be generated in the front end.
* PanelApp confidence has been moved to PaAttrs on the gene, along with the new moi attribute.
* Unit tests have been written for the new apis and utility functions, in python and jest.

### DIFF
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -147,11 +147,6 @@ def _add_discovery_tags(variants, discovery_tags):
 
 def _add_pa_attrs(locus_list_gene, locus_list_guid, gene_json):
     if hasattr(locus_list_gene, 'palocuslistgene'):
-        # Keeping locusListConfidence for backwards compatibility for now
-        if not gene_json.get('locusListConfidence'):
-            gene_json['locusListConfidence'] = {}
-        gene_json['locusListConfidence'][locus_list_guid] = locus_list_gene.palocuslistgene.confidence_level
-
         if not gene_json.get('locusListPaAttrs'):
             gene_json['locusListPaAttrs'] = {}
         gene_json['locusListPaAttrs'][locus_list_guid] = {
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -105,14 +105,11 @@ def _add_locus_lists(projects, genes, add_list_detail=False, user=None, is_analy
     for interval in _get_json_for_models(intervals, nested_fields=[{'fields': ('locus_list', 'guid')}]):
         locus_lists_by_guid[interval['locusListGuid']]['intervals'].append(interval)
 
-    for locus_list_gene in LocusListGene.objects.filter(locus_list__in=locus_lists, gene_id__in=genes.keys()).prefetch_related('locus_list', 'palocuslistgene'):
+    for locus_list_gene in LocusListGene.objects.filter(locus_list__in=locus_lists, gene_id__in=genes.keys()).prefetch_related('locus_list', 'locus_list__palocuslist', 'palocuslistgene'):
         gene_json = genes[locus_list_gene.gene_id]
         locus_list_guid = locus_list_gene.locus_list.guid
         gene_json['locusListGuids'].append(locus_list_guid)
-        if hasattr(locus_list_gene, 'palocuslistgene'):
-            if not gene_json.get('locusListConfidence'):
-                gene_json['locusListConfidence'] = {}
-            gene_json['locusListConfidence'][locus_list_guid] = locus_list_gene.palocuslistgene.confidence_level
+        _add_pa_attrs(locus_list_gene, locus_list_guid, gene_json)
 
     return locus_lists_by_guid
 
@@ -146,6 +143,21 @@ def _add_discovery_tags(variants, discovery_tags):
             if not variant.get('discoveryTags'):
                 variant['discoveryTags'] = []
             variant['discoveryTags'] += [tag for tag in tags if tag['savedVariant']['familyGuid'] not in variant['familyGuids']]
+
+
+def _add_pa_attrs(locus_list_gene, locus_list_guid, gene_json):
+    if hasattr(locus_list_gene, 'palocuslistgene'):
+        # Keeping locusListConfidence for backwards compatibility for now
+        if not gene_json.get('locusListConfidence'):
+            gene_json['locusListConfidence'] = {}
+        gene_json['locusListConfidence'][locus_list_guid] = locus_list_gene.palocuslistgene.confidence_level
+
+        if not gene_json.get('locusListPaAttrs'):
+            gene_json['locusListPaAttrs'] = {}
+        gene_json['locusListPaAttrs'][locus_list_guid] = {
+            'confidence': locus_list_gene.palocuslistgene.confidence_level,
+            'moi': locus_list_gene.palocuslistgene.mode_of_inheritance,
+        }
 
 
 LOAD_PROJECT_TAG_TYPES_CONTEXT_PARAM = 'loadProjectTagTypes'
--- a/ui/shared/components/panel/variants/VariantGene.jsx
+++ b/ui/shared/components/panel/variants/VariantGene.jsx
@@ -6,9 +6,10 @@ import { NavLink } from 'react-router-dom'
 import { Label, Popup, List, Header, Segment, Divider, Table, Button, Loader } from 'semantic-ui-react'
 
 import { getGenesById, getLocusListsByGuid, getFamiliesByGuid } from 'redux/selectors'
+import { panelAppUrl, moiToMoiTypes } from '../../../utils/panelAppUtils'
 import {
   MISSENSE_THRESHHOLD, LOF_THRESHHOLD, PANEL_APP_CONFIDENCE_LEVEL_COLORS,
-  PANEL_APP_CONFIDENCE_DESCRIPTION,
+  PANEL_APP_CONFIDENCE_DESCRIPTION, PANEL_APP_MODE_OF_INHERITANCE_INITIALS,
 } from '../../../utils/constants'
 import { camelcaseToTitlecase } from '../../../utils/stringUtils'
 import { HorizontalSpacer, VerticalSpacer } from '../../Spacers'
@@ -83,7 +84,10 @@ GeneLabel.propTypes = {
 }
 
 const BaseLocusListLabels = React.memo((
-  { locusListGuids, locusListsByGuid, locusListConfidence, compact, containerStyle, ...labelProps },
+  {
+    locusListGuids, locusListsByGuid, locusListPaAttrs,
+    geneSymbol, compact, containerStyle, ...labelProps
+  },
 ) => (
   compact ? (
     <GeneDetailSection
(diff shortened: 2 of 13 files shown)


hanars (changes requested):

seqr/views/utils/variant_utils.py
```diff
@@ -105,14 +105,11 @@ def _add_locus_lists(projects, genes, add_list_detail=False, user=None, is_analy
     for interval in _get_json_for_models(intervals, nested_fields=[{'fields': ('locus_list', 'guid')}]):
         locus_lists_by_guid[interval['locusListGuid']]['intervals'].append(interval)
 
-    for locus_list_gene in LocusListGene.objects.filter(locus_list__in=locus_lists, gene_id__in=genes.keys()).prefetch_related('locus_list', 'palocuslistgene'):
+    for locus_list_gene in LocusListGene.objects.filter(locus_list__in=locus_lists, gene_id__in=genes.keys()).prefetch_related('locus_list', 'locus_list__palocuslist', 'palocuslistgene'):
```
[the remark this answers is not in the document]
tommyli: Agree, it was left over from previous implementation when url was done in backend.

ui/shared/utils/panelAppUtils.js
```diff
@@ -0,0 +1,51 @@
+export const moiToMoiTypes = (rawMoi) => {
```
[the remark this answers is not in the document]
tommyli: Yeah, especially since we're not even showing initials for all of them.  This PR will already provide a reference for what each raw mode_of_inheritance value means (which is somewhat arbitrary in the first place) and we'll refer to it if we want more initials in the future.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -98,27 +102,59 @@ const BaseLocusListLabels = React.memo((
...
       {locusListGuids.map((locusListGuid) => {
-        const panelAppConfidence = locusListConfidence && locusListConfidence[locusListGuid]
+        const paAttrs = locusListPaAttrs[locusListGuid]
+        const { confidence, moi } = paAttrs
         let { description } = locusListsByGuid[locusListGuid] || {}
-        if (panelAppConfidence) {
+        let initials = ''
+        let label = (locusListsByGuid[locusListGuid] || {}).name
+
+        if (paAttrs) {
```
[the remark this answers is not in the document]
david-ma: Ok yes, I'll move this logic to a helper function in panelAppUtils.js

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -98,27 +102,59 @@ const BaseLocusListLabels = React.memo((
...
+          const { panelAppId, url } = locusListsByGuid[locusListGuid].paLocusList
+          const fullUrl = panelAppUrl(url, panelAppId, geneSymbol)
+          const initialsArray = moiToMoiTypes(moi)
+            .map(moiType => PANEL_APP_MODE_OF_INHERITANCE_INITIALS[moiType])
+            .filter(moiType => moiType)
+
+          if (initialsArray.length > 0) {
+            initials = `(${initialsArray.join(', ')})`
+            let allowedLabelSpace = 14
```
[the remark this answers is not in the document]
david-ma: I tried a few CSS things but couldn't get it done nicely before. I didn't know about content-detail in Semantic UI though, so I'll look into it and see if I can do it.

ui/shared/utils/panelAppUtils.js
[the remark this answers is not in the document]
david-ma: Ah yes, the MOI Type was originally used because I was expecting to put in Symbols or Emojis instead of initials. And then when they weren't all being shown, I thought it would be more accurately testable with the MOI Type & more future proof if we were to show the MOI Type elsewhere.

But yes, it is an uneccessary step now and I'll remove it for a moiToMoiInitials function.

david-ma:
We have made all these changes. Notes:

* I used flex css for the labels, I added 3 components to StyledComponents.jsx, but from the outside it should look quite simple. Only FlexLabel is exposed, it is an extension of the ColoredLabel and takes a `hint` as well as a `label`. The hint is placed on the right side of the label, and does not shrink. Giving us the visual that we want.

* I'm not sure if I've put the helper function `getPaProps` in the right place.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -34,12 +34,19 @@ const PADDED_INLINE_STYLE = {
...
+const BaseGeneLabelContent = styled(({
+  color, customColor, hint, label, maxWidth, containerStyle, dispatch, ...props
+}) => {
   const labelProps = {
     ...props,
     size: 'mini',
     content: label,
   }
+
+  if (hint && customColor) {
```
hanars, line 46: so I think you misunderstood what I meant by using the `detail` prop for the label. If you pass through a prop named `detail` directly to the label (pass it directly to `GeneDetailSection` and it will get passed through) and then it will render inside the label in its own div. If you use that prop for the initials, you can use this styled component to style that detail separately from the rest of the label content. You will not need to add any new components or conditional rendering. The easiest way I found to do this is to wrap the main label content in a span in the line above this (`content: label` -> `content: <span>{label}</span>`) and then change this styling below to the following:
```
margin: ${props => props.margin || '0px .5em .8em 0px'} !important;
white-space: nowrap;
   
span {
     display: inline-block;
     overflow: hidden;
     text-overflow: ellipsis;
     max-width: ${props => props.maxWidth || 'none'};
}

.detail {
     margin-left: 0.5em !important;
     
     &::before {
      content: "(";
     }
     &::after {
      content: ")";
     }
 }
```
This means the max width and overflow are applied to the label content and not the initials, which is I think better behavior (although it means we probably want a small value for max width here, maybe 9em)

hanars (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: Oh! I didn't realise you could pass in selectors in StyledComponents. I couldn't figure out how to target the Label.Detail when I tried it before, but yes, this is much cleaner.

I also like using the before/after content to show the (), that works much better with the "hint".

I've cleaned the FlexLabel component to use `Label.Detail` and a normal `span`, and added a simple render test.

hanars: So I think you misuderstood what I was saying. You should be adding no new styled components as part of this PR. The `FlexLabel` is unnecessary, the code I posted will work with no new components and was intended to be put here, in this already existing  `BaseGeneLabelContent` component. I did describe it in a few pieces which was maybe unclear, here is the full version of what the code should look like:

```
const BaseGeneLabelContent = styled(({ color, customColor, label, maxWidth, containerStyle, dispatch, hint, ...props })`
 const labelProps = {
    ...props,
    size: 'mini',
    content: <span>label<span>, // update here to wrap the content so it can be styled
    detail: hint, // add this to pass the correctly named prop to the label. I think it would be cleaner to just name the prop "detail' rather than rename it here, but this will work
 }
 return customColor ?
    <ColoredLabel {...labelProps} color={customColor} /> : <Label {...labelProps} color={color || 'grey'} />
 })`
  margin: ${props => props.margin || '0px .5em .8em 0px'} !important;
  white-space: nowrap;
   
  span {
     display: inline-block;
     overflow: hidden;
     text-overflow: ellipsis;
     max-width: ${props => props.maxWidth || 'none'};
  }

  .detail {
     margin-left: 0.5em !important;
     
     &::before {
      content: "(";
     }
     &::after {
      content: ")";
     }
 }
`
```

david-ma: Ah sorry, I didn't realise passing it in as a prop would work like that.

In that case, renaming hint to detail makes more sense, I was avoiding the word detail when I made the other components, that didn't use the native detail.

I've used your code, made a few fixes, and removed the extra component.

david-ma:
A screenshot of how the design looks, for posterity.

A screenshot of how the design looks, for posterity.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -34,20 +34,39 @@ const PADDED_INLINE_STYLE = {
...
-const BaseGeneLabelContent = styled(({ color, customColor, label, maxWidth, containerStyle, dispatch, ...props }) => {
+const BaseGeneLabelContent = styled(({
+  color, customColor, detail, label, maxWidth, containerStyle, dispatch, ...props
+}) => {
   const labelProps = {
     ...props,
     size: 'mini',
-    content: label,
+    content: <span>{label}</span>,
+    detail: detail || null,
```
hanars, line 44: If the prop is not being renamed then we should just let it pass through in the `... props`, no need to destructure it at all

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -98,28 +174,30 @@ const BaseLocusListLabels = React.memo((
...
-              <br />
-              <br />
-              <b>PanelApp gene confidence: &nbsp;</b>
-              {PANEL_APP_CONFIDENCE_DESCRIPTION[panelAppConfidence]}
-            </div>
-          )
+        const locusList = locusListsByGuid[locusListGuid]
+        const panelAppDetails = panelAppDetail[locusListGuid]
+        const { name: label, description: locusListDescription, paLocusList } = locusList
+        const { description, initials, customColor } = panelAppDetails && paLocusList ? getPaProps({
```
hanars, line 180: not strictly necessary, but I think parenthesis would make the control flow here a little cleaner. The `&& ? :` is a little hard to read

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -98,28 +174,30 @@ const BaseLocusListLabels = React.memo((
...
+          panelAppDetails,
+          panelAppPanel: locusListDescription || '',
```
hanars, line 182: why not name this prop `locusListDescription ` to make it more readable

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -98,28 +174,30 @@ const BaseLocusListLabels = React.memo((
...
+
         return (
           <GeneDetailSection
             key={locusListGuid}
             color="teal"
-            customColor={panelAppConfidence && PANEL_APP_CONFIDENCE_LEVEL_COLORS[panelAppConfidence]}
-            maxWidth="7em"
+            customColor={customColor}
+            detail={initials}
+            maxWidth="12em"
```
hanars, line 197: so especially with the initials not counting toward the max width, this is too wide for us. I would accept `8em` or `9em`.  Its also okay if we agree to disagree here and our forks differ a little

ui/shared/utils/panelAppUtils.js
```diff
@@ -0,0 +1,54 @@
...
+      return ['XR', 'XD']
+    }
+  }
+  if (moi.startsWith('BIALLELIC')) {
+    return ['AR']
+  }
+  if (moi.startsWith('BOTH')) {
+    return ['AD', 'AR']
+  }
+  if (moi.startsWith('MITOCHONDRIAL')) {
```
hanars, line 31: all if conditionals from here down are unnecessary, you should just use the fallback `return []`

hanars (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: Detail needs to be transformed to null, or it will be added to every label and the brackets () will be on all of them. I guess it could be set to null earlier and this line could be removed.

ui/shared/utils/panelAppUtils.js
david-ma: Ok

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: Ah the initials don't count that maxWidth? I thought it looked a bit wider. I agree it's too wide in some cases.

I would have liked the widest ones without initials to be the same width as the ones showing initials.

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: The function that is receiving this property, PanelAppHoverOver() wants to know the PanelApp Panel (which is different to the PanelApp Name).

I thought the word description was becoming a bit too overloaded and wanted to be clear what is happening in that function, since it is returning a description.

I'm happy to use `locusListDescription`, `panelAppPanel` could be a bit confusing for a reader unfamiliar with the PanelApp datasource.

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: Ok 👍  I thought the linter would have added that, but it didn't. I'll put it in.

ui/shared/components/panel/variants/VariantGene.jsx
david-ma: `8em` is quite readable on my laptop screen (1680x1050). I've added this to the PR.

Screen Shot 2022-05-23 at 3 38 30 pm

hanars (approved):

PR #2729, opened 2022-05-20
Deprecate nginx deployment

sjahl:
Now that argo is managing the ingress-nginx controller installation, this PR removes the servctl bits and the controller manifests that we were using to deploy it.

Note that the `Ingress` manifest itself is specific to seqr, so it should be deployed alongside the app rather than with the controller. So, with that in mind, I've moved the ingress definition to seqr.gcloud.yaml. The forthcoming helm chart for seqr will bundle seqr's ingress in a similar fashion.

Closes https://github.com/broadinstitute/seqr-private/issues/1109


hanars:
So I'm pretty sure this isn't an issue in kubernetes but just want to confirm: If we update part of the yaml file unrelated to the ingress (say add a new environment variable to the seqr pod) and run `kubectl apply` (which is what the `servctl deploy` ends up running) the ingress controller will not update, correct? I just wouldn't want to accidentally break ingress due when we are updating the rest of seqr

sjahl:
Yeah, if we changed something on the seqr pod, and left that Ingress section untouched, the `kubectl apply` would still "apply" the ingress since it's in the file. But, if there's no difference between what we apply and what's already on the cluster, the kube API will just return that it's "unchanged" and not do anything.

sjahl:
per the discussion in standup, I confirmed that this is a no-op in dev:

`2022-05-24 15:18:44,083 INFO     ingress.networking.k8s.io/ingress-nginx unchanged`

Though, the dev seqr pod seems to be unhappy for other reasons (but I don't think they're related to my changes here):

`  Warning  Failed     71s (x5 over 93s)  kubelet            Error: couldn't find key airflow_api_audience in Secret gcloud-dev/seqr-secrets`

hanars:
okay so based on the standup discussion we should change the base of this to master, fill out the release checklist, and then merge this to master and run `servctl deploy` just so the noop is done. Does that sound right to you?

sjahl:
- [ ] All unit test are passing and coverage has not substantively dropped
- [ ] No new dependency vulnerabilities have been introduced
- [ ] Any changes to the docker image have been vetted for potential vulnerabilities
- [ ] If any python requirements are updated, they are noted and here and will be updated before deploying: 
- [ ] Any database migrations are noted here, and will be run before deploying: 
- [ ] All major changes are recorded in the changelog, and the changelog has been updated to reflect the latest release
- [ ] Any new endpoints are explicitly tested to ensure they are only accessible to correctly permissioned users
- [ ] No secrets have been committed
- [ ] Infrastructure changes: 
   
  OR 
  
  - [ ] Any chages to non-seqr docker images have been built and pushed to the container registry
  - [ ] All these changes have been vetted for potential vulnerabilities

hanars:
can you put x in all the relevant checkboxes to indicate that they are done?

sjahl:
> can you put x in all the relevant checkboxes to indicate that they are done?

Yup, sorry, just pasting stuff in between meetings 😓

hanars (approved):

PR #2734, opened 2022-05-23
fix saved variants when no panel app

hanars:


sjahl (approved):
Assuming this is to correctly handle the case where panelappdetail is undefined, this looks good to me 👍

PR #2741, opened 2022-05-24
fix: downgrade hijack back to 2.x

tommyli:
Reverted django-hijack to version 2.x.  3.x added breaking changes and will require backend and UI changes.


hanars:
I'm sorry but we can't downgrade to unsupported major versions. If you need to maintain the hijack functionality please feel free to submit a request that updates the hijack button to work with the v3 URLS

tommyli:
PR #2746 raised.

PR #2750, opened 2022-05-25
update vulnerable dependencies

hanars:
Updated packages to fix vulnerable package versions, used the following commands:
```
pip-compile --upgrade-package hail
pip-compile --upgrade-package pyjwt[crypto]
```


sjahl (approved):
The snyk test hasn't finished as of this review, but this looks good to me if snyk's happy.

hanars:
the snyk integration is showing that the security test failed, but when you click into the details of the failure it actually shows a success, and I double checked that the success is associated with the commit from this PR. Therefore I think this is safe to merge

PR #2763, opened 2022-05-31
Fix sorting in management console

david-ma:
Low Priority

The tables in the Elastic Search Managment Console are not being sorted properly.
https://seqr.broadinstitute.org/data_management/elasticsearch_status

I dressed them up a bit and made them more usable.

* Simple numbers (shards) are parsed so they sort properly
* Disk sizes are stored as bytes to be sortable, and then formatted to be human readble
* The more complex Projects array of objects is sorted on number of projects, then project name.

I used this low priority task to understand the tables used in seqr, since other tables in seqr use the same DataTable component and similar patterns would be used for modifying them.


hanars:
so unfortunatley this is a case where you are doing in the front end what should be done in the backend. You can modify the query to elasticsearch by adding a `bytes=b` query parameter which will cause the API to return the size in bytes instead of in the human-readable size format. The `formatBytes` function would still belong in the UI

PR #2769, opened 2022-06-02
Remove linkerd and eck operator installs

sjahl:
Since we're now deploying the ECK operator and linkerd using argocd, this removes the steps that installed them from the servctl script.

Closes https://github.com/broadinstitute/seqr-private/issues/1151

Does the first half of https://github.com/broadinstitute/seqr-private/issues/1108 -- the rest of that can be done once our seqr helm chart is responsible for the Elasticsearch manifests that deploy the ES cluster itself.

### DIFF
--- a/deploy/kubernetes/shared-settings.yaml
+++ b/deploy/kubernetes/shared-settings.yaml
@@ -1,7 +1,6 @@
 #####################################################
 ####### SETTINGS THAT MAY NEED TO BE MODIFIED #######
 ELASTICSEARCH_VERSION: 7.16.3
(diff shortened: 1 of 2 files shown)


hanars:
looks good to me but I would rather this got merged to dev and then went to master as part of a regular release. If you feel strongly about it going to master then please fill out the release checklist

sjahl:
With the small caveat, if we update only prod elasticsearch with the `_set_elasticsearch_kubernetes_resources()` still present in servctl, it'll interfere with the operator deployment that argo is managing.

hanars (approved):

PR #2778, opened 2022-06-05
fix: Show gene lists filter input field due to bad margin

tommyli:
* Suspect this was caused by the new maxHeight prop in DataTable introduced in https://github.com/broadinstitute/seqr/pull/2532


hanars (changes requested):
There are two issues with this fix:
1) The display is currently working on the summary data gene lsits page, and this would break that display
2) This does not put the filter in the right place - it belongs inline with the title, the way it currently looks on the summary data page

Any fix needs to work for both the summary data page and the modal for the project page

hanars:
closed due to inactivity

PR #2787, opened 2022-06-21
Fix rich text whitespace

hanars:
We changed js libraries a while ago for markdown rendering and added some css to keep an old-style hack for RGP datstat notes working. However, it turns out from the related bug that the whitespace hack actually causes all other notes to look wrong as described in the linked issue. This PR takes out the old hack for properly indenting the datstat notes and instead migrates all the notes to use indented lists instead


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -525,9 +523,9 @@ def _relative_list_summary(row, relative, all_affected=False):
     if relative_list is None:
         return 'None'
 
-    divider = '\n{tab}{tab}'.format(tab=DSMConstants.TAB)
+    divider = '\n* * '
     return '{divider}{relatives}'.format(
-        divider=divider,
+        divider='\n* * ',
```
ShifaSZ, line 528: You may want to replace the placeholder `divider` with `\n* * ` in the string.

ShifaSZ (changes requested):
Mostly looks good. There is one issue you may want to fix it.

PR #2789, opened 2022-06-21
Gene search selected transcript

hanars:
Changes behavior for defining the selected transcript id to prioritize transcripts in genes specified in location search


seqr/utils/elasticsearch/es_search.py
```diff
@@ -228,7 +229,10 @@ def filter_variants(self, inheritance=None, genes=None, intervals=None, rs_ids=N
         self._filter_custom(custom_query)
 
         if has_location_filter:
-            self._filter(_location_filter(genes, intervals, locus))
+            exclude_locations = locus and locus.get('excludeLocations')
+            self._filter(_location_filter(genes, intervals, exclude_locations))
+            if genes and not exclude_locations:
+                self._filtered_gene_ids = set(genes.keys())
```
ShifaSZ, line 235: Is it necessary to cast `genes.keys()` to a set type?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_search.py
hanars: yes

ShifaSZ (approved):
Looks good.

PR #2801, opened 2022-06-29
fix browser console errors

hanars:
Fixes some errors that show in the browser console when react is rendering

### DIFF
--- a/ui/shared/components/panel/family/FamilyLayout.jsx
+++ b/ui/shared/components/panel/family/FamilyLayout.jsx
@@ -26,17 +26,19 @@ const FamilyLayout = React.memo((
   <div>
     {annotation}
     <FamilyGrid annotation={annotation} offset={offset} compact fixed={!useFullWidth}>
-      <Table.Row verticalAlign="top">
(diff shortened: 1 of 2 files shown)


ui/shared/components/panel/family/FamilyLayout.jsx
```diff
@@ -26,17 +26,19 @@ const FamilyLayout = React.memo((
...
-        {compact ? (fields || []).map(
-          field => <Table.Cell width={field.colWidth || 1} key={field.id}>{fieldDisplay(field)}</Table.Cell>,
-        ) : (
-          <Table.Cell width={getContentWidth(useFullWidth, leftContent, rightContent)}>
-            {(fields || []).map(field => fieldDisplay(field))}
-          </Table.Cell>
-        )}
-        {rightContent && <Table.Cell width={3}>{rightContent}</Table.Cell>}
-      </Table.Row>
+      <Table.Body>
```
hanars, line 29: I added this `Table.Body` content around the row to fix the dom nesting error we were getting, but nothing else changed except for the indentation

PR #2807, opened 2022-06-30
better dropdown category display

hanars:
Changes the display for dropdown category headers. Will be used for https://github.com/broadinstitute/seqr-private/issues/1158 but is generally a better display for everything


ShifaSZ (approved):
Looks good to me.

PR #2810, opened 2022-07-05
Mito seqr search

ShifaSZ:
Adding mitochondria frequency and quality search filters.


seqr/utils/elasticsearch/constants.py
```diff
@@ -351,10 +351,13 @@ def get_prediction_response_key(key):
 SHARED_QUALITY_FIELDS = {'gq': 5}
 SNP_QUALITY_FIELDS = {'ab': 5}
 SNP_QUALITY_FIELDS.update(SHARED_QUALITY_FIELDS)
+MITO_QUALITY_FIELDS = {'hl': 5}
+MITO_QUALITY_FIELDS.update(SHARED_QUALITY_FIELDS)
 SV_QUALITY_FIELDS = {'qs': 10}
 SV_QUALITY_FIELDS.update(SHARED_QUALITY_FIELDS)
 QUALITY_QUERY_FIELDS.update(SNP_QUALITY_FIELDS)
 QUALITY_QUERY_FIELDS.update(SV_QUALITY_FIELDS)
+QUALITY_QUERY_FIELDS.update(MITO_QUALITY_FIELDS)
```
hanars, line 360: Don't you also need to update the `POPULATIONS` constant to support the new frequency filter?

hanars (changes requested):

ui/shared/components/panel/search/constants.js
```diff
@@ -350,6 +350,23 @@ const SV_FREQUENCIES = [
   },
 ]
 
+const MITO_FREQUENCIES = [
```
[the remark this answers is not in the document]
hanars: Mor sepcifically right now the `PANEL_MAP` in `VariantSearchFormContent.jsx` handles 2 dataset types: `ALL_DATASET_TYPE` and `DATASET_TYPE_VARIANT_CALLS`. You will need to add `DATASET_TYPE_VARIANT_MITO` and `DATASET_TYPE_VARIANT_SV` to handle when the dataset type is regular variants and mito or when it is regular variants and SVs (the `ALL_DATASET_TYPE` will now refer to all 3 types). For both of the 2 new types you will need to update `DATASET_TYPE_PANEL_PROPS ` to override the fields for the quality and frequency panels, as well as update the existing configuration for `DATASET_TYPE_VARIANT_CALLS` for those panels. Let me know if you have any questions

seqr/utils/elasticsearch/constants.py
ShifaSZ: The `POPULATIONS` has been updated for displaying the mito frequencies. No more updates are required for the frequency filters.

ShifaSZ:
The latest frequency and quality filters:
ALL_DATASET_TYPE:
image
DATASET_TYPE_VARIANT_MITO:
image

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -181,14 +185,31 @@ const DATASET_TYPE_PANEL_PROPS = {
       fields: SNP_QUALITY_FILTER_FIELDS,
     },
   },
+  [DATASET_TYPE_VARIANT_MITO]: {
```
hanars, line 188: I think all the other panels for `DATASET_TYPE_VARIANT_MITO` need to have the same panels as `DATASET_TYPE_VARIANT_CALLS`. I think the cleanest way to do that is to do the assignment for this panel type after the dict has been created:
```
const DATASET_TYPE_PANEL_PROPS = {
  [DATASET_TYPE_VARIANT_CALLS]: {...},
  [DATASET_TYPE_VARIANT_SV]: {...},
}
DATASET_TYPE_PANEL_PROPS[DATASET_TYPE_VARIANT_MITO] = {
  ...DATASET_TYPE_PANEL_PROPS[DATASET_TYPE_VARIANT_CALLS],
  FREQUENCY_PANEL.name]: {
  ...
}
```

ui/shared/components/panel/search/constants.js
```diff
@@ -476,12 +485,22 @@ export const SNP_QUALITY_FILTER_FIELDS = [
   },
 ]
 
+export const MITO_QUALITY_FILTER_FIELDS = [
+  {
```
hanars, line 489: can we use a `DividedFormField` component for this similar to how we do SVs so it has the vertical divider?

hanars (changes requested):

ui/pages/Search/components/VariantSearchFormContent.jsx
ShifaSZ: Do we have an `SV` only case? If yes, I have to update a little bit to my last commit.

hanars: no, we do not have an only SV or an only MITO case, but thanks for checking

ui/shared/components/panel/search/constants.js
ShifaSZ: Added.

hanars (changes requested):

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -183,12 +187,32 @@ const DATASET_TYPE_PANEL_PROPS = {
...
+  ...DATASET_TYPE_PANEL_PROPS[DATASET_TYPE_VARIANT_CALLS],
+  [FREQUENCY_PANEL.name]: {
+    fields: SNP_FREQUENCIES.concat(MITO_FREQUENCIES),
+  },
+  [QUALITY_PANEL.name]: {
+    fields: SNP_QUALITY_FILTER_FIELDS.concat(MITO_QUALITY_FILTER_FIELDS),
+  },
+}
+
+DATASET_TYPE_PANEL_PROPS[DATASET_TYPE_VARIANT_SV] = {
```
hanars, line 200: it is cleaner to keep this in the definition for `DATASET_TYPE_PANEL_PROPS` instead of adding it after like this (so add this in the original `const DATASET_TYPE_PANEL_PROPS = {...}` definition). The only reason to add it after is if you need to reference the existing object the way you do for `DATASET_TYPE_VARIANT_MITO`.

PR #2817, opened 2022-07-06
Anvil loading wizard

hanars:
Splits the anvil loading request from a single form into a wizard with 3 steps, with a sever side action on each step submission. This allows us to break apart the loading which means we can spend more time on each step without running into browser timeout issues. This will allow future enhancements such as adding more robust validation and giving users a file list to choose from


ui/shared/components/form/FormWizard.jsx
```diff
@@ -0,0 +1,59 @@
...
+    onSubmit: PropTypes.func.isRequired,
+    pages: PropTypes.arrayOf(PropTypes.object),
+  }
+
+  state = { pageIndex: 0, asyncValues: {} }
+
+  navigateNext = () => {
+    this.setState(prevState => ({
+      pageIndex: prevState.pageIndex + 1,
+    }))
```
ShifaSZ, line 18: It would be better to support navigating backward as well.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -442,7 +518,7 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_f
     "active_projects": [
         "{guid}"
     ],
-    "vcf_path": "gs://test_bucket/test_path.vcf.gz",
+    "vcf_path": "gs://test_bucket/test_path.vcf",
```
ShifaSZ, line 521: Is there any reason to remove `.gz`?

ShifaSZ (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -92,7 +108,46 @@ def anvil_workspace_page(request, namespace, name):
...
+def grant_workspace_access(request, namespace, name):
+    request_json = json.loads(request.body)
+    if not request_json.get('agreeSeqrAccess'):
+        error = 'Must agree to grant seqr access to the data in the associated workspace.'
+        return create_json_response({'error': error}, status=400, reason=error)
+
+    # Add the seqr service account to the corresponding AnVIL workspace
+    added_account_to_workspace = add_service_account(request.user, namespace, name)
+    if added_account_to_workspace:
+        _wait_for_service_account_access(request.user, namespace, name)
```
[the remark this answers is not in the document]
hanars: No. The whole point of the wizard is that each step is independent of the others. Once they click "grant access", access is granted. If they really really want to revoke access they can do that from AnVIL

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -92,7 +108,46 @@ def anvil_workspace_page(request, namespace, name):
...
+        return create_json_response({'error': error}, status=400, reason=error)
+    if not does_file_exist(data_path, user=request.user):
+        error = 'Data file or path {} is not found.'.format(path)
+        return create_json_response({'error': error}, status=400, reason=error)
+
+    # Validate the VCF to see if it contains all the required samples
+    samples = get_vcf_samples(data_path)
+    if not samples:
+        return create_json_response(
+            {'error': 'No samples found in the provided VCF. This may be due to a malformed file'}, status=400)
```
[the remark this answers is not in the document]
hanars: A VCF with no samples is invalid, we absolutely do require samples. We want to keep the validation the same as it was before and we want to do as much of the validation in this validate VCF step as possible.

ui/shared/components/form/FormWizard.jsx
hanars: Thats a good point

seqr/views/apis/anvil_workspace_api_tests.py
hanars: yes, it was weird that we use different success request bodies and it made it hard to read/ maintain. This changes it so the successful requests use the same success body, not the body that was previously used to test a failure case

ui/shared/components/form/FormWizard.jsx
hanars: added

ui/shared/components/form/FormWizard.jsx
```diff
@@ -42,11 +48,17 @@ class FormWizard extends React.PureComponent {
       submitButtonText: 'Next',
       submitButtonIcon: 'angle double right',
     }
+    const backButtonProps = pageIndex === 0 ? {} : {
+      onCancel: this.navigateBack,
+      cancelButtonText: 'Back',
+      cancelButtonIcon: 'angle double left',
```
ShifaSZ, line 54: The double left angle looks like returning to the beginning. Should a single angle left be better?

hanars: No, I would like to keep this as is to keep it consistent with the "Next" button

ShifaSZ (approved):

PR #2831, opened 2022-07-12
fix inheritance mode display

hanars:


ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -68,7 +68,7 @@ const INHERITANCE_PANEL = {
           return null
         }
         const { affected, genotype, ...coreFilter } = val.filter
-        return INHERITANCE_MODE_LOOKUP[JSON.stringify(coreFilter)]
+        return INHERITANCE_MODE_LOOKUP[val.mode] || INHERITANCE_MODE_LOOKUP[JSON.stringify(coreFilter)]
```
ShifaSZ, line 71: It seems to me that `INHERITANCE_MODE_LOOKUP[val.mode]` is always defined. Does it make sense to have an `||`?

ShifaSZ (approved):

PR #2833, opened 2022-07-12
fix vulnerable ui dependencies

hanars:


ShifaSZ:
Why does it just change the lock file? No changes to the package.json?

PR #2837, opened 2022-07-15
Anvil new data

ShifaSZ:
Create a draft PR first since new unit tests are still missing.


ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -104,13 +104,23 @@ const createProjectFromWorkspace = postWorkspaceValues(
...
-  { fields: [AGREE_CHECKBOX], onPageSubmit: postWorkspaceValues('grant_access') },
-  { fields: [DATA_BUCK_FIELD, SAMPLE_TYPE_FIELD, REQUIRED_GENOME_FIELD], onPageSubmit: postWorkspaceValues('validate_vcf') },
-  { fields: [PROJECT_DESC_FIELD, UPLOAD_PEDIGREE_FIELD] },
-]
-
-const LoadWorkspaceDataForm = React.memo(({ params }) => (
+const addDataFromWorkspace = projectGuid => values => (
+  new HttpRequestHelper(`/api/project/${projectGuid}/add_workspace_data`, (responseJson) => {
+    if (responseJson.success) {
+      window.location.href = `/project/${projectGuid}/project_page`
```
ShifaSZ, line 110: The Wizard is created in a Modal. I use a redirect to jump to the project page after submission. It may be more user-friendly by just closing the Modal.  But it needs some changes to the Wizard component. Should I make the change?

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -165,6 +165,7 @@ def _process_result(result, project):
             'isMmeEnabled': result['isMmeEnabled'] and not result['isDemo'],
             'canEdit': has_project_permissions(project, user, can_edit=True),
             'userIsCreator': project.created_by == user,
+            'isAnalystProject': any(c.name == ANALYST_PROJECT_CATEGORY for c in project.projectcategory_set.all())
```
hanars, line 168: in order to prevent this having a performance impact, we need to change the prefetch behavior below to always run  `prefetch_related_objects(projects, 'projectcategory_set')` (currently it is only conditionally run)

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
```diff
@@ -104,13 +104,23 @@ const createProjectFromWorkspace = postWorkspaceValues(
...
+  }).post({ ...values, uploadedFileId: values.uploadedFile?.uploadedFileId }))
+
+const GRANT_ACCESS_PAGE = { fields: [AGREE_CHECKBOX], onPageSubmit: postWorkspaceValues('grant_access') }
+const VALIDATE_VCF_PAGE = { fields: [DATA_BUCK_FIELD, SAMPLE_TYPE_FIELD, REQUIRED_GENOME_FIELD], onPageSubmit: postWorkspaceValues('validate_vcf') }
+const ADD_DATA_VALIDATE_VCF_PAGE = { ...VALIDATE_VCF_PAGE, fields: [DATA_BUCK_FIELD] }
+const UPLOAD_PED_PAGE = { fields: [PROJECT_DESC_FIELD, UPLOAD_PEDIGREE_FIELD] }
+const ADD_DATA_UPLOAD_PED_PAGE = { fields: [UPLOAD_PEDIGREE_FIELD] }
+
+const FORM_WIZARD_PAGES = [GRANT_ACCESS_PAGE, VALIDATE_VCF_PAGE, UPLOAD_PED_PAGE]
+const FORM_ADD_DATA_WIZARD_PAGES = [GRANT_ACCESS_PAGE, ADD_DATA_VALIDATE_VCF_PAGE, ADD_DATA_UPLOAD_PED_PAGE]
```
hanars, line 121: This is kind of difficult to read, I think there are a lot more constants defined than there really need to be. Can we do the following instead:

```
const GRANT_ACCESS_PAGE = { fields: [AGREE_CHECKBOX], onPageSubmit: postWorkspaceValues('grant_access') }
const VALIDATE_VCF_PAGE = { fields: [DATA_BUCK_FIELD], onPageSubmit: postWorkspaceValues('validate_vcf') },
const NEW_PROJECT_PAGES = [
  GRANT_ACCESS_PAGE,
  { ... VALIDATE_VCF_PAGE, fields: [...VALIDATE_VCF_PAGE.fields, SAMPLE_TYPE_FIELD, REQUIRED_GENOME_FIELD] },
  { fields: [PROJECT_DESC_FIELD, UPLOAD_PEDIGREE_FIELD] },
]
const ADD_DATA_PAGES = [
  GRANT_ACCESS_PAGE,
  VALIDATE_VCF_PAGE,
  { fields: [UPLOAD_PEDIGREE_FIELD] }
]
```

ui/pages/AnvilWorkspace/components/LoadWorkspaceDataForm.jsx
hanars: So I didn't actually think this through on the ticket, so I'm sorry for not being more clear upfront. Thinking about this now what I think we actually want is after submission succeeds to keep the modal open and show a message telling them that their request was submitted and they should get their data in about a week, so they don't get confused about why there is no data. The exact message I would use is "Your request to load data has been submitted. Loading data from AnVIL to seqr is a slow process, and generally takes a week. You will receive an email letting you know once your new data is available."

I think the best way to do this is to modify the `FormWizard` component to take an optional `successMessage` prop and if that prop exists to show that message instead of the form after submission succeeds. I would add a field to the state like `formSubmitSuceeded` and then use the `onSubmitSucceeded` prop for the final page to set that state field

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -282,10 +285,26 @@ const Dataset = React.memo(({ project, samplesByType, user }) => {
...
+  Object.assign(params, { namespace: project.workspaceNamespace, name: project.workspaceName })
+
+  const addAnvilDataButton = (
+    <Modal
+      modalName="addAnvilDataset"
+      title="Add Anvil Dataset"
+      size="small"
+      trigger={<ButtonLink>Request Additional Data</ButtonLink>}
+    >
+      <LoadWorkspaceDataForm params={params} projectGuid={project.projectGuid} />
```
hanars, line 297: constructing an object is generally a bad idea. For `LoadWorkspaceDataForm` you should change the param field names to `workspaceNamespace` and `workspaceName` and then you can pass `project` as `params` here

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -34,6 +34,7 @@ import {
 import EditFamiliesAndIndividualsButton from './edit-families-and-individuals/EditFamiliesAndIndividualsButton'
 import EditIndividualMetadataButton from './edit-families-and-individuals/EditIndividualMetadataButton'
 import EditDatasetsButton from './EditDatasetsButton'
+import LoadWorkspaceDataForm from '../../AnvilWorkspace/components/LoadWorkspaceDataForm'
```
hanars, line 37: We never import components across different pages. This component should be moved to be a `shared` component. To make the diff more readable, it might be best to make a branch that just moves the `LoadWorkspaceDataForm` component to shared, and then have this branch off of that branch so the changes to the component will be easily readable

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -282,10 +285,26 @@ const Dataset = React.memo(({ project, samplesByType, user }) => {
...
+      modalName="addAnvilDataset"
+      title="Add Anvil Dataset"
+      size="small"
+      trigger={<ButtonLink>Request Additional Data</ButtonLink>}
+    >
+      <LoadWorkspaceDataForm params={params} projectGuid={project.projectGuid} />
+    </Modal>
+  )
+
+  const dataButton = !project.isAnalystProject && project.canEdit ? addAnvilDataButton :
```
hanars, line 301: not necessary, but adding parenthesis around `(!project.isAnalystProject && project.canEdit)` would make this more readable

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -193,29 +194,79 @@ def create_project_from_workspace(request, namespace, name):
...
+        return create_json_response({'error': error}, status=400, reason=error)
+    request_json['genomeVersion'] = project.genome_version
+    request_json['sampleType'] = Sample.objects.filter(individual__family__project=project)[0].sample_type
+
+    # Parse families/individuals in the uploaded pedigree file
+    json_records = load_uploaded_file(request_json['uploadedFileId'])
+    pedigree_records, _ = parse_pedigree_table(json_records, 'uploaded pedigree file', user=request.user, fail_on_warnings=True)
+
+    indivIds = {ped['individualId'] for ped in pedigree_records}
+    pedigree_records += [ped for ped in get_project_active_pedigrees(project) if ped['individualId'] not in indivIds]
```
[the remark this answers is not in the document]
ShifaSZ: When getting the previous samples, we filter the samples with `dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS` as your suggestion. Does it mean we don't support other dataset type like `SV` or `MITO`?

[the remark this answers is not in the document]
hanars: no, we do not support other data types for AnVIL users in general, and the automated loading pipeline kicked off by this request is variant loading. It will be a lot of engineering work on our end to make SV and MITO loading available for AnVIL data, which is out of scope of this feature as we need to add the loading to airflow and add better file format documentation and validation before we can make it broadly available

ui/pages/Project/components/ProjectOverview.jsx
ShifaSZ: I see what you mean. But if I change the param field names, I have to change the other use of `LoadWorkspaceDataForm` in the `LoadWorkspaceData.jsx`. In that case, I have to construct an object for the params.

hanars: To change the param names for the main form you just need to change the actual names of the parameters in the route:
```
<Route path="/create_project_from_workspace/:namespace/:name" component={LoadWorkspaceData} />
```
changes to 
```
<Route path="/create_project_from_workspace/: workspaceNamespace/: workspaceName" component={LoadWorkspaceData} />
```

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -104,16 +104,37 @@ const createProjectFromWorkspace = postWorkspaceValues(
...
 
-const FORM_WIZARD_PAGES = [
-  { fields: [AGREE_CHECKBOX], onPageSubmit: postWorkspaceValues('grant_access') },
-  { fields: [DATA_BUCK_FIELD, SAMPLE_TYPE_FIELD, REQUIRED_GENOME_FIELD], onPageSubmit: postWorkspaceValues('validate_vcf') },
+const addDataFromWorkspace = projectGuid => (values, onSuccess) => (
+  new HttpRequestHelper(`/api/project/${projectGuid}/add_workspace_data`, (responseJson) => {
+    if (responseJson.success) {
+      onSuccess()
+    }
+  }).post({ ...values, uploadedFileId: values.uploadedFile?.uploadedFileId }))
```
hanars, line 112: this shares too much logic with `postWorkspaceValues`/`createProjectFromWorkspace` to warrant being a separate function. Just modify `postWorkspaceValues` to take an optional third argument of `formatUrl` and then change the url logic to 
```
formatUrl ? formatUrl(values) : `/api/create_project_from_workspace/${workspaceNamespace}/${workspaceName}/${path}`
```
Then make a shared helper for the 2 submit functions:
```
const postSubmitValues =  formatUrl => postWorkspaceValues(
  'submit', ({ uploadedFile, ...values }) => ({ ...values, uploadedFileId: uploadedFile.uploadedFileId }), formatUrl,
)
const createProjectFromWorkspace = postSubmitValues()((responseJson) => {
  window.location.href = `/project/${responseJson.projectGuid}/project_page`
})
const addDataFromWorkspace = postSubmitValues(({ ...projectGuid }) => `/api/project/${projectGuid}/add_workspace_data`)
```

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -125,11 +146,13 @@ const LoadWorkspaceDataForm = React.memo(({ params }) => (
       {WARNING_BANNER ? <Message error compact header={WARNING_HEADER} content={WARNING_BANNER} /> : null}
     </Segment>
     <FormWizard
-      onSubmit={createProjectFromWorkspace}
-      pages={FORM_WIZARD_PAGES}
+      {...props}
+      onSubmit={params.projectGuid ? addDataFromWorkspace(params.projectGuid) : createProjectFromWorkspace}
```
hanars, line 150: since `params` is being set as the `initialValues`, `projectGuid` will be available in the `values` passed to the submit function so you don't need to bind it here

hanars (changes requested):

seqr/views/utils/orm_to_json_utils.py
ShifaSZ: The condition for `prefetch_related_objects(projects, 'projectcategory_set'` is removed.

ui/pages/AnvilWorkspace/LoadWorkspaceData.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Message, Segment } from 'semantic-ui-react'
 
-import LoadWorkspaceDataForm from './components/LoadWorkspaceDataForm'
+import LoadWorkspaceDataForm from 'shared/components/panel/LoadWorkspaceDataForm'
 
 export const WorkspaceAccessError = ({ match }) => (
```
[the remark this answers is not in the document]
ShifaSZ: I think it is changed this time.

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -104,16 +104,37 @@ const createProjectFromWorkspace = postWorkspaceValues(
...
 ]
 
-const LoadWorkspaceDataForm = React.memo(({ params }) => (
+const ADD_DATA_WIZARD_PAGES = [
+  GRANT_ACCESS_PAGE,
+  { ...VALIDATE_VCF_PAGE, fields: [DATA_BUCK_FIELD] },
+  { fields: [UPLOAD_PEDIGREE_FIELD] },
+]
+
+const successMessage = 'Your request to load data has been submitted. Loading data from AnVIL to seqr is a slow process, and generally takes a week. You will receive an email letting you know once your new data is available.'
```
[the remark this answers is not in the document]
ShifaSZ: I found the success messages for both the new Anvil project and new data are the same. So I added the message to the `LoadWorkspaceDataForm` so that it can be shared by both cases.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -143,6 +148,77 @@
...
+PROJECT_GUID = 'R0001_1kg'
+ADD_DATA_UPDATED_ANVIL_VARIABLES = {
+    "key": "AnVIL_WES",
+    "value": json.dumps({
+        "active_projects": [PROJECT_GUID],
+        "vcf_path": "gs://test_bucket/test_path.vcf",
+        "project_path": "gs://seqr-datasets/v02/GRCh37/AnVIL_WES/{guid}/v1".format(guid=PROJECT_GUID),
+        "projects_to_run": [PROJECT_GUID] })
+}
+ADD_DATA_UPDATE_DAG_TASKS_RESP = {
```
ShifaSZ, line 160: Maybe these new test codes are unnecessary because they are testing the identical codes as creating a new Anvil project. What do you think?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -143,6 +148,77 @@
...
+                ],
+            "total_entries": 6
+        }
+
+AIRTABLE_TRACKING_URL = f'{MOCK_AIRTABLE_URL}/appUelDNM3BnWaR7M/AnVIL%20Seqr%20Loading%20Requests%20Tracking'
+
+
+def _setup_responses(responses, update_dag_tasks_resp):
+    # Set up api responses
+    responses.add(responses.POST, AIRTABLE_TRACKING_URL, status=400)
```
ShifaSZ, line 189: Maybe these changes are unnecessary because they are testing the identical codes as creating a new Anvil project. What do you think?

seqr/views/utils/test_utils.py
```diff
@@ -252,7 +252,7 @@ def get_initial_page_json(self, response):
         'test_user_manager@test.com': {
             "accessLevel": "WRITER",
             "pending": False,
-            "canShare": True,
+            "canShare": False,
```
ShifaSZ, line 255: It is changed for testing a manager user who doesn't have workspace permission for adding workspace data.

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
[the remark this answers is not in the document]
hanars: there should never be a success message shown on the create page. It should redirect to the project page before anything can be shown

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
hanars: okay, but we don't want to delay form closing. The modal should never automatically close. When the form is successfully submitted it just changes the content from a form to a success message and the modal stays open. I don't see a good reason to add all the complexity of connecting this form to the modal when it should never interact with the modal

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -193,29 +188,94 @@ def create_project_from_workspace(request, namespace, name):
...
+    """
+    Add data from an AnVIL workspace.
+
+    :param request: Django request object
+    :param project_guid: Django request object
+    :return success if no exceptions
+
+    """
+    project = get_project_and_check_permissions(project_guid, request.user, can_edit=True)
+    check_workspace_perm(request.user, CAN_EDIT, project.workspace_namespace, project.workspace_name, can_share=True)
```
hanars, line 207: this is redundant, `get_project_and_check_permissions` will call `has_workspace_perm` if there isn't other access, as will `check_workspace_perm` so there is no case where `check_workspace_perm` is True but `get_project_and_check_permissions` fails. It would be better to just get the project directly without checking permission and the call `check_workspace_perm`

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -193,29 +188,94 @@ def create_project_from_workspace(request, namespace, name):
...
+    json_records = load_uploaded_file(request_json['uploadedFileId'])
+    pedigree_records, _ = parse_pedigree_table(json_records, 'uploaded pedigree file', user=user, fail_on_warnings=True)
+
+    missing_samples = [record['individualId'] for record in pedigree_records
+                       if record['individualId'] not in request_json['vcfSamples']]
+
+    error = 'The following samples are included in the pedigree file but are missing from the VCF: {}'.format(
+                ', '.join(missing_samples)) if missing_samples else None
+
+    if error:
```
hanars, line 248: it would be cleaner to do `if missing_samples` here and define the error message in the if block

seqr/views/utils/test_utils.py
hanars: wouldn't it make more sense for the manager user to be able to share and the collaborator user to not be able to share? If that makes the tests way more complicated no need to change it, but if it can be done either way then swapping it makes sense

ui/pages/Project/components/EditDatasetsButton.jsx
```diff
@@ -170,20 +171,33 @@ const PANES = [
...
-))
+const EditDatasetsButton = React.memo(({ project, user }) => {
+  const showLoadWorkspaceData = !project.isAnalystProject && project.canEdit
+  const title = showLoadWorkspaceData ? 'Request Additional Data from Anvil Workspace' : 'Datasets'
+  const text = showLoadWorkspaceData ? 'Request Additional Data' : 'Edit Datasets'
+  return (
+    (user.isDataManager || user.isPm || showLoadWorkspaceData) ? (
+      <Modal
+        modalName={MODAL_NAME}
+        title={title}
```
hanars, line 182: it would be more readable to define the variable inline here rather than as a constant above. Same for `text`. In geenral, if you are only using a variable once its more readable not to make it a separate variable

ui/shared/components/form/FormWizard.jsx
```diff
@@ -30,24 +32,56 @@ class FormWizard extends React.PureComponent {
...
 
+  setSubmitSucceeded = responseJson => this.setState({ responseJson, formSubmitSucceeded: true })
+
   onFormSubmit = (values) => {
-    const { onSubmit } = this.props
-    const { asyncValues } = this.state
-    return onSubmit({ ...asyncValues, ...values })
+    const { onSubmit, onClose } = this.props
+    const { asyncValues, formSubmitSucceeded, responseJson } = this.state
+    if (!formSubmitSucceeded) {
```
hanars, line 40: the change to this function add unneeded complexity and make it incredibly hard to read. There should be no changes to how submission on the create page works. The change here should be conditional on whether or not there is a success message to show, which as I've said only happens on the add data form not the create form. All you need here is:
```
let submitPromise = onSubmit({ ...asyncValues, ...values })
if (successMessage) {
  submitPromise = submitPromise.then(this.setSubmitSucceeded)
}
return submitPromise
```

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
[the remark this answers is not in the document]
ShifaSZ: okay

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -93,27 +93,48 @@ const DATA_BUCK_FIELD = {
...
+
+const postSubmitValues = formatUrl => postWorkspaceValues(
+  'submit', ({ uploadedFile, ...values }) => ({ ...values, uploadedFileId: uploadedFile.uploadedFileId }), formatUrl,
+)
+
+const createProjectFromWorkspace = postSubmitValues()
+
+const addDataFromWorkspace = postSubmitValues(({ projectGuid }) => (`/api/project/${projectGuid}/add_workspace_data`))
+
+// eslint-disable-next-line no-return-assign
```
hanars, line 113: we didn't need to supress linting before, so theres no reason we should have to now

hanars (changes requested):
I haven't had tie to review the tests, but here is the rest of my review and I'll review the tests later

seqr/views/apis/anvil_workspace_api_tests.py
hanars: I don't fully understand the question. You need to test both the create endpoint and the add data endpoint. While they do call the same helper function, we need to test the full behavior for both. I think abstracting this into a helper function is a good way to do that, although I would give it a more descriptive name like `_setup_dag_trigger_responses`

hanars: also, since ` _test_responses` is a function on the class it might be better to make this a class function as well, just a `@staticmethod`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -311,6 +387,49 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
         self.assertEqual(response.url,
                          '/login/google-oauth2?next=/api/create_project_from_workspace/my-seqr-billing/anvil-no-project-workspace1/validate_vcf')
 
+    def _test_responses(self, responses, updated_anvil_variables, update_dag_tasks_resp, sample_number, guid):
```
hanars, line 390: maybe a more descriptive name like `_assert_expected_dag_trigger_requests`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -552,6 +598,118 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
         self.assertEqual(response.status_code, 302)
         self.assertEqual(response.url, '/login/google-oauth2?next=/api/create_project_from_workspace/my-seqr-billing/anvil-no-project-workspace2/submit')
 
+    @mock.patch('seqr.models.Project._compute_guid', lambda project: f'P_{project.name}')
```
hanars, line 601: why would we need to mock project guid, you should never create a project in this tes?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -552,6 +598,118 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+    @mock.patch('seqr.views.apis.anvil_workspace_api.ANVIL_LOADING_DELAY_EMAIL', None)
+    @mock.patch('seqr.views.apis.anvil_workspace_api.datetime')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.logger')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.load_uploaded_file')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.safe_post_to_slack')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs')
+    @mock.patch('seqr.views.apis.anvil_workspace_api.tempfile.NamedTemporaryFile')
+    @responses.activate
+    def test_add_workspace_data(self, mock_tempfile, mock_mv_file, mock_slack, mock_send_email, mock_load_file,
```
hanars, line 616: a ton of this function is copy  pasted from the create test. It would be better to make more helper functions and share more code. I think a good idea might be to actually put these 2 functions in a different test class than the other test, something like `class LoadAnvilDataAPITest(AnvilAuthenticationTestCase):` . Then you could add all the shared mocks to the class instead of both functions, you could write a custom `setUp` function that does the responses set up, and the scope for the helper functions would be neater

hanars (changes requested):

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
ShifaSZ: Okay, I removed the modal. For creating a new project, it will jump to the project page after submission. For adding new data, the modal keeps opening after displaying the success message until the user clicks on the `x` of the modal or anywhere outside of the modal.

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: I see what you mean. Change to get the project directly.

seqr/views/utils/test_utils.py
ShifaSZ: Yes. It makes more sense but is not necessary. Before, I was trying to create a scenario where a user passes the local permission check but doesn't have enough permission on Anvil. So I had to change the manager user. Since we have changed the permission check, I can use the collaborator user rather than the manager user to do the test.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I think `_setup_dag_trigger_responses` can be a class function but `_test_responses` not because it needs to access object data.

hanars: I was saying that all the functions should be nested in the class, not top-level in the test file. I don't care whether they are staticmethod, classmethod, or regular methods with access to self use whatever is appropriate in each case

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -93,27 +93,45 @@ const DATA_BUCK_FIELD = {
...
+    formatUrl ? formatUrl(values) : `/api/create_project_from_workspace/${workspaceNamespace}/${workspaceName}/${path}`,
+    onSuccess,
+  ).post(formatVals ? formatVals(values) : values)
+)
+
+const postSubmitValues = formatUrl => postWorkspaceValues(
+  'submit', ({ uploadedFile, ...values }) => ({ ...values, uploadedFileId: uploadedFile.uploadedFileId }), formatUrl,
+)
+
+const createProjectFromWorkspace = postSubmitValues()((resp) => { window.location.href = `/project/${resp.projectGuid}/project_page` })
```
hanars, line 109: its more readable if you leave `resp` as `responseJson`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -311,28 +312,114 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+    @mock.patch('seqr.views.apis.anvil_workspace_api.datetime')
     @mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.safe_post_to_slack')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.tempfile.NamedTemporaryFile')
     @responses.activate
-    def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_slack,
-                                           mock_send_email, mock_load_file, mock_api_logger, mock_datetime,
-                                           mock_airtable_logger, mock_utils_logger):
+    def test_create_project_from_workspace_loading_delay_email(
```
hanars, line 323: You misunderstood my last PR feedback. The `LoadAnvilDataAPITest` class needs to include this test as well as the new test so they share behavior, and all the shared behavior is scoped to that class level. It is also not at all okay to change the fixtures for this test. It is crucial to test that the behavior for this endpoint has not changed

ui/shared/components/form/FormWizard.jsx
```diff
@@ -30,24 +31,31 @@ class FormWizard extends React.PureComponent {
...
+    const { pages, onSubmit, successMessage, ...props } = this.props
+    const { pageIndex, formSubmitSucceeded } = this.state
 
     const { fields, onPageSubmit } = pages[pageIndex]
 
-    const formProps = pageIndex === pages.length - 1 ? { onSubmit: this.onFormSubmit } : {
+    const formProps = (pageIndex === pages.length - 1) ? {
+      onSubmit: this.onFormSubmit,
+      onSubmitSucceeded: this.setSubmitSucceeded,
+      successMessage: formSubmitSucceeded ? successMessage : null,
```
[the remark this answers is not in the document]
ShifaSZ: Hmm..., I didn't think about that.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I can move the `test_create_project_from_workspace_loading_delay_email` into the `LoadAnvilDataAPITest` class and add similar test for the `add_workspace_data` API. I don't fully understand what you mean by `to change the fixtures for this test`. Which fixtures are changed in my tests?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -311,28 +312,74 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+@mock.patch('seqr.views.apis.anvil_workspace_api.BASE_URL', 'http://testserver/')
+@mock.patch('seqr.views.utils.permissions_utils.logger')
+@mock.patch('seqr.views.utils.airtable_utils.logger')
+@mock.patch('seqr.views.apis.anvil_workspace_api.load_uploaded_file')
+@mock.patch('seqr.views.apis.anvil_workspace_api.safe_post_to_slack')
+@mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs')
+@mock.patch('seqr.views.apis.anvil_workspace_api.tempfile.NamedTemporaryFile')
+@mock.patch('seqr.views.apis.anvil_workspace_api.logger')
+@mock.patch('seqr.views.apis.anvil_workspace_api.datetime')
+@mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email')
```
hanars, line 375: we end up passing a lot of these mocks arounds as arguments to various functions, which works but is hard to read/maintain. I think it would be better to define the patchers in the `setUp` method similar to what we do for `AnvilAuthenticationTestCase` so that the mocks are actually class properties. For example
```
        patcher = mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email'')
        self.mock_send_email = patcher.start()
        self.addCleanup(patcher.stop)
```
Then `self._assert_sent_slack_message` would no longer take `mock_send_email` as an argument but would instead access `self.mock_send_email` from inside the function call

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -412,40 +452,191 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+
+    @mock.patch('seqr.views.apis.anvil_workspace_api.ANVIL_LOADING_DELAY_EMAIL', None)
+    @responses.activate
+    def test_add_workspace_data(self, mock_send_email, mock_datetime, mock_api_logger, mock_tempfile, mock_mv_file,
+                                mock_slack, mock_load_file, mock_airtable_logger, mock_utils_logger):
+        self._setup_dag_trigger(responses, ADD_DATA_UPDATE_DAG_TASKS_RESP)
+
+        # Test requesting to load data from a workspace without an existing project
+        url = reverse(add_workspace_data, args=['no_PROJECT1_GUID'])
+        self.client.force_login(self.collaborator_user)
```
hanars, line 504: Why aren't you `self.check_collaborator_login` in this function? We always start our tests with a check login call on a real URL, and then do the error tests

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -412,40 +452,191 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(
+            response.json()['error'],
+            'In order to add new data to this project, new samples must be joint called in a single VCF with all previously'
+            ' loaded samples. The following samples were previously loaded in this project but are missing from the VCF:'
+            ' HG00731, HG00732, HG00733, NA19675_1, NA20870, NA20874')
+
+        # Test a valid operation
+        mock_load_file.return_value = LOAD_SAMPLE_DATA
```
hanars, line 536: this is the same as line 526 above, what does assigning it again do?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -412,40 +452,191 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+    def _test_missing_required_fields(self, url, fields, workspace_name):
+        # Test missing required fields in the request body
+        response = self.client.post(url, content_type='application/json', data=json.dumps({}))
+        self.assertEqual(response.status_code, 400)
+        field_str = ', '.join(fields)
+        self.assertEqual(response.reason_phrase, f'Field(s) "{field_str}" are required')
+        self.mock_get_ws_access_level.assert_called_with(self.manager_user, TEST_WORKSPACE_NAMESPACE, workspace_name)
+
+    def _test_sample_data_error(self, mock_load_file, url):
+        """ Test parsing sample data error """
```
hanars, line 590: you are inconsistently using docstrings and regular comments for these helper functions. In general, small helper functions are self documenting and don't need any sort of comment here at all. Unless there is information contained in a comment or docstring that adds more detail than the name of the helper function, it is better to take the comment out completely

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -454,61 +645,45 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
-        # Test triggering dag exception
-        responses.replace(responses.GET,
-                      '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/dagRuns'.format(MOCK_AIRFLOW_URL),
-                      json=DAG_RUNS_RUNNING)
-
-        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
-        self.assertEqual(response.status_code, 200)
-        project2 = Project.objects.get(workspace_namespace=TEST_WORKSPACE_NAMESPACE, workspace_name=TEST_NO_PROJECT_WORKSPACE_NAME2)
-
+    def _assert_loading_req_exceptions(self, mock_api_logger, mock_airtable_logger, mock_slack, mock_send_email,
```
hanars, line 678: it would be more readable if you use `request` instead of `req`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -540,49 +716,56 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
-        # Remove created project to allow future requests
-        project = Project.objects.get(
-            workspace_namespace=TEST_WORKSPACE_NAMESPACE, workspace_name=TEST_NO_PROJECT_WORKSPACE_NAME)
-        project.workspace_name = None
-        project.save()
-
         # Test after anvil email date
         mock_datetime.now.side_effect = lambda: datetime(2021, 9, 1, 0, 0, 0)
-        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
+        response = self.client.post(url2, content_type='application/json', data=json.dumps(request_body2))
```
hanars, line 768: why does this need a different url now? Before this test worked if you called it with the same url

hanars (changes requested):
One big concern I have is after the user exits the modal, they will be on the project page but will they actually see all the newly created individuals and families? I think the request actually needs to return all the updated families and individuals and we need to update the redux state to show them

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I need to login first to test the behavior for the non-existing project. But I found that I should do the `insufficient Anvil permissions` test first. Then no explicit `force_login` is required.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: It is redundant. I'll delete it.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Okay

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Because we use two URLs to use two projects for the `not yet anvil email date` and `after anvil email date` tests, it can avoid some cleaning up of the project before starting the second test.

hanars: I'd really rather not change test behavior unless actual behavior changed for this PR, so we can ensure that the functionality dud not actually change

ShifaSZ:
> One big concern I have is after the user exits the modal, they will be on the project page but will they actually see all the newly created individuals and families? I think the request actually needs to return all the updated families and individuals and we need to update the redux state to show them

You are right. Do we also need to return the updated notes for the individuals in the pedigree file?

ui/shared/components/form/FormWizard.jsx
```diff
@@ -30,31 +31,39 @@ class FormWizard extends React.PureComponent {
...
+
     const backButtonProps = pageIndex === 0 ? {} : {
       onCancel: this.navigateBack,
       cancelButtonText: 'Back',
       cancelButtonIcon: 'angle double left',
     }
 
-    return (
+    return (formSubmitSucceeded ? (
+      <b>{successMessage}</b>
```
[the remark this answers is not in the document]
ShifaSZ: The screenshot after successfully submitted.
image

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I can create two separate sharing tests for the `before` and `after` email dates. And then keep the test behavior unchanged for the `create__project_from_workspace` API.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -311,28 +312,74 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+@mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs')
+@mock.patch('seqr.views.apis.anvil_workspace_api.tempfile.NamedTemporaryFile')
+@mock.patch('seqr.views.apis.anvil_workspace_api.logger')
+@mock.patch('seqr.views.apis.anvil_workspace_api.datetime')
+@mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email')
+class LoadAnvilDataAPITest(AnvilAuthenticationTestCase):
+    fixtures = ['users', 'social_auth', '1kg_project']
+
+    @staticmethod
+    def _setup_dag_trigger(responses, update_dag_tasks_resp):
```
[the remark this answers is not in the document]
ShifaSZ: The requests to the endpoint use the `GET` method. And they are identical. There are two consecutive accesses to the same endpoint in each test. The second accesses in the two tests have different response data. I will move the second one out from the `setUp` and add it to the beginning of each test.

hanars:
> You are right. Do we also need to return the updated notes for the individuals in the pedigree file?

You should return the full family json and individual json for all the families/individuals in the pedigree file

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -311,28 +312,74 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
         # Set up api responses
-        airtable_tracking_url = f'{MOCK_AIRTABLE_URL}/appUelDNM3BnWaR7M/AnVIL%20Seqr%20Loading%20Requests%20Tracking'
-        responses.add(responses.POST, airtable_tracking_url, status=400)
+        responses.add(responses.POST, AIRTABLE_TRACKING_URL, status=400)
```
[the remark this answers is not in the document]
ShifaSZ: There is actually another use of the constant. But since that one is for result assertion and it should be replaced with a literal string, so I remove the constant.

PR #2849, opened 2022-07-19
remove unneccessary x-linked query

hanars:
The x-linked recessive query is more restrictive than the homozygous recessive query, so searching "X-LINKED OR HOM_RECESSIVE" is actually the same results as just HOM_RECESSIVE. This removes the unnecessary query to make the logic cleaner/ easier to read

### DIFF
--- a/seqr/utils/elasticsearch/es_utils_tests.py
+++ b/seqr/utils/elasticsearch/es_utils_tests.py
@@ -1088,28 +1088,14 @@ RECESSIVE_INHERITANCE_QUERY = {
             {'bool': {
                 '_name': 'F000002_2',
                 'must': [
-                    {'bool': {
-                        'should': [
-                            {'bool': {
-                                'must_not': [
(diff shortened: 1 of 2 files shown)


seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -1088,28 +1088,14 @@
...
-                        ]
+                        {'bool': {
+                            'must_not': [
+                                {'term': {'samples_no_call': 'HG00732'}},
+                                {'term': {'samples_num_alt_2': 'HG00732'}},
+                                {'term': {'samples_no_call': 'HG00733'}},
+                                {'term': {'samples_num_alt_2': 'HG00733'}}
+                            ],
+                            'must': [{'term': {'samples_num_alt_2': 'HG00731'}}]
                     }},
```
ShifaSZ, line 1099: You may want to align the indents for these lines.

ShifaSZ (approved):

PR #2850, opened 2022-07-20
Moi selector bug

hanars:


ui/shared/components/panel/search/PaMoiSelector.jsx
```diff
@@ -56,7 +59,8 @@ export default class PaMoiDropdown extends React.PureComponent {
         inline
         onChange={this.handleMOIselect}
         placeholder="Showing all MOIs as listed in Panel App"
-        options={this.moiOptions()}
+        disabled={disabled}
+        options={disabled ? EMPTY_LIST : this.moiOptions()}
```
ShifaSZ, line 63: It seems to me that disabling the dropdown list doesn't fix the bug. I think the list still needs to be shown. The problem is why the `locus.locusList.items` becomes empty after a gene list is selected without clearing the previous one.

ShifaSZ (changes requested):

ui/shared/components/panel/search/PaMoiSelector.jsx
hanars: the issue was about the component not updating on props updates because the should update functionality was not detecting a change to the options. This actually does fix the issue because it forces the component to re-render once the detailed items are loaded

hanars: the reason the items becomes empty briefly is because the gene list is in the process of being loaded from the server, so the state goes "items from old list" -> "no items (currently loading)" -> "items from new list"

ShifaSZ (approved):

PR #2865, opened 2022-07-25
migrate servctl to python 3

hanars:


deploy/servctl_utils/shell_utils.py
```diff
@@ -45,26 +50,28 @@ def run(command,
...
+                    line_buffer.write(out)
                     if out.endswith('\r') or (out.endswith('\n') and previous_is_slash_r):
                         sys.stdout.write(line_buffer.getvalue())
                         sys.stdout.flush()
                         line_buffer = StringIO()
                         previous_is_slash_r = True
                     elif out.endswith('\n'):
-                        logger.info(line_buffer.getvalue().rstrip('\n'))
+                        line_content = line_buffer.getvalue().rstrip('\n')
+                        if all(ignore_str not in line_content for ignore_str in ALWAYS_IGNORE_WARNINGS):
```
ShifaSZ, line 73: You may want to make a helper function for checking a line that doesn't contain the ignored warnings.

ShifaSZ (changes requested):

sjahl (approved):
This looks good to me. I slightly agree on the change Shifa requested, but I don't have a strong preference given that we're trying to get rid of servctl shortly.

hanars:
I agree with Steve, since we are getting rid of this so soon I'd rather not invest effort into maintainability

PR #2867, opened 2022-07-27
Add anvil file dropdown selection.

ShifaSZ:
As the title.


seqr/utils/file_utils.py
```diff
@@ -80,3 +80,17 @@ def mv_file_to_gs(local_path, gs_path, user=None):
...
         raise Exception('Run command failed: ' + ' '.join(errors))
+
+
+def get_gs_file_list(gs_path, user=None):
+    if not is_google_bucket_file_path(gs_path):
+        raise Exception('A Google Storage path is expected.')
+    gs_path = gs_path.rstrip('/')
+    process = _run_gsutil_command('ls', f'{gs_path}/**', user=user)
+    if process.wait() != 0:
+        errors = [line.decode('utf-8').strip() for line in process.stdout]
```
hanars, line 91: the error handling here looks copy-pasted from `mv_file_to_gs`, you should move it to a helper function and share it instead of copying

seqr/utils/file_utils.py
```diff
@@ -80,3 +80,17 @@ def mv_file_to_gs(local_path, gs_path, user=None):
...
+        raise Exception('Run command failed: ' + ' '.join(errors))
+    for line in process.stdout:
+        line = line.decode('utf-8').rstrip('\n')
+        if line.startswith(gs_path):
```
hanars, line 95: out of curiosity, what are the other lines it would be returning?

hanars (changes requested):
started the review, but based on the ticket discussion it looks like we are going a different direction for the component which will require a new endpoint, so holding off on the rest of the review for now

seqr/utils/file_utils.py
ShifaSZ: There could be some empty lines and something like that below.
```
Updates are available for some Cloud SDK components.  To install them,
please run:
  $ gcloud components update
```

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -125,6 +126,17 @@ def grant_workspace_access(request, namespace, name):
 
     return create_json_response({'success': True})
 
+
+@anvil_workspace_access_required(meta_fields=['workspace.bucketName'])
+def get_anvil_vcf_list(request, namespace, name, workspace_meta):
+    bucket_name = workspace_meta['workspace']['bucketName']
+    bucket_path = 'gs://{bucket}'.format(bucket=bucket_name.rstrip('/'))
+    data_path_list = [path.replace(bucket_path, '', 1) for path in get_gs_file_list(bucket_path)
```
hanars, line 134: is there a reason you are adding the `count` parameter in the `replace` call here? In general its best to just use default behavior for optional parameters unless there is a strong reason not to, as it keeps things more readable

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -355,6 +355,22 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+    def test_get_anvil_vcf_list(self, mock_get_file_list, mock_utils_logger):
+        # Requesting to load data from a workspace without an existing project
+        url = reverse(get_anvil_vcf_list, args=[TEST_WORKSPACE_NAMESPACE, TEST_WORKSPACE_NAME1])
+        self.check_manager_login(url, login_redirect_url='/login/google-oauth2')
+        mock_utils_logger.warning.assert_called_with('User does not have sufficient permissions for workspace {}/{}'
+                                                     .format(TEST_WORKSPACE_NAMESPACE, TEST_WORKSPACE_NAME1),
+                                                     self.collaborator_user)
+
+        # Test valid operation
+        mock_get_file_list.return_value = ['gs://test_bucket/test.vcf', 'gs://test_bucket/data/test.vcf.gz']
```
hanars, line 368: you should add some non-VCF files to the mock list to test that they are properly filtered out

ui/shared/components/form/AnvilFileSelector.jsx
```diff
@@ -0,0 +1,53 @@
...
+        this.setState({ loading: false, dataPathList: responseJson.dataPathList || [] })
+      },
+      (e) => {
+        this.setState({ loading: false, dataPathList: [e.message] })
+      }).get()
+  }
+
+  render() {
+    const { dataPathList, loading } = this.state
+    const options = dataPathList?.map(file => ({ name: file, value: file })) || []
```
hanars, line 35: We need to show the user an error if there are no valid files to choose from - right now they would just have a dropdown with no options to choose from which would be a kind of bad experience. Lets do the following:
```
<Message
    error
    header="No Joint Called VCF found in workspace"
    content="There are no joint called VCFs in the Files section of this workspace. VCFs must have a .vcf, .vcf.gz, or .vcf.bgz file extension. Please add a VCF to your workspace before proceeding with loading."
  />
```

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -89,7 +90,8 @@ const DATA_BUCK_FIELD = {
   name: 'dataPath',
   label: 'Path to the Joint Called VCF',
   labelHelp: 'File path for a joint called VCF available in the workspace "Files".',
-  placeholder: '/path-under-Files-of-the-workspace',
+  component: AnvilFileSelector,
+  placeholder: 'Click to select',
```
hanars, line 94: we don't need a placeholder here, people know how dropdowns work

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: I use `1` for the `count` parameter to avoid accidentally replacing more than one occurrence of the pattern. But maybe you are right. It doesn't look like that `gs://` could have more occurrences in a path. I can take out the `count`.

ui/shared/components/form/AnvilFileSelector.jsx
```diff
@@ -0,0 +1,63 @@
...
+        this.setState({ loading: false, dataPathList: responseJson.dataPathList })
+        if (responseJson.dataPathList?.length === 0) {
+          this.setState({
+            errorHeader: 'No Joint Called VCF found in workspace',
+            error: 'There are no joint called VCFs in the Files section of this workspace. VCFs must have a .vcf, .vcf.gz, or .vcf.bgz file extension. Please add a VCF to your workspace before proceeding with loading.',
+          })
+        }
+      },
+      (e) => {
+        this.setState({ loading: false, errorHeader: 'Exception Error', error: e.message })
```
hanars, line 38: "Exception Error" is not user friendly. Change this to "Error loading workspace files"

ui/shared/components/form/AnvilFileSelector.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+  load = () => {
+    const { namespace, name } = this.props
+    this.setState({ loading: true })
+    new HttpRequestHelper(`/api/create_project_from_workspace/${namespace}/${name}/get_vcf_list`,
+      (responseJson) => {
+        this.setState({ loading: false, dataPathList: responseJson.dataPathList })
+        if (responseJson.dataPathList?.length === 0) {
+          this.setState({
+            errorHeader: 'No Joint Called VCF found in workspace',
```
hanars, line 32: The capitalization here is inconsistent. Either capitalize all the words, or just the first word

hanars (changes requested):

ShifaSZ:
Some screenshots.
No VCF file found
image

Something wrong
image

After loading the file list successfully
image

hanars (approved):
Looks good, just make sure you merge your PRs in the correct order so everything ends up in dev

PR #2875, opened 2022-08-08
Update the `appVersion` in the seqr Helm Chart when there is a push the branches [dev, master]

ShifaSZ:
The newly added Github workflow creates a PR in the [seqr-helm repo](https://github.com/broadinstitute/seqr-helm) to update the `appVersion` in the seqr Helm Chart when there is a push the release branch. Here is an example of the PR: https://github.com/broadinstitute/seqr-helm/pull/9

### DIFF
--- a/.github/workflows/update-seqr-helm-version.yml
+++ b/.github/workflows/update-seqr-helm-version.yml
@@ -8,23 +8,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: get seqr version
+      - name: collect seqr version info
         id: seqr
         run:
+          echo "::set-output name=seqr_version::$(grep "SEQR_VERSION[ ]*=[ ]*'.*'" ./settings.py | sed -n "s/^.*'\(.*\)'/\1/p")
           echo "::set-output name=commit_id::$(git rev-parse --short HEAD)"
       - uses: actions/checkout@v3
         with:
           repository: broadinstitute/seqr-helm
       - name: Update appVersion in the seqr HelmChart Chart.yaml
         uses: fjogeleit/yaml-update-action@main
+        pre:
+          echo "SEQR_VERSION=${{ steps.seqr.outputs.seqr_version }}-${{ steps.seqr.outputs.commit_id }}" >> $GITHUB_ENV
         with:
           valueFile: 'charts/seqr/Chart.yaml'
           propertyPath: 'appVersion'
-          value: ${{ steps.seqr.outputs.commit_id }}
+          value: ${{ env.SEQR_VERSION }}
           repository: broadinstitute/seqr-helm
           masterBranchName: main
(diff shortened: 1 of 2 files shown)


.github/workflows/update-seqr-helm-version.yml
```diff
@@ -0,0 +1,39 @@
+name: Update seqr Version
+on:
+  push:
+    branches:
+      - seqr-helm-version
```
ShifaSZ, line 5: It needs to be updated to `master` eventually.

.github/workflows/update-seqr-helm-version.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/checkout@v3
+        with:
+          repository: broadinstitute/seqr-helm
+          ref: main
+      - name: Update appVersion in seqr Chart file
+        uses: mikefarah/yq@v4.22.1
+        with:
+          cmd: >
+            yq -i '.appVersion = "${{ env.SEQR_VERSION }}"' charts/seqr/Chart.yaml
+      - name: Create Pull Request
```
ShifaSZ, line 25: Should we push to the main branch directly rather than creating a PR so that it is fully automated?

.github/workflows/update-seqr-helm-version.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: get seqr version
+        run: |
+          seqr_version=$(grep "SEQR_VERSION[ ]*=[ ]*'.*'" ./settings.py | sed -n "s/^.*'\(.*\)'/\1/p")
+          commit_id=$(git rev-parse --short HEAD)
+          echo "SEQR_VERSION=$seqr_version-$commit_id" >> $GITHUB_ENV
```
sjahl, line 15: This is cool, I didn't know you could add items to the runner's environment variables like this!

.github/workflows/update-seqr-helm-version.yml
sjahl: Yes, I think pushing to main (the default branch is "main" rather than "master" on this repo) is probably the right thing to do. If we had a fully automated rollout to prod, then we'd probably want a PR, but I think the plan is to have production pin to a specific commit of the helm chart, so it should be fine to push right to main for this.

sjahl (approved):

sjahl:
Looks good to me so far, I can re-review once the suggestion to push to main rather than opening a PR is done

sjahl (commented):
This all looks pretty good! I have some work to do with regards to automating updates to the dev environment, and I'll need to use this branch/PR for testing before we merge. I'll take this one off your hands @ShifaSZ

hanars:
@sjahl should we merge this?

sjahl:
No, we should probably close it actually. I'm going to use some of what Shifa put together, but it's going to end up as part of different github workflow configurations

PR #2887, opened 2022-08-16
Anvil new data

ShifaSZ:
This new feature allows requesting new datasets to be loaded into seqr.


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -405,32 +477,144 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+
+        # Test requesting to load data from a workspace without an existing project
+        url = reverse(add_workspace_data, args=['no_PROJECT1_GUID'])
+        response = self.client.post(url, content_type='application/json', data=json.dumps({}))
+        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response.json()['error'], 'Project matching query does not exist.')
+
+        # Test missing required fields in the request body
+        url = reverse(add_workspace_data, args=[PROJECT1_GUID])
+        self.client.force_login(self.manager_user)
```
hanars, line 532: why are we using `check_collaborator_login` but requiring a manager user here? shouldn't we just use `check_manager_login` instead?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -405,32 +477,144 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+            'In order to add new data to this project, new samples must be joint called in a single VCF with all previously'
+            ' loaded samples. The following samples were previously loaded in this project but are missing from the VCF:'
+            ' HG00731, HG00732, HG00733, NA19675_1, NA20870, NA20874')
+
+        # Test a valid operation
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_ADD_DATA))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertEqual(sorted(response_json.keys()), ['familiesByGuid', 'familyNotesByGuid', 'individualsByGuid'])
+        self.assertEqual(sorted(response_json['individualsByGuid'].keys()), ['I0000019_hg00735', 'I000001_na19675', 'I000002_na19678'])
```
hanars, line 555: this is an anti-pattern, for keys we always use `assertSetEqual` instead of sorted lists

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -405,32 +477,144 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+        self.assertEqual(sorted(response_json['familiesByGuid'].keys()), ['F000001_1', 'F000015_21'])
+        self.assertEqual(sorted(response_json['familyNotesByGuid'].keys()), ['FAN000004_21_c_a_new_family'])
+        self.mock_api_logger.error.assert_not_called()
+
+        self.mock_tempfile.assert_called_with(mode='wb', delete=False)
```
hanars, line 560: this seems like its copy-pasted from the above test, so it should be a helper function

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -405,32 +477,144 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+
+        self._assert_dag_trigger_requests(ADD_DATA_UPDATED_ANVIL_VARIABLES, 10, PROJECT1_GUID)
+
+        self._assert_sent_slack_message(Project.objects.get(guid=PROJECT1_GUID), "GRCh37")
+
+        url = reverse(add_workspace_data, args=[PROJECT2_GUID])
+        self._test_mv_file_and_triggering_dag_exception(url, {'guid': PROJECT2_GUID}, PROJECT2_SAMPLES, 'GRCh37', REQUEST_BODY_ADD_DATA2)
+
+        # Test logged in locally
+        remove_token(self.manager_user)  # The user will look like having logged in locally after the access token is removed
```
hanars, line 577: this seems to be copy pasted from the above test, so it should be a helper function

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -444,71 +628,76 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
         self.assertEqual(responses.calls[3].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
         self.assertEqual(responses.calls[3].response.json(), UPDATE_DAG_TASKS_RESP)
 
+        call_cnt = 4
```
hanars, line 631: this would be cleaner if you didn't abbreviate t and also set it explicitly here instead of in the if block: 
```
call_count = 5 if test_add_data else 4
```

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -444,71 +628,76 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
         # trigger dag
-        self.assertEqual(responses.calls[4].request.url, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/dagRuns'.format(MOCK_AIRFLOW_URL))
-        self.assertEqual(responses.calls[4].request.method, 'POST')
-        self.assertDictEqual(json.loads(responses.calls[4].request.body), {})
-        self.assertEqual(responses.calls[4].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+        self.assertEqual(responses.calls[call_cnt].request.url, '{}/api/v1/dags/seqr_vcf_to_es_AnVIL_WES_v0.0.1/dagRuns'.format(MOCK_AIRFLOW_URL))
+        self.assertEqual(responses.calls[call_cnt].request.method, 'POST')
+        self.assertDictEqual(json.loads(responses.calls[call_cnt].request.body), {})
+        self.assertEqual(responses.calls[call_cnt].request.headers['Authorization'], 'Bearer {}'.format(MOCK_TOKEN))
+        call_cnt += 1
```
hanars, line 644: this makes the latter part of the  test really hard to read - it would be better if you just explicitly use `call_count + 1`

ui/shared/components/form/FormWizard.jsx
```diff
@@ -30,31 +31,39 @@ class FormWizard extends React.PureComponent {
...
+
     const backButtonProps = pageIndex === 0 ? {} : {
       onCancel: this.navigateBack,
       cancelButtonText: 'Back',
       cancelButtonIcon: 'angle double left',
     }
 
-    return (
+    return (formSubmitSucceeded ? (
+      <b>{successMessage}</b>
```
hanars, line 65: This I think needs better formatting. I think using a Header would be good to get some padding and also a success icon - let me know what you think of this:
```
<Header
    icon="check circle"
    content="Request Submitted"
    subheader={successMessage}
  />
```

hanars (changes requested):

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -193,29 +189,110 @@ def create_project_from_workspace(request, namespace, name):
...
+        return create_json_response({
+            'error': 'In order to add new data to this project, new samples must be joint called in a single VCF with all previously loaded samples.'
+                     ' The following samples were previously loaded in this project but are missing from the VCF: {}'.format(
+                ', '.join(sorted(missing_loaded_samples)))}, status=400)
+
+    updated_individuals, updated_families, updated_notes = _trigger_add_workspace_data(
+        project, pedigree_records, request.user, request_json['fullDataPath'], previous_samples.first().sample_type,
+        previous_loaded_ids=previous_loaded_individuals)
+
+    individuals_by_guid = {
```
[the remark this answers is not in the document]
ShifaSZ: Agree. Should we also share the `_update_and_parse_individuals_and_families` function from individual_api.py?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -405,32 +477,144 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
+        url = reverse(add_workspace_data, args=[PROJECT2_GUID])
+        self._test_mv_file_and_triggering_dag_exception(url, {'guid': PROJECT2_GUID}, PROJECT2_SAMPLES, 'GRCh37', REQUEST_BODY_ADD_DATA2)
+
+        # Test logged in locally
+        remove_token(self.manager_user)  # The user will look like having logged in locally after the access token is removed
+        response = self.client.post(url)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.url, f'/login/google-oauth2?next=/api/project/{PROJECT2_GUID}/add_workspace_data')
+
+    def _test_missing_fields_in_request(self, url, fields, workspace_name):
```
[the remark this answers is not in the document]
ShifaSZ: No. They can be one helper.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Okay, I didn't notice that `check_manager_login` also tests the collaborator doesn't have sufficient permissions.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: We don't support logging in locally under Anvil mode. So I'm going to remove these tests.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Move them into the common assertion helper function.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -540,55 +694,60 @@ def test_create_project_from_workspace(self, mock_tempfile, mock_mv_file, mock_s
...
-    @mock.patch('seqr.views.apis.anvil_workspace_api.mv_file_to_gs', lambda *args, **kwargs: True)
-    @mock.patch('seqr.views.apis.anvil_workspace_api.load_uploaded_file', lambda *args, **kwargs: LOAD_SAMPLE_DATA)
-    @mock.patch('seqr.views.apis.anvil_workspace_api.logger')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.datetime')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.send_html_email')
     @responses.activate
-    def test_create_project_from_workspace_loading_delay_email(
-            self, mock_send_email, mock_datetime, mock_api_logger, mock_utils_logger):
+    def test_create_project_from_workspace_loading_delay_email(self):
+        responses.reset()  # Bypass dag operations
```
hanars, line 707: I think it would be a more readable test if instead of erasing all the dag responses you explicitly added an error response from the first endpoint

seqr/views/utils/individual_utils.py
```diff
@@ -185,3 +186,24 @@ def get_parsed_feature(feature):
             feature_json[field] = feature[field]
 
     return feature_json
+
+
+def get_updated_response_json(updated_individuals, updated_families, updated_notes, user):
```
hanars, line 191: this is a very unclear function name. Maybe do something more like `get_updated_pedigree_json`

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -144,6 +171,11 @@ const LoadWorkspaceDataForm = React.memo(({ params }) => (
 
 LoadWorkspaceDataForm.propTypes = {
   params: PropTypes.object.isRequired,
+  onSubmit: PropTypes.func.isRequired,
+}
+
+const mapDispatchToProps = {
```
hanars, line 177: I don't like that we are connecting this component for both usages when it only needs to be connected for one of them. I think it would be better to leave this as a plain component and connect the component on the project page

hanars (changes requested):
Can you add an updated screenshot for the success screen

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I'm going to add another task query response to allow all normal dag operations.

ui/shared/components/panel/LoadWorkspaceDataForm.jsx
ShifaSZ: How about creating a connected component dedicated for loading a new dataset? It is a little tricky to connect the component on the project page while using a reducer function (addDataFromWorkspace) from this file.

hanars (changes requested):
This overall looks good to me, I just want to see a screenshot of what the success page looks like now

ShifaSZ:
The success page is as below.
image

hanars (approved):

PR #2893, opened 2022-08-17
Consent code

hanars:


seqr/views/apis/project_api_tests.py
```diff
@@ -93,6 +80,44 @@ def test_create_update_and_delete_project(self):
...
+        updated_project = Project.objects.get(guid=PROJECT_GUID)
+        self.assertEqual(updated_project.description, 'updated project description')
+        self.assertEqual(updated_project.genome_version, '37')
+        self.assertEqual(updated_project.workspace_name, expected_workspace_name)
+
+        # test consent code
+        response = self.client.post(update_project_url, content_type='application/json', data=json.dumps(
+            {'consentCode': 'G'}
+        ))
+        self.assertEqual(response.status_code, 403)
```
ShifaSZ, line 112: You may want to test the error message as well.

ui/shared/components/buttons/EditProjectButton.jsx
```diff
@@ -4,17 +4,27 @@ import PropTypes from 'prop-types'
...
+  { ...MATCHMAKER_CONTACT_URL_FIELD, name: 'mmeContactUrl' },
+].map(({ label, ...field }) => ({ ...field, label: `Matchmaker ${label}` }))
 
-const EDITABLE_FIELD_KEYS = ['projectGuid', ...MATCHMAKER_PROJECT_FIELDS.map(({ name }) => name)]
+const FIELD_LOOKUP = {
+  true: {
+    true: [...PM_EDITABLE_PROJECT_FIELDS, ...MATCHMAKER_PROJECT_FIELDS],
+    false: [...EDITABLE_PROJECT_FIELDS, ...MATCHMAKER_PROJECT_FIELDS],
+  },
+  false: { true: PM_EDITABLE_PROJECT_FIELDS, false: EDITABLE_PROJECT_FIELDS },
```
ShifaSZ, line 24: I was confused by the `true` and `false` keys at the beginning. It could be helpful by adding comments for an example usage such as `FIELD_LOOKUP[isMmeEnabled][isPm]`.

ShifaSZ (changes requested):

seqr/models.py
```diff
@@ -173,6 +173,9 @@ class Project(ModelWithGUID):
     can_view_group = models.ForeignKey(Group, related_name='+', on_delete=models.PROTECT)
 
     genome_version = models.CharField(max_length=5, choices=GENOME_VERSION_CHOICES, default=GENOME_VERSION_GRCh37)
+    consent_code = models.CharField(max_length=1, null=True, blank=True, choices=[
+        (c[0], c) for c in ['HMB', 'GRU', 'Other']
```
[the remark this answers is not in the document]
hanars: Not really, constants are useful when you are reusing strings but these are only ever used once here so a constant would not reduce the risk of typos and would just make this a bit les readable. And I have no idea what they stand for, I just took them from the ticket which said those were the codes we needed 🤷

seqr/views/apis/project_api_tests.py
hanars: So the way our middleware works for security reason is any 403 permission denied errors we log the message but only return the error "Permission Denied" to the user, we don't return the message. That code is tested pretty extensively already so I don't think adding another test for it here is valuable
https://github.com/broadinstitute/seqr/blob/master/seqr/utils/middleware.py#L109

ShifaSZ (approved):

PR #2896, opened 2022-08-19
add a dev release workflow for testing

sjahl:
This adds a docker build workflow that is triggered off of successful completions of the Unit Tests workflow on the seqr-helm-version branch. This will allow me to test our dev release workflow when I push changes to that PR/branch. Our normal dev/master docker builds are still in place, and this should have no effect on them.

This has to go into the master branch, because apparently workflow_run triggers only work if the workflow you're defining exists on the repo's default branch (see the Note [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)).

### DIFF
--- a/.github/workflows/dev-release.yaml
+++ b/.github/workflows/dev-release.yaml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.workflow_run.head_branch }}
 
       - name: authenticate to google cloud
         id: 'auth'
--- a/.github/workflows/dev-release.yaml
+++ b/.github/workflows/dev-release.yaml
@@ -1,0 +1,34 @@
+name: seqr dev release
+on:
+  workflow_run:
+    workflows: ["Unit Tests"]
+    types:
+      - completed
+    branches:
+      - seqr-helm-version # TODO: trigger this flow on dev
+
+permissions:
+  id-token: write
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-release.yaml
```diff
@@ -0,0 +1,36 @@
+name: seqr dev release
+on:
+  workflow_run:
+    workflows: ["Unit Tests"]
+    types:
+      - completed
+    branches:
+      - seqr-helm-version # TODO: trigger this flow on dev
```
hanars, line 8: should this TODO still be in here

hanars (commented):
Since this is going directly to master can you fill out the release checklist: https://github.com/broadinstitute/seqr-private/blob/master/release_pr_checklist.md

.github/workflows/dev-release.yaml
sjahl: yes, because we're not going to trigger this workflow on changes to dev yet

sjahl:
- [x] Any changes to the docker image have been vetted for potential vulnerabilities (N/A)
- [x] If any python requirements are updated, they are noted and here and will be updated before deploying: (N/A)
- [x] Any database migrations are noted here, and will be run before deploying: (N/A)
- [x] All major changes are recorded in the changelog, and the changelog has been updated to reflect the latest release (N/A)
- [x] Any new endpoints are explicitly tested to ensure they are only accessible to correctly permissioned users (N/A)
- [x] Infrastructure changes: 
  - [x] No changes to the required seqr infrastructure are included in this change 
   
  OR 
  
  - [ ] Any chages to non-seqr docker images have been built and pushed to the container registry
  - [ ] All these changes have been vetted for potential vulnerabilities

hanars (approved):

PR #2898, opened 2022-08-19
Add a comment to mito CN.

ShifaSZ:
Add a `comment` field to the mito cn entry in the `GENOTYPE_DETAIL`. Update the `formattedGenotypeDetails` to show the comment.

The screenshot is as below picture.
image

### DIFF
--- a/ui/shared/components/panel/variants/VariantIndividuals.jsx
+++ b/ui/shared/components/panel/variants/VariantIndividuals.jsx
@@ -193,7 +193,12 @@ const GENOTYPE_DETAILS = [
   { title: 'Filter', variantField: 'genotypeFilters', shouldHide: val => (val || []).length < 1 },
   { title: 'Phred Likelihoods', field: 'pl' },
   { title: 'Quality Score', field: 'qs' },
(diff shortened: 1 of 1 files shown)


ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -209,13 +214,19 @@ const SV_GENOTYPE_DETAILS = [
   },
 ]
 
+const Comment = styled.p`
```
hanars, line 217: Its really best to use a semantic ui component or existing styled component where possible as it ensures that there will be consistent styles throughout the application. Styled components should ideally only be used when complicated style requirements leave us no other options. In this case, I think using the `subheader` prop of small or tiny `Header` would be best, or if we wanted to keep it really simple you could use a `ColoredDiv` from the `StyledComponents` helpers. The ticket doesn't specify it has to be indented so I think the added complexity isn't really worth it

hanars (changes requested):

ui/shared/components/panel/variants/VariantIndividuals.jsx
ShifaSZ: Okay

hanars (approved):

PR #2904, opened 2022-08-22
add logging when ls fails

hanars:


seqr/views/apis/igv_api_tests.py
```diff
@@ -21,12 +21,18 @@ class IgvAPITest(AuthenticationTestCase):
...
     @mock.patch('seqr.views.apis.igv_api.safe_redis_get_json')
     @mock.patch('seqr.views.apis.igv_api.safe_redis_set_json')
-    def test_proxy_google_to_igv(self, mock_set_redis, mock_get_redis, mock_subprocess):
-        mock_subprocess.return_value.stdout = iter([b'token1\n', b'token2\n'])
-        mock_subprocess.return_value.wait.side_effect = [-1, 0, 0]
+    def test_proxy_google_to_igv(self, mock_set_redis, mock_get_redis, mock_subprocess, mock_file_logger):
+        mock_ls_subprocess = mock.MagicMock()
+        mock_access_token_subprocess = mock.MagicMock()
+        mock_subprocess.side_effect = [mock_ls_subprocess, mock_access_token_subprocess]
+        mock_ls_subprocess.stdout = iter([b'CommandException: One or more URLs matched no objects.'])
```
ShifaSZ, line 32: We should also add error logging to the operation for getting the access token. But maybe it is out of scope for this PR.

ShifaSZ (changes requested):

seqr/views/apis/igv_api_tests.py
hanars: way out of scope

ShifaSZ (approved):

PR #2912, opened 2022-08-23
Update the frequency display for the mitochondria variants.

ShifaSZ:


ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -48,7 +48,7 @@ const getFreqLinkPath = ({ chrom, pos, variant, value }) => {
   } else if (isRegion) {
     const posInt = parseInt(pos, 10)
     const endOffset = variant.end ? variant.end - variant.pos : 0
-    coords = `${chrom}-${posInt - 100}-${posInt + endOffset + 100}`
+    coords = `${chrom}-${Math.max(posInt - 100, 1)}-${posInt + endOffset + 100}`
```
ShifaSZ, line 51: It was a bug that if `posInt` was smaller than 100 then the generated coordination was wrong.

hanars (changes requested):
This is the wrong approach. Have callset_het and callset_hom be separate frequencies returned form the server and render them using the existing `FreqSummary` component, just write the correct configs for them. Since we now want them showing up as 2 distinct populations, we need to actually treat them as 2 distinct populations

seqr/utils/elasticsearch/constants.py
```diff
@@ -109,25 +107,34 @@
...
+        'AN': 'helix_AN',
+        'AC': 'helix_AC_het',
+        'AF': 'helix_AF_het',
+    }
 }
-POPULATIONS.update(MITO_POPULATION)
 
 POPULATION_FIELD_CONFIGS = {
-    'AF': {'format_value': float},
+    'AF': {'format_value': lambda val: float(val) if val is not None else None, 'default_value': None},
```
hanars, line 130: This will completely break every single population in seqr. If a field exists in the index but is missing from a particular variant, that means that variant has an AF of 0. This change can not be integrated. Can you please explain why this change was necessary for mito so I can figure out how to make it work better?

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -70,6 +71,7 @@ const FreqSummary = React.memo((props) => {
   return (
     <div>
       {titleContainer ? titleContainer(props) : fieldTitle}
+      {isMito && (isHet ? ' heteroplasmy' : ' homoplasmy')}
```
hanars, line 74: you need to set the config that is being passed in to have the correct `fieldTitle` or `titleContainer` for the mito populations, not add mito specific logic to the generic helper component

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,33 +191,64 @@ const POPULATIONS = [
...
+  const isMito = POPULATIONS.some(pop => pop.isHet &&
+      (populations[pop.field]?.af !== null && populations[pop.field]?.af !== undefined))
+  const freqContent = (
+    <div>
+      {POPULATIONS.map(pop => <FreqSummary key={pop.field} variant={variant} isMito={isMito} {...pop} />)}
+    </div>
+  )
 
-  const hasAcPops = POPULATIONS.filter(pop => populations[pop.field] && populations[pop.field].ac)
+  const hasAcPops = POPULATIONS.filter(pop => !pop.isHet && populations[pop.field]?.ac)
```
hanars, line 237: the `isHet` specific logic here makes this really hard to read and maintain. I think it would be better if you made the logic more generic so it could be extended easily and we wouldn't have mito-specific logic hard coded in what it supposed to be a generic component. Add something like `detailSection: 'Hetroplasmy'` to the population configs, and then here you would add a check for `!pop.detailSection`. Then below you could do something to get a list of relevant detail sections as follows and then render them:
```
const detialSections = POPULATIONS.reduce((acc, pop) => {
  if (pop.detailSection && populations[pop.field]) {
    if (!acc[pop.detailSection]) {
      acc[pop.detailSection] = []
    }
    acc[pop.detailSection].push(populations[pop.field])
  }
  return acc
}, {})
```

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
ShifaSZ: I was hesitating to make this change, but forgot to ask. The are two reasons that we need to keep None, one is to avoid the null value gnomad_mito_af/gnomad_mito_af_het/helix_af/helix_af_het to become 0, the other is that we use af_het to detect if it is a mitochondrial calltype. If don't add `'default_value': None` then we can't use the af_het field to detect wether it is a mito call type because all the null value af_hets become 0.
Is there any way to fix the other populations' issue caused by the change?

hanars: But for mito variants, doesn't a null  gnomad_mito_af/gnomad_mito_af_het/helix_af/helix_af_het MEAN that the value is 0? We've joined with our reference data so if the variant is missing in that reference data set it means its not there aka has an AF of 0

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: If I understand correctly, I will use the name `sectionHeader` for the `detailSection` you mentioned.

seqr/utils/elasticsearch/constants.py
ShifaSZ: I thought that when a variant is missing in a reference data set, it means the data set doesn't have the info about the variant. Does it also mean the reference population doesn't have the variant?

hanars: If a reference data set does not have a variant, that means the variant is not in the reference data set. If the variant is not in a population frequency data set, that means the frequency is 0. If the frequency was anything other than 0, it would be in the datatset

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +207,115 @@ const POPULATIONS = [
...
-            <div key={pop.field}>
-              {`${pop.fieldTitle} max observed heteroplasmy:
-              ${populations[pop.field].max_hl.toPrecision(pop.precision || 2)}`}
-            </div>
-          ))}
-        </Popup.Content>
+        {catalogs.map((catalog, i) => (
+          <div key={catalog}>
+            {(i > 0) && <VerticalSpacer height={5} />}
+            <Popup.Header content={catalog} />
```
ShifaSZ, line 302: The `Popup.Header` component's style is changed as below picture.

hanars: The issue here is you are putting this all inside a `div` which is not the correct nesting for the styling to work. The element grouping should be unnecessary, you should be able to change this to render a top-level list by using a reduce instead of a map:
```
{catalogs.reduce((acc, catalog, i) => ([
  ...acc, 
  i > 0 ? <VerticalSpacer key={`spacer-${category}`} height={5} /> : null,
  <Popup.Header key={catalog} content={catalog} />,
  ...popupPoputionByCat[catalog].map((pop) => ...),
  ...etc
]), []}
```

Also, as a high level thing, "catalog" is a really confusing word in this context. I am really unsure what you meant for it to mean here, but I think "detailSection" would be a much better word, and for the list you could use the word "sections"

ShifaSZ: Thank you. Update pushed.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -141,9 +141,47 @@ gnomadLink.propTypes = {
...
+const SECTIONS = [GLABLE_AF_SECTION, AC_SECTION, HOM_SECTION, HET_SECTION]
+
 const POPULATIONS = [
   { field: 'sv_callset', fieldTitle: 'This Callset', acDisplay: 'AC', helpMessage: SV_CALLSET_CRITERIA_MESSAGE },
-  { field: 'callset', fieldTitle: 'This Callset', acDisplay: 'AC' },
+  {
+    field: 'callset',
+    fieldTitle: 'This Callset',
+    acDisplay: 'AC',
+    titleContainer: sectionTitle,
```
hanars, line 174: this will break the display for non-mito variants

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -185,69 +223,98 @@ const POPULATIONS = [
...
+    acc[key] = []
+  }
+  acc[key].push(value)
+}
+
 const Frequencies = React.memo(({ variant }) => {
   const { populations = {} } = variant
+  const callsetHetPop = populations.callset_heteroplasmy
+  const isMito = callsetHetPop && callsetHetPop.af !== null && callsetHetPop.af !== undefined
+  const popupDetailsBySection = POPULATIONS.reduce((acc, pop) => {
```
hanars, line 266: I think it would be better for the section constants to define their own display logic instead of having so much hardcoded conditional if/else render logic. Also, since we always show different sections for mito/ non-mito I think we should explicitly use different section constants for those instead of if/else logic. I also think it would be neater to use one constant rather than separate constants for the sections and the population sections. So something more like

```
const DETAIL_SECTIONS = [{
  name: 'Global AFs', 
  hasDetail: pop => pop. filter_af && pop. filter_af !== pop.af, 
  display: pop => ([{ value: pop.af.toPrecision(3) }]),
}, 
...
]
MITO_DETAIL_SECTIONS =  [{
  name: HET_SECTION, 
  hasDetail: pop => pop. ac || pop.max_hl 
  display: pop => ([
    pop.ac && { value: `${pop.ac} out of ${pop.an}` },
    pop.max_hl && { subtitle: ' max observed heteroplasmy', value: pop.max_hl.toPrecision(3) },
  ].filter(val => val)),
}, 
...

const sections = (isMito ? MITO_HOVER_SECTIONS : HOVER_SECTIONS).map(({ hasDetail, ...section }) => ({
  ...section,
  populations: POPULATIONS.filter(pop => (pop.section === undefined || pop.section === section.title) &&  populations[pop.field] && hasDetail(populations[pop.field]))
}), []).filter(({ populations }) => populations.length)
...

    {sections.reduce((acc, section, i) => ([
          ...acc,
          (i > 0) && <VerticalSpacer key={section.name} height={5} />,
          <Popup.Header key={`${section.name}_header`} content={section.name} />,
          ...section.populations.reduce((acc, pop) => ([
             ...acc, ...section.display(populations[pop.field]).map(({ subtitle, value }) => (
              <Popup.Content key={`${section.name}-${pop.field}${subtitle}`}>
                {`${pop.fieldTitle}${subtitle || ''}: ${value`}
              </Popup.Content>
            ))]), []),
        ]), [])}
```

hanars (changes requested):

seqr/utils/elasticsearch/constants.py
```diff
@@ -109,13 +107,24 @@
...
+        'AF': 'gnomad_mito_AF_het',
+    },
     'helix': {},
+    'helix_heteroplasmy': {
+        'AN': 'helix_AN',
+        'AC': 'helix_AC_het',
+        'AF': 'helix_AF_het',
+    }
 }
-POPULATIONS.update(MITO_POPULATION)
```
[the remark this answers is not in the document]
ShifaSZ: I merged all the mito populations into the existing `POPULATIONS`. A separate mito population constant is not helpful.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -141,9 +141,41 @@ gnomadLink.propTypes = {
...
+  section: PropTypes.string,
+}
+
 const POPULATIONS = [
   { field: 'sv_callset', fieldTitle: 'This Callset', acDisplay: 'AC', helpMessage: SV_CALLSET_CRITERIA_MESSAGE },
-  { field: 'callset', fieldTitle: 'This Callset', acDisplay: 'AC' },
+  {
+    field: 'callset_heteroplasmy',
+    fieldTitle: 'This Callset',
+    titleContainer: sectionTitle,
```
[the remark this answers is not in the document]
ShifaSZ: Because we also use the `fieldTitle` for the popup details, we don't need the second part of the titles in the popup cases.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +214,101 @@ const POPULATIONS = [
...
   {
     field: 'helix',
     fieldTitle: 'Helix mito',
+    titleContainer: sectionTitle,
+    section: 'Homoplasmy',
     precision: 3,
   },
 ]
 
+const pushDefaultKey = (acc, key, value) => {
```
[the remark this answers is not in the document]
ShifaSZ: I was trying to mimic the `defaultdict` in Python. Either `addListValue` or `pushArrayValue` doesn't reflect the action of creating the new array element if it doesn't exist. How about use a name like `pushObjectArrayElement`?

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Oh, yes. It used to have an `isMito` parameter for the `sectionTitle` function. I removed the parameter at some point.

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Using section constants is a good idea.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -141,9 +141,41 @@ gnomadLink.propTypes = {
...
+    titleContainer: sectionTitle,
+    section: 'Heteroplasmy',
+    acDisplay: 'AC',
+    precision: 3,
+  },
+  {
+    field: 'callset',
+    fieldTitle: 'This Callset',
+    acDisplay: 'AC',
+    titleContainer: sectionTitle,
```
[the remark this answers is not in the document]
hanars: bumping this. I see no reason for this new `sectionTitle` at all, we should just clearly write out what the field title is

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -185,69 +224,116 @@ const POPULATIONS = [
...
+  },
+]
+
+const acDisplay = (populations, popConfig) => ([
+  `${popConfig.fieldTitle}: ${populations[popConfig.field].ac} out of ${populations[popConfig.field].an}`,
+])
+const DETAIL_SECTIONS = [
+  {
+    name: GLOBAL_AF_SECTION,
+    hasDetail: (populations, popConfig) => populations[popConfig.field] && populations[popConfig.field].filter_af &&
```
hanars, line 262: the only part of `populations` and `popConfig` in any of the `hasDetail` functions is `populations[popConfig.field]` it would be much more readable/ maintainable if you called it with `hasDetail(populations[popConfig.field])`

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -185,69 +224,116 @@ const POPULATIONS = [
...
+      (populations[popConfig.field].filter_af !== populations[popConfig.field].af),
+    display: (populations, popConfig) => ([
+      `${popConfig.fieldTitle} : ${populations[popConfig.field].af.toPrecision(popConfig.precision || 2)}`,
```
hanars, line 265: I really dislike having `${popConfig.fieldTitle}:` copy pasted into every single `display` function, it makes the maintenance on this nearly impossible. Shared behavior needs to be shared not copy pasted 5 times. If you for some reason don't like the approach I proposed where `display` returns a list of `subfield` and `value` maybe try something like this instead:

```
const DETAIL_SECTIONS = [{
  ...,
  detailFields: ['af' , 'max_hl'], 
} ...]

const FIELD_SUBTITLES = { max_hl: ' max observed heteroplasmy' }
const FIELD_DISPLAY_VALUE = {
  ac: pop => `${pop.ac} out of ${pop.an}}`,
  af: pop, popConfig => pop.af.toPrecision(popConfig.precision || 2),
  ...
}

...
            const population = populations[popConfig.field]
            accDisplay.push(...section.detailFields.filter(field => population[field]).map(field => (
              <Popup.Content key={`${section.name}${field}`}>
                 {`${pop.fieldTitle}${FIELD_SUBTITLES[field] || ''}: ${FIELD_DISPLAY_VALUE[field](population, popConfig)}`}
              </Popup.Content>
            )))
```

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,33 +191,64 @@ const POPULATIONS = [
...
 ]
 
 const Frequencies = React.memo(({ variant }) => {
   const { populations = {} } = variant
-  const freqContent = <div>{POPULATIONS.map(pop => <FreqSummary key={pop.field} variant={variant} {...pop} />)}</div>
+  const isMito = POPULATIONS.some(pop => pop.isHet &&
+      (populations[pop.field]?.af !== null && populations[pop.field]?.af !== undefined))
+  const freqContent = (
+    <div>
+      {POPULATIONS.map(pop => <FreqSummary key={pop.field} variant={variant} isMito={isMito} {...pop} />)}
```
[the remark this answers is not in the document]
hanars: still a problem

hanars (changes requested):

ui/shared/components/panel/variants/Frequencies.jsx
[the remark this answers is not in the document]
ShifaSZ: We need to display the title on the page as `This Callset Heteroplasmy` while displaying it as `This Callset` in the popup details.

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: > If you for some reason don't like the approach I proposed where `display` returns a list of `subfield` and `value` maybe try something like this instead

Your `subtitle` and `value` suggestion looks good. I just didn't understand when I first looked at it.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -141,6 +141,27 @@ gnomadLink.propTypes = {
...
+
+const sectionTitle = ({ fieldTitle, section }) => (
+  <span>
+    {fieldTitle}
+    &nbsp;
+    {section.toLowerCase()}
+  </span>
+)
+
+sectionTitle.propTypes = {
```
hanars, line 157: functions should not have prop types, react components should. Either make this a react component stylistically by using a capitalized component name, or keep it as a simple function with the current name but remove the prop types

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +202,125 @@ const POPULATIONS = [
     queryParams: { [GENOME_VERSION_37]: 'dataset=gnomad_sv_r2_1' },
     helpMessage: GNOMAD_SV_CRITERIA_MESSAGE,
   },
+]
+
+const MITO_POPULATIONS = [
+  {
+    field: 'callset',
+    fieldTitle: 'This Callset',
+    acDisplay: 'AC',
```
hanars, line 211: As ever, copy paste is bad. You should define a `CALLSET_POP` constant and use it in the main `POPULATIONS` and then extend it here: `{ ...CALLSET_POP, titleContainer: sectionTitle, section: HOM_SECTION }`

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +202,125 @@ const POPULATIONS = [
...
+    titleContainer: sectionTitle,
+    section: HOM_SECTION,
+  },
+  {
+    field: 'callset_heteroplasmy',
+    fieldTitle: 'This Callset',
+    acDisplay: 'AC',
+    titleContainer: sectionTitle,
+    section: HET_SECTION,
+    precision: 3,
```
hanars, line 221: use the same precision for callset and callset_heteroplasmy, there is no reason for them to be different

hanars (changes requested):
Can you please provide an updated screenshot here for how this looks now

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Okay. To be consistent in this file, should I also change `gnomadLink`?

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: Good catch. It shouldn't have `precision`.

ui/shared/components/panel/variants/Frequencies.jsx
hanars: no thats out of scope

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +199,126 @@ const POPULATIONS = [
...
+    section: HET_SECTION,
     precision: 3,
   },
 ]
 
+const DETAIL_SECTIONS = [
+  {
+    name: 'Global AFs',
+    hasDetail: pop => pop && pop.filter_af && (pop.filter_af !== pop.af),
+    display: () => [{ subTitle: '', value: 'af' }],
```
hanars, line 253: Its not a blocking change, but needing to set `subTitle: ''` everywhere adds unnecessary overhead for maintaining these and adding new sections. It would be better if the rendering could handle adding a default so we only need to define subtitle when its actually defined

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +199,126 @@ const POPULATIONS = [
...
+  const isMito = callsetHetPop && callsetHetPop.af !== null && callsetHetPop.af !== undefined
+  const popConfigs = isMito ? MITO_POPULATIONS : POPULATIONS
+  const sections = (isMito ? MITO_DETAIL_SECTIONS : DETAIL_SECTIONS).reduce(
+    (acc, section) => ([
+      ...acc,
+      {
+        name: section.name,
+        details: popConfigs.map(popConfig => (section.hasDetail(populations[popConfig.field]) &&
+          (!popConfig.section || popConfig.section === section.name) &&
+          section.display(populations[popConfig.field], popConfig).map(({ subTitle, value }) => (
```
hanars, line 296: the `display` functions all only take one argument, so it is confusing that you pass in `popConfig`

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -182,72 +199,126 @@ const POPULATIONS = [
...
+            <Popup.Content key={`${section.name}${popConfig.field}${subTitle}`}>
+              {`${popConfig.fieldTitle}${subTitle}: ${DISPLAY_VALUE[value](populations[popConfig.field], popConfig)}`}
```
hanars, line 298: `value` is a confusing name for this property, as it is actually the `valueField` for looking up the `value` to display, not the value itself so I would rename it

hanars (changes requested):
Please add an updated screenshot

ShifaSZ:
The screenshot is as the below picture.

hanars (approved):

PR #2913, opened 2022-08-23
Add actions caching, update actions node to match docker

### LINKED ISSUE #2895: Cache pip dependency installs in the seqr unit tests
Right now it looks like our unit test pipeline in github actions is pip installing from scratch every time it runs. We can probably save a good deal of time if we use an [actions cache](https://github.com/actions/cache) to save the pip (and potentially npm) installations

sjahl:
This adds some caching steps to the pip and npm installs to hopefully shave some time off the unit tests pipeline.

I also noticed that we were running a much older version of nodejs in our actions tests than we do in the seqr docker image, so I updated that to version 14 to match.

Last change was updating the `npm install || true` to a [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci), which we also do in docker. I think this is the preferred way to run NPM in automated environments, according to the docs, since it only installs explicitly what's in the lockfile, rather than upgrading everything in the dependency tree like `install` does. It's also much faster.

Fixes #2895

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - master
       - dev
-      - add-actions-caching # TODO: remove before merge
   pull_request:
     types: [opened, synchronize, reopened]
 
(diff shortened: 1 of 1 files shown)


sjahl:
Old:
Screen Shot 2022-08-23 at 4 21 26 PM

New:
Screen Shot 2022-08-23 at 4 25 33 PM

The cache savings on pip are slightly smaller, but the real bloat in this workflow was the npm install.

hanars (approved):

Issue #1976, opened 2021-07-22
Make awesomebar search easier to use by interchangeably matching "-", "_" and "." in IDs


bw2:
For a lot of sample IDs, I often need to manually experiment with -, _ and . chars and remove leading/trailing space before seqr finds a match. It would simplify things if the search treated these chars as interchangeable.

hanars:
behavior has been updated

PR #2914, opened 2022-08-23
add unit tests for awesomebar

hanars:
When starting on https://github.com/broadinstitute/seqr/issues/1976 I realized that our unit tests for the awesomebar don't really test the actual contents of the returned results. This adds those tests so that it will be clearer what behavior changes when I implement that ticket


seqr/views/apis/awesomebar_api_tests.py
```diff
@@ -25,13 +27,104 @@ def test_awesomebar_autocomplete_handler(self):
         self.login_collaborator()
         response = self.client.get(url + "?q=1")
         self.assertEqual(response.status_code, 200)
+        matches = response.json()['matches']
         self.assertSetEqual(
-            set(response.json()['matches'].keys()), {'projects', 'families', 'analysis_groups', 'individuals', 'genes'}
+            set(matches.keys()), {'projects', 'families', 'analysis_groups', 'individuals', 'genes'}
         )
```
ShifaSZ, line 33: You may want to put these lines into one line.

seqr/fixtures/1kg_project.json
```diff
@@ -295,7 +295,7 @@
         "last_modified_date": "2017-03-12T19:27:08.773Z",
         "project": 1,
         "family_id": "10",
-        "display_name": "10",
+        "display_name": "10_a",
```
ShifaSZ, line 298: Is there any reason to change this one?

ShifaSZ (changes requested):

seqr/fixtures/1kg_project.json
hanars: So right now, the results are sorted strictly by the length of the title, so when the query is "1" then  "2_1" is sorted before "10_a" because of length. In the coming ticket I am going to change the behavior to prioritize first character match wherever possible, so "10_a" would come before "2_1" so I needed to add fixture data so that change would be testable. I updated the tests to explicitly check the order all the results are returned in so that this will be clearer

hanars: also, I realized I actually needed a different field for this so I changed which family has the long display name, its family 12 now

PR #2916, opened 2022-08-24
Awesomebar charachter sensitivity

hanars:


seqr/views/apis/awesomebar_api_tests.py
```diff
@@ -136,6 +136,13 @@ def test_awesomebar_autocomplete_handler(self):
             'category': 'HP:0033127',
         })
 
+        # Test fuzzy matching
+        response = self.client.get(url + "?q=2-&categories=families")
+        self.assertEqual(response.status_code, 200)
+        families = response.json()['matches']['families']['results']
+        self.assertEqual(len(families), 3)
+        self.assertListEqual([f['title'] for f in families], ['12-a', '2_1', '42'])
```
ShifaSZ, line 144: Is it okay for `2-` to match `42`?

seqr/views/apis/awesomebar_api_tests.py
```diff
@@ -26,7 +26,7 @@ def test_awesomebar_autocomplete_handler(self):
         )
 
         self.login_collaborator()
-        response = self.client.get(url + "?q=1")
+        response = self.client.get(url + "?q=%201")
```
ShifaSZ, line 29: It may be better to test blanks on both sides.

ShifaSZ (changes requested):

seqr/views/apis/awesomebar_api_tests.py
hanars: yes, it is supposed to match the middle of a word not just the start. However, we have adjusted the sort in this PR so matches that start with the query are sorted above those that don't, which is why this is last even though it is shorter

ShifaSZ (approved):

PR #2922, opened 2022-08-25
Handle get VCF error on empty bucket

hanars:
If you run `gsutil ls gs://bucket/**` on an empty bucket it throws an error instead of just returning an empty list, which causes unfriendly error to be raised if a user tries to load data from a fully empty workspace. This happened in production and was a bit hard to debug so I added some additional logging here that would have helped me


ShifaSZ (approved):
LGTM.

PR #2937, opened 2022-08-31
fix edit project categories js performance

hanars:


ShifaSZ:
I'm curious that the new code looks equivalent to the old one. How is the performance solved?

hanars:
Is not the same, it is now using a selector function that is wrapped in the reselect `createSelector` decorator instead of just using a function.Every time render was called on that component it was recreating the list of categories, which in turn meant that the child components detected a change in props and re-rendered, and this was happening dozens of  times every time the component was interacted with. Now, because the function is wrapped by reselect, it does not recompute unless the input to the function changes, which means the categories prop being passed in is the same list object, and not a new list with the same values. Because it is actually the same list, the child component realizes that no props have changed and nothing rerenders

ShifaSZ (approved):

PR #2945, opened 2022-09-09
Add more VCF file validations.

ShifaSZ:
Create a draft PR for a basic idea review. Tests haven't changed yet, just some modifications for Github auto tests.

### DIFF
--- a/seqr/utils/vcf_utils.py
+++ b/seqr/utils/vcf_utils.py
@@ -1,0 +1,96 @@
+import re
+
+from collections import defaultdict
+
+from seqr.utils.middleware import ErrorsWarningsException
+from seqr.utils.file_utils import file_iter
+
+BLOCK_SIZE = 65536
+
+EXPECTED_META_FIELDS ={
+    'INFO': {
+        'AC': {
+            'Number': 'A',
+            'Type': 'Integer',
+        },
+        'AN': {
+            'Number': '1',
+            'Type': 'Integer',
+        },
+        'AF': {
+            'Number': 'A',
+            'Type': 'Float',
+        }
+    },
+    'FORMAT': {
+        'GQ': {
+            'Number': '1',
+            'Type': 'Integer',
+        },
+        'GT': {
+            'Number': '1',
+            'Type': 'String',
+        }
+    }
+}
+
+REQUIRED_HEADERS = ['#CHROM', 'POS', 'ID', 'REF', 'ALT', 'QUAL', 'FILTER', 'INFO', 'FORMAT']
+
+
+def _validate_vcf(header, meta, first_line):
+
+    missing_headers = [h for h in REQUIRED_HEADERS if h not in header ]
+    if missing_headers:
+        missing_fields = ','.join(missing_headers)
+        raise ErrorsWarningsException([f'Required VCF header fields {missing_fields} are missing.'], [])
+
+    for field, sub_field_meta in EXPECTED_META_FIELDS.items():
+        for sub_field, metadata in sub_field_meta.items():
+            for key, value in metadata.items():
+                if meta.get(field, {}).get(sub_field, {}).get(key) != value:
+                    error = f'VCF header field {field}.{sub_field} and meta information {key}={value} is expected.'
+                    raise ErrorsWarningsException([error], [])
+
+    row_data = first_line.split('\t')
+    info_keys = [k.split('=')[0] for k in row_data[7].split(';')]
+    missing_info_fields = [f for f in EXPECTED_META_FIELDS['INFO'].keys() if f not in info_keys ]
+    if missing_info_fields:
+        missing_fields = ','.join(missing_info_fields)
+        raise ErrorsWarningsException([f'Missing INFO field(s) {missing_fields} in the row data'], [])
+
+    format_keys = [k for k in row_data[8].split(':')]
+    missing_format_fields = [f for f in EXPECTED_META_FIELDS['FORMAT'].keys() if f not in format_keys]
+    if missing_format_fields:
+        missing_fields = ','.join(missing_format_fields)
+        raise ErrorsWarningsException([f'Missing FORMAT field(s) {missing_fields} in the row data'], [])
+
+
+def _get_vcf_meta_data(line):
+    r = re.search(r'##(.*?)=<(.*)>$', line)
+    if r:
+        field, value = r.groups()
+        value = re.sub(r'".*"', '""', value)  # remove description string which could have '=' or ',' in it
+        sub_field_meta = {}
+        for k, v in map(lambda s: s.split('='), value.split(',')):
+            sub_field_meta[k] = v
+        sub_field = sub_field_meta.pop('ID')
+        return field, {sub_field: sub_field_meta}
+    return None, None
+
+
+def validate_vcf_and_get_samples(vcf_filename):
+    byte_range = None if vcf_filename.endswith('.vcf') else (0, BLOCK_SIZE)
+    samples = {}
+    meta = defaultdict(dict)
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            _validate_vcf(header[0:9], meta, line)
+            break
+        field, sub_field_meta = _get_vcf_meta_data(line)
+        if field:
+            meta[field].update(sub_field_meta)
+        if line.startswith('#CHROM'):
+            header = line.rstrip().split('\t')
+            samples = set(header[9:])
+
+    return samples
--- a/seqr/views/apis/anvil_workspace_api_tests.py
+++ b/seqr/views/apis/anvil_workspace_api_tests.py
@@ -313,8 +313,8 @@ class AnvilWorkspaceAPITest(AnvilAuthenticationTestCase):
                          '/login/google-oauth2?next=/api/create_project_from_workspace/my-seqr-billing/anvil-no-project-workspace1/grant_access')
 
     @mock.patch('seqr.views.apis.anvil_workspace_api.does_file_exist')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.file_iter')
-    def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_logger):
+    @mock.patch('seqr.views.apis.anvil_workspace_api.validate_vcf_and_get_samples')
+    def test_validate_anvil_vcf(self, mock_vcf_validate, mock_file_exist, mock_utils_logger):
         # Requesting to load data from a workspace without an existing project
         url = reverse(validate_anvil_vcf,
                       args=[TEST_WORKSPACE_NAMESPACE, TEST_NO_PROJECT_WORKSPACE_NAME])
@@ -345,23 +345,25 @@ class AnvilWorkspaceAPITest(AnvilAuthenticationTestCase):
                          'Invalid VCF file format - file path must end with .vcf or .vcf.gz or .vcf.bgz')
 
         mock_file_exist.return_value = True
-        mock_file_iter.return_value = ['##fileformat=VCFv4.2\n',
-                                       '#CHROM	POS	ID	REF	ALT	QUAL']  # incomplete header line
+        # mock_file_iter.return_value = ['##fileformat=VCFv4.2\n',
+        #                                '#CHROM	POS	ID	REF	ALT	QUAL']  # incomplete header line
+        mock_vcf_validate.return_value = {}
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_GZ_DATA_PATH))
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.json()['error'],
                          'No samples found in the provided VCF. This may be due to a malformed file')
-        mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf.gz', byte_range=(0, 65536))
-        mock_file_exist.assert_called_with('gs://test_bucket/test_path.vcf.gz', user=self.manager_user)
+        # mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf.gz', byte_range=(0, 65536))
+        # mock_file_exist.assert_called_with('gs://test_bucket/test_path.vcf.gz', user=self.manager_user)
 
         # Test valid operation
         mock_file_exist.return_value = True
-        mock_file_iter.return_value = FILE_DATA
+        # mock_file_iter.return_value = FILE_DATA
+        mock_vcf_validate.return_value = {'NA19675', 'NA19678', 'HG00735'}
         response = self.client.post(url, content_type='application/json', data=json.dumps(VALIDATE_VCF_BODY))
         self.assertEqual(response.status_code, 200)
         self.assertDictEqual(response.json(), VALIDATE_VFC_RESPONSE)
         mock_file_exist.assert_called_with('gs://test_bucket/test_path.vcf', user=self.manager_user)
-        mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf', byte_range=None)
+        # mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf', byte_range=None)
 
         # Test logged in locally
         remove_token(
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -27,7 +27,8 @@ from seqr.views.utils.terra_api_utils import add_service_account, has_service_ac
 from seqr.views.utils.pedigree_info_utils import parse_pedigree_table
 from seqr.views.utils.individual_utils import add_or_update_individuals_and_families, get_updated_pedigree_json
 from seqr.utils.communication_utils import safe_post_to_slack, send_html_email
-from seqr.utils.file_utils import does_file_exist, file_iter, mv_file_to_gs, get_gs_file_list
+from seqr.utils.file_utils import does_file_exist, mv_file_to_gs, get_gs_file_list
+from seqr.utils.vcf_utils import validate_vcf_and_get_samples
 from seqr.utils.logging_utils import SeqrLogger
 from seqr.utils.middleware import ErrorsWarningsException
 from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm, login_and_policies_required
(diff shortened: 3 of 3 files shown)


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -313,8 +313,8 @@ def test_grant_workspace_access(self, mock_add_service_account, mock_has_service
                          '/login/google-oauth2?next=/api/create_project_from_workspace/my-seqr-billing/anvil-no-project-workspace1/grant_access')
 
     @mock.patch('seqr.views.apis.anvil_workspace_api.does_file_exist')
-    @mock.patch('seqr.views.apis.anvil_workspace_api.file_iter')
-    def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_logger):
+    @mock.patch('seqr.views.apis.anvil_workspace_api.validate_vcf_and_get_samples')
+    def test_validate_anvil_vcf(self, mock_vcf_validate, mock_file_exist, mock_utils_logger):
```
ShifaSZ, line 317: No real update yet.

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            _validate_vcf(header[0:9], meta, line)
+            break
+        field, sub_field_meta = _get_vcf_meta_data(line)
+        if field:
+            meta[field].update(sub_field_meta)
+        if line.startswith('#CHROM'):
+            header = line.rstrip().split('\t')
+            samples = set(header[9:])
```
hanars, line 95: Hardcoding the indices for the header row is really fragile. We only should be validating that all the columns we care about are there, we should not be failing if extra columns exist. I think it would be better to pass the whole header row to an explicit `_validate_vcf_header` function here as the extra sample columns shouldn't matter, and then we should explicitly find the column where the sample IDs start similar to what we have now
```
header = line.rstrip().split('\t')
_validate_vcf_header(header)
sample_col_index = next(i for i, col in enumerate header if col == 'FORMAT') + 1
samples = set(header[sample_col_index:])
```

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    for field, sub_field_meta in EXPECTED_META_FIELDS.items():
+        for sub_field, metadata in sub_field_meta.items():
+            for key, value in metadata.items():
+                if meta.get(field, {}).get(sub_field, {}).get(key) != value:
+                    error = f'VCF header field {field}.{sub_field} and meta information {key}={value} is expected.'
+                    raise ErrorsWarningsException([error], [])
+
+    row_data = first_line.split('\t')
+    info_keys = [k.split('=')[0] for k in row_data[7].split(';')]
+    missing_info_fields = [f for f in EXPECTED_META_FIELDS['INFO'].keys() if f not in info_keys ]
```
hanars, line 56: I'm a bit confused by whats happening here. You seem to be validating that INFO and FORMAT exist by checking both the `meta` and the `first_line` for the exact same values. Per the discussion in the ticket, it would be better if we use the meta to do validation, if we can check everything in the meta we should not check the `first_line` at all

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,97 @@
...
+            'Number': '1',
+            'Type': 'String',
+        }
+    }
+}
+
+REQUIRED_HEADERS = ['#CHROM', 'POS', 'ID', 'REF', 'ALT', 'QUAL', 'FILTER', 'INFO', 'FORMAT']
+
+
+def _validate_vcf(header, meta, first_line):
```
hanars, line 40: It's hard to read/ maintain that you validate the header and meta in one function, especially since neither depends on the other. It would be much cleaner to have explicit functions for each: `_validate_vcf_header(header)` and `_validate_vcf_meta(meta)`

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: Even if the subfields exist in the `meta`, they are still possibly missing in the data lines. So it is safer to validate both. But since each data line can be different, it is not a complete validation to check the `first_line`. I'm okay with taking out the validation for the first line.

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,91 @@
...
+def validate_vcf_and_get_samples(vcf_filename):
+    byte_range = None if vcf_filename.endswith('.vcf') else (0, BLOCK_SIZE)
+    samples = {}
+    header = []
+    meta = defaultdict(dict)
+    for line in file_iter(vcf_filename, byte_range=byte_range):
+        if line[0] != '#':
+            _validate_vcf_header(header[0:len(REQUIRED_HEADERS)])
+            _validate_vcf_meta(meta)
+            _validate_vcf_data_row(line)
```
hanars, line 82: this is a really confusing control flow. It would be much more readable to just have the break statement here, and then do the validation outside of the loop entirely

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,91 @@
...
+            break
+        field, sub_field_meta = _get_vcf_meta_data(line)
+        if field:
+            meta[field].update(sub_field_meta)
+        if line.startswith('#CHROM'):
```
hanars, line 87: I think it would make more sense to move this if statement up and put the metadata parsing in an else block, since it is an either/or situation of whether its a header or metadata

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,91 @@
...
+def _validate_vcf_meta(meta):
+    for field, sub_field_meta in EXPECTED_META_FIELDS.items():
+        for sub_field, metadata in sub_field_meta.items():
+            for key, value in metadata.items():
+                if meta.get(field, {}).get(sub_field, {}).get(key) != value:
+                    error = f'VCF header field {field}.{sub_field} and meta information {key}={value} is expected.'
+                    raise ErrorsWarningsException([error], [])
+
+
+def _validate_vcf_data_row(first_line):
```
hanars, line 43: per the ticket and our previous discussion on this PR, we will not be validating both the meta and the first row. Pick one or the other, doing both is redundant, confusing, and hard to maintain

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    missing_format_fields = [f for f in EXPECTED_META_FIELDS['FORMAT'].keys() if f not in format_keys]
+    if missing_format_fields:
+        missing_fields = ','.join(missing_format_fields)
+        raise ErrorsWarningsException([f'Missing FORMAT field(s) {missing_fields} in the row data'], [])
+
+
+def _get_vcf_meta_data(line):
+    r = re.search(r'##(.*?)=<(.*)>$', line)
+    if r:
+        field, value = r.groups()
```
hanars, line 63: it is better to used named groups to be explicit about what you are parsing from where:
using `r'##(?P<field>.*?)=<(?P<value>.*)>$'` if you call `r.groupdict()` you will get a dictionary keys `field` and `value`

I also think there is a better way to write this regex so that it can capture the subfields we care about directly. I think we only want to validate on ID and Type, so using `r'##(?P<field>.*?)=<ID=(?P<id>[^,]*).*Type=(?P<type>[^,]*).*>$'` will explicitly capture those subfields

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: Okay.

seqr/utils/vcf_utils.py
ShifaSZ: It is great to use `groupdict`. But I'm unsure whether it works for picking up the `ID` and `Type` fields because their order is not determined.

ShifaSZ: But the `ID` must be the first one.

hanars: if ID must be first than the regex I listed will work regardless of what other fields are there. If thats not actually true of the metadata then we can rethink this, but I think it would still be better to write an explicit regex for ID and Type than parsing out all the subfields

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    if missing_headers:
+        missing_fields = ', '.join(missing_headers)
+        raise ErrorsWarningsException([f'Missing required VCF header field(s) {missing_fields}.'], [])
+
+
+def _validate_vcf_meta(meta):
+    for field, sub_field_meta in EXPECTED_META_FIELDS.items():
+        for sub_field, field_type in sub_field_meta.items():
+            if meta.get(field, {}).get(sub_field) != field_type:
+                error = f'VCF header field {field}.{sub_field} and meta information Type={field_type} is expected.'
```
hanars, line 38: I think you want to be clear about whether the fields are missing completely or if the type is mismatched, and I think we want to make sure to send all the field errors in one validation response instead of one at at time. So we want a list of errors, one for all the missing subfields per group, and one for each mismatched type. An example of the errors we would return is this:
```
Missing required INFO field(s) AN
Missing required FORMAT field(s) AD, GT
Incorrect meta Type for INFO.AF - expected "Float", got "Integer"
Incorrect meta Type for FORMAT.GQ - expected "Integer", got "String"
Incorrect meta Type for FORMAT.DP - expected "Integer", got "Float"
```

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,70 @@
...
+                raise ErrorsWarningsException([error], [])
+
+
+def _get_vcf_meta_data(line):
+    r = re.search(r'##(?P<field>.*?)=<ID=(?P<id>[^,]*).*Type=(?P<type>[^,]*).*>$', line)
+    if r:
+        matches = r.groupdict()
+        return matches['field'], {matches['id']: matches['type']}
```
hanars, line 46: splitting this into a tuple isn't so helpful, I think it would be neater to just have it return the dict and then do the formatting bellow with `meta[meta_row['field']].update({meta_row['id']: meta_row['type']})`

seqr/utils/vcf_utils_tests.py
```diff
@@ -0,0 +1,76 @@
+import mock
```
hanars, line 1: I really dislike having util-file specific tests. It is incredible fragile because we don't really care how the utility function operates, what we actually care about is how the api calls that use those functions operate. If we decide to split a helper function in two or change its function signature in a way that has no impact on any of the functions calling it, we don't want to update a test. Conversely, if the way a utility function is called changes in a breaking way but we are testing the utility function separately from the actual function calls, our tests could easily miss that.

These tests all belong in the `test_validate_anvil_vcf` function in the anvil_workspace_api_tests file. That test and that tets alone should be sufficient to get you to 100% coverage of this new file

While we do have some legacy files that are explicitly testing util functions, we should not add more

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: Okay, it is doable.

seqr/utils/vcf_utils_tests.py
ShifaSZ: Okay, I'll merge the util tests to the `test_validate_anvil_vcf`.

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,80 @@
...
+                samples = set(header[len(REQUIRED_HEADERS):])
+            else:
+                field, sub_field_type = _get_vcf_meta_data(line)
+                if field:
+                    meta[field].update(sub_field_type)
+        else:
+            break
+
+    errors = _validate_vcf_header(header[0:len(REQUIRED_HEADERS)])
+    errors += _validate_vcf_meta(meta)
```
hanars, line 76: sorry I was unclear her, but if `_validate_vcf_header` fails, it should hard fail and not continue with validating the meta. Each of these functions should raise an exception if there are any error

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: In a VCF file, the meta info part comes first and doesn't depend on the header. So the meta info and the header validations are paralleled. It should be okay to provide all the errors at once.

hanars: I just think its a different level of broken. If the header itself is wrong, the file is just completely useless, there is just no way you have a joint called VCF with any of those fields missing. If the meta info is wrong you probably do have a real VCF but maybe with some formatting issues. If you have a completely broken file, all the meta errors are just confusing to the user

ShifaSZ: Okay, I'll raise the exceptions in the validation functions.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -344,19 +385,31 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
 
+        # test header errors
         mock_file_exist.return_value = True
-        mock_file_iter.return_value = ['##fileformat=VCFv4.2\n',
-                                       '#CHROM	POS	ID	REF	ALT	QUAL']  # incomplete header line
+        mock_file_iter.return_value = BASIC_META + BAD_INFO_META + BAD_FORMAT_META + BAD_HEADER_LINE + DATA_LINES
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_GZ_DATA_PATH))
         self.assertEqual(response.status_code, 400)
-        self.assertEqual(response.json()['error'],
-                         'No samples found in the provided VCF. This may be due to a malformed file')
```
hanars, line 353: we still need to test the case where there is a valid VCF header but no samples

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: Okay

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: In the original code, `no samples` is part of error handling which means that bad VCF could return no samples. The new code raises exceptions if there are any errors. How about I change the `validate_anvil_vcf` API function not to detect `no samples`? We can do that because we will have a missing sample checking in the following step.

hanars: sure but it still needs to be an explicit "no samples" error and not a "missing header column" error

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        'GT': 'String'
+    }
+}
+
+REQUIRED_HEADERS = ['#CHROM', 'POS', 'ID', 'REF', 'ALT', 'QUAL', 'FILTER', 'INFO', 'FORMAT']
+
+
+def _validate_vcf_header(header):
+    errors = []
+    if len(header) <= len(REQUIRED_HEADERS):
```
hanars, line 29: so as I said previously in my review of this PR, we allow extra headers to be present, and we don't want to detect and extra column as a sample. You need to maintain the current behavior for determining the samples. Please see my proposed solution from my previous comment: https://github.com/broadinstitute/seqr/pull/2945#discussion_r968555579

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -344,19 +385,40 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
         mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf.gz', byte_range=(0, 65536))
         mock_file_exist.assert_called_with('gs://test_bucket/test_path.vcf.gz', user=self.manager_user)
 
+        # test header errors
+        mock_file_iter.return_value = BASIC_META + BAD_INFO_META + BAD_FORMAT_META + BAD_HEADER_LINE + DATA_LINES
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_GZ_DATA_PATH))
+        self.assertEqual(response.status_code, 400)
+        self.assertListEqual(response.json()['errors'], [
+            'No samples found in the provided VCF.',
+            'Missing required VCF header field(s) POS, FILTER, INFO, FORMAT.'
```
hanars, line 405: so as per my previous review, when the header is invalid it means the file is totally malformed and we should do no other validation. You need to do 3 distinct steps of validation: validate the header, validate the samples, validate the metadata. If any of those validations fail you dont continue on to the next stage of validation and you only return the error s for that validation step, you do not run more validation and combine the errors

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: I think the `FORMAT` is not a robust delimitation.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I agree with the 3-step validation. We can extract samples after the header is validated.

seqr/utils/vcf_utils.py
ShifaSZ: Okay, I checked the VCF standard, and it says a `FORMAT` column follows any additional columns, then an arbitrary number of sample IDs.

hanars: we are going to maintain current behavior, which considers all columns after the FORMAT column to be samples

seqr/utils/vcf_utils.py
```diff
@@ -0,0 +1,82 @@
...
+                format_index = format_indices[0] + 1 if format_indices else len(header_cols)
+                header = header_cols[0:format_index]
+                samples = set(header_cols[format_index:])
+                break
+            else:
+                meta_info = _get_vcf_meta_info(line)
+                if meta_info:
+                    meta[meta_info['field']].update({meta_info['id']: meta_info['type']})
+        else:
+            raise ErrorsWarningsException(['No header found in the VCF file.'], [])
```
hanars, line 75: I like this addition, it makes for a much better user experience

hanars (approved):

PR #2958, opened 2022-09-14
support service account authorization for anvil access

hanars:
For https://github.com/broadinstitute/seqr-private/issues/1176 we will need the ability to query a list of group members as the seqr service account. This is because users need to be able to see the list of seqr analysts even if they are not themselves an analyst/ an admin on the analyst group. In order to use google OAuth's recommended service account credentials library, I reworked how we do the initial service account oauth in the setting file so I wanted to have this reviewed separately from the coming-soon much larger analyst reworking PR

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -405,11 +405,12 @@ AIRFLOW_API_AUDIENCE = os.environ.get('AIRFLOW_API_AUDIENCE')
 AIRFLOW_WEBSERVER_URL = os.environ.get('AIRFLOW_WEBSERVER_URL')
 
 if TERRA_API_ROOT_URL:
-    if not os.path.exists('/.config/service-account-key.json'):
+    service_account_file = '/.config/service-account-key.json'
+    if not os.path.exists(service_account_file):
         raise Exception('Error starting seqr - gcloud auth is not properly configured')
 
     SERVICE_ACCOUNT_CREDENTIALS = service_account.Credentials.from_service_account_file(
-        '/.config/service-account-key.json', scopes=SOCIAL_AUTH_GOOGLE_OAUTH2_SCOPE)
+        service_account_file, scopes=SOCIAL_AUTH_GOOGLE_OAUTH2_SCOPE)
     SERVICE_ACCOUNT_FOR_ANVIL = SERVICE_ACCOUNT_CREDENTIALS.service_account_email
     if not SERVICE_ACCOUNT_FOR_ANVIL:
         raise Exception('Error starting seqr - gcloud auth credentials are not properly configured')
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -398,25 +399,32 @@
...
+    service_account_file = '/.config/service-account-key.json'
+    if not os.path.exists(service_account_file):
+        raise Exception('Error starting seqr - gcloud auth is not properly configured')
 
-            SERVICE_ACCOUNT_FOR_ANVIL = re.findall(r'\[(.*)\]', auth_output)[0]
+    SERVICE_ACCOUNT_CREDENTIALS = service_account.Credentials.from_service_account_file(
+        service_account_file, scopes=SOCIAL_AUTH_GOOGLE_OAUTH2_SCOPE)
+    SERVICE_ACCOUNT_FOR_ANVIL = SERVICE_ACCOUNT_CREDENTIALS.service_account_email
+    if not SERVICE_ACCOUNT_FOR_ANVIL:
+        raise Exception('Error starting seqr - gcloud auth credentials are not properly configured')
```
sjahl, line 416: Are there cases where the call to `from_service_account_file` succeeds, but we can't pull `service_account_email` off of the resulting object? If not, I feel like these lines might be cleaner in a try/except block. But, I don't feel too strongly and defer to your judgement here.

sjahl (approved):
These changes look fine to me -- just the one inline comment.

I think longer term, we'll want to use [workload identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) for attaching the pod to a service account, or at least support it in addition to using a service account key. But, we can worry about that when it's more appropriate to move that up higher in the priority stack (it'll require some changes to the GKE cluster).

settings.py
hanars: I don't think that it is possible actually, so I will go with your suggestion

settings.py
```diff
@@ -398,25 +399,33 @@
...
+SERVICE_ACCOUNT_CREDENTIALS = None
 
 AIRFLOW_API_AUDIENCE = os.environ.get('AIRFLOW_API_AUDIENCE')
 AIRFLOW_WEBSERVER_URL = os.environ.get('AIRFLOW_WEBSERVER_URL')
 
 if TERRA_API_ROOT_URL:
-    SERVICE_ACCOUNT_FOR_ANVIL = subprocess.run(['gcloud auth list --filter=status:ACTIVE --format="value(account)"'],
+    service_account_file = '/.config/service-account-key.json'
+    if not os.path.exists(service_account_file):
+        raise Exception('Error starting seqr - gcloud auth is not properly configured')
```
ShifaSZ, line 410: It seems unnecessary to check the file's existence because it will be an exception in the following getting credential operation.

ShifaSZ (changes requested):
A minor issue you may want to solve.

ShifaSZ:
There are no tests yet for the new code. Suppose you'll add them later.

ShifaSZ (approved):

PR #2968, opened 2022-09-20
use generic load options select class

hanars:
makes the VCF loader select generic so it can be used for any dropdown that should load its own options without adding them to the redux state. This will be user more broadly in my upcoming PR, but this is a stand alone change so it wil be easier to review it separately


ShifaSZ (approved):
Good generalization.

PR #2969, opened 2022-09-20
No analyst project access

hanars:
This fundamentally changes how analysts receive access to seqr projects. Currently, there is a special project category that is used to track projects that analysts should have access to, and analysts by default have manager access to these projects. This PR removes that behavior and instead requires that an analyst user group is explicitly added to a project if it needs access. For ANVIL, and the Broad seqr instance, adding user groups to workspaces is already supported and the necessary user groups have already been added to our workspaces. For local installs, this Pr adds the ability to add groups directly to projects in the same way as adding users. By default analyst projects are disabled for local installs, but if they did enable that feature they would need to do a one-time manual update to add the analyst group to their projects. 

There were some interesting consequences to pulling out this access. Here is a high level summary of the big changes in this PR:

- Analysts no longer have access to porjects unless the user group is explicitly added
- Local installs can add collaborator groups to projects in addition to directly adding collaborators. This is not limited to analysts, they can create any groups they want.
  - This PR intentionally does not add any new UI for managing which users go in which groups. That will need to be done by administrators in the djnago admin. This is already how they will have had to manage analysts if they were using that functionality so it is okay to continue to use this approach
- For most places where we were checking the analyst group list of projects we can replace that check by simply using the full list of projects a user has access to
- There were a small number places in the code where it actually was important to distinguish between "internal" Broad projects and "external" AnVIL projects. In order to support that we now use the associated workspace namespace  to determine which projects are part of an internal namespace and which aren't. Note that this will still not grant users access to projects in the namespaces.
  - If AnVIL is not enabled there is no real way there would be both internal and external data in a seqr instance, so for local installs all projects are considered internal 
  - Due to these changes, the way we broke down projects on the seqr stats page didn't really make sense, exspecially to local users, so I reworked the breakdowns for that pae and made the behavior different for local and AnVIL instances


ShifaSZ:
> * There were a small number places in the code where it actually was important to distinguish between "internal" Broad projects and "external" AnVIL projects.

What are the relations between the "Analyst Projects" and the "Internal Broad Projects"?

hanars:
> What are the relations between the "Analyst Projects" and the "Internal Broad Projects"?

I mean there is no like formal enforced relationship between them, but right now using the values I set in our kubernestes deployment for this will result in 100% parity between the projects that are currently in the analyst project group and those that are associated with the new internal workspaces

hanars:
One other change - I updated the code to pull the anvil analysts from the group in anvil instead of the local group

ui/pages/Report/components/SeqrStats.jsx
```diff
@@ -3,28 +3,43 @@ import { connect } from 'react-redux'
...
 import { loadSeqrStats } from '../reducers'
 
-const SeqrStats = React.memo(({ stats, error, loading, load }) => (
+const DEMO_COLUMN = { title: 'Demo Projects', key: 'demo' }
+const COLUMN_MAP = {
+  [true]: [
+    { title: 'Internal Projects', key: 'internal' },
+    { title: 'External AnVIL Projects', key: 'external' },
+    DEMO_COLUMN,
+    { title: 'No Access Projects', key: 'no_access' },
```
ShifaSZ, line 18: Does the `No Access Projects` mean `inaccessible projects`? Consider renaming it.

ShifaSZ (changes requested):
It is an incredible PR.

seqr/views/apis/family_api_tests.py
```diff
@@ -98,7 +102,6 @@ def test_family_page_data(self, mock_analyst_group):
 
         response_json = response.json()
```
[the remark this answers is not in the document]
hanars: that feels out of scope

ui/pages/Report/components/SeqrStats.jsx
hanars: they aren't inaccessible because superusers and data managers do still have access, theres just no way to grant onyone else access. I think 'No AnVIL" will be clearer and will match the proposed API change

seqr/views/apis/family_api_tests.py
```diff
@@ -89,17 +88,22 @@ def test_family_page_data(self, mock_analyst_group):
...
+        family_fields.update(CASE_REVIEW_FAMILY_FIELDS)
+        internal_family_fields = deepcopy(family_fields)
+        internal_family_fields.update(INTERNAL_FAMILY_FIELDS)
+        individual_fields.update(CASE_REVIEW_INDIVIDUAL_FIELDS)
+        internal_individual_fields = deepcopy(individual_fields)
+        internal_individual_fields.update(INTERNAL_INDIVIDUAL_FIELDS)
+        self.assertSetEqual(set(response_json['familiesByGuid'][FAMILY_GUID].keys()), internal_family_fields)
+        self.assertSetEqual(set(next(iter(response_json['individualsByGuid'].values())).keys()), internal_individual_fields)
+
+        self.mock_analyst_group.__str__.return_value = ''
```
ShifaSZ, line 102: This line of code makes the `analyst_user` a non-analyst user. Right? Adding a comment or logging in with a different test user is better.

seqr/views/utils/test_utils.py
```diff
@@ -205,9 +216,17 @@ def get_initial_page_window(self, key, response):
     def get_initial_page_json(self, response):
         return self.get_initial_page_window('initialJSON', response)
 
+    def check_no_analyst_no_access(self, url, get_response=None):
```
ShifaSZ, line 219: The `url` can be optional.

ShifaSZ:
Some APIs don't have tests under the Anvil mode yet. These APIs include the `family_api`, `individual_api`, `gene_api`, `locus_list_api`, `project_categories_api`, `igv_api`, and `data_manager_api`. Do you think it is a risk?

seqr/views/apis/family_api_tests.py
hanars: Not at all. This is testing seqr's behavior when no `ANALYST_USER_GROUP` is configured, which is the default setting for local installs. It is is crucial to test that seqr works as expected when the group is entirely disabled, which is very different than testing the behavior for a non-analyst user in a seqr instance that does have analysts

seqr/views/utils/test_utils.py
hanars: I'd rather not, it makes it harder to read/maintain and its best practice to include it

hanars:
> Some APIs don't have tests under the Anvil mode yet. These APIs include the `family_api`, `individual_api`, `gene_api`, `locus_list_api`, `project_categories_api`, `igv_api`, and `data_manager_api`. Do you think it is a risk?

Not really, and that is so out of scope for this PR. If that was a risk it should have been a blocker to take AnVIL live at all

ShifaSZ (approved):

PR #2977, opened 2022-09-28
do not cache unlimited data manager project guids

hanars:
Currently, if `get_project_guids_user_can_view` is called for a data manager with `limit_data_manager=True` the limited projects are cached. Then if it is called again before the cache expires with `limit_data_manager=False` it returns the cached value which is the limited list and not the full list, which is an incorrect list of values. For the no limit case, theres no need to cache at all as there is no external requests or processing being done

### DIFF
--- a/seqr/views/utils/permissions_utils.py
+++ b/seqr/views/utils/permissions_utils.py
@@ -228,8 +228,7 @@ def check_multi_project_permissions(obj, user):
 
 
 def get_project_guids_user_can_view(user, limit_data_manager=True):
-    is_data_manager = user_is_data_manager(user)
-    if is_data_manager and not limit_data_manager:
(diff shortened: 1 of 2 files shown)


seqr/views/utils/permissions_utils.py
```diff
@@ -228,24 +228,25 @@ def check_multi_project_permissions(obj, user):
 
 
 def get_project_guids_user_can_view(user, limit_data_manager=True):
+    is_data_manager = user_is_data_manager(user)
+    if is_data_manager and not limit_data_manager:
```
ShifaSZ, line 232: Is it neater to inline `user_is_data_manager` into the `if` condition and remove `is_data_manager` variable.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #2989, opened 2022-10-05
Load LIRICAL/Exomiser data

ShifaSZ:
Add loading phenotype prioritization data function


seqr/views/utils/dataset_utils.py
```diff
@@ -449,3 +450,115 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+    count = 0
+    for line in tqdm(f, unit=' rows'):
+        row = dict(zip(header, _parse_tsv_row(line)))
+        record = _parse_phenotype_pri_row(row)
+        sample_id = record.pop('sample_id', None)
+        project = record.pop('project', None)
+        if not sample_id or not project:
+            raise ValueError('Both sample ID and project fields are required.')
+        if data_by_id[sample_id]:
+            if project != data_by_id[sample_id]['project']:  # a sample must belong to a single project
```
hanars, line 499: sample IDs are not unique across projects, and I see no reason why we shouldn't let you include the same sample ID for multiple projects in the same file. I think you should change around the structure of how you are storing this so it is a dictionary keyed by project, and then the values are a dictionary keyed by sample with a list of records. So you would have your structure be ` defaultdict(lambda: defaultdict(list))` and then all you would do here is 
```
data_by_id[project][sample_id].append(record)
```

seqr/views/utils/dataset_utils.py
```diff
@@ -449,3 +450,115 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+def load_phenotype_pri(file_path, user, ignore_extra_samples):
+    count, data_by_id = _load_phenotype_pri_file(file_path)
+
+    message = f'Parsed {count} LIRICAL/Exomiser data records in {len(data_by_id)} samples'
+    info = [message]
+    logger.info(message, user)
+
+    indivs = Individual.objects.filter(individual_id__in=data_by_id.keys())
+    prefetch_related_objects(indivs, 'family__project')
+    existing_indivs_by_id = {ind.individual_id: ind for ind in indivs
```
hanars, line 519: `individual_id` is not unique across projects, and I think the code in this file could have serious unintended consequences for data in unrelated projects because of that. While it is slightly less performant, I think it is safer to do this all on a per-project basis, and if multiple projects are represented in a file you would do the updates multiple times:
```
individual_records = {}
for project, sample_records in data_by_id.items():
    indivs = Individual.objects.filter(
        project__name=project, individual_id__in=sample_records.keys()).prefetch_related('family__project')
    existing_indivs_by_id = {ind.individual_id: ind for ind in invids}
    ...
   individual_records[existing_indivs_by_id[sample_id]] = sample_records[sample_id]
```

ui/pages/DataManagement/DataManagement.jsx
```diff
@@ -10,6 +10,7 @@ import ElasticsearchStatus from './components/ElasticsearchStatus'
 import RnaSeq from './components/RnaSeq'
 import SampleQc from './components/SampleQc'
 import Users from './components/Users'
+import PhenotypePri from './components/PhenotypePri'
```
hanars, line 13: as above, `Pri` sounds like an acronym and not an abbreviation so its confusing. `PhentoypePriority` is not that long of a class name

ui/pages/DataManagement/DataManagement.jsx
```diff
@@ -22,6 +23,7 @@ export const DATA_MANAGEMENT_PAGES = [
   { path: 'sample_qc', component: SampleQc },
   { path: 'rna_seq', component: RnaSeq },
   { path: 'users', component: Users },
+  { path: 'lirical_exomiser', component: PhenotypePri },
```
hanars, line 26: we do not want to hardcode lirical or expmiser anywhere

ui/pages/DataManagement/components/PhenotypePri.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { validators } from 'shared/components/form/FormHelpers'
+import { BooleanCheckbox } from 'shared/components/form/Inputs'
+import UploadFormPage from 'shared/components/page/UploadFormPage'
+
+import { getPhenoPriUploadStats } from '../selectors'
+import { uploadPhenoPri } from '../reducers'
+
+const mapStateToProps = state => ({
+  fields: [
```
hanars, line 11: You need to define a `FIELDS` constant, otherwise every time the component gets the chance to rerender it will recreate this list, think it is different, and rerender the whole thing

ui/pages/DataManagement/reducers.js
```diff
@@ -75,11 +76,27 @@ export const uploadRnaSeq = values => (dispatch) => {
...
 
+export const uploadPhenoPri = values => (dispatch) => {
+  let successResponseJson = null
+  return new HttpRequestHelper(
+    '/api/data_management/load_phenotype_pri_data',
+    (responseJson) => {
+      successResponseJson = responseJson
+    },
+    (e) => {
+      successResponseJson = { warnings: [e.message] }
```
hanars, line 87: There is no reason for you to write your own error handling here, it would be better if this behaved similar to how `uploadQcPipelineOutput` works

hanars (changes requested):

seqr/models.py
```diff
@@ -1053,3 +1053,26 @@ class Meta:
...
 
         json_fields = ['gene_id', 'tpm']
+
+
+class PhenotypePrioritization(models.Model):
+    TOOL_CHOICES = (
+        ('E', 'exomiser'),
+        ('L', 'lirical')
+    )
+    TOOL_LOOKUP = {v: k for k, v in TOOL_CHOICES}
```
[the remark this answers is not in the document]
ShifaSZ: Okay. I misunderstood your comments on Ben's suggestions.

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +395,30 @@ def load_rna_seq_sample_data(request, sample_guid):
     return create_json_response({'success': True})
 
 
+@data_manager_required
+def load_phenotype_pri_data(request):
+    request_json = json.loads(request.body)
+
+    file_name = request_json['file']
+    ignore_extra_samples = request_json.get('ignoreExtraSamples', False)
```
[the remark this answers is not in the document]
ShifaSZ: okay

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +395,30 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+
+    logger.info(f'Loading phenotype prioritization data from {file_name}', request.user)
+    records, info, warnings = load_phenotype_pri(file_name, request.user, ignore_extra_samples)
```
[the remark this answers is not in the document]
ShifaSZ: Agree. I was trying to implement `load_phenotype_pri` in the same way as `load_rna_seq_outlier` and `load_rna_seq_tpm`. So these functions share the same util functions in the `dataset_utils.py`. To have better modularity, I put `load_phenotype_pri` in the util file. Since it is implemented differently, I can move it back to the API file.

seqr/views/utils/dataset_utils.py
ShifaSZ: Good to know that.

ui/pages/DataManagement/DataManagement.jsx
ShifaSZ: Change it to 'phenotype_prioritization`.

seqr/views/utils/dataset_utils.py
```diff
@@ -449,3 +450,40 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
 
     return samples_to_load, info, warnings
+
+
+PHENOTYPE_PRI_HEADER = ['tool', 'project', 'sampleId', 'rank', 'geneId', 'diseaseId', 'diseaseName', 'scoreName1', 'score1']
+MAX_SCORES = 100
+
+
+def _parse_phenotype_pri_row(row):
+    record = {_to_snake_case(key): row[key] for key in PHENOTYPE_PRI_HEADER[:-2]}
```
hanars, line 460: It would be cleaner to define separate lists of validating missing columns and for parsing the row
```
PHENOTYPE_PRI_HEADER = [...]
PHENOTYPE_PRI_REQUIRED_HEADER = PHENOTYPE_PRI_HEADER + ['scoreName1', 'score1']
```

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
     return create_json_response({'success': True})
 
 
+def _load_phenotype_prioritization(file_path, user):
+    data_by_id = load_phenotype_prioritization_data_file(file_path)
+
+    all_samples = [sample for project_samples in data_by_id.values() for sample in project_samples.values()]
+    all_records = [rec for sample_records in all_samples for rec in sample_records]
+    message = f'Parsed {len(all_records)} LIRICAL/Exomiser data records in {len(all_samples)} samples'
```
hanars, line 404: computing out the totals is hard to read and doesn't add much value. I think it would be better to log here how many projects were parsed, and then have a separate log command for each project with the counts

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
     return create_json_response({'success': True})
 
 
+def _load_phenotype_prioritization(file_path, user):
+    data_by_id = load_phenotype_prioritization_data_file(file_path)
```
hanars, line 400: it would be better to use a more descriptive name, maybe `data_by_project_sample_id`

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+
+    all_samples = [sample for project_samples in data_by_id.values() for sample in project_samples.values()]
+    all_records = [rec for sample_records in all_samples for rec in sample_records]
+    message = f'Parsed {len(all_records)} LIRICAL/Exomiser data records in {len(all_samples)} samples'
+    info = [message]
+    logger.info(message, user)
+
+    for project, project_samples in data_by_id.items():
```
hanars, line 408: Instead of `project_samples` maybe something more accurate like `sample_records` or `records_by_sample`

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+        existing_indivs_by_id = {ind.individual_id: ind for ind in indivs}
+
+        tool_sample_id_set = set()
+        for sample_id, records in project_samples.items():
+            if existing_indivs_by_id[sample_id]:
+                for rec in records:
+                    rec['individual'] = existing_indivs_by_id[sample_id]
+                    tool_sample_id_set.add(f'{rec["tool"]}{sample_id}')
+            else:
+                raise ValueError(f'Individual {sample_id} doesn\'t exist in project {project}')
```
hanars, line 419: It would be neater to do this before starting iteration, you could get the list of all the missing samples
```
missing_individuals = set(project_samples.keys()) -  set(existing_indivs_by_id.keys()) 
```

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+
+        # Delete old data
+        to_delete = PhenotypePrioritization.objects.annotate(
+            tool_ind=Concat('tool', 'individual__individual_id', output_field=TextField())
+        ).filter(
+            tool_ind__in=tool_sample_id_set,
```
hanars, line 425: It is okay for us to assume that if a tool is in a file for some samples we are reloading that tool for all samples, so we can just filter for the tool and individuals separately

seqr/views/apis/data_manager_api.py
```diff
@@ -395,6 +396,69 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+        )
+        if to_delete:
+            deleted, _ = PhenotypePrioritization.bulk_delete(user, to_delete, parent='individual')
+            message = f'Deleted {deleted} existing phenotype-based prioritization records from project {project}'
```
hanars, line 429: we should use the tool names here instead of "phenotype-based prioritization"

hanars (changes requested):

seqr/utils/logging_utils.py
```diff
@@ -73,11 +73,14 @@ def log_model_update(logger, model, user, update_type, update_fields=None):
...
 
 
-def log_model_bulk_update(logger, models, user, update_type, update_fields=None):
+def log_model_bulk_update(logger, models, user, update_type, update_fields=None, parent=None):
     if not models:
         return []
     db_entity = type(models[0]).__name__
-    entity_ids = [o.guid for o in models]
+    if parent:
+        entity_ids = list({getattr(o, parent).guid for o in models})
```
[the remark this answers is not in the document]
ShifaSZ: Okay. I will add the `parentEntityIds` field to the `db_update` to align with the existing code. The new code will be as below.
```
    entity_ids = [o.guid if hasattr(o, 'guid') else o.id for o in models]
    db_update = {
        'dbEntity': db_entity, 'entityIds': entity_ids, 'updateType': 'bulk_{}'.format(update_type),
    }
    if parent:
        db_update['parentEntityIds'] = list({getattr(model, parent).guid for model in models})
```

seqr/models.py
```diff
@@ -45,6 +45,28 @@ def get_audit_field_names(audit_field):
     return list(_get_audit_fields(audit_field).keys())
 
 
+class BulkOperationBase:
```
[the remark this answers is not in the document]
ShifaSZ: Okay

seqr/views/apis/data_manager_api.py
ShifaSZ: Okay.

seqr/views/apis/data_manager_api.py
ShifaSZ: Good point.

seqr/models.py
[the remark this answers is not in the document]
ShifaSZ: Can you send me a test RNA-seq data file for uploading? I need to test the code after changing the bulk operation code.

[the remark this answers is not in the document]
hanars: the unit tests should be sufficient

seqr/models.py
```diff
@@ -1011,19 +1011,36 @@ def __unicode__(self):
...
+        """Helper bulk create method that logs the creation"""
+        for model in new_models:
+            model.created_by = user
+        models = cls.objects.bulk_create(new_models)
+        log_model_bulk_update(logger, models, user, 'create', parent=parent)
+        return models
 
     @classmethod
-    def bulk_delete(cls, user, queryset=None, **filter_kwargs):
+    def bulk_delete(cls, user, queryset=None, parent=None, **filter_kwargs):
```
hanars, line 1026: per my previous review, I do not think we should support "parent" here. That should be a different helper function if need be

seqr/utils/logging_utils.py
```diff
@@ -73,14 +73,16 @@ def log_model_update(logger, model, user, update_type, update_fields=None):
     logger.info('{} {} {}'.format(update_type, db_entity, entity_id), user, db_update=db_update)
 
 
-def log_model_bulk_update(logger, models, user, update_type, update_fields=None):
+def log_model_bulk_update(logger, models, user, update_type, update_fields=None, parent=None):
     if not models:
         return []
     db_entity = type(models[0]).__name__
-    entity_ids = [o.guid for o in models]
+    entity_ids = [o.guid if hasattr(o, 'guid') else o.id for o in models]
```
hanars, line 80: this is not the logging we care about here. for models without guids, write a different helper function, there is no need to log a list of database IDs just log the number of models

seqr/utils/logging_utils.py
```diff
@@ -73,14 +73,16 @@ def log_model_update(logger, model, user, update_type, update_fields=None):
...
     db_update = {
         'dbEntity': db_entity, 'entityIds': entity_ids, 'updateType': 'bulk_{}'.format(update_type),
     }
+    if parent:
+        db_update['parentEntityIds'] = list({getattr(model, parent).guid for model in models})
```
hanars, line 85: This does not belong in a generic helper method. For the one use case where you care about this, just write a regular looger.info message that says something like `priortization updated for 47 samples` or something like that, we don't need it to be a structured audit data.

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,68 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+def _load_phenotype_prioritization(file_path, user):
+    tool, data_by_project_sample_id = load_phenotype_prioritization_data_file(file_path)
+
+    info = []
+    _log_append_info(user, info, f'Parsed {tool.upper()} data for project(s): {", ".join(data_by_project_sample_id.keys())}')
+
+    all_records = []
+    to_delete = None
+    for project, records_by_sample in data_by_project_sample_id.items():
+        indivs = Individual.objects.filter(family__project__name=project, individual_id__in=records_by_sample.keys())
```
hanars, line 409: project names are not unique. Use the project model itself - `family__project=project`

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,68 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+
+
+@data_manager_required
+def load_phenotype_prioritization_data(request):
+    request_json = json.loads(request.body)
+
+    file_name = request_json['file']
+
+    logger.info(f'Loading phenotype-based prioritization data from {file_name}', request.user)
+    records, info, tool = _load_phenotype_prioritization(file_name, request.user)
```
hanars, line 440: This helper function doesn't really help. All the logic is essentially in the helper function and its only used once, so its just a bit less readable. A better way to do a helper function would be to make the part inside the for loop the helper and hkeep the main login here

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -281,6 +281,21 @@
...
 RNA_FILENAME_TEMPLATE = 'rna_sample_data__{}__2020-04-15T00:00:00.json.gz'
 
+PHENOTYPE_PRIORITIZATION_HEADER = ['tool\tproject\tsampleId\trank\tgeneId\tdiseaseId\tdiseaseName\tscoreName1\tscore1\tscoreName2\tscore2\tscoreName3\tscore3']
+PHENOTYPE_PRIORITIZATION_MISS_HEADER = ['tool\tsampleId\trank\tgeneId\tdiseaseName\tscoreName1\tscore1\tscoreName2\tscore2\tscoreName3\tscore3']
+LIRICAL_NO_PROJECT_DATA = ['lirical']
+LIRICAL_NO_EXIST_INDV_DATA = [
+    'lirical\tCMG_Beggs_WGS\tNA19678\t1\tENSG00000105357\tOMIM:618460\tKhan-Khan-Katsanis syndrome\tpost_test_probability\t0\tcompositeLR\t0.066',
+    'lirical\tCMG_Beggs_WGS\tNA19679\t1\tENSG00000105357\tOMIM:618460\tKhan-Khan-Katsanis syndrome\tpost_test_probability\t0\tcompositeLR\t0.066',
+]
+LIRICAL_DATA = [
```
hanars, line 291: these long strings are throwing codacy errors. Define the rows as lists and then run `'\t'.join` on them

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -743,11 +760,45 @@ def test_load_rna_seq_sample_data(self, mock_logger, mock_open, mock_os):
...
+        mock_file_iter.return_value = iter(PHENOTYPE_PRIORITIZATION_MISS_HEADER)
+        response = self.client.post(url, content_type='application/json', data=json.dumps({'file': 'lirical_data.tsv.gz'}))
+        self.assertEqual(response.status_code, 500)
+        self.assertEqual(response.json()['error'], 'Invalid file: missing column(s) project, diseaseId')
+        mock_logger.info.assert_called_with('Loading phenotype-based prioritization data from lirical_data.tsv.gz', self.data_manager_user)
+        mock_file_iter.assert_called_with('lirical_data.tsv.gz')
+
+        mock_file_iter.return_value = iter(PHENOTYPE_PRIORITIZATION_HEADER + LIRICAL_NO_PROJECT_DATA)
+        response = self.client.post(url, content_type='application/json', data=json.dumps({'file': 'lirical_data.tsv.gz'}))
+        self.assertEqual(response.status_code, 500)
```
hanars, line 789: never test a 500 error. 500 errors happen in an unhandled situation that gives me an urgent push notification, they should only happen when something is broken. If there is something that we would actually expect to happen as part of regular application usage, it needs to be a 400 error. If we really don't think this will happen, don't. test it

seqr/views/utils/dataset_utils.py
```diff
@@ -449,3 +445,47 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+        row = dict(zip(header, _parse_tsv_row(line)))
+        for row_dict in _parse_phenotype_pri_row(row):
+            sample_id = row_dict.pop('sample_id', None)
+            project = row_dict.pop('project', None)
+            if not sample_id or not project:
+                raise ValueError('Both sample ID and project fields are required.')
+            data_by_project_sample_id[project][sample_id].append(row_dict)
+            if tool and tool != row_dict['tool']:
+                raise ValueError(f'Multiple tools found {tool} and {row_dict["tool"]}. Only one in a file is supported.')
+            if not tool:
```
hanars, line 488: the control flow would be a bit neater here if it was 
```
if not tool:
    ...
elif tool != row_dict['tool']:
```

hanars (changes requested):

seqr/views/apis/data_manager_api.py
ShifaSZ: Oh, you are right. So, shouldn't the `project` column be the project name? When two projects with the same name have individuals with identical IDs, it will be trouble.

seqr/utils/logging_utils.py
ShifaSZ: I mimic the existing logging of RNA-seq data and log the `parent` models. What information should be included in the `db_update`?

seqr/views/apis/data_manager_api.py
hanars: discussed offline, need to first get the list of projects, check permission, see that the projects actually exist, and validate that there is only one project per name

seqr/utils/logging_utils.py
hanars: discussed offline, but this function should not change at all. You have 2 options:
1) Make a different logging helper for bulk updating non-guid models. It will not include any parent info or the `entityIds` list, probably add an `entityCount` instead
2) Don't treat the  data models as the thing that is being updated, log it as if the parents themselves are being updated. So instead of passing a list of  RNA or prioritization models here, pass the Individual or sample models directly, and pass the name of the set being updated in `update_fields` (i.e `update_fields=['phenotypeprioritization]`). In this case you would not put the logging in the bulk update helper function, you would call it explicitly afterwards

ShifaSZ: I will do the option 1.

hanars:
in the other PR we decided to process the score names at loading time. That change should be made in this PR

seqr/views/utils/dataset_utils.py
```diff
@@ -449,3 +445,47 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+
+def _parse_phenotype_pri_row(row):
+    record = {_to_snake_case(key): row.get(key) for key in PHENOTYPE_PRIORITIZATION_HEADER}
+
+    scores = {}
+    for i in range(1, MAX_SCORES):
+        score_name = row.get(f'scoreName{i}')
+        if not score_name:
+            break
+        scores[_to_camel_case(_to_snake_case(score_name))] = float(row[f'score{i}'])
```
ShifaSZ, line 463: We have score fields like both `compositeLR` and `post_test_probability`. So I convert everything to snake case first, then to camel case.

hanars: add a comment explaining this

seqr/utils/logging_utils.py
```diff
@@ -85,4 +85,14 @@ def log_model_bulk_update(logger, models, user, update_type, update_fields=None)
...
     logger.info(
         '{} {} {}s'.format(update_type, len(entity_ids), db_entity), user, db_update=db_update)
-    return entity_ids
\ No newline at end of file
+    return entity_ids
+
+
+def log_model_no_guid_bulk_update(logger, models, user, update_type):
+    if not models:
+        return []
```
hanars, line 93: this function has no return value, why would it return something here?

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,75 @@ def load_rna_seq_sample_data(request, sample_guid):
         data_by_gene = json.loads(row.split('\t\t')[1])
 
     model_cls = RNA_DATA_TYPE_CONFIGS[data_type]['model_class']
-    models = model_cls.objects.bulk_create([model_cls(sample=sample, **data) for data in data_by_gene.values()])
-    logger.info(f'create {len(models)} {model_cls.__name__}', request.user, db_update={
-        'dbEntity': model_cls.__name__, 'numEntities': len(models), 'parentEntityIds': [sample_guid], 'updateType': 'bulk_create',
```
hanars, line 392: I think ultimately we do want to maintain the `parentEntityIds` part of this log. You can add it back to the helper by either making `parent_ids ` an argument to `bulk_<create/delete>` and/or by adding a PARENT_FIELD property to anything that inherits from `BulkOperationBase` and the updating the logging helper to compute it:
```
def log_model_no_guid_bulk_update(logger, models, user, update_type):
    db_entity = type(models[0]).__name__
    prefetch_related_objects(models, db_entity.PARENT_FIELD)
    parent_ids = {getattr(model, db_entity.PARENT_FIELD).guid for model in models}
    ...
```

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,75 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+    logger.info(message, user)
+
+
+@data_manager_required
+def load_phenotype_prioritization_data(request):
+    request_json = json.loads(request.body)
+
+    file_path = request_json['file']
+
+    info = []
```
hanars, line 406: You are misusing `info` in this function. The user is going to get all the messages at once, at the very end as a summary. So intermediate messages are just unhelpful clutter - the user wants to see summary statistics not "Attempting to load..", "Loaded ...". These are all potentially useful as log messages because they are logged in real time not at the end, but its probably a bit overkill especially since the bulk create/delete helpers are logging whats happening as well. The end result for info probably should look something like this:
```
- Loaded {tool} data in {num_projects} projects from {file_path}
- Project {project_a}: loaded {x} records, deleted {y} existing records
- Project {project_b}: loaded {n} records, deleted {m} existing records
- etc
```

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -687,14 +719,11 @@ def mock_write(content):
...
-                        mock.call(f'delete {deleted_count} {model_cls.__name__}s', self.data_manager_user, db_update={
-                            'dbEntity': model_cls.__name__, 'numEntities': deleted_count, 'parentEntityIds': mock.ANY, 'updateType': 'bulk_delete',
-                        }),
-                    ], any_order=True
+                mock_logger.info.assert_has_calls([mock.call(info_log, self.data_manager_user) for info_log in info])
+                mock_model_logger.info.assert_called_with(
+                    f'delete {model_cls.__name__}s', self.data_manager_user,
+                    db_update={'dbEntity': model_cls.__name__, 'numEntities': deleted_count, 'updateType': 'bulk_delete'}
                 )
-                self.assertTrue(RNA_SAMPLE_GUID in mock_logger.info.call_args_list[1].kwargs['db_update']['parentEntityIds'])
```
hanars, line 697: this should be added back

seqr/views/utils/dataset_utils.py
```diff
@@ -418,12 +419,7 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
     individual_db_ids = {s.individual_id for s in samples}
     to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
     if to_delete:
-        prefetch_related_objects(to_delete, 'sample')
-        logger.info(f'delete {len(to_delete)} {model_cls.__name__}s', user, db_update={
-            'dbEntity': model_cls.__name__, 'numEntities': len(to_delete), 'updateType': 'bulk_delete',
-            'parentEntityIds': list({model.sample.guid for model in to_delete}),
-        })
-        to_delete.delete()
+        model_cls.bulk_delete(user, to_delete, parent='sample')
```
hanars, line 422: `parent` currently doesn't do anything?

seqr/views/utils/dataset_utils.py
hanars: would it be easier to just store everything as snake case? Theres no real advantage to camel case...

hanars (changes requested):

seqr/utils/logging_utils.py
ShifaSZ: Good catch. It is modified from `log_model_bulk_update`, and I forgot to change it after removing the `entity_ids`.

seqr/views/apis/data_manager_api.py
ShifaSZ: The `PARENT_FIELD` field sounds better to me.

seqr/views/apis/data_manager_api.py
ShifaSZ: Okay

seqr/views/utils/dataset_utils.py
ShifaSZ: If we convert to snake case only, then `compositeLR` will become `composite_l_r`. The frontend function `snakecaseToTitlecase` will convert it to `Composite L R`. That is not what we want.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -743,11 +800,115 @@ def test_load_rna_seq_sample_data(self, mock_logger, mock_open, mock_os):
...
+class LocalDataManagerAPITest(AuthenticationTestCase, DataManagerAPITest):
+    fixtures = ['users', '1kg_project']
+
+
+# Test for permissions from AnVIL only
+class AnvilDataManagerAPITest(AnvilAuthenticationTestCase, DataManagerAPITest):
+    fixtures = ['users', 'social_auth', '1kg_project']
+
+    @mock.patch('seqr.views.utils.permissions_utils.INTERNAL_NAMESPACES', ['my-seqr-billing'])
+    def test_load_phenotype_prioritization_data(self, *args):
```
hanars, line 913: unless you are actually going to explicitly test different behavior in the anvil and non-anvil cases, there is no point in having separate test cases for them

hanars (changes requested):

seqr/utils/logging_utils.py
```diff
@@ -85,4 +86,16 @@ def log_model_bulk_update(logger, models, user, update_type, update_fields=None)
         db_update['updateFields'] = list(update_fields)
     logger.info(
         '{} {} {}s'.format(update_type, len(entity_ids), db_entity), user, db_update=db_update)
-    return entity_ids
\ No newline at end of file
+    return entity_ids
+
+
+def log_model_no_guid_bulk_update(logger, models, user, update_type):
+    if models:
```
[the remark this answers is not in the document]
ShifaSZ: It could be done. But to keep the caller function simple and with no changes to the return values of the caller function, it looks neater to check the models here.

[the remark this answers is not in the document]
hanars: having the entire body of a function nested inside an if block is not neat. If you really need to do this check in this function, you should explicitly return when there are no models and then have the actual function code executed at the top level

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,70 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+    request_json = json.loads(request.body)
+
+    file_path = request_json['file']
+
+    try:
+        tool, data_by_project_sample_id = load_phenotype_prioritization_data_file(file_path)
+    except ValueError as e:
+        return create_json_response({'error': str(e)}, status=400)
+
+    info = [f'Loaded {tool.upper()} data from {file_path}']
```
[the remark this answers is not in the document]
ShifaSZ: Okay

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,70 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+
+    if error:
+        return create_json_response({'error': error}, status=400)
+
+    if to_delete:
+        PhenotypePrioritization.bulk_delete(request.user, to_delete)
+
+    PhenotypePrioritization.bulk_create(request.user, [PhenotypePrioritization(**data) for data in all_records])
+
+    logger.info('\n'.join(info), request.user)
```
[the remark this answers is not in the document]
ShifaSZ: okay

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,70 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+    except ValueError as e:
+        return create_json_response({'error': str(e)}, status=400)
+
+    info = [f'Loaded {tool.upper()} data from {file_path}']
+
+    all_records = []
+    to_delete = PhenotypePrioritization.objects.none()
+    error = None
+    for project_name, records_by_sample in data_by_project_sample_id.items():
+        projects = [p for p in Project.objects.filter(name=project_name) if is_internal_project(p)]
```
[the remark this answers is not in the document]
ShifaSZ: Sounds good.

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,70 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+        indivs = Individual.objects.filter(family__project=projects[0], individual_id__in=records_by_sample.keys())
+        existing_indivs_by_id = {ind.individual_id: ind for ind in indivs}
+
+        missing_individuals = set(records_by_sample.keys()) - set(existing_indivs_by_id.keys())
+        if missing_individuals:
+            error = f'Can\'t find individuals {", ".join(sorted(list(missing_individuals)))}'
+            break
+        for sample_id, records in records_by_sample.items():
+            for rec in records:
+                rec['individual'] = existing_indivs_by_id[sample_id]
```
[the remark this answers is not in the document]
ShifaSZ: I see.

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: I added the anvil tests to cover a branch in the new `is_internal_project` function. Since we remove the function, there is no need for Anvil test cases.

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: The `RNA_SAMPLE_GUID` has been added to the `parentEntityIds` field of the predicted data.

seqr/models.py
```diff
@@ -1011,19 +1011,48 @@ def __unicode__(self):
...
 
-class DeletableSampleMetadataModel(models.Model):
 
-    sample = models.ForeignKey('Sample', on_delete=models.CASCADE, db_index=True)
-    gene_id = models.CharField(max_length=20)  # ensembl ID
+class BulkOperationBase(models.Model):
+
+    @classmethod
+    def log_model_no_guid_bulk_update(cls, models, user, update_type):
+        if models:
```
hanars, line 1019: per my previous review, it is not readable to have the entire contents of a function nested inside an `if` block. You should explicitly return if there are no models and leave the rest of the function at top level

seqr/models.py
```diff
@@ -1011,19 +1011,48 @@ def __unicode__(self):
...
+            db_entity = type(models[0]).__name__
```
hanars, line 1020: You can use `cls` everywhere you are currently using `models[0]`

seqr/views/apis/data_manager_api.py
```diff
@@ -387,14 +387,81 @@ def load_rna_seq_sample_data(request, sample_guid):
...
+                indiv_records.append(rec)
+
+        exist_records = PhenotypePrioritization.objects.filter(tool=tool, individual__in=indivs)
+
+        delete_info = f'deleted {len(exist_records)} record(s), ' if exist_records else ''
+        info.append(f'Project {project_name}: {delete_info}loaded {len(indiv_records)} record(s)')
+
+        logger.info(
+            f'Project {project_name}: deleting {len(exist_records)} record(s), loading {len(indiv_records)} record(s)',
```
hanars, line 444: If you are going to log and put in the info field the exact same message except one is "loading" and one is "loaded" just log the same message as the info, the complexity is not worth it

hanars (changes requested):

seqr/models.py
ShifaSZ: Sounds good.

seqr/views/apis/data_manager_api.py
ShifaSZ: Could it be misleading to say `loaded` but not loaded because of missing individuals later on?

hanars: then move the log message to a place where it is true. Again, its really not important to have logs and info messages line up - honestly since we log the bulk creates and deletes its probably not necessary to log these summary stats at all

hanars (approved):

PR #2994, opened 2022-10-07
handle case sensitivity for anvil groups

hanars:
`permission_levels` are always lower cased but the analyst group name in AnVIL is not all lower case so we are not correctly fetching anvil users for that group


ShifaSZ (approved):
LGTM

PR #3008, opened 2022-10-13
Pop db links

hanars:
This adds a couple of links to variants, and also cleans up the code for generating that list of links as it was hard to work with

### DIFF
--- a/ui/shared/components/panel/variants/Annotations.jsx
+++ b/ui/shared/components/panel/variants/Annotations.jsx
@@ -112,19 +112,22 @@ const isTrnaOrRrna = genesById => genesById &&
     Object.values(genesById).some(({ gencodeGeneType }) => gencodeGeneType === 'Mt-tRNA' || gencodeGeneType === 'Mt-rRNA')
 
 const BaseSearchLinks = React.memo(({ variant, mainTranscript, genesById }) => {
-  const isSv = !!variant.svType
+  const {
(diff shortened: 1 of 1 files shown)


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -157,47 +157,54 @@ const BaseSearchLinks = React.memo(({ variant, mainTranscript, genesById }) => {
...
       key="seqr-search"
-      trigger={<SearchResultsLink key="seqr" buttonText="seqr" {...seqrLinkProps} />}
-      content={`Search for this variant across all your seqr projects${variant.svType ? '. Any structural variant with ≥20% reciprocal overlap will be returned.' : ''}`}
+      trigger={(
+        <SearchResultsLink
+          buttonText="seqr"
+          genomeVersion={genomeVersion}
+          svType={svType}
+          variantId={isSv ? null : variantId}
+          location={svType && (
```
ShifaSZ, line 199: Use `isSv` instead of `svType`?

ShifaSZ (changes requested):

ui/shared/components/panel/variants/Annotations.jsx
hanars: if its not an sv we want this to be null not false

ShifaSZ (approved):

PR #3011, opened 2022-10-14
Project permission performance

hanars:
remove unnecessary calls to check per-project workspace access for long lists of projects. If all we care about is view access, we should just use the list of projects users can view. For the dashboard page, 99% of the time it doesn't matter if a user can edit a project r not so don't check the permission until its needed by the user


ShifaSZ (approved):
Great PR.

PR #3012, opened 2022-10-18
Add Lirical/Exomiser data display

ShifaSZ:


seqr/views/apis/data_manager_api.py
```diff
@@ -427,7 +427,7 @@ def _load_phenotype_prioritization(file_path, user):
         deleted, _ = PhenotypePrioritization.bulk_delete(user, to_delete, parent='individual')
         _log_append_info(user, info, f'Deleted {deleted} existing {tool.upper()} records')
 
-    return all_records, info
+    return all_records, info, tool
```
hanars, line 430: this looks like a merge issue?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -592,12 +618,15 @@ BaseVariantGene.propTypes = {
...
   compoundHetToggle: PropTypes.func,
   hasRnaTpmData: PropTypes.bool,
-  rnaSeqData: PropTypes.object,
+  sampleGeneData: PropTypes.object,
 }
 
 const getRnaSeqProps = (state, ownProps) => ({
   hasRnaTpmData: getFamiliesByGuid(state)[ownProps.variant.familyGuids[0]]?.hasRnaTpmData,
-  rnaSeqData: getRnaSeqOutilerDataByFamilyGene(state)[ownProps.variant.familyGuids[0]],
+  sampleGeneData: {
```
hanars, line 626: this should not be done in the UI. You should change the server side code to return `sampleGeneData` instead of rnaSeqData and phePriData

hanars (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
hanars: no, we need it grouped by individual for the family page `RnaSeqOutliers` detail. You should just change the behavior of the `getRnaSeqOutilerDataByFamilyGene` selector to instead be `getScoresDataByFamilyGene` and have it include both rna seq and prioritization data. Thinking about it more, its probably better to leave the sever side changes you made as is and only update this one selector

ShifaSZ:
Current UI display:
Exomiser:
image

LIRICAL:
image

hanars (changes requested):

seqr/views/utils/variant_utils.py
```diff
@@ -220,4 +236,7 @@ def get_variants_response(request, saved_variants, response_variants=None, add_a
         if families_by_guid:
             _add_family_has_rna_tpm(families_by_guid)
 
+    if include_phenotype_prioritization:
+        response['phePriData'] = _get_phenotype_prioritization(genes.keys(), families)
```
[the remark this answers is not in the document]
ShifaSZ: Okay. I like `phenotypeGeneScores` more.

ui/redux/rootReducer.js
```diff
@@ -319,6 +319,7 @@ const rootReducer = combineReducers({
   mmeResultsByGuid: createObjectsByIdReducer(RECEIVE_DATA, 'mmeResultsByGuid'),
   genesById: createObjectsByIdReducer(RECEIVE_DATA, 'genesById'),
   rnaSeqDataByIndividual: createObjectsByIdReducer(RECEIVE_DATA, 'rnaSeqData'),
+  phePriDataByIndividual: createObjectsByIdReducer(RECEIVE_DATA, 'phePriData'),
```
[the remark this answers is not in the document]
ShifaSZ: How about changing it to `phenotypeGeneScoresByIndividual`?

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -314,7 +314,32 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+    ...PHENOTYPE_GENE_INFO_COLUMNS,
+    { name: 'scores.post_test_probability', content: 'Posttest-Probability', width: 3, format: ({ scores }) => (scores.post_test_probability.toPrecision(3)) },
+    { name: 'scores.compositeLR', content: 'Composite-LR', width: 3, format: ({ scores }) => (scores.compositeLR.toPrecision(3)) },
+  ],
+  exomiser: [
+    ...PHENOTYPE_GENE_INFO_COLUMNS,
+    { name: 'scores.exomiser_score', content: 'Exomiser-Score', width: 3, format: ({ scores }) => (scores.exomiser_score.toPrecision(3)) },
+    { name: 'scores.phenotype_score', content: 'Phenotype-Score', width: 3, format: ({ scores }) => (scores.phenotype_score.toPrecision(3)) },
+    { name: 'scores.variant_score', content: 'Variant-Score', width: 3, format: ({ scores }) => (scores.variant_score.toPrecision(3)) },
+  ],
```
ShifaSZ, line 341: Can we use static score fields for LIRICAL and Exomiser tools?

hanars: no

ShifaSZ: So, can we assume all the score fields are the same for records of a type of tool and generate the columns for the DataTable dynamically?

hanars: yes

ShifaSZ:
UI after updating.
RNA_Seq outliers
image

LIRICAL:
image

Exomiser:
image

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
+    description: 'Phenotype Prioritization',
+    lable: 'PhenotypeGene',
+    showDetails: (gene, { phenotypeGeneScores }) => phenotypeGeneScores && phenotypeGeneScores[gene.geneId],
+    detailsDisplay: (gene, { phenotypeGeneScores }) => (Object.entries(phenotypeGeneScores[gene.geneId]).map(
+      ([tool, data]) => {
+        PHENOTYPE_GENE_SCORE_COLUMNS[tool] = [
+          ...PHENOTYPE_GENE_INFO_COLUMNS,
+          ...Object.keys(data[0].scores).map(score => ({
+            name: score,
+            content: snakecaseToTitlecase(camelcaseToTitlecase(score)).replace(' ', '-'),
```
ShifaSZ, line 408: We use both `snakecaseToTitlecase` and `camelcaseToTitlecase` because we have both snake case and camelcase for the score names. For example, `compositeLR` and `post_test_probability`. Can we stick to just the snake case in the input file?

hanars: I agree that it would be a good idea to process the scores when we first load them to make them all consistent. Its up to you if they should all be snake case or all be title case in the database

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -867,3 +867,10 @@ def _process_result(data, model):
         data['isSignificant'] = data['pAdjust'] < model.SIGNIFICANCE_THRESHOLD
 
     return _get_json_for_models(models, process_result=_process_result, **kwargs)
+
+
+def get_json_for_phenotype_prioritization(models, **kwargs):
+    def _process_result(data, model):
+        data['scores'] = {_to_camel_case(score): value for score, value in data['scores'].items()}
```
hanars, line 874: I think we should do this at loading time and store things in the database the way we want them, instead of post processing them here

seqr/views/utils/variant_utils.py
```diff
@@ -131,6 +131,20 @@ def _get_rna_seq_outliers(gene_ids, families):
     return data_by_individual_gene
 
 
+def _get_phenotype_prioritization(gene_ids, families):
+    data_by_individual_gene = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
+
+    data_dicts = _get_json_for_models(
```
hanars, line 137: Is there a reason you defined a `get_json_for_phenotype_prioritization` helper function and don't use it here?

seqr/views/utils/variant_utils.py
```diff
@@ -161,8 +175,10 @@ def _add_pa_detail(locus_list_gene, locus_list_guid, gene_json):
 LOAD_PROJECT_TAG_TYPES_CONTEXT_PARAM = 'loadProjectTagTypes'
 LOAD_FAMILY_CONTEXT_PARAM = 'loadFamilyContext'
 
+
 def get_variants_response(request, saved_variants, response_variants=None, add_all_context=False, include_igv=True,
-                          add_locus_list_detail=False, include_rna_seq=True, include_project_name=False):
+                          add_locus_list_detail=False, include_rna_seq=True, include_project_name=False,
+                          include_phenotype_prioritization=True):
```
hanars, line 181: Is there any place in the code where we want one of rna seq and phenotype prioritization but not both? If not, we should change `include_rna_seq` to be `include_individual_gene_scores` and use one flag to control the behavior for this

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -314,7 +315,20 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
   },
 ]
 
-const RNA_SEQ_DETAIL_FIELDS = ['zScore', 'pValue', 'pAdjust']
+const RNA_SEQ_COLUMNS = [
```
hanars, line 318: It would be much cleaner to use `map` to set the values for this rather than do all this copy-paste
```
const RNA_SEQ_COLUMNS = [
  { name: 'individual', content: '', width: 3 },
  ...['zScore', 'pValue', 'pAdjust'].map(name => (
    { name, content: camelcaseToTitlecase(name).replace(' ', '-'), width: 3, format: row => row[name].toPrecision(3) }
  ),
]
```

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -314,7 +315,20 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+  { name: 'individual', content: '', width: 3 },
+  { name: 'zScore', content: 'Z-Score', width: 3, format: ({ zScore }) => (zScore ? zScore.toPrecision(3) : null) },
+  { name: 'pValue', content: 'P-Value', width: 3, format: ({ pValue }) => (pValue ? pValue.toPrecision(3) : null) },
+  { name: 'pAdjust', content: 'P-Adjust', width: 3, format: ({ pAdjust }) => (pAdjust ? pAdjust.toPrecision(3) : null) },
+]
+
+const PHENOTYPE_GENE_INFO_COLUMNS = [
+  { name: 'individual', content: '', width: 3 },
+  { name: 'rank', content: 'Rank', width: 3 },
+  { name: 'diseaseName', content: 'Disease', width: 3, format: ({ diseaseName, diseaseId }) => `${diseaseName} (${diseaseId})` },
```
hanars, line 328: It would look better to have disease come before rank

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
     color: 'pink',
     description: 'RNA-Seq Outlier',
     label: 'RNA-Seq',
-    showDetails: (gene, rnaSeqData) => rnaSeqData && rnaSeqData[gene.geneId],
-    detailsDisplay: (gene, rnaSeqData) => (
-      <div>
-        This gene is flagged as an outlier for RNA-Seq in the following samples
```
hanars, line 376: This can not be removed, it is not an acceptable change. Add it back

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
-                )}
-              </Table.Row>
-            ))}
-          </Table.Body>
-        </Table>
-      </div>
+    showDetails: (gene, { rnaSeqData }) => rnaSeqData && rnaSeqData[gene.geneId],
+    detailsDisplay: (gene, { rnaSeqData }) => (
+      <DataTable
+        basic="very"
```
hanars, line 390: why did you remove `compact="very"`

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
+        data={rnaSeqData[gene.geneId]}
+        idField="geneId"
```
hanars, line 392: gene Id is not unique - it is literally the same for every single entry in this table as it is a per gene table. To maintain behavior this needs to be `individual` the way it was before

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -314,7 +315,20 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
   },
 ]
 
-const RNA_SEQ_DETAIL_FIELDS = ['zScore', 'pValue', 'pAdjust']
+const RNA_SEQ_COLUMNS = [
+  { name: 'individual', content: '', width: 3 },
+  { name: 'zScore', content: 'Z-Score', width: 3, format: ({ zScore }) => (zScore ? zScore.toPrecision(3) : null) },
+  { name: 'pValue', content: 'P-Value', width: 3, format: ({ pValue }) => (pValue ? pValue.toPrecision(3) : null) },
+  { name: 'pAdjust', content: 'P-Adjust', width: 3, format: ({ pAdjust }) => (pAdjust ? pAdjust.toPrecision(3) : null) },
```
hanars, line 322: `DataTable` will set the column widths equally by default if you don't specify a width so unless you want different sizes you should leave the width prop out

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
     ),
   },
+  {
+    color: 'orange',
+    description: 'Phenotype Prioritization',
+    lable: 'PhenotypeGene',
+    showDetails: (gene, { phenotypeGeneScores }) => phenotypeGeneScores && phenotypeGeneScores[gene.geneId],
+    detailsDisplay: (gene, { phenotypeGeneScores }) => (Object.entries(phenotypeGeneScores[gene.geneId]).map(
+      ([tool, data]) => {
+        PHENOTYPE_GENE_SCORE_COLUMNS[tool] = [
```
hanars, line 404: The complexity here is not worth it. Make a single "scores" column, and then have the content for each cell render the scores - something like 
```
format: ({ scores }) => Object.keys(scores).sort().map(scoreName => ([
  <div key={scoreName}>
    <b>{camelcaseToTitlecase(score)}</b>: {scores[scoreName]}
  </div>
])
```

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -421,10 +449,18 @@ const OmimSegments = styled(Segment.Group).attrs({ size: 'tiny', horizontal: tru
...
+    {
+      ...sectionConfig,
+      detail: showDetails(gene, individualGeneData) && detailsDisplay(gene, individualGeneData),
+    }),
+).reduce((acc, config) => (Array.isArray(config.detail) ?
+  [
+    ...acc,
+    ...config.detail.map(([tool, detail]) => ({ ...config, label: tool.toUpperCase(), detail })),
+  ] : [...acc, config]),
+[]).filter(({ detail }) => detail).map(({ detail, expandedDisplay, ...sectionConfig }) => (
```
hanars, line 463: you should do the filter before the reduce, or do the filtering as part of the reduce

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -421,10 +449,18 @@ const OmimSegments = styled(Segment.Group).attrs({ size: 'tiny', horizontal: tru
...
-    { ...sectionConfig, detail: showDetails(gene, rnaSeqData) && detailsDisplay(gene, rnaSeqData) }),
-).filter(({ detail }) => detail).map(({ detail, expandedDisplay, ...sectionConfig }) => (
+    {
+      ...sectionConfig,
+      detail: showDetails(gene, individualGeneData) && detailsDisplay(gene, individualGeneData),
+    }),
+).reduce((acc, config) => (Array.isArray(config.detail) ?
+  [
+    ...acc,
+    ...config.detail.map(([tool, detail]) => ({ ...config, label: tool.toUpperCase(), detail })),
```
hanars, line 461: this is hard to read - make each item in `config.detail` an object, in this case with keys `detail` and `label` and then here you can do 
```
...config.detail.map(detail=> ({ ...config, ...detail }))
```

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
-              <Table.HeaderCell />
-              {RNA_SEQ_DETAIL_FIELDS.map(
-                field => <Table.HeaderCell key={field}>{camelcaseToTitlecase(field).replace(' ', '-')}</Table.HeaderCell>,
-              )}
-            </Table.Row>
-          </Table.Header>
-          <Table.Body>
-            {Object.entries(rnaSeqData[gene.geneId]).map(([individual, data]) => (
-              <Table.Row key={individual}>
-                <Table.HeaderCell>{individual}</Table.HeaderCell>
```
hanars, line 389: Since you can no longer use `HeaderCell` for the individual column, you need to have the format function set it to be wrapped in a `<b>` component

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +384,47 @@ const GENE_DETAIL_SECTIONS = [
...
+            format: ({ scores }) => (scores[score].toPrecision(3)),
+          })),
+        ]
+        return ([
+          tool,
+          (
+            <DataTable
+              basic="very"
+              data={data}
+              idField="diseaseId"
```
hanars, line 419: you may need to construct a composite id column. There is no reason to assume you couldn't have the same disease for multiple individuals

ui/shared/components/panel/variants/selectors.js
```diff
@@ -15,21 +15,41 @@ import {
...
-        (acc2, [geneId, data]) => (data.isSignificant ?
-          { ...acc2, [geneId]: { ...(acc2[geneId] || {}), [displayName]: data } } : acc2
-        ), acc[familyGuid] || {},
-      )
-      return acc
-    }, {},
+  getPhenotypeGeneScoresByIndividual,
+  (individualsByGuid, rnaSeqDataByIndividual, phenotypeGeneScoresByIndividual) => (
+    Object.entries(individualsByGuid).reduce((acc, [individualGuid, { familyGuid, displayName }]) => {
+      const rnaSeqData = rnaSeqDataByIndividual && rnaSeqDataByIndividual[individualGuid]?.outliers
```
hanars, line 27: `rnaSeqDataByIndividual` and `phenotypeGeneScoresByIndividual` will never be null, they are initialized in the state as empty objects

ui/shared/components/panel/variants/selectors.js
```diff
@@ -15,21 +15,41 @@ import {
...
+      const phenotypeGeneScores = phenotypeGeneScoresByIndividual && phenotypeGeneScoresByIndividual[individualGuid]
+      if (!rnaSeqData && !phenotypeGeneScores) {
+        return acc
+      }
+      return {
+        ...acc,
+        [familyGuid]: {
+          rnaSeqData: Object.entries(rnaSeqData || {}).reduce(
```
hanars, line 35: We want these to be null or undefined when there is no data, we do not want empty dictionaries as javascript treats those as truthy. Its probably better to handle these explicitly in if blocks:
```
acc[familyGuid] = acc[familyGuid] || {}
if (rnaSeqData) {
  acc[familyGuid].rnaSeqData = Object.entries(rnaSeqData).reduce(
...
}
if (phenotypeGeneScores) {
  acc[familyGuid].phenotypeGeneScores = Object.entries(phenotypeGeneScores).reduce(
...
}
```

ui/shared/components/panel/variants/selectors.js
```diff
@@ -15,21 +15,41 @@ import {
...
+            (acc2, [geneId, data]) => (data.isSignificant ? {
+              ...acc2,
+              [geneId]: [...(acc2[geneId] || []), { ...data, individual: displayName }],
```
hanars, line 38: its a little confusing to have this key be `individual`, it implies that it is the full individual object. `individualId` or `individualName` would be better

hanars (changes requested):

seqr/views/utils/variant_utils.py
ShifaSZ: The only thing that the `get_json_for_phenotype_prioritization` does is formating the score field names. I think we can get rid of the helper function in the end.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay, I'll add the message back.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Do you mean the prop for the `DataTable`? I can't find the `compact` in the prop list of DataTable.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: There is no `individual` field in the data. Can we add an `individual` field in the selector? Or should we remove the `pop('individualGuid')` in the backend?

ui/shared/components/panel/variants/VariantGene.jsx
hanars: it passes through props to the table so you can pass any table prop to it. Since you have to maintain the existing functionality, you need to pass the same props

ui/shared/components/panel/variants/VariantGene.jsx
hanars: you add it in the selector already??

ShifaSZ: Oh, yes.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Oh, yes. I thought `individual-gene-disease` was unique. But we can have multiple individuals with the same diseaseId. Since all data belong to a single gene, I'm going to compose individual and `diseaseId` into a new column for the `idField`.

ui/shared/components/panel/variants/selectors.js
ShifaSZ: Okay. Since it is the `displayName`, I'll use `individualName`.

ShifaSZ:
The column widths are not perfect. I'm still looking for a better way to change them.
image

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -275,6 +285,8 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts, mock_erro
         response_json = response.json()
         self.assertSetEqual(set(response_json.keys()), set(EXPECTED_SEARCH_RESPONSE.keys()))
         self.assertDictEqual(response_json, EXPECTED_SEARCH_RESPONSE)
+        lirical_data = response_json['phenotypeGeneScores']['I000001_na19675']['ENSG00000268903']['lirical']
+        self.assertListEqual(sorted(lirical_data, key=lambda d: d['diseaseId']), EXPECTED_LIRICAL_DATA)
```
hanars, line 289: just have the lirical data be sorted in the response so you can deterministically test this

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -316,6 +317,33 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+
+const RNA_SEQ_COLUMNS = [
+  INDIVIDUAL_NAME_COLUMN,
+  ...RNA_SEQ_DETAIL_FIELDS.map(name => (
+    { name, content: camelcaseToTitlecase(name).replace(' ', '-'), format: row => row[name].toPrecision(3) }
+  )),
+]
+
+const PHENOTYPE_GENE_INFO_COLUMNS = [
+  { ...INDIVIDUAL_NAME_COLUMN, width: 4 },
```
hanars, line 330: the width in the screenshot is not the width you are specifying in this config, its just setting each column to its own minimum width because of the `singleLine` and `fixedWdith` props. It looks good as it is, but you should take the unused `width` props out here

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -316,6 +317,33 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+  { name: 'diseaseName', content: 'Disease', width: 5, format: ({ diseaseName, diseaseId }) => `${diseaseName} (${diseaseId})` },
```
hanars, line 331: it will look better with the disease ID on a different line from the disease name. Maybe 
```
<div>
  {diseaseName}
  <br />
  <i>{diseaseId}<i>
</div>
```

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -370,33 +398,40 @@ const GENE_DETAIL_SECTIONS = [
...
-                <Table.HeaderCell>{individual}</Table.HeaderCell>
-                {RNA_SEQ_DETAIL_FIELDS.map(
-                  field => <Table.Cell key={field}>{data[field].toPrecision(3)}</Table.Cell>,
-                )}
-              </Table.Row>
-            ))}
-          </Table.Body>
-        </Table>
+        <DataTable
+          basic="very"
```
hanars, line 406: Please maintain previous behavior by adding back the `compact="very"` prop

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -421,10 +456,18 @@ const OmimSegments = styled(Segment.Group).attrs({ size: 'tiny', horizontal: tru
...
-).filter(({ detail }) => detail).map(({ detail, expandedDisplay, ...sectionConfig }) => (
+    {
+      ...sectionConfig,
+      detail: showDetails(gene, individualGeneData) && detailsDisplay(gene, individualGeneData),
+    }),
+).reduce((acc, config) => (Array.isArray(config.detail) ?
+  [
+    ...acc,
+    ...config.detail.map(detail => ({ ...config, ...detail })),
+  ] : (config.detail && [...acc, config]) || acc),
```
hanars, line 469: `x && y || z` is not readable/standard syntax - that should be a ternary expression `x ? y : z`. If it is throwing an error about nested ternarys, the solution is either to  not do this in one line or to add in an explicit call to `filter`, not to replace the nested ternary with an unreadable line of code

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -592,12 +635,12 @@ BaseVariantGene.propTypes = {
...
   compoundHetToggle: PropTypes.func,
   hasRnaTpmData: PropTypes.bool,
-  rnaSeqData: PropTypes.object,
+  individualGeneData: PropTypes.object,
 }
 
 const getRnaSeqProps = (state, ownProps) => ({
   hasRnaTpmData: getFamiliesByGuid(state)[ownProps.variant.familyGuids[0]]?.hasRnaTpmData,
-  rnaSeqData: getRnaSeqOutilerDataByFamilyGene(state)[ownProps.variant.familyGuids[0]],
+  individualGeneData: getIndividualGeneDataByFamilyGene(state)[ownProps.variant.familyGuids[0]] || {},
```
hanars, line 643: Defining a new object outside of a selector that will be passed as a prop causes unnecessary re-renders and is bad practice. Change the component to handle the case where this is missing

ui/shared/components/panel/variants/selectors.js
```diff
@@ -15,21 +15,41 @@ import {
...
-        (acc2, [geneId, data]) => (data.isSignificant ?
-          { ...acc2, [geneId]: { ...(acc2[geneId] || {}), [displayName]: data } } : acc2
-        ), acc[familyGuid] || {},
-      )
+  getPhenotypeGeneScoresByIndividual,
+  (individualsByGuid, rnaSeqDataByIndividual, phenotypeGeneScoresByIndividual) => (
+    Object.entries(individualsByGuid).reduce((acc, [individualGuid, { familyGuid, displayName }]) => {
+      const rnaSeqData = rnaSeqDataByIndividual[individualGuid]?.outliers
+      const phenotypeGeneScores = phenotypeGeneScoresByIndividual[individualGuid]
+      acc[familyGuid] = acc[familyGuid] || {}
```
hanars, line 29: if there is not any data for this family, you should return the `acc` without adding an empty object for the family

hanars (changes requested):

seqr/views/apis/variant_search_api_tests.py
ShifaSZ: I'm hesitant to add sorting to the API just for testing.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: okay

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Nested ternaries throw an error. I will add a filter.

seqr/views/apis/variant_search_api_tests.py
hanars: We do it all over the place in the code. It is better to have deterministic responses anyways even if the client code doesn't really on it, because it's better for the same query to always return the same response. You can use an order statement on the database query itself to keep the order deterministic here

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay. I'll move the `|| {}` and add null checking in `showDetails` and `detailsDisplay`.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay. I'll take out all the widths.

ui/shared/components/panel/variants/selectors.js
ShifaSZ: Okay. I'll move the `acc[familyGuid]` initialization into the both `rnaSeqData` and `phenotypeGeneScores` conditional branches.

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -274,6 +284,8 @@ def test_query_variants(self, mock_get_variants, mock_get_gene_counts, mock_erro
         response_json = response.json()
         self.assertSetEqual(set(response_json.keys()), set(EXPECTED_SEARCH_RESPONSE.keys()))
         self.assertDictEqual(response_json, EXPECTED_SEARCH_RESPONSE)
+        self.assertListEqual(response_json['phenotypeGeneScores']['I000001_na19675']['ENSG00000268903']['lirical'],
+                             EXPECTED_LIRICAL_DATA)
```
hanars, line 288: just put `EXPECTED_LIRICAL_DATA` into `EXPECTED_SEARCH_RESPONSE`, there is no reason to test it separatley

hanars (changes requested):
Can you add screenshots for the latest version of the UI

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -858,7 +858,10 @@ def test_load_phenotype_prioritization_data(self, mock_model_logger, mock_logger
         self.assertEqual(response.json()['error'], "Can't find individuals NA19678x, NA19679x")
 
         # Test a successful operation
+        exist_data = PhenotypePrioritization.objects.filter()
+        PhenotypePrioritization.bulk_delete(self.data_manager_user, exist_data)  # clear existing data
```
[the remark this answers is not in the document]
ShifaSZ: Make sense

ShifaSZ:
The lastest displays are as blow pictures.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -316,6 +317,51 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+      </div>
+    ),
+  },
+  { name: 'rank', content: 'Rank' },
+  {
+    name: 'scores',
+    content: 'Scores',
+    format: ({ scores }) => Object.keys(scores).sort().map(scoreName => (
+      <div key={scoreName}>
+        <b>{camelcaseToTitlecase(scoreName).replace(' ', '-')}</b>
```
hanars, line 348: it would look better without the dashes, so I would get rid of the `replace` here

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -316,6 +317,51 @@ const GENE_DISEASE_DETAIL_SECTIONS = [
...
+        : &nbsp;
+        { scores[scoreName].toPrecision(3) }
+      </div>
+    )),
+  },
+]
+
+const getDataTable = props => (
```
hanars, line 356: It is really not good practice in react to have functions that return rendered components, its better to have it be a component and render it directly. In this case however, since all you are doing is adding a few props to the component, the cleanest way to do this would be to define a styled component with `attrs`. So something like 
```
const HoverDetailTable = styled(DataTable).attrs({ basic: 'very', compact: 'very', singleLine: true })
```

hanars (changes requested):

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Styling DataTable is not working on my first try. But I found it simpler to define a common prop constant.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -322,7 +322,7 @@ const INDIVIDUAL_NAME_COLUMN = { name: 'individualName', content: '', format: ({
 const RNA_SEQ_COLUMNS = [
   INDIVIDUAL_NAME_COLUMN,
   ...RNA_SEQ_DETAIL_FIELDS.map(name => (
-    { name, content: camelcaseToTitlecase(name).replace(' ', '-'), format: row => row[name].toPrecision(3) }
+    { name, content: camelcaseToTitlecase(name), format: row => row[name].toPrecision(3) }
```
hanars, line 325: RNA behavior needs to stay the same, this should not change. I requested this change for the phenotype hover onl

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -345,22 +345,15 @@ const PHENOTYPE_GENE_INFO_COLUMNS = [
...
 
-const getDataTable = props => (
-  <DataTable
-    basic="very"
-    compact="very"
-    singleLine
-    {...props}
-  />
-)
+const hoverTableProps = { basic: 'very', compact: 'very', singleLine: true }
```
hanars, line 356: consts like this need to be all caps, camel case means a function

hanars (changes requested):

hanars (commented):
generally looks good, but it has merge conflicts

hanars (approved):

PR #3018, opened 2022-10-20
Anvil sample subset bug

hanars:
Our pedigree utility function did not correctly detect new individuals with no sex or affected status when geenrating the list of updated individuals. This fixes that bug, and also requires AnVIL users to include sex and affected status in the pedigree file  since they really should be providing that (although they can leave it blank of they really don't know)

### DIFF
--- a/seqr/views/utils/json_to_orm_utils.py
+++ b/seqr/views/utils/json_to_orm_utils.py
@@ -68,7 +68,8 @@ def update_model_from_json(model_obj, json, user, allow_unknown_keys=False, immu
             raise ValueError('Cannot edit field {}'.format(orm_key))
         if allow_unknown_keys and not hasattr(model_obj, orm_key):
             continue
-        if getattr(model_obj, orm_key) != value:
+        model_value = getattr(model_obj, orm_key)
+        if (model_value or value) and model_value != value:
             if orm_key in internal_fields and not user_is_analyst(user):
                 raise PermissionDenied('User {0} is not authorized to edit the internal field {1}'.format(user, orm_key))
             updated_fields.add(orm_key)
--- a/seqr/views/utils/json_to_orm_utils.py
+++ b/seqr/views/utils/json_to_orm_utils.py
@@ -68,8 +68,7 @@ def update_model_from_json(model_obj, json, user, allow_unknown_keys=False, immu
             raise ValueError('Cannot edit field {}'.format(orm_key))
         if allow_unknown_keys and not hasattr(model_obj, orm_key):
             continue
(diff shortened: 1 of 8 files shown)


seqr/views/utils/json_to_orm_utils.py
```diff
@@ -68,7 +68,8 @@ def update_model_from_json(model_obj, json, user, allow_unknown_keys=False, immu
             raise ValueError('Cannot edit field {}'.format(orm_key))
         if allow_unknown_keys and not hasattr(model_obj, orm_key):
             continue
-        if getattr(model_obj, orm_key) != value:
+        model_value = getattr(model_obj, orm_key)
+        if (model_value or value) and model_value != value:
```
ShifaSZ, line 72: It means no update from `Null` to `[]` or '{}` and vice versa. Is that what we want?

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -163,10 +163,12 @@ def _convert_fam_file_rows_to_json(rows):
...
         # validate
-        if not json_record.get(JsonConstants.FAMILY_ID_COLUMN):
-            raise ValueError("Family Id not specified in row #%d:\n%s" % (i+1, json_record))
-        if not json_record.get(JsonConstants.INDIVIDUAL_ID_COLUMN):
-            raise ValueError("Individual Id not specified in row #%d:\n%s" % (i+1, json_record))
+        columns = [JsonConstants.FAMILY_ID_COLUMN, JsonConstants.INDIVIDUAL_ID_COLUMN]
+        if required_columns:
+            columns += required_columns
+        missing_cols = [col for col in columns if not json_record.get(col)]
+        if missing_cols:
```
ShifaSZ, line 170: It may be neater to change to use the `any` function like `if any(not json_record.get(col) for col in columns)`.

ShifaSZ (changes requested):

seqr/views/utils/json_to_orm_utils.py
hanars: yes

hanars: oh but I just realized we do care about `None` vs `False` for nullable booleans. I'll change this back

seqr/views/utils/pedigree_info_utils.py
hanars: no because we use the `missing_cols` list to format the error message so we would need to compute this list anyways

ShifaSZ (approved):

PR #3037, opened 2022-10-31
Remove 1kg

hanars:


seqr/utils/elasticsearch/es_search.py
```diff
@@ -325,6 +325,7 @@ def _filter_by_in_silico(self, in_silico_filters):
             self._filter(_in_silico_filter(in_silico_filters))
 
     def _filter_by_frequency(self, frequencies, clinvar_terms=None):
+        frequencies = {pop: v for pop, v in (frequencies or {}).items() if pop in POPULATIONS}
```
ShifaSZ, line 328: Is it necessary? The frontend won't provide search filters with `g1k` anymore. Do we have to remove it here?

hanars: saved searches will still have them, especially since for years if you selected the AF adjuster for all the populations that would include this population.  We should not cause every single user to have to remake every single saved search, its much better behavior to just ignore this

ShifaSZ (approved):

seqr/utils/elasticsearch/es_search.py
hanars: Not really. There are over 100,000 searches in the database - there is no real value to migrating them when this is a cleaner and more robust solution

PR #3043, opened 2022-11-04
do not include no data individuals in gregor report

hanars:


seqr/views/apis/report_api.py
```diff
@@ -707,7 +707,9 @@ def gregor_export(request, consent_code):
     individuals = Individual.objects.filter(
         family__project__consent_code=consent_code[0],
         family__project__in=projects,
-    ).prefetch_related('family__project', 'mother', 'father')
+        sample__elasticsearch_index__isnull=False,
+    ).distinct().prefetch_related('family__project', 'mother', 'father')
```
ShifaSZ, line 711: I don't understand why the `distinct` is necessary.

hanars: because otherwise it return one individual row per sample, duplicating the same individual multiple times

ShifaSZ (approved):

PR #3045, opened 2022-11-04
update cn sensititvity details

hanars:


ui/shared/components/panel/genes/GeneDetail.jsx
```diff
@@ -146,9 +146,20 @@ const DosageSensitivity = ({ gene, clingenField, scoreFields, sensitivityType, t
       scores={gene.cnSensitivity}
       fields={scoreFields}
       rankDescription="intolerant of LoF mutations"
-      note={`These are a score under development by the Talkowski lab that predict whether a gene is ${sensitivityType}
-      based on large chromosomal microarray data set analysis. Scores >${threshold} are considered to have high
-      likelihood to be ${sensitivityType}.`}
+      note={(
+        <span>
+          These are a score developed by the Talkowski lab [
```
ShifaSZ, line 151: The text differs from the requested text "This metric was developed ..."

ShifaSZ (changes requested):

ui/shared/components/panel/genes/GeneDetail.jsx
hanars: yes, because as I am the technical lead for seqr I am the final arbiter of what actually goes into the seqr UI, and I think that maintaining constancy and accuracy (this is a  score not a metric, and this wording is consistent with what users are used to, and this is consistent with the existing code) is more important than aligning with a proposed wording change. I determine which parts of change requests are worth doing and which are not, and for small wording changes the analysts don't actually care enough to make tit worth while to discuss this on the ticket, I get to just make the executive decision to tweak wording.

ShifaSZ: Okay. But `There are` is plural while `a score` is single. Is it an issue?

hanars: I am maintaining consistency with the original request

ShifaSZ (approved):

PR #3048, opened 2022-11-07
Delete servctl artifacts

sjahl:
This deletes the servctl code, YAML settings files, the now-unused kubernetes manifests, the secrets directory, and the dockerfiles for curator and redis.

This should just leave the dockerbuild for seqr, kubectl helpers, and some documentation in the deploy directory.

We had a jinja dependency listed in the requirements-dev.in file -- I removed that line and regenerated the requirements-dev.txt file, which is where the reqs changes came from.

Let me know if this looks good, and I will attach the prod release checklist. There's no deployable changes here, but I think we'll want the checklist if we merge to master.

Closes broadinstitute/seqr-private#1107, closes broadinstitute/seqr-private#1108, closes broadinstitute/seqr-private#1110, closes broadinstitute/seqr-private#1111, closes broadinstitute/seqr-private#1038, closes broadinstitute/seqr-private#1069


requirements-dev.txt
```diff
@@ -35,14 +35,10 @@ idna==2.10
     # via
     #   -c requirements.txt
     #   requests
-jinja2==3.0.3
+importlib-metadata==5.0.0
```
hanars, line 38: why are dev dependencies being added when we remove a requiremnt?

hanars (changes requested):
Is there a reason for this to go to master on its own, instead of to dev and then to master as part of a regular release?

requirements-dev.txt
sjahl: Not sure -- but they should be there either way. importlib-metadata is a package that gets pulled in on python installation versions less than 3.8. It comes in unpinned (as far as I can tell) [from the click setup.py](https://github.com/pallets/click/blob/8.0.x/setup.py#L7). I don't know why it wasn't there before, but it could be that the different release version requirement didn't exist when i last generated the requirement file, or having a newer version of python 3.7 in my venv caused it to resolve differently, click may have republished a version in the interim, etc.

sjahl:
> Is there a reason for this to go to master on its own, instead of to dev and then to master as part of a regular release?

No, only real reason is that I typically get asked to change the base branch these no-code releases when I PR them against dev.

hanars:
> No, only real reason is that I typically get asked to change the base branch these no-code releases when I PR them against dev.

In my mind, when there are changes to the deployments we want to do them as their own explicit changes instead of a mixed infrastructure/feature deployment, but for this code change theirs no change to our deployments. I think the distinction here will be less going forward now that we have moved most of the deployment code out of the repo. Up to you which branch you want to merge this to

hanars (approved):

PR #3053, opened 2022-11-07
Path quality override

hanars:


seqr/utils/elasticsearch/es_utils_tests.py
```diff
@@ -2052,8 +2056,16 @@ def test_compound_het_get_es_variants_secondary_annotation(self):
             {'terms': {'transcriptConsequenceTerms': ['frameshift_variant', 'intron']}},
         ]}}
 
+        inheritance_query = deepcopy(COMPOUND_HET_INHERITANCE_QUERY)
+        for fam_q in inheritance_query['bool']['should']:
+            fam_quality_q = fam_q['bool']['must'][1]
+            fam_quality_q['bool'] = {'should': [
+                deepcopy(fam_quality_q),
+                {'terms': {'clinvar_clinical_significance': ['Pathogenic', 'Pathogenic/Likely_pathogenic']}}
+            ]}
```
ShifaSZ, line 2065: Are there any reasons you don't define another constant for the `inheritance_query`?

ShifaSZ (changes requested):

seqr/utils/elasticsearch/es_utils_tests.py
hanars: its only used once and our constant namespace is pretty polluted, so I feel like inline definitions are more readable at this point. But if you feel strongly I can make it a constant

ShifaSZ: Our rule is to use constants/literals for the expected data if possible. This one looks more straightforward with constant or literal data.

ShifaSZ (approved):

PR #3061, opened 2022-11-09
update docker-compose redis

hanars:
Now that we no longer are maintaining our own redis docker image, we shouldn't reference it in the local docker compose


sjahl (approved):
This looks good to me 👍

Issue #3025, opened 2022-10-21
Add ability to see Lirical and Exomiser results on the Family Page


lynnpais:
**Is your feature request related to a problem? Please describe.**
We'd like to quickly review results

**Describe the solution you'd like**
Add ability to see Lirical and Exomiser results on the Family Page. Include rank number and disease association. Also include the ability to search for variants in these genes. 

**Describe alternatives you've considered**
NA

**Additional context**
Previously discussed

hanars:
We will add a link to the family page under the "Show RNA-Seq OUTRIDER" link that says "Show Phenotype Prioritized Genes". When clicked it will open a modal which displays a table of all the genes we have phenotype prioritization scores for(we only load data for high-ranked genes). The table itself should have a first column column for "Gene", the display should use the inline gene detail display we use for SV genes that shows labels for omim/constraint/etc. We should also have a gene search link for this family/ gene to that column. The other columns in the table should share the display for the table we show on the hover for the label in search results. Table should be sorted by default by rank. Above the table should be a "Search for variants in all genes" link, similar to what we show for outrider.

ShifaSZ:
> the display should use the inline gene detail display we use for SV genes that shows labels for omim/constraint/etc

I think that the inline gene detail display module is for all genes, not just SV ones. Is it right?

hanars:
yes but we want the `showInlineDetails` to be true, which is only for SVs

ShifaSZ:
> we only load data for high-ranked genes

How do we know a phenotype prioritized gene is high-ranked?

hanars:
> > we only load data for high-ranked genes
> 
> How do we know a phenotype prioritized gene is high-ranked?

Ben only gives us high ranked genes in the lirical/exomiser output, we don't need to do any additional filtering for this

ShifaSZ:
![image](https://user-images.githubusercontent.com/33550059/201169149-9b9e0b5a-0247-4a8e-be57-229fc776ced8.png)
@hanars Is this display what you mean?

hanars:
no I want the display like on SVs, where it shows the labels next to the gene, although obviously without the "show gene expression". I also think it would be better if OMIM was just an orange label, not the box

<img width="344" alt="Screen Shot 2022-11-10 at 1 31 33 PM" src="https://user-images.githubusercontent.com/24598672/201177783-eb7b5637-27d5-486f-aa51-76c3594af8d8.png">

lynnpais:
Looks good!

PR #3066, opened 2022-11-14
Add high ranked gene display on the Family page

ShifaSZ:
See this [issue](https://github.com/broadinstitute/seqr/issues/3025).


ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -32,11 +32,13 @@ import {
   ONSET_AGE_OPTIONS, INHERITANCE_MODE_OPTIONS, INHERITANCE_MODE_LOOKUP, AR_FIELDS,
 } from '../../constants'
 import { getCurrentProject } from '../../selectors'
-
+import PhenotypePrioritizedGenes from '../PhenotypePrioritizedGenes'
 import CaseReviewStatusDropdown from './CaseReviewStatusDropdown'
 
 const RnaSeqOutliers = React.lazy(() => import('../RnaSeqOutliers'))
 
+// const PhenotypePrioritizedGenes = React.lazy(() => import('../PhenotypePrioritizedGenes'))
```
ShifaSZ, line 40: To be updated

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -0,0 +1,118 @@
...
+    <DataTable
+      data={Object.entries(phenotypeGeneScores || {}).reduce((acc, [geneId, toolData]) => ([
+        ...acc,
+        ...Object.values(toolData).reduce((acc2, data) => ([
+          ...acc2,
+          ...data.map(d => ({
+            ...d,
+            geneId,
+            gene: genes[geneId],
+            variant: { variantId: `modalId-${geneId}` },
```
ShifaSZ, line 84: A fake variant data to provide a modal ID for the gene detail modal.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -387,7 +387,7 @@ const GENE_DETAIL_SECTIONS = [
     color: 'red',
     description: 'HaploInsufficient',
     label: 'HI',
-    showDetails: gene => gene.cnSensitivity.phi && gene.cnSensitivity.phi > HI_THRESHOLD,
+    showDetails: gene => gene.cnSensitivity?.phi && gene.cnSensitivity.phi > HI_THRESHOLD,
```
ShifaSZ, line 390: We don't fetch the `cnSensitivity` field of the gene for the phenotype-prioritized gene data. So it can be undefined.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -464,7 +464,7 @@ const OmimSegments = styled(Segment.Group).attrs({ size: 'tiny', horizontal: tru
   }
 `
 
-const getDetailSections = (configs, gene, compact, labelProps, individualGeneData) => configs.map(
+const getDetailSections = (configs, gene, compact, labelProps, individualGeneData, noExpand) => configs.map(
```
ShifaSZ, line 467: Add the `noExpand` parameter to disable the `expandedDisplay` config for the OMIM details. Not sure if the logic is correct or not.

seqr/urls.py
```diff
@@ -174,6 +175,7 @@
     'individual/(?P<individual_guid>[\w.|-]+)/update_case_review_discussion': update_case_review_discussion,
     'individual/(?P<individual_guid>[\w.|-]+)/update_case_review_status': update_case_review_status,
     'individual/(?P<individual_guid>[\w.|-]+)/rna_seq_data': get_individual_rna_seq_data,
+    r'individual/(?P<individual_guid>[\w.|-]+)/phenotype_gene_scores': get_individual_phenotype_gene_scores,
```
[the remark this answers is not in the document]
hanars: consistency is more important than codacy in this case, take it out

seqr/utils/gene_utils.py
```diff
@@ -24,6 +24,10 @@ def get_genes_for_variant_display(gene_ids):
     return _get_genes(gene_ids, gene_fields=VARIANT_GENE_DISPLAY_FIELDS)
 
 
+def get_genes_for_prioritized_display(gene_ids):
```
hanars, line 27: we want all the labels we show for SVS, I was just giving omim and constraint as examples. You should be loading the same data as for vartiants

seqr/views/apis/family_api.py
```diff
@@ -48,6 +49,9 @@ def family_page_data(request, family_guid):
     if sample_models.filter(sample_type=Sample.SAMPLE_TYPE_RNA).exclude(rnaseqtpm=None):
         response['familiesByGuid'][family_guid]['hasRnaTpmData'] = True
 
+    if PhenotypePrioritization.objects.filter(individual__guid=individual_guid):
```
hanars, line 52: this is inefficient, you are running one query per individual when you should be running one query total for the family

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -133,6 +135,21 @@ ShowRnaSeqOutliers.propTypes = {
   sample: PropTypes.object,
 }
 
+const ShowPhenotypePrioritizedGenes = ({ individual, hasPhenotypeGeneScores, ...props }) => (hasPhenotypeGeneScores ? (
+  <Modal
```
hanars, line 139: this is copy-paste from `ShowRnaSeqOutliers`, which means you need a shared base class, or you can define a list of links and do a map, i.e.
```
SHOW_DATA_DETAILS.filter(({ individual, shouldShowField }) => individual[shouldShowField]).map(({ individual, shouldShowField, modalTitle, linkText, elementType, ...props  }) => (
 <Modal
    modalName={toSnakecase(modalTitle)}
    title={modalTitle}
    trigger={<ButtonLink padding="1em 0 0 0" content={linkText} />}
  >
    <React.Suspense fallback={<Loader />}>
      {React.createElement(elementType, props)}
    </React.Suspense>
  </Modal>
))
```

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -0,0 +1,118 @@
...
+    format: ({ geneId, gene, variant, familyGuid }) => (
+      <div>
+        <BaseVariantGene
+          geneId={geneId}
+          gene={gene}
+          variant={variant}
+          compact
+          showInlineDetails
+        />
+        <GeneSearchLink
```
hanars, line 28: unless this is coming from `BaseVariantGene` there is no need for this, the all genes link is fine

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -0,0 +1,118 @@
...
+  <DataLoader content={phenotypeGeneScores} contentId={individual.individualGuid} load={load} loading={loading}>
+    <GeneSearchLink
+      buttonText="Search for variants in high-ranked genes"
+      icon="search"
+      location={Object.values(phenotypeGeneScores || {}).map(({ geneId }) => geneId).join(',')}
+      familyGuid={familyGuid}
+      floated="right"
+    />
+    <DataTable
+      data={Object.entries(phenotypeGeneScores || {}).reduce((acc, [geneId, toolData]) => ([
```
hanars, line 76: you can never create arrays or objects directly as props, it causes react to unneccessarily re-render. You need to make a selector for this

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
hanars: update the component to accept a `modalId` prop and use that instead of the variant if it is passed in

hanars: this will also not be a unique ID, you need to use the unique `rowId` in the modal ID

ui/shared/components/panel/variants/VariantGene.jsx
hanars: as I said above, we need to fetch the same data as for variants, we want to show everything

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -475,11 +475,11 @@ const getDetailSections = (configs, gene, compact, labelProps, individualGeneDat
     ...acc,
     ...config.detail.map(detail => ({ ...config, ...detail })),
   ] : [...acc, config]),
-[]).map(({ detail, expandedDisplay, ...sectionConfig }) => (
-  (expandedDisplay && !compact) ? (
+[]).map(({ detail, expandedDisplay, expandedLabel, ...sectionConfig }) => (
+  (expandedDisplay && !compact && !noExpand) ? (
```
hanars, line 479: it looks like `compact` is only used in this function in this line, so wouldn't it be cleaner to change what you pass as `compact` for this function instead of adding another prop that is also only used here?

hanars (changes requested):

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Okay

seqr/views/apis/family_api.py
ShifaSZ: The data is fetched when the users click the `Show Phenotype Prioritized Genes` button. Do the users usually click and see the high-ranked genes for multiple members in the family?

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
ShifaSZ: Sounds good.

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
ShifaSZ: Okay

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
ShifaSZ: Okay

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: The `compact` is also used in a few lines below. It doesn't look okay if `compact` is set to `true`.
Is there any scenario in which we need to expand the `OMIM` details, and `showInlineDetails` is `true`? It seems not to me. So we could use `showInlineDetails` to disable expanding the `OMIM` details.

seqr/views/apis/family_api.py
hanars: this is an API endpoint returning all the data for a family. You need to return all the data a user would potentially see in that family. That means you need to return the data for all individuals in that family. How elese will you know which individuals should even have the button?

ShifaSZ: Okay. It seems that my implementation of `hasPhenotypeGeneScores` isn't correct.

ShifaSZ: Do you think we should have per individual `hasPhenotypeGeneScores`?  So that the individuals who don't have data won't show the button.

hanars: yes

ui/shared/components/panel/variants/selectors.test.js
```diff
@@ -95,7 +96,6 @@ const RNA_SEQ_PHENOTYPE_PRIORITIZATION_STATE = {
         lirical: [{
           diseaseId: 'OMIM:618460',
           diseaseName: 'Khan-Khan-Katsanis syndrome',
-          geneId: 'ENSG00000228198',
```
ShifaSZ, line 98: No `geneId` field in the data returning from the backend.

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -500,8 +500,9 @@ export const GeneDetails = React.memo((
   { gene, compact, showLocusLists, showInlineDetails, individualGeneData, ...labelProps },
 ) => {
   const geneDetails = getDetailSections(GENE_DETAIL_SECTIONS, gene, compact, labelProps, individualGeneData)
-  const geneDiseaseDetails = getDetailSections(GENE_DISEASE_DETAIL_SECTIONS, gene, compact, labelProps)
-  const hasLocusLists = showLocusLists && gene.locusListGuids.length > 0
+  const geneDiseaseDetails = getDetailSections(GENE_DISEASE_DETAIL_SECTIONS, gene, compact, labelProps,
+    null, showInlineDetails)
```
hanars, line 504: this seems like it would break behavior in cases where `showInlineDetails` is already being passed in, can you confirm that in cases where that was already being used nothing is changing?

hanars (changes requested):

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -171,11 +176,27 @@ const DataDetails = React.memo(({ loadedSamples, individual, mmeSubmission }) =>
...
+        return (
+          <Modal
+            key={modalName(data[dataType])}
+            modalName={modalName(data[dataType])}
+            title={title(data[dataType])}
+            trigger={<ButtonLink padding="1em 0 0 0" content={linkText} />}
+          >
+            <React.Suspense fallback={<Loader />}>
+              {React.createElement(component,
+                { familyGuid: individual.familyGuid, individualGuid: data[dataType].individualGuid }) }
```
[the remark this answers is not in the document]
ShifaSZ: It should also work. I kept the existing logic of the `RnaSeqOutliers` component.

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -171,11 +176,27 @@ const DataDetails = React.memo(({ loadedSamples, individual, mmeSubmission }) =>
...
       ) : <MmeStatusLabel title="Submitted to MME" dateField="lastModifiedDate" color="violet" individual={individual} mmeSubmission={mmeSubmission} />
     )}
-    <ShowRnaSeqOutliers
-      familyGuid={individual.familyGuid}
-      sample={loadedSamples.find(({ sampleType, isActive }) => isActive && sampleType === SAMPLE_TYPE_RNA)}
-      hasRnaOutlierData={individual.hasRnaOutlierData}
-    />
+    {SHOW_DATA_MODAL_CONFIG.filter(({ shouldShowField }) => individual[shouldShowField]).map(
+      ({ dataType, modalName, title, linkText, component }) => {
+        const data = {
```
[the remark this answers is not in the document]
ShifaSZ: Make sense.

ui/shared/components/panel/variants/VariantGene.jsx
ShifaSZ: Or I could reuse the `geneModalId` parameter, which is only used by the `Phenotype Prioritized Genes` modal.

hanars: I think I slightly prefer to add an explicit `noExpand` parameter

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -722,7 +728,7 @@ class VariantGenes extends React.PureComponent {
         {!mainGeneId && (
           <div>
             {[...GENE_DISEASE_DETAIL_SECTIONS, ...GENE_DETAIL_SECTIONS].map(
-              ({ showDetails, detailsDisplay, ...sectionConfig }) => {
+              ({ showDetails, detailsDisplay, expandedDisplay, expandedLabel, ...sectionConfig }) => {
```
ShifaSZ, line 731: Add `expandedDisplay` and `expandedLabel` to avoid passing them to GeneDetailSection through ...sectionConfig and get rid of an unexpected parameter warning.

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -171,11 +172,25 @@ const DataDetails = React.memo(({ loadedSamples, individual, mmeSubmission }) =>
...
     )}
-    <ShowRnaSeqOutliers
-      familyGuid={individual.familyGuid}
-      sample={loadedSamples.find(({ sampleType, isActive }) => isActive && sampleType === SAMPLE_TYPE_RNA)}
-      hasRnaOutlierData={individual.hasRnaOutlierData}
-    />
+    {SHOW_DATA_MODAL_CONFIG.filter(({ shouldShowField }) => individual[shouldShowField]).map(
+      ({ modalName, title, linkText, component }) => {
+        const sample = loadedSamples.find(({ sampleType, isActive }) => isActive && sampleType === SAMPLE_TYPE_RNA)
+        const titleIds = { sampleId: sample.sampleId, individualId: individual.individualId }
```
hanars, line 178: what if there is no matched `sample`? Then `sample.sampleId` will error, so we need to do `sample?.sampleId``

hanars (changes requested):

hanars (approved):

PR #3078, opened 2022-11-28
update hail downstream dependencies

hanars:
Manually recreates the fixes attempted by snyk's automated update: https://github.com/broadinstitute/seqr/pull/3077

The dependency tree was a bit too complex for Snyk to patch, so needed to rerun pip compile to resolve patch versions


requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.7
+# This file is autogenerated by pip-compile with python 3.10
```
sjahl, line 2: Since we run python 3.7, we probably want to generate this reqs file with pip-compile in a 3.7 venv (and in the more general sense, it's probably a good idea to have local dev venvs match the python version that seqr actually runs with)

sjahl (changes requested):

requirements.txt
hanars: updated, although there were no changes to the dependencies in this case

sjahl: Cool, thank you! It's definitely good to keep an eye on this, because projects can specify to install different module versions in their setup.py based on the python version that's running the install 👍

sjahl (approved):

PR #3079, opened 2022-11-28
load genome version in variant project context

hanars:
IGC depends on having the project's genome version to load the correct reference context


ShifaSZ (approved):
LGTM

PR #3089, opened 2022-11-30
Project Overview performance

hanars:


ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -359,16 +359,16 @@ const mapAnalysisStatusStateToProps = (state, ownProps) => ({
...
     <Grid.Column width={5}>
       {familiesLoading ? <Dimmer inverted active><Loader /></Dimmer> : <FamiliesIndividualsOverview {...props} />}
       <VerticalSpacer height={10} />
-      <MatchmakerOverview {...props} />
+      {overviewLoading ? <Dimmer inverted active><Loader /></Dimmer> : <MatchmakerOverview {...props} />}
     </Grid.Column>
     <Grid.Column width={5}>
       <DetailSection title="Genome Version" content={GENOME_VERSION_LOOKUP[props.project.genomeVersion]} />
-      <DatasetOverview {...props} />
+      {overviewLoading ? <Dimmer inverted active><Loader /></Dimmer> : <DatasetOverview {...props} />}
```
ShifaSZ, line 371: Only the MME and the `DatasetOveriew` are left for the project overview loading, how about loading these data separately?

ShifaSZ (changes requested):
Mostly looks good.

ui/pages/Project/components/ProjectPageUI.jsx
```diff
@@ -82,7 +82,7 @@ const NO_DETAIL_FIELDS = [
 
 const ProjectPageUI = React.memo(props => (
   <Grid stackable>
-    <DataLoader load={props.load} loading={props.loading} content>
+    <DataLoader load={props.load} loading={false} content>
```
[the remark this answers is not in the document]
hanars: it is, otherwise the content would never be loaded

ui/pages/Project/components/ProjectOverview.jsx
hanars: There are also some other stuff that are displayed on different parts of the page/ different components. It might be worth it at some point to further split out the overview, but I time boxed this PR in order to prevent myself from spending too long. My plan is after this is live to see how the performance is in production, and then if its still slow to do another iteration where I split this endpoint further

ShifaSZ (approved):

PR #3104, opened 2022-12-06
Resolve float conversion issue.

ShifaSZ:
See this [ticket](https://github.com/broadinstitute/seqr-private/issues/1203).


hanars:
PRS always need to be against dev not master. I've fixed it here

hanars:
Can you add a screenshot of what it looks like in both the saved variants/ search results hover and the family page where you have a a gene where a score is missing for some phenotypes but not t=others

ShifaSZ:
In family page:
image

hanars (approved):

PR #3105, opened 2022-12-06
Correct a miss-leading error message when the permission is insufficient.

ShifaSZ:
See this [ticket.](https://github.com/broadinstitute/seqr-private/issues/1202)

### DIFF
--- a/seqr/views/apis/data_manager_api_tests.py
+++ b/seqr/views/apis/data_manager_api_tests.py
@@ -814,30 +814,38 @@ class DataManagerAPITest(AuthenticationTestCase):
     def _join_data(cls, data):
         return iter(['\t'.join(line) for line in data])
 
+    @mock.patch('seqr.utils.file_utils.subprocess.Popen')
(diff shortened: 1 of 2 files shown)


seqr/views/apis/data_manager_api_tests.py
```diff
@@ -814,53 +814,61 @@ def test_load_rna_seq_sample_data(self, mock_model_logger, mock_logger, mock_ope
     def _join_data(cls, data):
         return iter(['\t'.join(line) for line in data])
 
+    @mock.patch('seqr.utils.file_utils.subprocess.Popen')
     @mock.patch('seqr.views.utils.dataset_utils.file_iter')
```
hanars, line 818: for consistency, if we are mocking `file_iter` at the `dataset_utils` level we should also mock `does_file_exist` at the `dataset_utils`. Alternatively, we could mock `file_utils.subprocess` and use that mock for both the file exist and file iter

hanars (changes requested):

hanars (approved):

Issue #3110, opened 2022-12-08
Filters for LIRICAL feature


lynnpais:
**Is your feature request related to a problem? Please describe.**
LIRICAL's 'Show Phenotype Prioritized Genes' feature is returning too many variants for an analyst to review

**Describe the solution you'd like**
Reduce the number of genes included in the 'Show Phenotype Prioritized Genes' feature to the top 10 ranked genes only. 

Also, please include the option to sort by LIRICAL results on the variant results page

**Describe alternatives you've considered**
Discussed by the analysts

**Additional context**
NA

hanars:
> Also, please include the option to sort by LIRICAL results on the variant results page

This is a totally different feature and requires its own ticket

ShifaSZ:
> Reduce the number of genes included in the 'Show Phenotype Prioritized Genes' feature to the top 10 ranked genes only.

Do we also show a tip message if more genes/phenotypes exist beyond the top 10?

lynnpais:
For now, change 'Search for Variants in High Ranked Genes' to 'Search for Variants in Top 10 Ranked Genes' and only return variants for top 10 ranked genes. 

Also, can we restrict variants in these genes to <1% AF? I'll put in a separate ticket for this and the sort by feature

hanars:
so the sort by feature is a separate feature from changes to the "search for variants link" This ticket will encompass any changes in the search link, so changes to the default AF cutoff for that link belong in this ticket, not the sort ticket

ShifaSZ:
We can paginate the data table to display the top 10 genes. And more genes are accessible by simply clicking on the page numbers at the end of the page. See the picture below. Does it work for this requirement?
![image](https://user-images.githubusercontent.com/33550059/206546686-286d24b4-887a-41d6-a94e-f5534449dbc0.png)

hanars:
I would rather we only show the top 10, and actually only fetch the top 10 from the database. This should also be true when loading lirical labels for search results, we shoudl only query the database for the top 10, and this should share code

PR #3114, opened 2022-12-08
Show top 10 ranked genes on the 'Show Phenotype Prioritized Genes' pop window

ShifaSZ:
See the details in this [ticket](https://github.com/broadinstitute/seqr/issues/3110).


hanars:
per my comment on the ticket, this is not at all the approach we should take here

PR #3115, opened 2022-12-09
Show top 10 ranked phenotype prioritized genes

ShifaSZ:
See this [ticket](https://github.com/broadinstitute/seqr/issues/3110) for details.

### DIFF
--- a/seqr/views/apis/variant_search_api_tests.py
+++ b/seqr/views/apis/variant_search_api_tests.py
@@ -60,9 +60,11 @@ expected_detail_saved_variant['mainTranscriptId'] = mock.ANY
 
 EXPECTED_EXOMISER_DATA = [
     {'diseaseId': 'OMIM:219800', 'diseaseName': 'Cystinosis, nephropathic', 'rank': 2,
-     'scores': {'compositeLR': 0.003, 'post_test_probability': 0}},
+     'scores': {'exomiser_score': 0.969347946, 'phenotype_score': 0.443567539,
+                'variant_score': 0.999200702}},
     {'diseaseId': 'OMIM:618460', 'diseaseName': 'Khan-Khan-Katsanis syndrome', 'rank': 1,
-     'scores': {'compositeLR': 0.066, 'post_test_probability': 0}},
+     'scores': {'exomiser_score': 0.977923765, 'phenotype_score': 0.603998205,
+                'variant_score': 1}},
 ]
 
 EXPECTED_LIRICAL_DATA = [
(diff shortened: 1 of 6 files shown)


seqr/views/utils/variant_utils.py
```diff
@@ -147,9 +147,12 @@ def get_phenotype_prioritization(families, gene_ids=None):
     data_by_individual_gene = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
 
     gene_filter = {'gene_id__in': gene_ids} if gene_ids is not None else {}
+    phenotype_genes = PhenotypePrioritization.objects.none()
+    for indiv in Individual.objects.filter(family__in=families):
+        phenotype_genes |= PhenotypePrioritization.objects.filter(individual=indiv, **gene_filter).order_by('rank')[:10]
```
hanars, line 152: this contains many unnecessary queries. In general, you always want as few database queries as possible. In this case, instead of ordering by rank and picking the top 10 for each individual, why not just explicitly filter for rank:
```
PhenotypePrioritization.objects.filter(
  individual__family__in=families, rank__lte=10,  **gene_filter,
).order_by('disease_id')
```

hanars (changes requested):

seqr/views/utils/variant_utils.py
ShifaSZ: Oh, I didn't know that we could use the value of the `rank` field. But I'm curious since we never return records with a `rank` of more than 10, why do we still load them?

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -96,7 +96,6 @@
     'rnaSeqData': {'I000001_na19675': {'outliers': {'ENSG00000268903': mock.ANY}}},
     'phenotypeGeneScores': {
         'I000001_na19675': {'ENSG00000268903': {'exomiser': EXPECTED_EXOMISER_DATA}},
-        'I000002_na19678': {'ENSG00000268903': {'lirical': EXPECTED_LIRICAL_DATA}},
```
hanars, line 99: Can we change the rank on the other entry fixture, or add a new fixture with lower rank? I do still want to test lirical data in search results

seqr/views/utils/variant_utils.py
```diff
@@ -6,7 +6,7 @@
 from matchmaker.models import MatchmakerSubmissionGenes, MatchmakerSubmission
 from reference_data.models import TranscriptInfo
 from seqr.models import SavedVariant, VariantSearchResults, Family, LocusList, LocusListInterval, LocusListGene, \
-    RnaSeqOutlier, RnaSeqTpm, PhenotypePrioritization
+    RnaSeqOutlier, RnaSeqTpm, PhenotypePrioritization, Individual
```
hanars, line 9: this is unused now

hanars (changes requested):

seqr/views/apis/variant_search_api_tests.py
ShifaSZ: Sure

hanars (approved):

PR #3117, opened 2022-12-14
Anvil loading delay settings

hanars:
Re-enables the loading warnings we had last year, and changes the mechanism to be a single env variable to turn on/off the message rather than requiring a full code deployment every year to turn this on. Text is based on what we did last year: https://github.com/broadinstitute/seqr/commit/518a1acb0eda7eaf24f50ddc7ab4577dcf76970d


seqr/views/react_app.py
```diff
@@ -43,11 +44,15 @@ def render_app_html(request, additional_json=None, include_user=True, status=200
         ui_version = 'local'
         html = html.replace('</head>', '<script defer="defer" src="/app.js"></script></head>')
 
+    should_show_loading_delay = bool(ANVIL_LOADING_DELAY_EMAIL_DATE)
+    if should_show_loading_delay:
+        should_show_loading_delay = datetime.strptime(ANVIL_LOADING_DELAY_EMAIL_DATE, '%Y-%m-%d') < datetime.now()
```
ShifaSZ, line 49: How about moving the time checking to the front end?

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -33,14 +33,13 @@
 from seqr.utils.middleware import ErrorsWarningsException
 from seqr.views.utils.permissions_utils import is_anvil_authenticated, check_workspace_perm, login_and_policies_required
 from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL, POLICY_REQUIRED_URL, API_POLICY_REQUIRED_URL,\
-    SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, AIRFLOW_API_AUDIENCE, AIRFLOW_WEBSERVER_URL, SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL
+    SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL, AIRFLOW_API_AUDIENCE, AIRFLOW_WEBSERVER_URL, ANVIL_LOADING_DELAY_EMAIL_DATE, \
```
ShifaSZ, line 36: If you don't think it is too long, it is less confusing by changing the constant name to `ANVIL_LOADING_DELAY_EMAIL_START_DATE`.

ShifaSZ (changes requested):
Two non-critical comments.

seqr/views/react_app.py
hanars: I'm not in love with javascript's date processing and I'd rather leave this sort of business logic server side

ShifaSZ (approved):

PR #3132, opened 2023-01-03
Show MME tags in saved variant bar

hanars:


seqr/views/apis/family_api.py
```diff
@@ -72,11 +72,13 @@ def family_variant_tag_summary(request, family_guid):
...
     response = families_discovery_tags([{'familyGuid': family_guid}])
 
-    family_tag_type_counts = VariantTag.objects.filter(saved_variants__family=family).values(
-        'variant_tag_type__name').annotate(count=Count('*'))
+    tags = VariantTag.objects.filter(saved_variants__family=family)
+    family_tag_type_counts = tags.values('variant_tag_type__name').annotate(count=Count('*'))
     response['familyTagTypeCounts'] = {
         family_guid: {c['variant_tag_type__name']: c['count'] for c in family_tag_type_counts},
     }
+    response['familyTagTypeCounts'][family_guid]['MME Submission'] = tags.filter(
```
ShifaSZ, line 80: Use a tag name constant for `MME Submission`?

seqr/views/utils/project_context_utils.py
```diff
@@ -141,6 +141,15 @@ def add_project_tag_types(projects_by_guid):
             project_guid = vtt.project.guid if vtt.project else None
             project_tag_types[project_guid].append(variant_tag_types_by_guid[vtt.guid])
 
+    project_tag_types[None].append({
```
ShifaSZ, line 144: Why is the new tag added to the key `None` list?

ui/pages/Project/selectors.js
```diff
@@ -229,8 +233,13 @@ export const getSavedVariantTagTypeCountsByFamily = createSelector(
 
 export const getSavedVariantTagTypeCounts = createSelector(
   getVariantTagsByGuid,
-  variantTagsByGuid => Object.values(variantTagsByGuid).reduce(
-    (acc, { name }) => ({ ...acc, [name]: (acc[name] || 0) + 1 }), {},
+  getSavedVariantsByGuid,
+  (variantTagsByGuid, savedVariantsByGuid) => Object.values(variantTagsByGuid).reduce(
+    (acc, { name }) => ({ ...acc, [name]: (acc[name] || 0) + 1 }), {
+      [MME_TAG_NAME]: Object.values(savedVariantsByGuid).filter(
+        ({ mmeSubmissions = [] }) => mmeSubmissions.length > 0,
```
ShifaSZ, line 240: Just curious about if this line is equivalent to `({ mmeSubmissions }) => mmeSubmissions?.length > 0,`

ShifaSZ (changes requested):

seqr/views/utils/project_context_utils.py
hanars: because this is a dict that groups tags by project, and uses None as the key for tags that go to every project, which is what we want in this case

ui/pages/Project/selectors.js
hanars: yes

ShifaSZ (approved):

Issue #3123, opened 2022-12-15
Updates to seqr-Anvil loading page


lynnpais:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when ... external users ask silly questions. 

**Describe the solution you'd like**
Based on recent questions, can we make the following changes-
1. Add an example pedigree file with 1-2 families. In one family, show at least two affecteds. 
2. To the error message (when info is missing in the pedigree file), request the user to add a row for the missing individual. 
3.  Add a PHI checkbox (which includes examples such as DOB, names, etc.) that a user must agree to before proceeding to loading data in seqr.

**Describe alternatives you've considered**
NA

**Additional context**
As discussed at the product meeting this afternoon.

PR #3135, opened 2023-01-04
Shared phi component

hanars:
Makes no real functionality changes, but breaks out the PHI confirmation logic from the pedigree image component into a sharable base component that will be used in https://github.com/broadinstitute/seqr/issues/3123

This also cleans up how error handling is managed in the base file uploader field to allow the pedigree image component to be greatly simplified

### DIFF
--- a/ui/shared/components/form/XHRUploaderWithEvents.jsx
+++ b/ui/shared/components/form/XHRUploaderWithEvents.jsx
@@ -4,6 +4,7 @@
 import React from 'react'
 import Cookies from 'js-cookie'
 import PropTypes from 'prop-types'
+import { Message } from 'semantic-ui-react'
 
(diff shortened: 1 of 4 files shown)


ui/shared/components/form/XHRUploaderWithEvents.jsx
```diff
@@ -46,6 +48,10 @@ class XHRUploaderWithEvents extends XHRUploader {
     )
   }
 
+  renderButton() {
```
hanars, line 51: currently, the default logic in the base component renders an empty space for an "upload" button that we do not show, this instead uses that space to show an error message if relevant

hanars: this class extends an imported library, so it is overriding the base class implementation

ShifaSZ (approved):

PR #3137, opened 2023-01-04
Add 'LIRICAL' sort

ShifaSZ:
Draft PR for the UI part of the change.


hanars (commented):
this looks fine, although as discussed it still needs the server side sort change

ShifaSZ:
Regarding the server side, I have the following questions.
- I will do it with the script based sorting as the in_omim does. To prepare the params for the scripts, I need to fetch the rank info from the PhenotypePrioritization model. It may be impossible to include all the ranking info into the params because there are too many (currently around 70k records). Including the records for the families in the search could be better. Should we add family info while calling the params value functions in the _sort_variants function of the EsSearch class?
- My draft idea for the sorting is like this: generate params of a list of (sample, gene, rank) tuples, create a script to get the samples from doc[‘genotypes’][‘sample_id’], and the genes from doc[‘geneIds’], and return the minimum rank in the tuples that match the samples and genes. Do you think it is the right way to do it?
- There are gene IDs in both ‘geneIds’ and ‘mainTranscript_gene_id’. Do we need to use the gene IDs from both these fields?

hanars:
> * I will do it with the script based sorting as the in_omim does. To prepare the params for the scripts, I need to fetch the rank info from the PhenotypePrioritization model. It may be impossible to include all the ranking info into the params because there are too many (currently around 70k records). Including the records for the families in the search could be better. Should we add family info while calling the params value functions in the _sort_variants function of the EsSearch class?

the `EsSearch` class has a bunch of `self` fields that track the samples/ families for the search. You should use those

> * My draft idea for the sorting is like this: generate params of a list of (sample, gene, rank) tuples, create a script to get the samples from doc[‘genotypes’][‘sample_id’], and the genes from doc[‘geneIds’], and return the minimum rank in the tuples that match the samples and genes. Do you think it is the right way to do it?

No. This search will only work on a single family, as I said on the ticket. Just use the min rank for the family as the rank for the gene, and do not bother to look at genotypes in the sort at all, only the gene

> * There are gene IDs in both ‘geneIds’ and ‘mainTranscript_gene_id’. Do we need to use the gene IDs from both these fields?

Use `geneIds`

ShifaSZ:
@hanars I added the server-side sorting for LIRICAL. Is it the right way to do it? And I still need to add the number of family checking. It could be checked at the front end or the backend. If we do it in the backend, we may raise an exception before creating the sort script params and handle the exception somewhere in the variant search API. Do you think it is okay?

seqr/utils/elasticsearch/constants.py
```diff
@@ -172,6 +173,16 @@
     }
 }
 
+
+def _get_ranks_by_gene(families):
```
hanars, line 177: I would name this `_get_phenotype_priority_ranks_by_gene` for clarity

seqr/utils/elasticsearch/constants.py
```diff
@@ -172,6 +173,16 @@
...
+    ranks_by_gene = defaultdict(list)
+    for record in PhenotypePrioritization.objects.filter(individual__family=families[0]):
```
hanars, line 179: from the ticket: `we should only sort on the top ranked genes, probably the top 100`
So you need to add more filtering here

hanars, line 179: In general you really want to do as much database interaction in the database ORM as possible, not query all the data and post-process it. In the case, you can use the following to pull the data out already grouped:
```
family_ranks = PhenotypePrioritization.objects.filter(
  individual__family=families[0], rank__lte=100).values('gene_id').annotate(min_rank= Min('rank'))
return {agg['gene_id']: agg['min_rank] for agg in family_ranks}
```

hanars (changes requested):
The overall approach for the server side filter looks correct, but there are some small changes I would make

hanars:
> And I still need to add the number of family checking

It should be added in the backed, by raising a `InvalidSearchException` in the `_get_ranks_by_gene` function

seqr/utils/elasticsearch/constants.py
ShifaSZ: Okay

seqr/utils/elasticsearch/constants.py
```diff
@@ -172,6 +174,17 @@
     }
 }
 
+
+def _get_phenotype_priority_ranks_by_gene(families, *args):
+    from seqr.utils.elasticsearch.utils import InvalidSearchException
+    if len(families) > 1:
+        raise InvalidSearchException('Sorting on multiple families search results is not supported.')
```
hanars, line 181: We need a clearer error message: `Phenotype sort is only supported for single-family search`

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -888,7 +888,7 @@ def test_load_phenotype_prioritization_data(self, mock_logger, mock_subprocess):
         mock_logger.info.assert_called_with('create PhenotypePrioritizations', self.data_manager_user, db_update=db_update)
         saved_data = _get_json_for_models(PhenotypePrioritization.objects.filter(tool='lirical'),
                                           nested_fields=[{'fields': ('individual', 'guid'), 'key': 'individualGuid'}])
-        self.assertListEqual(saved_data, EXPECTED_LIRICAL_DATA)
+        self.assertListEqual(sorted(saved_data, key=lambda d: d['individualGuid']), EXPECTED_LIRICAL_DATA)
```
hanars, line 891: would be cleaner to add an `order_by` to the above query for the underlying `PhenotypePrioritization` used for `saved_data`

hanars (changes requested):

hanars (approved):

PR #3144, opened 2023-01-10
All project inheritance search performance

hanars:
There are plenty of inheritance-based many-project searches that time out that can easily complete with no inheritance filter applied, where they run as a single search across all indices. This change leverages that by first running a query without sample-specific filters across all indices, and then only runs the sample-specific index searches on indices/variants that were possible matches.

The prefilter query consistently runs in less than a second, and in the test examples I was given it reduces the number of indices to search from >200 to ~50


seqr/utils/elasticsearch/es_search.py
```diff
@@ -701,14 +704,48 @@ def _execute_multi_search(self, **kwargs):
...
                 ms = ms.add(search)
+            for message in index_logs[index_name]:
+                logger.info(message, self._user)
 
         responses = self._execute_search(ms) if ms._searches else []
         parsed_responses = [self._parse_response(response) for response in responses]
         return self._process_multi_search_responses(parsed_responses, **kwargs)
 
+    def _get_possible_hit_indices(self):
+        no_inheritance_search = self._search.index(self.index_name).source('')[:200]
```
ShifaSZ, line 737: Use a constant for the number `200`?

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #3145, opened 2023-01-10
Sample metadata 414

hanars:
Split airtable requests with many OR filters into multiple requests in order to prevent the query URL from being too long


ShifaSZ (approved):
Looks good to me.

PR #3149, opened 2023-01-10
Add new SV data format support.

ShifaSZ:
The display for the gnomAD AN and AC, new gene consequence:
image

The new search annotation for the structural consequences:
image

### DIFF
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -288,6 +288,10 @@ class EsSearch(object):
         clinvar_terms, hgmd_classes = _parse_pathogenicity_filter(pathogenicity or {})
 
         annotations = {k: v for k, v in (annotations or {}).items() if v}
+        # Temporal code for migrating SV data file formats, will be delete after the migration finished
+        if annotations.get('structural_consequence') and 'MSV_EXON_OVERLAP' in annotations['structural_consequence']:
+            annotations['structural_consequence'].append('MSV_EXON_OVR')
+        # End of the temporal code
         new_svs = bool(annotations.pop(NEW_SV_FIELD, False))
         splice_ai = annotations.pop(SPLICE_AI_FIELD, None)
         screen = annotations.pop(SCREEN_KEY, None)
@@ -811,6 +815,11 @@ class EsSearch(object):
             {_to_camel_case(k): v for k, v in transcript.to_dict().items()}
             for transcript in hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
         ]
+        # Temporal code for SV data file format migration
+        for trans in sorted_transcripts:
+            if trans.get('majorConsequence') and trans['majorConsequence'] == 'MSV_EXON_OVR':
+                trans['majorConsequence'] = 'MSV_EXON_OVERLAP'
+        # End of temporal code
         transcripts = defaultdict(list)
         for transcript in sorted_transcripts:
             transcripts[transcript['geneId']].append(transcript)
--- a/seqr/utils/elasticsearch/es_search.py
+++ b/seqr/utils/elasticsearch/es_search.py
@@ -288,10 +288,14 @@ class EsSearch(object):
         clinvar_terms, hgmd_classes = _parse_pathogenicity_filter(pathogenicity or {})
 
         annotations = {k: v for k, v in (annotations or {}).items() if v}
-        # Temporal code for migrating SV data file formats, will be delete after the migration finished
-        if annotations.get('structural_consequence') and 'MSV_EXON_OVERLAP' in annotations['structural_consequence']:
-            annotations['structural_consequence'].append('MSV_EXON_OVR')
-        # End of the temporal code
+
+        # Temporary code, remove it after all the SV indices are reloaded
+        if annotations.get('structural_consequence'):
+            if 'MSV_EXON_OVERLAP' in annotations['structural_consequence']:
+                annotations['structural_consequence'].append('MSV_EXON_OVR')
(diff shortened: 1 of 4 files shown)


seqr/utils/elasticsearch/es_search.py
```diff
@@ -288,6 +288,10 @@ def _parse_annotation_overrides(self, annotations, pathogenicity):
         clinvar_terms, hgmd_classes = _parse_pathogenicity_filter(pathogenicity or {})
 
         annotations = {k: v for k, v in (annotations or {}).items() if v}
+        # Temporal code for migrating SV data file formats, will be delete after the migration finished
+        if annotations.get('structural_consequence') and 'MSV_EXON_OVERLAP' in annotations['structural_consequence']:
+            annotations['structural_consequence'].append('MSV_EXON_OVR')
+        # End of the temporal code
```
hanars, line 294: theres no need for an end comment, just leave a blank line at the end to indicate what the temporary code block is

seqr/utils/elasticsearch/es_search.py
```diff
@@ -811,6 +815,11 @@ def _parse_hit(self, raw_hit):
             {_to_camel_case(k): v for k, v in transcript.to_dict().items()}
             for transcript in hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
         ]
+        # Temporal code for SV data file format migration
```
hanars, line 818: as above, should say "temporary" and use blank lines instead of end comments

ui/shared/utils/constants.js
```diff
@@ -654,6 +656,26 @@ const VEP_SV_CONSEQUENCES = [
...
+  },
+  {
+    description: 'An SV is predicted to duplicate the transcription start site',
+    text: 'TSS DUP',
+    value: 'TSS_DUP',
+  },
+  {
+    description: 'The duplication SV has one breakpoint in the coding sequence',
+    text: 'Partial Exon DUP',
+    value: 'PARTIAL_EXON_DUP',
```
hanars, line 677: Per the discussion on the ticket, both this field and `INTRAGENIC_EXON_DUP` are mapped to the old `DUP_LOF` field. I think we should replace `DUP_LOF`  in the UI with these new fields now, and then in the backend if either of these new fields are included n search we should also return results with the old `DUP_LOF` annotation, similar to what we did for `MSV_EXON_OVERLAP`

Also, it looks like we removed the `INTERGENIC` consequences in the pipeline, so we should remove the UI filter for those here

hanars: Also, for `INTRAGENIC_EXON_DUP` lets add a sentence to the description that says `previously called "Loss of function via Duplication"`

ui/shared/utils/constants.js
```diff
@@ -654,6 +656,26 @@ const VEP_SV_CONSEQUENCES = [
...
     value: 'PROMOTER',
   },
+  {
+    description: 'An SV which the SV breakend is predicted to fall in an exon',
+    text: 'Breakend Exonic',
+    value: 'BREAKEND_EXONIC',
+  },
+  {
+    description: 'An SV which is predicted to result in intragenic exonic duplication without breaking any coding sequences',
+    text: 'Intragenic Exon DUP',
```
hanars, line 666: Shouly say the full "Duplication", not DUP in the text for all of these for consistency with the other consequences

ui/shared/utils/constants.js
```diff
@@ -654,6 +656,26 @@ const VEP_SV_CONSEQUENCES = [
...
+    value: 'INTRAGENIC_EXON_DUP',
+  },
+  {
+    description: 'An SV is predicted to duplicate the transcription start site',
+    text: 'TSS DUP',
```
hanars, line 671: should say "Transcription Start Site Duplication"

hanars (changes requested):

ui/shared/utils/constants.js
```diff
@@ -609,11 +609,6 @@ const VEP_SV_CONSEQUENCES = [
     text: 'Loss of function',
     value: 'LOF',
   },
-  {
```
hanars, line 612: you should put `INTRAGENIC_EXON_DUP` over here

hanars (changes requested):

ui/shared/utils/constants.js
ShifaSZ: Okay

ui/shared/utils/constants.js
```diff
@@ -625,9 +620,11 @@ const VEP_SV_CONSEQUENCES = [
     value: 'DUP_PARTIAL',
   },
   {
-    description: 'A multiallelic SV predicted to have a Loss of function, Loss of function via Duplication, Copy Gain, or Duplication Partial effect',
+    description: 'A multiallelic SV would be predicted to have a Loss of function, Intragenic Exon Duplication, Copy Gain,' +
+        ' Duplication Partial, Duplication at the Transcription Start Site (TSS_DUP), or Duplication with a breakpoint' +
+        ' in the coding sequence annotation if the SV were biallelic',
```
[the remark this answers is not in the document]
ShifaSZ: It's from the example VCF file they provided us.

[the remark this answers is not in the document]
ShifaSZ: The original text is like this `##INFO=<ID=PREDICTED_MSV_EXON_OVERLAP,Number=.,Type=String,Description="Gene(s) on which the multiallelic SV would be predicted to have a LOF, INTRAGENIC_EXON_DUP, COPY_GAIN, DUP_PARTIAL, TSS_DUP, or PARTIAL_EXON_DUP annotation if the SV were biallelic.">`

hanars (approved):

PR #3157, opened 2023-01-13
Family saved variant fix

hanars:
Fixes the reported bug and a new bug from my lasrt PR that returns duplicate family guids for saved variants


ShifaSZ (approved):
Looks good to me.

PR #3164, opened 2023-01-17
Family structure notation

hanars:


ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -158,10 +176,28 @@ const FamiliesIndividuals = React.memo(({ canEdit, hasCaseReview, familyCounts,
...
+          <br />
+          {`${individualsCount} Individuals${title || ''}`}
+        </span>
+      )}
       content={
-        sortBy(Object.keys(familySizeHistogram)).map(size => (
-          <div key={size}>{`${familySizeHistogram[size]} ${FAMILY_SIZE_LABELS[size](familySizeHistogram[size] > 1)}`}</div>
+        sortBy(Object.entries(familySizeHistogram)).map(([size, { total, withParents }]) => (
+          <div key={size}>
+            {`${total} famil${total === 1 ? 'y' : 'ies'} with ${FAMILY_SIZE_LABELS[size] || size} individual${size === '1' ? '' : 's'}`}
```
ShifaSZ, line 189: You may not want to display this line when the `size` is bigger than 2.

ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -41,12 +41,22 @@ const DetailContent = styled.div`
...
+  0: 'no',
+  5: '5+',
+}
+
+const FAMILY_STRUCTURE_SIZE_LABELS = {
+  2: plural => ` duo${plural ? 's' : ''}`,
   3: plural => ` trio${plural ? 's' : ''}`,
   4: plural => ` quad${plural ? 's' : ''}`,
-  5: plural => ` ${plural ? 'families' : 'family'} with 5+ individuals`,
+  5: plural => ` trio${plural ? 's' : ''}+`,
```
ShifaSZ, line 52: Is a `quad` also a `trio+`?

ui/pages/Project/selectors.js
```diff
@@ -105,16 +105,24 @@ export const getProjectAnalysisGroupFamiliesByGuid = createSelector(
 
 export const getProjectAnalysisGroupFamilyIndividualCounts = createSelector(
   getProjectAnalysisGroupFamiliesByGuid,
-  familiesByGuid => Object.values(familiesByGuid).map(family => (family.individualGuids || []).length),
+  familiesByGuid => Object.values(familiesByGuid).map(family => ({
+    size: (family.individualGuids || []).length,
+    hasMom: (family.maternalGuids || []).length === 1,
+    hasDad: (family.paternalGuids || []).length === 1,
```
ShifaSZ, line 111: Is it possible for a family to have 3 generations and one grandma and one dad? If yes, `hasMon` and `hasDad` may not work for deciding the family structure.

ShifaSZ (changes requested):

ui/pages/Project/components/ProjectOverview.jsx
hanars: We do want it

ui/pages/Project/components/ProjectOverview.jsx
hanars: no, its a special case

ui/pages/Project/selectors.js
hanars: Great catch. I will update the logic, I'll need to refactor what we are actually sending back from the server in order to make this work

hanars: Okay, previously the case you outlined here would have returned `{maternalGuids: ['grandma'], paternalGuids: ['dadd]}`. I have updated the server response to keep the association between parents so this will now come back as `{parents: [{maternalGuid: 'grandma', paternalGuid: null}, {maternalGuid: null, paternalGuid: 'dad'}]}` and then I updated the client side code to not to consider parents if there is more than one parent object returned as that means its a complex structure

ShifaSZ (approved):

Issue #3112, opened 2022-12-08
Add frequency filter for LIRICAL results


lynnpais:
**Is your feature request related to a problem? Please describe.**
Too many common variants are being returned in a search using LIRICAL's  Search for Variants in High-Ranked Genes feature. 

**Describe the solution you'd like**
Limit the variants in the search to <1% AF in gnomAD

**Describe alternatives you've considered**
Reviewed with analyst team.

**Additional context**
NA

hanars:
We should use the default `GeneSearchLink` to set behavior, which would apply a <1% AF and an "any affected" inheritance filter

ShifaSZ:
> We should use the default `GeneSearchLink` to set behavior, which would apply a <1% AF and an "any affected" inheritance filter

I can see that the `GeneSearchLink` applies the search settings of the below constants.
```
const INITIAL_GENE_SEARCH = {
  inheritance: { mode: ANY_AFFECTED },
  freqs: FREQUENCIES.filter(({ name }) => name !== THIS_CALLSET_FREQUENCY && name !== SV_CALLSET_FREQUENCY).reduce(
    (acc, { name }) => ({ ...acc, [name]: { af: 0.1 } }), {},
  ),
}
```
My question is when to apply the search settings for `LIRICAL's Search for Variants in High-Ranked Genes`. In the project variant search, since we don't know whether there are some LIRICAL High-ranked genes, we can't apply the settings by default. On the variant page and family page, both the "Gene Search" and the "Search for variants in high-ranked genes" have applied the <0.1 AF setting. Is there anything more that we need to do for this ticket?

hanars:
This ticket is about modifying the search link in the Lirical modal on the family page

ShifaSZ:
Since the search link in the Lirical modal on the family page uses the `GeneSearchLink` component, the initial gnomAD AF is set to 0.1. Can we simply change gnomAD AF from 0.1 to 0.01  in the initial settings of `GeneSearchLink`? Or should we pass in the initial gnomAD AF setting to the component?

hanars:
we should use whatever the default `GeneSearchLink` behavior is for the lirical link with no modifications, as I said above

ShifaSZ:
So I'm going to change the initial AF of `GeneSearchLink` to 0.01. But I don't know if there is any negative impact on the other use cases of the component.

hanars:
[...]

PR #3169, opened 2023-01-18
Add LIRICAL frequency filter

ShifaSZ:
See ticket #3112.

### DIFF
--- a/ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
+++ b/ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
@@ -64,6 +64,7 @@ const BasePhenotypePriGenes = React.memo((
       icon="search"
       location={(phenotypeGeneScores[individualGuid] || []).map(({ geneId }) => geneId).join(',')}
       familyGuid={familyGuid}
+      af={0.01}
       floated="right"
     />
     <DataTable
--- a/ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
+++ b/ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
 import DataLoader from 'shared/components/DataLoader'
-import { GeneSearchLink } from 'shared/components/buttons/SearchResultsLink'
+import { RareGeneSearchLink } from 'shared/components/buttons/SearchResultsLink'
 import DataTable from 'shared/components/table/DataTable'
 import { BaseVariantGene } from 'shared/components/panel/variants/VariantGene'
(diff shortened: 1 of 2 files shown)


ShifaSZ:
I was thinking about adding an `af` prop to the `SearchResultsLink` component, as discussed in the requirement ticket. But I later discovered that there are cases with no `initialSearch`. So we can't generate `initialSearch` by default.

ui/shared/components/buttons/SearchResultsLink.jsx
```diff
@@ -48,11 +48,17 @@ const ConnectedSearchResultsLink = connect(null, mapDispatchToProps)(SearchResul
 
 export default ConnectedSearchResultsLink
 
-const INITIAL_GENE_SEARCH = {
+const getGeneSearchProps = af => ({
```
hanars, line 51: What I recommended on the ticket had a very important purpose: pre-generating the dictionary so it will not be re-generated, resulting in unnecessary re-renders of the component. With this change, anytime any prop to the link changes this whole dictionary will be recomputed, and the link will constantly re-render

This also adds an unnecessary and dangerous amount of flexibility. We don't want every gene search link in seqr to be slightly different, we want consistant behavior. In this specific case we need it to be different, but we want that to be a very explicit different case. We don't want the possibility of every possible af in gene search, we want exactly 2 possibilities

If you prefer adding a prop instead of a new component, you can add a boolean `requireRareAf` and then stil precompute the 2 possible initial search constants and toggle between them
```
initialSearch={requireRareAf ? RARE_INITIAL_GENE_SEARCH : INITIAL_GENE_SEARCH}
```

hanars (changes requested):

ui/shared/components/buttons/SearchResultsLink.jsx
ShifaSZ: Okay, I see. A new rare gene search link component may be more straightforward.

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -59,11 +59,12 @@ const BasePhenotypePriGenes = React.memo((
   { individualGuid, phenotypeGeneScores, familyGuid, loading, load },
 ) => (
   <DataLoader content={phenotypeGeneScores[individualGuid]} contentId={individualGuid} load={load} loading={loading}>
-    <GeneSearchLink
+    <RareGeneSearchLink
       buttonText="Search for variants in high-ranked genes"
       icon="search"
       location={(phenotypeGeneScores[individualGuid] || []).map(({ geneId }) => geneId).join(',')}
       familyGuid={familyGuid}
+      af={0.01}
```
hanars, line 67: this is no longer needed

hanars (changes requested):

PR #3180, opened 2023-01-23
Gregor report - airtable data

hanars:


seqr/views/apis/report_api.py
```diff
@@ -558,53 +558,57 @@ def _get_discovery_rows(sample, parsed_variants, male_individual_guids):
...
 SINGLE_SAMPLE_FIELDS = ['Collaborator', 'dbgap_study_id', 'dbgap_subject_id', 'dbgap_sample_id']
 LIST_SAMPLE_FIELDS = ['SequencingProduct', 'dbgap_submission']
 
 
-def _get_sample_airtable_metadata(sample_ids, user, include_collaborator=False):
+def _get_airtable_samples(sample_ids, user, fields, list_fields=None):
+    list_fields = list_fields or []
     session = AirtableSession(user)
-    raw_records = {}
-    # Airtable does handle its own pagination, but the query URI has a max length so the filter formula needs to be truncated
```
hanars, line 570: this logic was moved directly into `AirtableSession` in a previous PR

seqr/views/apis/report_api.py
```diff
@@ -743,10 +754,16 @@ def gregor_export(request, consent_code):
...
+        # airtable data
+        if airtable_sample:
+            analyte_id = f'Broad_{airtable_sample[SMID_FIELD]}'
+            airtable_metadata = airtable_sample.get('metadata')
+            if airtable_metadata:
+                airtable_rows.append(dict(analyte_id=analyte_id, **airtable_metadata, **_get_experiment_ids(airtable_sample)))
 
-    airtable_rows = []  # TODO populate airtable data once new columns are confirmed
+        # analyte table
+        analyte_rows.append(dict(participant_id=participant_id, analyte_id=analyte_id, **_get_analyte_row(individual)))
```
ShifaSZ, line 766: It may be better to replace the field names like `participant_id` and `analyte_id` with constants. But it could be overkilled because of so many of them.

seqr/views/apis/report_api_tests.py
```diff
@@ -488,15 +539,28 @@ def test_sample_metadata_export(self, mock_google_authenticated):
         self.assertEqual(response.status_code, 403)
         self.assertEqual(response.json()['error'], 'Permission Denied')
 
+    @mock.patch('seqr.views.utils.airtable_utils.is_google_authenticated')
     @mock.patch('seqr.views.apis.report_api.datetime')
     @mock.patch('seqr.views.utils.export_utils.zipfile.ZipFile')
     @responses.activate
-    def test_gregor_export(self, mock_zip, mock_datetime):
+    def test_gregor_export(self, mock_zip, mock_datetime, mock_google_authenticated):
```
ShifaSZ, line 546: Don't you think the `test_gregor_export` also needs to mock and test `MAX_OR_FILTERS` like `test_sample_metadata_export`?

ShifaSZ (changes requested):

seqr/views/apis/report_api.py
```diff
@@ -782,7 +819,7 @@ def _get_participant_row(individual):
         'reported_race': GREGOR_ANCESTRY_MAP.get(individual.population, 'Unknown'),
         'ancestry_detail': GREGOR_ANCESTRY_DETAIL_MAP.get(individual.population),
         'reported_ethnicity': ANCESTRY_MAP[HISPANIC] if individual.population == HISPANIC else 'Unknown',
-        'recontactable': None,  # TODO populate airtable data once new columns are confirmed
+        'recontactable': airtable_sample.get('Recontactable') or 'Unknown',
```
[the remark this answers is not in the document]
hanars: no because if the value that comes back from airtable is `{'Recontactable': ''}` then your change would have us report it as blank when the way it is currently written has it report as 'Unknown'

seqr/views/apis/report_api.py
```diff
@@ -760,6 +777,26 @@ def gregor_export(request, consent_code):
     ], f'GREGoR Reports {consent_code}', file_format='tsv')
 
 
+def _get_gregor_airtable_data(individuals, user):
+    sample_records, session = _get_airtable_samples(
+        individuals.order_by('individual_id').values_list('individual_id', flat=True), user,
+        fields=[SMID_FIELD, 'CollaboratorSampleID', 'Recontactable'],
```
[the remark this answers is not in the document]
hanars: not really, its no more of a constant here than Recontactable

seqr/views/apis/report_api.py
hanars: its overkill

seqr/views/apis/report_api_tests.py
hanars: no you don't need to tests every possible path of  base utility functions every single time they are called. We have full coverage of that utility function, I think its actually better to test the case where there are fewer than the max samples

ShifaSZ (approved):

PR #3183, opened 2023-01-24
Support Sharded VCF Files

ShifaSZ:
Make the work visible before the final PR.

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -33,7 +33,7 @@ def get_google_project(gs_path):
     return 'anvil-datastorage' if gs_path.startswith('gs://fc-secure') else None
 
 
-def does_file_exist(file_path, user=None, return_outputs=False):
+def does_file_exist(file_path, user=None):
     if is_google_bucket_file_path(file_path):
         process = _run_gsutil_command('ls', file_path, user=user)
         success = process.wait() == 0
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -83,7 +83,7 @@ def mv_file_to_gs(local_path, gs_path, user=None):
 
 def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     gs_path = gs_path.rstrip('/')
-    command = '-q ls'
+    command = 'ls'
 
     # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
     subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
@@ -93,7 +93,9 @@ def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     if check_subfolders:
         gs_path = f'{gs_path}/**'
 
-    return _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
+    all_lines = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
+    path_prefix = gs_path.rsplit('/', 1)[0]
+    return [line for line in all_lines if line.startswith(path_prefix)]
 
 
 def _run_gsutil_with_wait(command, gs_path, user=None, get_stdout=False):
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -94,8 +94,7 @@ def get_gs_file_list(gs_path, user=None, check_subfolders=True):
         gs_path = f'{gs_path}/**'
 
     all_lines = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
-    path_prefix = gs_path.rsplit('/', 1)[0]
-    return [line for line in all_lines if line.startswith(path_prefix)]
+    return [line for line in all_lines if is_google_bucket_file_path(line)]
 
 
 def _run_gsutil_with_wait(command, gs_path, user=None, get_stdout=False):
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -85,12 +85,11 @@ def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     gs_path = gs_path.rstrip('/')
     command = 'ls'
 
-    # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
-    subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
-    if not subfolders:
-        return []
-
     if check_subfolders:
+        # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
+        subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
+        if not subfolders:
+            return []
         gs_path = f'{gs_path}/**'
 
     all_lines = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -37,11 +37,10 @@ def does_file_exist(file_path, user=None, return_outputs=False):
     if is_google_bucket_file_path(file_path):
         process = _run_gsutil_command('ls', file_path, user=user)
         success = process.wait() == 0
-        outputs = [line.decode('utf-8').strip() for line in process.stdout]
         if not success:
-            logger.info(' '.join(outputs), user)
-            return False
-        return outputs if return_outputs else True
+            errors = [line.decode('utf-8').strip() for line in process.stdout]
+            logger.info(' '.join(errors), user)
+        return success
     return os.path.isfile(file_path)
 
 
@@ -82,17 +81,14 @@ def mv_file_to_gs(local_path, gs_path, user=None):
     _run_gsutil_with_wait(command, gs_path, user)
 
 
-def get_gs_file_list(gs_path, user=None):
+def get_gs_file_list(gs_path, user=None, recursive=True):
     gs_path = gs_path.rstrip('/')
     command = 'ls'
 
-    # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
-    subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
-    if not subfolders:
-        return []
+    if recursive:
+        gs_path = f'{gs_path}/**'
 
-    all_lines = _run_gsutil_with_wait(command, f'{gs_path}/**', user, get_stdout=True)
-    return [line for line in all_lines if line.startswith(gs_path)]
+    return _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
 
 
 def _run_gsutil_with_wait(command, gs_path, user=None, get_stdout=False):
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -33,7 +33,7 @@ def get_google_project(gs_path):
     return 'anvil-datastorage' if gs_path.startswith('gs://fc-secure') else None
 
 
-def does_file_exist(file_path, user=None):
+def does_file_exist(file_path, user=None, return_outputs=False):
     if is_google_bucket_file_path(file_path):
         process = _run_gsutil_command('ls', file_path, user=user)
         success = process.wait() == 0
@@ -41,7 +41,7 @@ def does_file_exist(file_path, user=None):
         if not success:
             logger.info(' '.join(outputs), user)
             return False
-        return outputs
+        return outputs if return_outputs else True
     return os.path.isfile(file_path)
 
 
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -37,10 +37,11 @@ def does_file_exist(file_path, user=None):
     if is_google_bucket_file_path(file_path):
         process = _run_gsutil_command('ls', file_path, user=user)
         success = process.wait() == 0
+        outputs = [line.decode('utf-8').strip() for line in process.stdout]
         if not success:
-            errors = [line.decode('utf-8').strip() for line in process.stdout]
-            logger.info(' '.join(errors), user)
-        return success
+            logger.info(' '.join(outputs), user)
+            return False
+        return outputs
     return os.path.isfile(file_path)
 
 
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -81,11 +81,11 @@ def mv_file_to_gs(local_path, gs_path, user=None):
     _run_gsutil_with_wait(command, gs_path, user)
 
 
-def get_gs_file_list(gs_path, user=None, recursive=True):
+def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     gs_path = gs_path.rstrip('/')
-    command = 'ls'
+    command = '-q ls'
 
-    if recursive:
+    if check_subfolders:
         gs_path = f'{gs_path}/**'
 
     return _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -85,6 +85,11 @@ def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     gs_path = gs_path.rstrip('/')
     command = '-q ls'
 
+    # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
+    subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
+    if not subfolders:
+        return []
+
     if check_subfolders:
         gs_path = f'{gs_path}/**'
 
--- a/seqr/utils/file_utils_tests.py
+++ b/seqr/utils/file_utils_tests.py
@@ -1,7 +1,7 @@
 import mock
 
 from unittest import TestCase
-from seqr.utils.file_utils import mv_file_to_gs, get_gs_file_list
+from seqr.utils.file_utils import mv_file_to_gs
 
 
 class FileUtilsTest(TestCase):
@@ -30,34 +30,3 @@ class FileUtilsTest(TestCase):
         mock_subproc.Popen.assert_called_with('gsutil mv /temp_path gs://bucket/target_path', stdout=mock_subproc.PIPE, stderr=mock_subproc.STDOUT, shell=True)
         mock_logger.info.assert_called_with('==> gsutil mv /temp_path gs://bucket/target_path', None)
         process.wait.assert_called_with()
-
-    @mock.patch('seqr.utils.file_utils.subprocess')
-    @mock.patch('seqr.utils.file_utils.logger')
-    def test_get_gs_file_list(self, mock_logger, mock_subproc):
-        with self.assertRaises(Exception) as ee:
-            get_gs_file_list('/temp_path')
-        self.assertEqual(str(ee.exception),  'A Google Storage path is expected.')
-
-        process = mock_subproc.Popen.return_value
-        process.wait.return_value = -1
-        process.stdout = [b'-bash: gsutil: command not found.\n', b'Please check the path.\n']
-        with self.assertRaises(Exception) as ee:
-            get_gs_file_list('gs://bucket/target_path/', user=None)
-        self.assertEqual(str(ee.exception), 'Run command failed: -bash: gsutil: command not found. Please check the path.')
-        mock_subproc.Popen.assert_called_with('gsutil ls gs://bucket/target_path', stdout=mock_subproc.PIPE,
-                                              stderr=mock_subproc.STDOUT, shell=True)
-        mock_logger.info.assert_called_with('==> gsutil ls gs://bucket/target_path', None)
-        process.wait.assert_called_with()
-
-        mock_subproc.reset_mock()
-        mock_logger.reset_mock()
-        process.wait.return_value = 0
-        process.stdout = [b'\n', b'\n', b'Updates are available for some Cloud SDK components.  To install them,\n',
-                          b'please run:\n', b'  $ gcloud components update\n',
-                          b'gs://bucket/target_path/id_file.txt\n', b'gs://bucket/target_path/data.vcf.gz\n']
-        file_list = get_gs_file_list('gs://bucket/target_path', user=None)
-        mock_subproc.Popen.assert_called_with('gsutil ls gs://bucket/target_path/**', stdout=mock_subproc.PIPE,
-                                              stderr=mock_subproc.STDOUT, shell=True)
-        mock_logger.info.assert_called_with('==> gsutil ls gs://bucket/target_path/**', None)
-        process.wait.assert_called_with()
-        self.assertEqual(file_list, ['gs://bucket/target_path/id_file.txt', 'gs://bucket/target_path/data.vcf.gz'])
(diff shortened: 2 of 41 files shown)


seqr/utils/file_utils.py
```diff
@@ -37,10 +37,11 @@ def does_file_exist(file_path, user=None):
...
         process = _run_gsutil_command('ls', file_path, user=user)
         success = process.wait() == 0
+        outputs = [line.decode('utf-8').strip() for line in process.stdout]
         if not success:
-            errors = [line.decode('utf-8').strip() for line in process.stdout]
-            logger.info(' '.join(errors), user)
-        return success
+            logger.info(' '.join(outputs), user)
+            return False
+        return outputs
```
ShifaSZ, line 44: Change to return the file name or the file name list.
The purpose of the change is to fetch the file list for a path with `*` in it. To get the file list, there are a couple of ways: 1) we could store the file list when merging the file names; 2) we could re-fetch the file list with the `get_gs_file_list` function after the user selection; 3) re-fetch the file list with the `does_file_exist` function. #1 could be complicated because of the need to manage some extra data. For #2, using  `get_gs_file_list` to re-fetch the file list is straightforward. But the function was designed to get a file list under a folder, and no wildcard (*) in the path is supported. The change for #3 is small but may not be straightforward for the function itself.

seqr/utils/vcf_utils.py
```diff
@@ -53,7 +53,10 @@ def _get_vcf_meta_info(line):
     return None
 
 
-def validate_vcf_and_get_samples(vcf_filename):
+def validate_vcf_and_get_samples(vcf_filename, sharded_file_list=None):
+    if sharded_file_list:
+        # simply validate the first sharded VCF file, more validation can be added if needed.
+        vcf_filename = sharded_file_list[0]
```
ShifaSZ, line 59: Maybe we should validate all the sharded files have the same header.

seqr/utils/file_utils.py
```diff
@@ -81,17 +81,14 @@ def mv_file_to_gs(local_path, gs_path, user=None):
...
 
-def get_gs_file_list(gs_path, user=None):
+def get_gs_file_list(gs_path, user=None, recursive=True):
     gs_path = gs_path.rstrip('/')
     command = 'ls'
 
-    # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
-    subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
-    if not subfolders:
-        return []
```
ShifaSZ, line 91: This is unnecessary after the error handling of `_run_gsutil_with_wait` was improved.

seqr/utils/file_utils_tests.py
```diff
@@ -30,34 +30,3 @@ def test_mv_file_to_gs(self, mock_logger, mock_subproc):
         mock_subproc.Popen.assert_called_with('gsutil mv /temp_path gs://bucket/target_path', stdout=mock_subproc.PIPE, stderr=mock_subproc.STDOUT, shell=True)
         mock_logger.info.assert_called_with('==> gsutil mv /temp_path gs://bucket/target_path', None)
         process.wait.assert_called_with()
-
-    @mock.patch('seqr.utils.file_utils.subprocess')
-    @mock.patch('seqr.utils.file_utils.logger')
-    def test_get_gs_file_list(self, mock_logger, mock_subproc):
```
ShifaSZ, line 36: Remove the utility function test since it doesn't increase coverage.

seqr/utils/vcf_utils.py
hanars: per the discussion on the ticket, no. We will validate one header

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -449,4 +458,25 @@ def _make_airflow_api_request(endpoint, method='GET', timeout=90, **kwargs):
...
+    files_by_path = defaultdict(list)
+
+    for vcf_file in vcf_files:
+        subfolder_path, file = vcf_file.rsplit('/', 1)
+        files_by_path[subfolder_path].append(file)
+
+    # discover the sharded VCF files in each folder, replace the sharded VCF files with a single path with '*'
+    for subfolder_path, files in files_by_path.items():
+        prefix = _get_common_prefix(files)
+        postfix = files[0].replace(prefix, '').lstrip('_0123456789')
```
hanars, line 478: the opposite of prefix is suffix, not postfix. Also, this works but I think its a bit more explicit using regex:
```
if len(files) < 2:
    continue
prefix = os.path.commonprefix(files)
suffix = re.fullmatch(f'{prefix}\d+(?P<suffix>\D+)', files[0]).groupdict()['suffix']
if all(re.fullmatch(f'{prefix}\d+{suffix}', file) for file in files]):
    ...
```

hanars (changes requested):

seqr/utils/vcf_utils.py
ShifaSZ: Okay

seqr/utils/file_utils.py
ShifaSZ: The `_run_gsutil_with_wait` function raises an exception for the `process.wait() != 0`. Your comment reminds me to add exception capturing in the API. The concern will be that there may still be warning messages when the `process.wait() == 0`. Can we add a `-q` option to the `gsutil` command to silence any warnings?

seqr/utils/file_utils_tests.py
ShifaSZ: I'm updating the test cases for the `get_anvil_vcf_list` and `validate_anvil_vcf` APIs. I can mock the subprocess.wait.return_value to 1 and test the error case.

seqr/utils/file_utils.py
hanars: As the comment that you deleted here says `If a bucket is empty gsutil throws an error when running ls with **`. This has nothing to do with our error handling, and it has nothing to do with silencing warnings. Unless the gsutil behavior has changed and no longer throws an error in that case, we need to keep this code as is

ShifaSZ: The behavior doesn't change. But we should handle the case as an error because an empty bucket provides no data for analysis. So I added exception capturing in the API code to handle the error.

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -125,8 +128,14 @@ def grant_workspace_access(request, namespace, name):
...
     bucket_name = workspace_meta['workspace']['bucketName']
     bucket_path = 'gs://{bucket}'.format(bucket=bucket_name.rstrip('/'))
-    data_path_list = [path.replace(bucket_path, '') for path in get_gs_file_list(bucket_path, request.user)
-                      if path.endswith(VCF_FILE_EXTENSIONS)]
+    try:
+        data_path_list = [path.replace(bucket_path, '') for path in get_gs_file_list(bucket_path, request.user)
+                          if path.endswith(VCF_FILE_EXTENSIONS)]
+    except Exception as ee:
+        error = f'Loading VCF file list from {bucket_path} failed with error {str(ee)}'
+        logger.error(error, request.user)
```
ShifaSZ, line 136: It creates an error log because it is usually caused by a system error rather than a user operation error. But the exception case is to analyze a project with an empty bucket. Should we create a log differently for the empty bucket?

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -141,15 +150,25 @@ def validate_anvil_vcf(request, namespace, name, workspace_meta):
...
         error = 'Invalid VCF file format - file path must end with {}'.format(' or '.join(VCF_FILE_EXTENSIONS))
-        return create_json_response({'error': error}, status=400, reason=error)
-    if not does_file_exist(data_path, user=request.user):
+    elif '*' in data_path:
+        try:
+            sharded_file_list = get_gs_file_list(data_path, request.user, check_subfolders=False)
+            file_to_check = sharded_file_list[0]
+        except Exception as ee:
+            error = f'Loading VCF file list for {data_path} failed with error {str(ee)}'
+            logger.error(error, request.user)
```
ShifaSZ, line 163: It creates an error log because it can happen only when the frontend is hacked, or there are bugs in the code.

seqr/utils/file_utils.py
hanars: I really disagree with making this change. I do not understand why its necessary, it was added for a very specific reason and as far as I can tell there is no reason to believe its no longer needed. Please explain why you think we need this change because from my perspective this is a risky change that has no relationship to the functionality change this PR is supposed to be working on

ShifaSZ: We have to handle the exception of `process.wait() != 0` in the `validate_anvil_vcf` and `get_anvil_vcf_list` APIs. We should have added the exception handling when we added raising an exception for `process.wait() != 0`. Once we handle the exceptions in the APIs, we don't have to run an extra `gsutil` command to avoid exceptions.

hanars: > we don't have to run an extra gsutil command to avoid exceptions.

This is out of scope of this PR. This is an optimization that is risky that is having a cascading effect on a lot of other change and is making it really difficult to figure out what is changing because we actually need the change to support sharded VCFs, and what is changing because of this optimization. If you feel really strongly that we need this optimization, that needs to go in a separate PR. However, I do not think that is a good use of your time. The performance of this endpoint is adequate so improving the performance is not high priority

seqr/views/apis/anvil_workspace_api.py
hanars: Error handling should be happening in `get_gs_file_list` the way it was previously. I do not understand why this needs to change to support sharded VCFs

ShifaSZ: I agree that the error handling is not just for supporting sharded VCFs. But the sharded VCFs feature should have good error handling for files not found or insufficient permission for accessing the cloud bucket.

hanars: The sharded VCF should have identical error handling to regular VCFs, and we already have error handling that has been live for months and  is fine. Unless you can demonstrate a specific case that is actually different for a sharded VCF than a regular VCF, this PR should have no changes to error handling

ShifaSZ: We could keep the error handling as is for fetching the VCF file list from the Anvil. But we should add exception capturing for fetching sharded file list, which is unique for the sharded VCFs feature.

seqr/utils/file_utils.py
```diff
@@ -81,17 +81,19 @@ def mv_file_to_gs(local_path, gs_path, user=None):
     _run_gsutil_with_wait(command, gs_path, user)
 
 
-def get_gs_file_list(gs_path, user=None):
+def get_gs_file_list(gs_path, user=None, check_subfolders=True):
     gs_path = gs_path.rstrip('/')
-    command = 'ls'
+    command = '-q ls'
```
hanars, line 86: why is this changing?

seqr/utils/file_utils.py
```diff
@@ -81,17 +81,19 @@ def mv_file_to_gs(local_path, gs_path, user=None):
...
 
     # If a bucket is empty gsutil throws an error when running ls with ** instead of returning an empty list
     subfolders = _run_gsutil_with_wait(command, gs_path, user, get_stdout=True)
     if not subfolders:
         return []
 
-    all_lines = _run_gsutil_with_wait(command, f'{gs_path}/**', user, get_stdout=True)
-    return [line for line in all_lines if line.startswith(gs_path)]
```
hanars, line 94: the `if line.startswith(gs_path)` is necessary to omit any warnings that are logged to the console

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -141,15 +145,25 @@ def validate_anvil_vcf(request, namespace, name, workspace_meta):
...
+    file_to_check = data_path
     if not data_path.endswith(VCF_FILE_EXTENSIONS):
         error = 'Invalid VCF file format - file path must end with {}'.format(' or '.join(VCF_FILE_EXTENSIONS))
-        return create_json_response({'error': error}, status=400, reason=error)
-    if not does_file_exist(data_path, user=request.user):
+    elif '*' in data_path:
+        try:
+            sharded_file_list = get_gs_file_list(data_path, request.user, check_subfolders=False)
+            file_to_check = sharded_file_list[0]
+        except Exception as ee:
```
hanars, line 156: In general, catching broad Exceptions like this is bad form. Is there a specific error we expect to happen in the course of normal application usage? If so, we should ad explicit handling for that case. But if this can only happen in the event of something being really broken, we want to use our normal built-in handling for unhandled exceptions

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -449,4 +463,27 @@ def _make_airflow_api_request(endpoint, method='GET', timeout=90, **kwargs):
     return resp.json()
 
 
+def _get_common_prefix(files):
```
hanars, line 466: this is no longer used

seqr/utils/file_utils_tests.py
```diff
@@ -30,34 +30,3 @@ def test_mv_file_to_gs(self, mock_logger, mock_subproc):
...
-        with self.assertRaises(Exception) as ee:
-            get_gs_file_list('/temp_path')
-        self.assertEqual(str(ee.exception),  'A Google Storage path is expected.')
```
hanars, line 39: this case is no longer tested

seqr/utils/file_utils_tests.py
```diff
@@ -30,34 +30,3 @@ def test_mv_file_to_gs(self, mock_logger, mock_subproc):
...
-        self.assertEqual(str(ee.exception), 'Run command failed: -bash: gsutil: command not found. Please check the path.')
-        mock_subproc.Popen.assert_called_with('gsutil ls gs://bucket/target_path', stdout=mock_subproc.PIPE,
-                                              stderr=mock_subproc.STDOUT, shell=True)
-        mock_logger.info.assert_called_with('==> gsutil ls gs://bucket/target_path', None)
-        process.wait.assert_called_with()
-
-        mock_subproc.reset_mock()
-        mock_logger.reset_mock()
-        process.wait.return_value = 0
-        process.stdout = [b'\n', b'\n', b'Updates are available for some Cloud SDK components.  To install them,\n',
```
hanars, line 55: the case where gsutil is printing warnings is no longer tested

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -375,44 +378,72 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+            mock.call('==> gsutil ls gs://test_bucket/test_no_slash_path.vcf.bgz', self.manager_user),
+            mock.call('File not found', self.manager_user),
+        ])
 
         response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_BAD_DATA_PATH))
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.json()['error'],
                          'Invalid VCF file format - file path must end with .vcf or .vcf.gz or .vcf.bgz')
 
+        # Test bad sharded VCF file path
```
hanars, line 397: does this actually test a different code path than the above test? It seems like both are just testing the case when the call to `ls` fails

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -375,44 +378,72 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+        mock_file_logger.info.assert_has_calls([
+            mock.call('==> gsutil ls gs://test_bucket/test_path.vcf.gz', self.manager_user),
+            mock.call('==> gsutil cat -r 0-65536 gs://test_bucket/test_path.vcf.gz | gunzip -c -q - ', None),
         ])
-        mock_file_iter.assert_called_with('gs://test_bucket/test_path.vcf.gz', byte_range=(0, 65536))
-        mock_file_exist.assert_called_with('gs://test_bucket/test_path.vcf.gz', user=self.manager_user)
 
         # test header errors
-        mock_file_iter.return_value = BASIC_META + BAD_INFO_META + BAD_FORMAT_META + BAD_HEADER_LINE + DATA_LINES
+        mock_subprocess.return_value.stdout = [s.encode('utf-8') for s in
```
hanars, line 429: wouldn't it be better to update the fixture data to be encoded, instead of calling encode every time?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -451,16 +513,16 @@ def test_get_anvil_vcf_list(self, mock_subprocess, mock_file_logger, mock_utils_
 
         # Test empty bucket
         mock_subprocess.return_value.wait.return_value = 0
-        mock_subprocess.return_value.stdout = b''
+        mock_subprocess.return_value.stdout = []
         response = self.client.get(url, content_type='application/json')
         self.assertEqual(response.status_code, 200)
         self.assertDictEqual(response.json(), {'dataPathList': []})
-        mock_subprocess.assert_has_calls([
```
hanars, line 458: why did this change?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -469,16 +531,25 @@ def test_get_anvil_vcf_list(self, mock_subprocess, mock_file_logger, mock_utils_
...
         ])
         mock_file_logger.info.assert_has_calls([
-            mock.call('==> gsutil ls gs://test_bucket', self.manager_user),
-            mock.call('==> gsutil ls gs://test_bucket/**', self.manager_user),
+            mock.call('==> gsutil -q ls gs://test_bucket', self.manager_user),
+            mock.call('==> gsutil -q ls gs://test_bucket/**', self.manager_user),
         ])
 
+        # Test merging sharded VCF files
+        mock_subprocess.return_value.stdout = [
```
hanars, line 545: why is this a separate test case, why not just modify the above test to have both sharded and non-sharded VCFs?

hanars (changes requested):

seqr/utils/file_utils.py
ShifaSZ: Adding the `-q` option to the `gsutil` command disables reporting progress and other information.

seqr/utils/file_utils.py
ShifaSZ: Hm... you are right.

seqr/views/apis/anvil_workspace_api.py
ShifaSZ: The errors can be files not found or no permissions. The errors are not expected in normal application usage because the previous step generates the data path. The errors can only happen when the front end accidentally changes the data path by bugs or malicious operations.

hanars: then we should not add any extra error handling. Although I do think we should run `does_file_exist` whether or not it is sharded, and do the sharded list afterwards

seqr/utils/file_utils.py
ShifaSZ: But the filter doesn't work to get the sharded VCF files because no files have `*` in their names. We could filter by either starting with `gs://` or ending with VCF extensions. Maybe I should change the function name from `get_gs_file_list` to `get_vcf_file_list` and add filtering by the VCF extensions.

seqr/utils/file_utils_tests.py
ShifaSZ: In one of the `get_anvil_vcf_list` tests, the mocked stdout has a warning as the code below. The test covers this case.
```
        mock_subprocess.return_value.stdout = [
            b'Warning: some packages are out of date',
            b'gs://test_bucket/test.vcf', b'gs://test_bucket/data/test.vcf.gz', b'gs://test_bucket/test.tsv',
        ]
```

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Yes, it tests a different code path. If the data path has a `*` in it, the code will use `get_gs_file_list` to get sharded VCFs. Otherwise, it calls `does_file_exist` to check file existence.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: I don't want to test the `wait()` called and make the test cleaner.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Good point.

seqr/utils/file_utils.py
ShifaSZ: The `-q` option doesn't change the behaviors of the `ls` command. I will take it out.

seqr/utils/file_utils.py
hanars: ah I see. Lets just change `line.startswith(gs_path)` to ` if is_google_bucket_file_path(line)`

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -422,14 +436,47 @@ def test_validate_anvil_vcf(self, mock_file_iter, mock_file_exist, mock_utils_lo
...
+        mock_get_header_subproc.wait.return_value = 0
+        mock_file_exist_or_list_subproc.stdout = [b'gs://test_bucket/test_path-001.vcf.gz', b'gs://test_bucket/test_path-102.vcf.gz']
+        mock_get_header_subproc.stdout = BASIC_META + INFO_META + FORMAT_META + HEADER_LINE + DATA_LINES
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY_SHARDED_DATA_PATH))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), {'fullDataPath': 'gs://test_bucket/test_path-*.vcf.gz', 'vcfSamples': ['HG00735', 'NA19675', 'NA19678']})
+        mock_subprocess.assert_has_calls([
+            mock.call('gsutil ls gs://test_bucket/test_path-*.vcf.gz', stdout=-1, stderr=-2, shell=True),
+            mock.call('gsutil ls gs://test_bucket/test_path-*.vcf.gz', stdout=-1, stderr=-2, shell=True),
+            mock.call('gsutil ls gs://test_bucket/test_path-*.vcf.gz', stdout=-1, stderr=-2, shell=True),
```
hanars, line 471: why is `ls` called 3 times in a row?

hanars (changes requested):

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: One time for file existence, one for avoiding the empty bucket exception, and the other for fetching the sharded file list.

hanars: >  one for avoiding the empty bucket exception

The empty bucket exception occurs when you run `ls` with the `**` to check the nested directories, so for this were you are never checking the nested directories it should not be run.

ShifaSZ: Okay, I see. I can move the empty bucket detection into the `ls` with the `**` branch.

hanars (approved):

PR #3184, opened 2023-01-24
Sample metadata performance

hanars:
This improves performance for single-project sample metadata pages by reducing the total number of airtable calls. Most airtable records have a `CollaboratorSampleID` that maps to the seqr ID, but a relatively small subset have a different value for this, and the actual seqr ID is stored in `SeqrCollaboratorSampleID`. Since there is a URL query limit, we are limited in the number of or filters we can send per request, so in the previous approach of filtering for either one of those fields we essentially double the number of airtable requests we make (i.e if there are 400 samples at 200 filters per request we need to make 4 requests instead of only making 2 requests if we could only filter on one field). This change first runs the query for all the records that match the `CollaboratorSampleID`, and then makes a follow up query for any records that were not returned. In the worst-case scenario the number of requests stays the same, but in the best case scenario (which is true for several of our largest projects) the number of requests is halved

### DIFF
--- a/seqr/views/utils/airtable_utils.py
+++ b/seqr/views/utils/airtable_utils.py
@@ -8,7 +8,7 @@ from settings import AIRTABLE_API_KEY, AIRTABLE_URL
 
 logger = SeqrLogger(__name__)
 
-MAX_OR_FILTERS = 200
+MAX_OR_FILTERS = 270
 
 
 class AirtableSession(object):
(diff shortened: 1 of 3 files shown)


seqr/views/utils/airtable_utils.py
```diff
@@ -8,7 +8,7 @@
 
 logger = SeqrLogger(__name__)
 
-MAX_OR_FILTERS = 200
+MAX_OR_FILTERS = 270
```
hanars, line 11: I tested with real-world data and this higher cutoff still works for every single seqr project

Issue #3178, opened 2023-01-23
Time-sensitive - sample metadata page not loading


lynnpais:
**Describe the bug**
Related to issue: Time-sensitive - data for grant submission #3177

**Link to page(s) where bug is occurring**
https://seqr.broadinstitute.org/report/sample_metadata/all

Waited 15+ minutes but the page still hasn't loaded. 

<img width="1373" alt="Screen Shot 2023-01-23 at 2 41 17 PM" src="https://user-images.githubusercontent.com/37872186/214134052-faa93443-2a34-4f61-8bf8-85d15dea3123.png">

anneodonnell:
I also tried all day before asking Lynn to check. I did once get this error message - in that helps: "(504), could not connect to server: Connection refused Is the server running on host "127.0.0.1" and accepting TCP/IP connections on port 5432?" It happened when I asked Lynn to check so not sure if the 2nd request kicked out my request.

hanars:
note that this report is currently failing for cetain single projects, as well as for all projects (i.e. https://seqr.broadinstitute.org/report/sample_metadata/R0474_cmg_sherr_exomes)

hanars:
spoke with Sam and we are going to reconfigure this page a bit. There will be still be the ability to pull this report for single projects, we will replace the "all CMG" with "all GreGOR", and then we will add an "all Broad" option which is basically all the projects a user has access to. For the "all Broad" option we will not be pulling in data from airtable (such as dbgap_study_id) which will make this page substantially faster. If in the future there are any  airtable fields it would be helpful to view for all projects we can add them back in

hanars:
this page is now working again

hanars:
this has been fixed. It now take 2-3 minutes for the report to load

PR #3186, opened 2023-01-25
sample metadata performance - db queries

hanars:
optimizes db queries to improve performance for the sample metadata report


ShifaSZ:
No issues were found in the code changes. But it only solves some problems in #3178. Is that right?

hanars:
> No issues were found in the code changes. But it only solves some problems in #3178. Is that right?

Yes that why I also previously made https://github.com/broadinstitute/seqr/pull/3184 and also have https://github.com/broadinstitute/seqr/pull/3187

ShifaSZ (approved):

PR #3201, opened 2023-02-06
VCI integration

ShifaSZ:
The updated classify button and popup look as below figure.
image

The issues discussed:
1. In Lynn's mocked popup, there are both "in seqr" (or a button for ACMG popup modal) and "in ClinGen VCI". But I have problems opening the modal in a popup. The popup can't be closed while showing the modal. Can we display both "in seqr" (or "Classify") and "ClinGen VCI" in the main window, as shown in the figure below?
image
2. I use the backend to call the Allele Register API. Since the API doesn't require authentication, it is possible to use the front end to call it. Any comments?
3. It will be nicer to call the API in the background. But I don't know how to do it without blocking or slowing down any un-relative front-end display.

### DIFF
--- a/seqr/views/apis/saved_variant_api.py
+++ b/seqr/views/apis/saved_variant_api.py
@@ -1,5 +1,6 @@
 import logging
 import json
+import requests
 from django.db.models import Q
 
 from seqr.models import SavedVariant, VariantTagType, VariantTag, VariantNote, VariantFunctionalData,\
@@ -332,3 +333,22 @@ def update_variant_main_transcript(request, variant_guid, transcript_id):
     update_model_from_json(saved_variant, {'selected_main_transcript_id': transcript_id}, request.user)
 
     return create_json_response({'savedVariantsByGuid': {variant_guid: {'selectedMainTranscriptId': transcript_id}}})
+
+
+@login_and_policies_required
+def get_clingen_allele_id(request, hgvsc):
+    if not hgvsc:
+        create_json_response({'clinGenAlleleIdByHgvsc': {}})
+
+    url = f'http://reg.genome.network/allele?hgvs={hgvsc}'
+    response = requests.get(url)
+
+    if response.status_code != 200:
+        error = 'Error: called ClinGen Allele Registry API: GET {} got status: {} with a reason: {}'.format(
+            url, response.status_code, response.reason)
+        logger.warning(error)
+        create_json_response({'error': error}, status=400, reason=error)
+
+    logger.info(f'Called ClinGen Allele Registry API: GET {url} {response.status_code} {len(response.text)}')
+
+    return create_json_response({'clinGenAlleleIdByHgvsc': {hgvsc: {'@id': response.json()['@id'].split('/')[-1]}}})
--- a/seqr/views/apis/saved_variant_api.py
+++ b/seqr/views/apis/saved_variant_api.py
@@ -1,6 +1,5 @@
 import logging
 import json
-import requests
 from django.db.models import Q
 
 from seqr.models import SavedVariant, VariantTagType, VariantTag, VariantNote, VariantFunctionalData,\
@@ -333,22 +332,3 @@ def update_variant_main_transcript(request, variant_guid, transcript_id):
     update_model_from_json(saved_variant, {'selected_main_transcript_id': transcript_id}, request.user)
 
     return create_json_response({'savedVariantsByGuid': {variant_guid: {'selectedMainTranscriptId': transcript_id}}})
-
-
-@login_and_policies_required
-def get_clingen_allele_id(request, hgvsc):
-    if not hgvsc:
-        create_json_response({'clinGenAlleleIdByHgvsc': {}})
-
-    url = f'http://reg.genome.network/allele?hgvs={hgvsc}'
-    response = requests.get(url)
-
-    if response.status_code != 200:
-        error = 'Error: called ClinGen Allele Registry API: GET {} got status: {} with a reason: {}'.format(
-            url, response.status_code, response.reason)
-        logger.warning(error)
-        create_json_response({'error': error}, status=400, reason=error)
-
-    logger.info(f'Called ClinGen Allele Registry API: GET {url} {response.status_code} {len(response.text)}')
-
-    return create_json_response({'clinGenAlleleIdByHgvsc': {hgvsc: {'@id': response.json()['@id'].split('/')[-1]}}})
--- a/ui/redux/rootReducer.js
+++ b/ui/redux/rootReducer.js
@@ -46,7 +46,6 @@ const UPDATE_USER = 'UPDATE_USER'
 const REQUEST_HPO_TERMS = 'REQUEST_HPO_TERMS'
 const RECEIVE_HPO_TERMS = 'RECEIVE_HPO_TERMS'
 const REQUEST_FAMILY_DETAILS = 'REQUEST_FAMILY_DETAILS'
-const REQUEST_CLINGEN_ALLELE_ID = 'REQUEST_CLINGEN_ALLELE_ID'
 
 // action creators
 
@@ -311,29 +310,6 @@ export const updateLocusList = values => (dispatch) => {
     }).post(values)
 }
 
-export const loadClinGeneAlleleId = hgvsc => (dispatch, getState) => {
-  if (!hgvsc) {
-    return
-  }
-  const alleleId = getState().clinGenAlleleIdByHgvsc[hgvsc]
-  if (!alleleId) {
-    dispatch({ type: REQUEST_CLINGEN_ALLELE_ID })
-    new HttpRequestHelper('http://reg.genome.network/allele',
-      (responseJson) => {
-        dispatch({
-          type: RECEIVE_DATA,
-          updatesById: { clinGenAlleleIdByHgvsc: { [hgvsc]: { alleleId: responseJson['@id'].split('/').pop() } } },
-        })
-      },
-      (e) => {
-        dispatch({
-          type: RECEIVE_DATA,
-          updatesById: { clinGenAlleleIdByHgvsc: { [hgvsc]: { alleleId: e.message } } },
-        })
-      }).get({ hgvs: hgvsc }, 'omit')
-  }
-}
-
 // root reducer
 const rootReducer = combineReducers({
   projectCategoriesByGuid: createObjectsByIdReducer(RECEIVE_DATA, 'projectCategoriesByGuid'),
@@ -373,8 +349,6 @@ const rootReducer = combineReducers({
   savedSearchesByGuid: createObjectsByIdReducer(RECEIVE_SAVED_SEARCHES, 'savedSearchesByGuid'),
   savedSearchesLoading: loadingReducer(REQUEST_SAVED_SEARCHES, RECEIVE_SAVED_SEARCHES),
   transcriptsById: createObjectsByIdReducer(RECEIVE_DATA, 'transcriptsById'),
-  clinGenAlleleIdByHgvsc: createObjectsByIdReducer(RECEIVE_DATA, 'clinGenAlleleIdByHgvsc'),
-  clinGenAlleleIdLoading: loadingReducer(REQUEST_CLINGEN_ALLELE_ID, RECEIVE_DATA),
   user: createSingleObjectReducer(UPDATE_USER),
   newUser: zeroActionsReducer,
   userOptionsByUsername: createSingleValueReducer(RECEIVE_USER_OPTIONS, {}),
--- a/ui/redux/rootReducer.js
+++ b/ui/redux/rootReducer.js
@@ -318,13 +318,16 @@ export const loadClinGeneAlleleId = hgvsc => (dispatch, getState) => {
   const alleleId = getState().clinGenAlleleIdByHgvsc[hgvsc]
   if (!alleleId) {
     dispatch({ type: REQUEST_CLINGEN_ALLELE_ID })
-    new HttpRequestHelper(`/api/clingen/get_allele_id/${hgvsc}`,
+    new HttpRequestHelper('http://reg.genome.network/allele',
       (responseJson) => {
-        dispatch({ type: RECEIVE_DATA, updatesById: responseJson })
+        dispatch({
+          type: RECEIVE_DATA,
+          updatesById: { clinGenAlleleIdByHgvsc: { [hgvsc]: { alleleId: responseJson['@id'].split('/').pop() } } },
+        })
       },
       (e) => {
         dispatch({ type: RECEIVE_DATA, error: e.message, updatesById: {} })
-      }).get()
+      }).get({ hgvs: hgvsc }, 'omit')
   }
 }
 
--- a/ui/redux/rootReducer.js
+++ b/ui/redux/rootReducer.js
@@ -326,7 +326,10 @@ export const loadClinGeneAlleleId = hgvsc => (dispatch, getState) => {
         })
       },
       (e) => {
-        dispatch({ type: RECEIVE_DATA, error: e.message, updatesById: {} })
+        dispatch({
+          type: RECEIVE_DATA,
+          updatesById: { clinGenAlleleIdByHgvsc: { [hgvsc]: { alleleId: e.message } } },
+        })
       }).get({ hgvs: hgvsc }, 'omit')
   }
 }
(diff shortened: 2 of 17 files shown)


seqr/views/apis/saved_variant_api.py
```diff
@@ -332,3 +333,22 @@ def update_variant_main_transcript(request, variant_guid, transcript_id):
     update_model_from_json(saved_variant, {'selected_main_transcript_id': transcript_id}, request.user)
 
     return create_json_response({'savedVariantsByGuid': {variant_guid: {'selectedMainTranscriptId': transcript_id}}})
+
+
+@login_and_policies_required
+def get_clingen_allele_id(request, hgvsc):
```
ShifaSZ, line 339: I don't know which file the new backend API should go to. I put it in the `saved_variant_api` for now.

hanars:
> 1. ... I have problems opening the modal in a popup. The popup can't be closed while showing the modal. Can we display both "in seqr" (or "Classify") and "ClinGen VCI" in the main window, as shown in the figure below?

No, please keep the original single button design. Use the `PopupWithModal` class for the popup. The fact that the popup stays open behind the modal is fine

> 2. I use the backend to call the Allele Register API. Since the API doesn't require authentication, it is possible to use the front end to call it. Any comments?

I think it would be neater to just use the front end for this, we can always move it to the backend later if we really have to

> 3. It will be nicer to call the API in the background. But I don't know how to do it without blocking or slowing down any un-relative front-end display.

Lets just use the front end for now, theres no advantage to the backend for the current use case

ui/shared/components/panel/clingen/ClinGenVciLink.jsx
```diff
@@ -0,0 +1,81 @@
...
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+import { Icon } from 'semantic-ui-react'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
+import { getClinGeneAlleleIdIsLoading, getClinGenAlleleIdByHgvsc } from 'redux/selectors'
+import { loadClinGeneAlleleId } from 'redux/rootReducer'
+import DataLoader from '../../DataLoader'
+import { getVariantMainTranscript } from '../../../utils/constants'
+
+class BaseClinGenAlleleId extends React.PureComponent {
```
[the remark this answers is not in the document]
hanars: Since this link is shown only on the variant, we want it in `shared/components/panel/variants`

ui/redux/rootReducer.js
```diff
@@ -310,6 +311,26 @@ export const updateLocusList = values => (dispatch) => {
...
+    new HttpRequestHelper('http://reg.genome.network/allele',
+      (responseJson) => {
+        dispatch({
+          type: RECEIVE_DATA,
+          updatesById: { clinGenAlleleIdByHgvsc: { [hgvsc]: { alleleId: responseJson['@id'].split('/').pop() } } },
+        })
+      },
+      (e) => {
+        dispatch({ type: RECEIVE_DATA, error: e.message, updatesById: {} })
+      }).get({ hgvs: hgvsc }, 'omit')
```
ShifaSZ, line 330: Add `omit` credentials to allow CORS going through.

ui/shared/components/panel/acmg/AcmgModal.jsx
```diff
@@ -30,7 +31,7 @@ const AcmgModal = (props) => {
       size="fullscreen"
       modalName={modalName}
       trigger={
-        <Button as={Label} color={buttonBackgroundColor} content={`Classify ${classify || ''}`} horizontal basic={!classify} size="small" />
+        <ButtonLink color={buttonBackgroundColor} content={`- In seqr ${classify || ''}`} />
```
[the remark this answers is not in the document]
hanars: I think showing `classify` here will be confusing. Lets show any current classification in the popup header, and then have this just say "In seqr"

hanars (changes requested):

ui/redux/rootReducer.js
ShifaSZ: The `https://gtexportal.org` responses with `access-control-allow-origin: https://seqr.broadinstitute.org` (or access-control-allow-origin: http://localhost:3000 for my local seqr) while `http://reg.genome.network` responses with `Access-Control-Allow-Origin: *`. According to the [`fetch()` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included) of the browser, Access-Control-Allow-Origin is prohibited from using a `*` for requests with credentials: 'include'.
Calling the API with credentials: 'omit' is safer because no authentication information, such as cookies, will be included in the requests.

ShifaSZ:
Current screenshots:
- **With `in seqr` classified information (pathogenic and benign)**






- **Without `in seqr` classified information**



- **CA id copied**

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,116 @@
...
+
+  static propTypes = {
+    variant: PropTypes.object.isRequired,
+    familyGuid: PropTypes.string,
+  }
+
+  state = {
+    copied: false,
+    alleleId: null,
+  };
```
ShifaSZ, line 58: Change from a function to the PureComponent class to store states. The content of the `Popup` component re-renders every time it is displayed. The states are lifted from the popup content components to maintain the states between two popup events.

hanars: This is unnecessarily complex and a premature optimization. It is better to refetch the Allele Id every time the Popup opens. The component that handles fetching and displaying the allele ID should be rendered inside the popup content. To prevent accidentally opening the popup, we should make this popup only triggered on click instead of hover which means its unlikely it will ever be opened except the one time someone wants to use it

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,116 @@
...
+  <CopyToClipboard
+    text={alleleId}
+    onCopy={onCopy}
+  >
+    <div>
+      <span>
+        {alleleId}
+        &nbsp;
+      </span>
+      {copied ? <ColoredIcon name="check circle" color="#00C000" /> : <ColoredIcon name="copy" link />}
```
hanars, line 25: the copy button should always be shown, just with a checkmark next to it on success

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,116 @@
...
+    const { variant, familyGuid } = this.props
+    const { copied, alleleId } = this.state
+    const { hgvsc } = getVariantMainTranscript(variant)
+    const { classify } = variant.acmgClassification || {}
+    const buttonBackgroundColor = getButtonBackgroundColor(classify)
+
+    return (
+      <PopupWithModal
+        onOpen={this.onOpenPopup(hgvsc)}
+        header={classify}
```
hanars, line 88: I know I told you to add this in my last review, but looking at the screenshots I don't think its necessary and I think its actually a bit confusing. I would just take it out entirely as you can still see the classification on the button while the popup is open

ui/shared/utils/httpRequestHelper.js
```diff
@@ -36,12 +36,12 @@ export class HttpRequestHelper {
...
    *   and append to the url as HTTP GET params (eg. "?gene=ENSG00012345&chrom=1")
    */
-  get = (urlParams = {}) => {
+  get = (urlParams = {}, credentials = 'include') => {
     const urlQueryString = getUrlQueryString(urlParams)
 
     const p = fetch(`${this.url}?${urlQueryString}`, {
       method: 'GET',
-      credentials: 'include',
+      credentials,
```
hanars, line 44: I think it would be clearer/more extensible if instead of having an optional `credentials` param we instead allow an optional dictionary of `fetchParams` that we add here - 
```
method: 'GET',
credentials: 'include',
...(fetchParams || {})
```

 This way the VCI request would have `{ credentials: 'omit' }` as the param it passes which is more readable

hanars (changes requested):

ui/shared/components/panel/variants/VariantClassify.jsx
ShifaSZ: Okay. I will change it to be triggered on click.

ui/shared/utils/httpRequestHelper.js
ShifaSZ: Sounds good.

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+  const { classify } = variant.acmgClassification || {}
+  const buttonBackgroundColor = getButtonBackgroundColor(classify)
+
+  return (
+    <PopupWithModal
+      content={
+        <div>
+          {hgvsc && (
+            <div>
+              <a href={CLINGEN_VCI_URL} target="_blank" rel="noreferrer">
```
hanars, line 88: As stated in the ticket and in a couple previous reviews, the whole link is not shown until the allele ID is loaded. You need to have a `VCILink` component that handles fetching the allele ID and then showing the link and Allele ID, you cannot have the link directly rendered here

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+import PopupWithModal from '../../PopupWithModal'
+import { getVariantMainTranscript } from '../../../utils/constants'
+
+const CLINGEN_ALLELE_REGISTRY_URL = 'http://reg.genome.network/allele'
+const CLINGEN_VCI_URL = 'https://curation.clinicalgenome.org/select-variant'
+
+class ClinGenAlleleId extends React.PureComponent {
+
+  static propTypes = {
+    hgvsc: PropTypes.string,
```
hanars, line 18: should be a required prop

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+  state = {
+    copied: false,
+    alleleId: null,
+  };
+
+  constructor(props) {
+    super()
+
+    const { hgvsc } = props
+    new HttpRequestHelper(CLINGEN_ALLELE_REGISTRY_URL,
```
hanars, line 30: you need to keep track of loading and show a spinner while the ID is fetching. Even if this happens very quickly for you locally, there might be a day when the allele registry is laggy. It probably best to use the `DataLoader` component 

```

load = (hgvsc) => {
  this.setState({ loading: true })
  new HttpRequestHelper(CLINGEN_ALLELE_REGISTRY_URL,
    ...
}

render() {
  const { hgvsc } = props
  const { alleleId, loading } = this.state
  return (
    <DataLoader contentId={hgvsc} content={alleleId} loading={loading} load={this.load}>
      ...
    </DataLoader>
  )
}

```

hanars (changes requested):

ui/shared/components/panel/variants/VariantClassify.jsx
ShifaSZ: I also think about hiding the link before the allele ID is ready. But we may still want the link available even if the ID is unavailable. So I made the link always available. What do you think about it?

hanars: No. As I stated in 3 separate reviews and the ticket, this is not the behavior we want

ShifaSZ: If we don't show the link while loading, the popup size is smaller, and the `In seqr` is covered by the `loading` spinner, as below figure.


After the allele ID is loaded, the popup size grows and covers the button, as shown in the figure below.


So I change it to show the popup under the button as shown below. But it still has the issue of the `In seqr` covered by the spinner before CA id is loaded.

hanars: It looks like the loading spinner is not properly setting its display so the popup is not resizing for it correctly. You can fix this in the CSS, either by fixing the spinner display or by just giving the popup a minimum hight

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button, Label } from 'semantic-ui-react'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
```
hanars, line 4: I think we want to use `React.lazy` for this library, as this is the only place it is used and therefore we can avoid adding it to our main JS bundle. Alternatively, we could move the`ClinGenAlleleId` component to its own file and lazily load it here

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  load = (hgvsc) => {
+    this.setState({ loading: true })
+    new HttpRequestHelper(CLINGEN_ALLELE_REGISTRY_URL,
+      (responseJson) => {
+        this.setState({ alleleId: responseJson['@id'].split('/').pop(), loading: false })
+      },
+      (e) => {
+        this.setState({ alleleId: e.message, loading: false })
```
hanars, line 35: can I see a screenshot of how it looks in this case?

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+import DataLoader from 'shared/components/DataLoader'
+import AcmgModal from '../acmg/AcmgModal'
+import { VerticalSpacer } from '../../Spacers'
+import PopupWithModal from '../../PopupWithModal'
+import { getVariantMainTranscript } from '../../../utils/constants'
+
+const CLINGEN_ALLELE_REGISTRY_URL = 'http://reg.genome.network/allele'
+const CLINGEN_VCI_URL = 'https://curation.clinicalgenome.org/select-variant'
+
+class ClinGenAlleleId extends React.PureComponent {
```
hanars, line 16: this is no longer an accurately named component as it has the allele ID but also the actual VCI link

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+      <div>
+        <DataLoader contentId={hgvsc} content={alleleId} loading={loading} load={this.load}>
+          <a href={CLINGEN_VCI_URL} target="_blank" rel="noreferrer">In ClinGen VCI</a>
+          <br />
+          <CopyToClipboard
+            text={alleleId}
+            onCopy={this.onCopy}
+          >
+            <div>
+              <span>
```
hanars, line 57: why is this span needed?

ui/shared/components/panel/variants/VariantClassify.jsx
```diff
@@ -0,0 +1,115 @@
...
+                {alleleId}
+                &nbsp;
+              </span>
+              <ColoredIcon name="copy" link />
+              {copied && <ColoredIcon name="check circle" color="#00C000" />}
```
hanars, line 62: It is better to use the regular icon component with one of the built in colors, in this case green

hanars (changes requested):

ui/shared/components/panel/variants/VariantClassify.jsx
ShifaSZ: Sure, it is as below figure. Maybe we should hide the copy button in case of an error.
image

ui/shared/components/panel/variants/VariantClassify.jsx
ShifaSZ: I changed it to ClinGenVciLink and moved the component into a new file.

ui/shared/components/panel/variants/VariantClassify.jsx
ShifaSZ: It is not needed. Removed.

ui/shared/components/panel/variants/ClinGenVciLink.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    const { hgvsc } = this.props
+    const { alleleId, copied, loading, error } = this.state
+
+    return (
+      <DataLoader contentId={hgvsc} content={alleleId || error} loading={loading} load={this.load}>
+        <a href={CLINGEN_VCI_URL} target="_blank" rel="noreferrer">In ClinGen VCI</a>
+        <br />
+        {error || (
```
hanars, line 47: probably better to use the `error` prop on the `DataLoader`

hanars (changes requested):

ui/shared/components/panel/variants/ClinGenVciLink.jsx
ShifaSZ: Do we still want to show the VCI link in case of fetching the CA id error? Because users may still want the link and use their own ways to find the CA id when seqr can't get it for them.

hanars: oh that is a good point

hanars (approved):

PR #3209, opened 2023-02-09
Rna seq loading slack message

hanars:


seqr/views/apis/data_manager_api_tests.py
```diff
@@ -742,40 +770,51 @@ def mock_write(content):
...
                 if params.get('extra_warnings'):
                     warnings = params['extra_warnings'] + warnings
                 file_name = RNA_FILENAME_TEMPLATE.format(data_type)
                 response_json = response.json()
-                self.assertDictEqual(response_json, {'info': info, 'warnings': warnings, 'sampleGuids': [mock.ANY], 'fileName': file_name})
+                self.assertDictEqual(response_json, {'info': info, 'warnings': warnings, 'sampleGuids': [RNA_SAMPLE_GUID, mock.ANY], 'fileName': file_name})
                 deleted_count = params.get('deleted_count', params['initial_model_count'])
-                mock_logger.info.assert_has_calls([mock.call(info_log, self.data_manager_user) for info_log in info])
+                info_log_calls = [mock.call(info_log, self.data_manager_user) for info_log in info]
+                if test_round == 0:
```
ShifaSZ, line 783: Would it be neater to use the `data_type` as the condition?

seqr/views/utils/dataset_utils.py
```diff
@@ -427,12 +432,16 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
     }
 
     prefetch_related_objects(list(samples_to_load.keys()), 'individual__family__project')
-    projects = {sample.individual.family.project.name for sample in samples_to_load}
-    project_names = ', '.join(sorted(projects))
-    message = f'Attempted data loading for {len(samples_to_load)} RNA-seq samples in the following {len(projects)} projects: {project_names}'
+    project_samples = defaultdict(list)
+    for sample in samples_to_load:
+        project_samples[sample.individual.family.project].append(sample)
```
ShifaSZ, line 437: Will the performance be better to use values_list() to get the `project.guid` and `project.name` for the samples directly?

seqr/views/apis/data_manager_api_tests.py
hanars: no because the model will be created by the first subtest that is run regardless of which data type it is, and the order is not deterministic

ShifaSZ (approved):

PR #3212, opened 2023-02-14
Splice outliers -- Dataset loading

ShifaSZ:
Add a database model for the RNA-seq splice junction outlier dataset and data loading UI.


seqr/views/utils/dataset_utils.py
```diff
@@ -304,6 +304,17 @@ def _parse_tsv_row(row):
...
 
+RNA_SPLICE_OUTLIER_COLUMNS = {
+                                 'geneID': 'gene_id',
+                                 'pValue': 'p_value',
+                                 'zScore': 'z_score',
+                                 'chrom': 'chrom',
+                                 'start': 'start',
+                                 'end': 'end',
+                                 'strand': 'strand',
+                                 'reads': 'read_count'
```
hanars, line 315: not super important, but if you change the model field to `read_count` instead of `reads` you could do `_to_camel_case` for the mapped key instead of needing to set it as an explicit mapping, and also it would be a more accurate field name

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## dev
 
+## 2/14/23
+* Add RnaSeqSpliceOutlier model (REQUIRES DB MIGRATION)
```
hanars, line 6: changelog entries go in the `dev` section, I will handle updating the release date when we actually do a release

hanars:
the overall approach here looks good so far!

ShifaSZ:
The splice junction data is slightly different from the other RNA-seq data because it has multiple records per individual and gene. We can have two ways to store the splice junction data. One is to store them in separate DB records. Another is to store the splice junctions belonging to the same gene of an individual in a single DB record with a JSONField. @hanars Which one is better?

hanars:
We need to store each junction as a separate DB record because we will want to be able to show whether a variant is in a particular junction or not.

ShifaSZ:
@hanars, I have a question about the `tissue` field of the splice junction. Do you think we should put the field in the Sample model? Is it possible to have conflicts with other RNA datasets belonging to the same individual?

hanars:
> Do you think we should put the field in the Sample model? Is it possible to have conflicts with other RNA datasets belonging to the same individual?

It is already in the sample model so yes, we should keep it there and not do a full rearchitecture of our databse relationships. We should already be supporting having multiple samples per individual with differing tissue types, this is already something that happens. In the event we have  2 different tissue types available we should have one Sample model per individual-tissue and then the related outlier/tpm/splice modesl should be associated with the sample of the desired tissue type

ShifaSZ:
> We should already be supporting having multiple samples per individual with differing tissue types

I can't find the code that creates different sample objects for each individual-tissue type. It seems to me that RnaSeqOutlier data don't have a tissue column, and RnaSeqTpm only uses a single tissue type and skips data from different tissues. https://github.com/broadinstitute/seqr/blob/5f4c91329f5cc9469214abb8fd95dc7ac22b827f/seqr/views/utils/dataset_utils.py#L351-L364

hanars:
>  RnaSeqTpm only uses a single tissue type and skips data from different tissues

That is when there is a single file that specifies different tissue types for the same sample. However, if you load TPM data once from one tissue and then again from another file for a different tissue it should just create a new sample model for the new tissue type. In the same way, if you have previously loaded either Splice junction or TPM data for one individual in a particular tissue tpe and you now have new data in a new tissue type, we should have a model for each tissue type

ShifaSZ:
> if you load TPM data once from one tissue and then again from another file for a different tissue it should just create a new sample model for the new tissue type

Loading a TPM file for a different tissue creates a new sample model only when the TPM file provides a new sample ID. In the case of Splice Junction Outlier data, we use an individual ID rather than a sample ID. So we can't have a different sample ID. We could twist the code slightly to create a new sample model with the same sample ID but a different tissue ID. Or we could generate a new sample ID for the new tissue type. Which is better?

hanars:
> We could twist the code slightly to create a new sample model with the same sample ID but a different tissue ID. 

Lets do this

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -767,7 +836,7 @@ def _set_file_iter_stdout(rows):
                 self.assertEqual(response.status_code, 400)
                 self.assertDictEqual(response.json(), {'error': mock.ANY})
                 self.assertTrue(response.json()['error'].startswith(
-                    f'Error in NA19675_D2 data for {mismatch_row[2]}: mismatched entries '))
+                    f'Error in {loaded_data_row[0]} data for {mismatch_row[2]}: mismatched entries '))
```
ShifaSZ, line 839: The `outlier` and `tpm` tests use `NA19675_D2` as the loaded sample id. But the `splice` can't use the same sample id because the `splice` data provide the individual id rather than the sample id, and there is no individual with `NA19675_D2` as the individual id in the fixture data. So the tests for `splice` data use a different sample id (`NA20889`).

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -846,8 +915,9 @@ def mock_write(content):
                     warnings = params['extra_warnings'] + warnings
                 deleted_count = params.get('deleted_count', params['initial_model_count'])
                 response_json, new_sample_guid = _test_basic_data_loading(
-                    params['new_data'], params["num_parsed_samples"], 2, 16, body,
-                    '1kg project nåme with uniçøde, Test Reprocessed Project', warnings=warnings,
+                    params['new_data'], params["num_parsed_samples"], params.get('num_loaded_samples', 2), 16, body,
+                    params.get('project_names', '1kg project nåme with uniçøde, Test Reprocessed Project'), warnings=warnings,
```
ShifaSZ, line 919: The tests for `splice` has a different number of `num_loaded_samples` and `project_names`.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -857,12 +927,16 @@ def mock_write(content):
                             'updateType': 'bulk_update', 'updateFields': ['data_source']}}),
                     ])
                 self.assertTrue(params['sample_guid'] in response_json['sampleGuids'])
-                self.assertEqual(mock_send_slack.call_count, 2)
-                mock_send_slack.assert_has_calls([
+                slack_call_count = params.get('slack_call_count', 2)
+                self.assertEqual(mock_send_slack.call_count, slack_call_count)
```
ShifaSZ, line 931: The tests for `splice` have different numbers of slack calls.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -871,7 +945,8 @@ def mock_write(content):
 
                 # test database models are correct
                 self.assertEqual(model_cls.objects.count(), params['initial_model_count'] - deleted_count)
-                sample_guid = self._check_rna_sample_model(individual_id=1, data_source='new_muscle_samples.tsv.gz',
+                sample_guid = self._check_rna_sample_model(individual_id=params.get('exit_sample_individual_id', 1),
+                                                           data_source='new_muscle_samples.tsv.gz',
```
ShifaSZ, line 949: The tests for `splice` have a different existing individual id.

seqr/models.py
```diff
@@ -1104,6 +1104,35 @@ class Meta:
...
+        ('-', '3′ to 5′ direction'),
+        ('*', 'Any direction'),
+    )
+
+    p_value = models.FloatField()
+    z_score = models.FloatField()
+    chrom = models.CharField(max_length=2)
+    start = models.IntegerField()
+    end = models.IntegerField()
+    strand = models.CharField(max_length=1, choices=STRAND_CHOICES)  # "+", "-", or "*"
```
hanars, line 1119: are we going to show this in seqr?

seqr/models.py
```diff
@@ -1104,6 +1104,35 @@ class Meta:
...
+    gene_name = models.CharField(max_length=20)
```
hanars, line 1120: the base class already has a gene_id field, that should be sufficient

seqr/models.py
```diff
@@ -1104,6 +1104,35 @@ class Meta:
...
+    type = models.CharField(max_length=12)
+    delta_psi = models.FloatField()
+    read_count = models.IntegerField()  # RNA-seq reads that span the splice junction
+    dot_size = models.FloatField()
+    rare_disease_samples_with_junction = models.IntegerField()
+    rare_disease_samples_total = models.IntegerField()
```
hanars, line 1126: wre should only be stroing fields that need displaying - I think it would be helpful to update the display ticket to be more explicit about what we will actually be needing

seqr/utils/xpos_utils.py
```diff
@@ -40,13 +40,8 @@
...
 MAX_POS = 3e8
 
-def get_xpos(chrom, pos):
-    """Compute single number representing this chromosome and position.
 
-    Args:
-        chrom (string): examples '1', 'Y', 'M'
-        pos (integer): genomic position on chromosome
-    """
+def get_chrom(chrom):
```
hanars, line 44: I think `format_chrom` would be a better name for this

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -823,7 +893,7 @@ def _test_basic_data_loading(data, num_parsed_samples, num_loaded_samples, indiv
                         'dbEntity': 'Sample', 'updateType': 'bulk_create',
                         'entityIds': response_json['sampleGuids'] if num_created_samples > 1 else
                             [response_json['sampleGuids'][-1]],
-                    }})] + (additional_logs or [])
+                    }})] + (additional_logs or []) if num_created_samples else additional_logs or []
```
hanars, line 896: the conditional belongs above when you create `additional_logs`, there is no reason to create a variable on the above line and then immediately follow it with a conditional where it is not used

seqr/views/apis/data_manager_api_tests.py
hanars: why not just load 2 samples for the splice test though?

seqr/views/apis/data_manager_api_tests.py
hanars: I really think you should be updating the same individual in the splice test, there is no reason not to

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -885,16 +960,18 @@ def mock_write(content):
...
-                mock_writes = []
-                _test_basic_data_loading(data, 2, 2, 20, body, '1kg project nåme with uniçøde, Test Reprocessed Project',
-                                         num_created_samples=2)
-                self.assertSetEqual(set([s.split('_', 1)[1] for s in mock_writes]), params['write_data'])
+                if 'duplicated_indiv_id_data' in params:
+                    data = params['duplicated_indiv_id_data']
+                    mock_writes = []
+                    _test_basic_data_loading(data, 2, 2, 20, body, '1kg project nåme with uniçøde, Test Reprocessed Project',
+                                             num_created_samples=2)
+                    self.assertSetEqual(set([s.split('_', 1)[1] for s in mock_writes]), params['write_data'])
```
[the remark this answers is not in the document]
hanars: What is the harm in testing that the shared code works the same in all cases? This test is too unreadable with the amount of conditionals you have added, if we want the new test case to share functionality with the previous tests it needs to be much closer to the previous tests

seqr/views/utils/dataset_utils.py
```diff
@@ -320,6 +321,38 @@ def _parse_tsv_row(row):
...
 TPM_HEADER_COLS = [SAMPLE_ID_COL, PROJECT_COL, GENE_ID_COL, TISSUE_COL, TPM_COL]
 
+CHROM_COL = 'chrom'
+START_COL = 'start'
+END_COL = 'end'
+STRAND_COL = 'strand'
+READ_COUNT_COL = 'read_count'
+DOT_SIZE_COL = 'dot_size'
+SPLICE_TYPE_COL = 'type'
+P_VALUE_COL ='p_value'
```
hanars, line 331: wouldn't the be shared with the existing outlier columns?

seqr/views/utils/dataset_utils.py
```diff
@@ -461,7 +520,7 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
 
 
 def _notify_rna_loading(model_cls, sample_projects):
-    data_type = 'Outlier' if model_cls == RnaSeqOutlier else 'Expression'
+    data_type = 'Outlier' if model_cls == RnaSeqOutlier else 'Expression' if model_cls == RnaSeqTpm else 'Splice Junction'
```
hanars, line 523: use a dictionary lookup, not a three tiered if else statement

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: Good catch. I added `num_loaded_samples` for the `splice` when I was trying to load for two projects. I changed to add two samples to one project in the end. I could change it to add two samples to two projects by configuring the individual id.

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: The existing sample ID for `outlier` and `TPM` is  `NA19675_D2` and the corresponding individual ID is `NA19675`. The `splice` dataset has an `individual_id` column but no `sample_id` column. We have no way to map the individual ID `NA19675` to the sample ID `NA19675_D2`. So we can't use the same existing sample IDs for the other RNA datasets.

I maybe can add another existing sample model belonging to the same individual as other RNA datasets for the `splice` data.

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: Now, it always loads two samples to two projects. No more params for `num_loaded_samples` and ` project_names`.

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: I added a fixture sample with a sample_id `NA19675_1` for the individual NA19675_1 (the same individual of the existing sample models of the Outlier and TPM data). It removes most differences between the `splice` and `outlier/tpm` tests.

seqr/views/apis/data_manager_api_tests.py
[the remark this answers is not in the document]
ShifaSZ: I changed to do the tests for all cases.

seqr/models.py
ShifaSZ: Yes. Ben mentioned, `hover over the tag should show info about the splice junction outlier (chrom/start/end/strand, z-score, pvalue, maybe tissue and read count also)`.

seqr/views/utils/dataset_utils.py
ShifaSZ: Splice Junction Outlier and existing Outlier have the `p_value` and `z_score` columns. But the Outlier has a `p_adjust` column which Splice Junction Outlier doesn't need. Otherwise, using the `Outlier` model as a base class for the `Splice Junction Outlier` model might be good.

hanars: lets keep them separate, I think they are different enough that the savings we get from sharing the model aren't worth it

seqr/models.py
ShifaSZ: I removed the `gene_name` field. Do we need to verify that the `gene_id` exists in the seqr DB and that the `gene_name` matches the DB info? Or do we not use the gene name for display?

seqr/views/utils/dataset_utils.py
```diff
@@ -366,15 +422,17 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
 
     sample_id_to_tissue_type = {}
     samples_with_conflict_tissues = defaultdict(set)
+    miss_matches = []
     for line in tqdm(f, unit=' rows'):
         row = dict(zip(header, _parse_tsv_row(line)))
         for sample_id, row_dict in parse_row(row):
             tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None))
             project = row_dict.pop(PROJECT_COL)
+            indiv_id = row_dict.pop(INDIV_ID_COL, None)
```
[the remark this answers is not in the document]
ShifaSZ: Because we need to pop the `individual id` field before miss-match detection and error message generation, otherwise, the row_dict could have an `individual_id` field, but the existing_data doesn't.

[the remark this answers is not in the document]
ShifaSZ: I updated the test to catch treating the identical rows as miss-matches.

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -922,14 +998,14 @@ def test_load_rna_seq_sample_data(self, mock_open, mock_os):
...
                 models = model_cls.objects.all()
                 self.assertEqual(models.count(), 2)
-                self.assertSetEqual({model.sample.guid for model in models}, {params['sample_guid']})
+                self.assertSetEqual({model.sample.guid for model in models}, {sample_guid})
 
                 mock_open.assert_called_with(file_name, 'rt')
 
                 self.assert_json_logs(self.data_manager_user, [
-                    ('Loading outlier data for NA19675_D2', None),
+                    (f'Loading outlier data for {sample_guid.split("_", 1)[1].upper()}', None),
```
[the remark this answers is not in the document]
ShifaSZ: It is a sample ID. I know it is not straightforward. I can change to use the sample ID in the `params["loaded_data_row"]`.

seqr/views/utils/dataset_utils.py
```diff
@@ -383,17 +442,22 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
-            existing_data = samples_by_id[(sample_id, project)].get(gene_id)
+            if get_unique_key:
+                gene_or_unique_id = get_unique_key(row_dict)
+            else:
+                gene_or_unique_id = row_dict['gene_id']
+            existing_data = samples_by_id[(sample_id, project)].get(gene_or_unique_id)
             if existing_data and existing_data != row_dict:
-                raise ValueError(
-                    f'Error in {sample_id} data for {gene_id}: mismatched entries {existing_data} and {row_dict}')
+                errors.append(f'Error in {sample_id} data for {existing_data["gene_id"]}: mismatched entries '
```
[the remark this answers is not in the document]
ShifaSZ: I know it could be more precise. But the users should be able to determine which data rows are mismatched based on the gene IDs.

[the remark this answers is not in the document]
ShifaSZ: I changed it to show the data row id as `ENSG00000106554-7-132885746-132886973-*-psi5`.

hanars (approved):

PR #3213, opened 2023-02-14
Individual row collapsed in case review

hanars:
Moves the behavior for toggling whether a FamilyLayout` component is collapsed or not into a helper component instead of implementing directly for Family table rows, so behavior can be easily shared for individual rows. This does require some tweaking for the display, but overall everything (except case review) looks the same

### DIFF
--- a/ui/pages/Project/components/FamilyTable/CollapsableLayout.jsx
+++ b/ui/pages/Project/components/FamilyTable/CollapsableLayout.jsx
@@ -1,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+class CollapsableLayout extends React.PureComponent {
+
+  static propTypes = {
+    children: PropTypes.node,
+  }
+
+  state = { showDetails: false }
+
+  toggle = () => {
+    this.setState(prevState => ({ showDetails: !prevState.showDetails }))
+  }
+
+  render() {
+    const { children } = this.props
+    const { showDetails } = this.state
+    return React.cloneElement(children, { showDetails, toggleDetails: this.toggle })
+  }
+
+}
+
+export default CollapsableLayout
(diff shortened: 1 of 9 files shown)


ui/pages/Project/components/FamilyTable/CollapsableLayout.jsx
```diff
@@ -0,0 +1,35 @@
...
+  static propTypes = {
+    layoutComponent: PropTypes.elementType.isRequired,
+    detailFields: PropTypes.arrayOf(PropTypes.object).isRequired,
+    noDetailFields: PropTypes.arrayOf(PropTypes.object),
+  }
+
+  state = { showDetails: false }
+
+  toggle = () => {
+    this.setState(prevState => ({ showDetails: !prevState.showDetails }))
```
ShifaSZ, line 15: Using an object parameter for the `setState` function should be sufficient.

ui/shared/components/panel/family/Family.jsx
```diff
@@ -189,15 +190,15 @@ class Family extends React.PureComponent {
               {`(${family.individualGuids.length})`}
             </span>
           ) : (
-            <div key="header">
+            <span key="header">
```
ShifaSZ, line 193: Is the `key` prop necessary?

ShifaSZ (changes requested):

ui/pages/Project/components/FamilyTable/CollapsableLayout.jsx
hanars: I don't understand what that means. This is adapted from a copy-paste

ui/shared/components/panel/family/Family.jsx
hanars: yes

ui/pages/Project/components/FamilyTable/CollapsableLayout.jsx
ShifaSZ: Using a function for the setState parameter allows access to the `latest` state value after previous calls to the `setState`. We don't have multiple calls to the `setState` before re-rendering. So it is safe and neater to use an object parameter like `this.setSate({ showDetails: !prevState.showDetails })`.

hanars: thats a violation of our linter and less stable than the function approach used here

PR #3221, opened 2023-02-15
allow skip airtable for sample metadata

hanars:
Allows users to opt to skip the airtable metadata in a single-project sample metadata table

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -183,8 +183,9 @@ def anvil_export(request, project_guid):
 
 @analyst_required
 def sample_metadata_export(request, project_guid):
-    omit_airtable = project_guid == 'all'
-    if omit_airtable:
+    is_all_projects = project_guid == 'all'
+    omit_airtable = is_all_projects or 'true' in request.GET.get('omitAirtable', '')
+    if is_all_projects:
         projects = get_internal_projects()
     else:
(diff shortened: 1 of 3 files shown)


seqr/views/apis/report_api.py
```diff
@@ -183,8 +183,9 @@ def anvil_export(request, project_guid):
 
 @analyst_required
 def sample_metadata_export(request, project_guid):
-    omit_airtable = project_guid == 'all'
-    if omit_airtable:
+    is_all_projects = project_guid == 'all'
+    omit_airtable = is_all_projects or 'true' in request.GET.get('omitAirtable', '')
```
ShifaSZ, line 187: Is it neater to change the condition after the `or` to `request.GET.get('omitAirtable', false)`?

ShifaSZ (changes requested):

seqr/views/apis/report_api.py
hanars: not really, it has the same effect and I think differentiating between missing and false is cleaner

ShifaSZ: Or, how about changing it to `'true' === request.GET.get('omitAirtable', '')`. The `in` operation seems to need more computation.

hanars: No because the param can be an array with a single string

ShifaSZ (approved):

PR #3224, opened 2023-02-17
Gregor sample manifest

hanars:


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -102,7 +103,7 @@ def parse_pedigree_table(parsed_file, filename, user, project=None, fail_on_warn
     except Exception as e:
         raise ErrorsWarningsException(['Error while converting {} rows to json: {}'.format(filename, e)], [])
 
-    warnings = validate_fam_file_records(json_records, fail_on_warnings=fail_on_warnings)
+    warnings = validate_fam_file_records(json_records, fail_on_warnings=fail_on_warnings, errors=errors)
```
ShifaSZ, line 106: It is weird that the `errors` don't belong to `validate_fam_file_records` but are handled in it. If we have to raise the exception for all `errors` together, how about letting `validate_fam_file_records` return errors? And an exception for all errors outside of the function.

ShifaSZ (changes requested):

seqr/views/utils/pedigree_info_utils.py
hanars: no thats much less clear. This validate function, like every other validate function in our repo, will throw an exception if validation fails. It just also accepts some already found errors that it can merge with any newly found validation errors so all validation errors can be reported at one

ShifaSZ (approved):

PR #3226, opened 2023-02-17
Sample manifest - set proband relationship

hanars:


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -354,6 +356,56 @@ def _parse_merged_pedigree_sample_manifest_format(rows, project):
...
+    for r in json_records:
+        records_by_family[r[JsonConstants.FAMILY_ID_COLUMN]].append(r)
+
+    family_relationships = {}
+    for family_id, records in records_by_family.items():
+        affected = [r for r in records if r[JsonConstants.AFFECTED_COLUMN] == 'A']
+        if len(affected) > 1:
+            affected_children = sorted(
+                [r for r in affected if r[JsonConstants.PATERNAL_ID_COLUMN] and r[JsonConstants.MATERNAL_ID_COLUMN]],
+                key=lambda r: bool(r[JsonConstants.PATERNAL_ID_COLUMN]) and bool(r[JsonConstants.MATERNAL_ID_COLUMN]),
```
ShifaSZ, line 370: All the sorting keys must be the same with a value `True` because all the elements in the array meet `r[JsonConstants.PATERNAL_ID_COLUMN] and r[JsonConstants.MATERNAL_ID_COLUMN]`. You may want to change the condition to `or` for filtering the elements.

ShifaSZ (approved):

PR #3234, opened 2023-02-21
Metadata entered filter

hanars:


ui/pages/Project/constants.js
```diff
@@ -262,11 +267,17 @@ const familyIsInReview = (family, individualsByGuid) => getFamilyCaseReviewStatu
...
+).map(({ field, subFields }) => (subFields ? subFields[0].field : field))
+
+const familyHasRequiredMetadata = (family, individualsByGuid) => {
   const individuals = family.individualGuids.map(
     individualGuid => individualsByGuid[individualGuid],
   ).filter(individual => individual)
-  return individuals.length ? individuals.some(({ features }) => (features || []).length > 0) : family.hasFeatures
+  return individuals.length ? individuals.some(individual => REQUIRED_METADATA_FIELDS.every(
+    field => individual[field] || individual[field] === false,
+  ) && individual.features.length > 0) : family.hasRequiredMetadata
```
ShifaSZ, line 280: Is the `individual.features` still possible to be null?

ui/shared/components/panel/view-fields/BaseFieldView.jsx
```diff
@@ -199,6 +200,7 @@ class BaseFieldView extends React.PureComponent {
           content="Only visible to internal users."
         />
       ),
+      isRequired && <Icon key="required" name="asterisk" size="small" />,
```
ShifaSZ, line 203: You may want to hide the asterisk when the field has a value.

ShifaSZ (changes requested):

ui/shared/components/panel/view-fields/BaseFieldView.jsx
hanars: no per the ticket we always want to show its required

ui/pages/Project/constants.js
hanars: no because `features` is included in `REQUIRED_METADATA_FIELDS` and we just ran the `REQUIRED_METADATA_FIELDS.every(field => individual[field])`

ShifaSZ (approved):

PR #3237, opened 2023-02-21
Upgrade to python 3.9

sjahl:
This should resolve the security alert for the old numpy version that won't run on anything newer than py 3.7. I upgraded to python 3.9, which is the latest supported python release that only gets security updates, and is supported until 2025. My reasoning here was that this version was likely to result in the fewest breaking changes for us.

Still confirming that the docker build works locally, but wanted to run the unit tests in parallel, I'll comment back when my local docker build looks good.


sjahl:
OK, local docker built fine. Snyk found some other problem on this branch... but I can't log in to see what it is. @hanars would you mind confirming what issue Snyk is raising?

hanars:
Snyk is complaining about this issue in `setuptools`: https://security.snyk.io/vuln/SNYK-PYTHON-SETUPTOOLS-3180412
Introduced through: gunicorn@20.1.0
Fixed in: setuptools@65.5.1

sjahl:
Was this ignored before? The update didn't change the version of gunicorn in the requirements file.

In general, we probably don't want to add a setuptools pin to our requirements file, because it can break the pip installation in virtualenvs. Gunicorn looks like it just has a minimum version floor for setuptools in its setup.py file. If we were concerned about this vulnerability in our deployment, i can add a step to the dockerfile which upgrades setuptools before doing a pip install, which should allow pip to decide the latest version of setuptools that it's compatible with. But that still won't fix Snyk...

hanars:
> we probably don't want to add a setuptools pin to our requirements file, because it can break the pip installation in virtualenvs... i can add a step to the dockerfile which upgrades setuptools before doing a pip install, which should allow pip to decide the latest version of setuptools that it's compatible with.

Would pinning a minimum version in our requirements file and then upgrading in the dockerfile solve the issue?

sjahl:
If we set a minimum requirement, upgrading setuptools explicitly in the dockerfile isn't necessary, since it'll just pull that minimum version when we pip install.

In my experience, you don't usually want to put any version constraints on setuptools, because it makes your resolved package set much more difficult to install in different scenarios where you can't control your version of pip for some reason. This is why pip-tools doesn't pin it by default, and leaves it in there with a comment/note. If we just want to silence the snyk alarm and we're OK with the possible implications of potential installation problems, we just need to run `pip-tools` with the `--allow-unsafe` flag, and it'll pin setuptools for us like any other package.

hanars:
snyk just threw an "unable to run" one-off error on that one. Looking at this and what you explained I think it would be better to not pin setuptools and add a check or upgrade in the docker file to prevent using a vulerable package version as you proposed earlier, and then I can tell snyk to ignore the error

sjahl:
Yeah, snyk is precisely one of those places that we can't really control pip versions 😄

I will revert that requirements pin and do a setuptools upgrade in the docker build

hanars (approved):

Issue #45, opened 2014-10-27
Show AC and AN along with AF


brettpthomas:
Clearly we need this for all the ExAC populations, but I'm not sure the best way to do it. Maybe we show when you hover over the allele frequncies in the variant list?

PR #3240, opened 2023-02-22
Gregor report - validate and upload

hanars:


seqr/views/apis/report_api.py
```diff
@@ -893,6 +932,112 @@ def _get_experiment_ids(airtable_sample, airtable_metadata):
...
+            )
+        if warn_missing:
+            warnings.append(
+                f'{error_template.format(issue="are missing recommended")}: {", ".join(sorted(warn_missing))}'
+            )
+
+
+def _get_row_id(row):
+    id_col = next(col for col in ['participant_id', 'experiment_sample_id', 'family_id'] if col in row)
+    return row[id_col]
```
ShifaSZ, line 1038: It may be better to return a row number when it doesn't have an ID column. But I wonder if it is necessary.

seqr/views/apis/report_api_tests.py
```diff
@@ -582,39 +630,91 @@ def test_sample_metadata_export(self, mock_google_authenticated):
...
+                'Unable to load data model for validation: 404 Client Error: Not Found for url: http://raw.githubusercontent.com/gregor_data_model.json',
+            ] + skipped_file_validation_warnings,
+        })
 
-        files = self._get_zip_files(mock_zip, [
-            'participant.tsv', 'family.tsv', 'phenotype.tsv', 'analyte.tsv', 'experiment_dna_short_read.tsv',
-            'aligned_dna_short_read.tsv', 'aligned_dna_short_read_set.tsv', 'called_variants_dna_short_read.tsv',
-        ])
+        self.assertListEqual(
+            mock_open.call_args_list, [mock.call(f'/mock/tmp/{file}.tsv', 'w') for file in expected_files])
```
ShifaSZ, line 713: It looks better than using the `assert_has_calls` function.

seqr/views/utils/export_utils.py
```diff
@@ -64,25 +65,42 @@ def export_table(filename_prefix, header, rows, file_format='tsv', titlecase_hea
...
+        content += '\n'.join([
+            DELIMITERS[file_format].join(row) for row in content_rows
+            if any(val != blank_value for val in row)
+        ])
+        content = str(content.encode('utf-8'), 'ascii', errors='ignore')  # Strip unicode chars in the content
+        parsed_files.append(('{}.{}'.format(filename, file_format), content))
+    return parsed_files
+
+
+def export_multiple_files(files, zip_filename, **kwargs):
```
ShifaSZ, line 89: Maybe, the Anvil report should also be saved to a bucket so that no file downloading is needed for the reports.

ui/pages/Report/components/Gregor.jsx
```diff
@@ -1,16 +1,53 @@
...
+      }
+      return Promise.reject(error)
+    }).post(values)
+
+  render() {
+    const { uploadStats } = this.state
+    return (
+      <div>
+        <Header size="medium" textAlign="center" content="Validate and Upload GREGoR Reports" />
+        <UploadFormPage fields={FIELDS} uploadStats={uploadStats} onSubmit={this.onSubmit} />
```
ShifaSZ, line 46: I wonder if it is okay to say `Upload` because the data doesn't leave the cloud.

ShifaSZ (changes requested):

seqr/views/apis/report_api.py
```diff
@@ -893,6 +932,112 @@ def _get_experiment_ids(airtable_sample, airtable_metadata):
...
+            col_summary = ', '.join(sorted(extra_columns))
+            warnings.append(
+                f'The following columns are included in the "{file_name}" table but are missing from the data model: {col_summary}'
+            )
+        missing_columns = set(table_validator.keys()).difference(columns)
+        if missing_columns:
+            col_summary = ', '.join(sorted(missing_columns))
+            warnings.append(
+                f'The following columns are included in the "{file_name}" data model but are missing in the report: {col_summary}'
+            )
```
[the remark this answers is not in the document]
hanars: We report warnings if the schema of the report seqr is generating differs from the one in the one in the data model, we only report errors when the data model and the report match up but the data in the report does not pass the data model validation .This line reports a wraning about a misalignment of the data model, and we report an error if any row is missing a required column below

seqr/views/apis/report_api.py
```diff
@@ -893,6 +932,112 @@ def _get_experiment_ids(airtable_sample, airtable_metadata):
...
+
+        for column in columns:
+            _validate_column_data(
+                column, file_name, data, column_validator=table_validator.get(column, {}),
+                warnings=warnings, errors=errors,
```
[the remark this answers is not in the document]
hanars: no? the report needs to stay the same as before

seqr/views/apis/report_api_tests.py
hanars: In general I agree, but that doesn't work because the `assert_has_calls` also checks for all the calls to `__enter__` and `write` and it is really unreadable once those are all included (in addition to being like 32 calls). I really want to check that open is called the exact way we intend for it to, which is what this test does

seqr/views/utils/export_utils.py
hanars: Tis ticket is about the Gregor reports which are delivered to the Gregor consortium via AnVIL workspaces so Sam put in a ticket to streamline that process, but the AnVIL reports are a totally different  thing that are used in a totally different way and noone asked us to move, so no

ui/pages/Report/components/Gregor.jsx
hanars: this is the name of an existing react component that is never exposed to the user and renaming it is out of scope

seqr/views/apis/report_api.py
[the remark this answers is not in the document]
ShifaSZ: You may know that a missing column won't be validated in the column data validation. So no error will be reported for the missing column afterward.

ui/pages/Report/components/Gregor.jsx
ShifaSZ: How about the `Upload` in the content prop of the `Header` component in line #45?

hanars: It uploads the reports to the given AnVIL bucket

seqr/views/apis/report_api.py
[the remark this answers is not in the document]
hanars: If the column is missing in the data model we obviously can't validate the data in that column it because there is no model to validate against. If a column is missing in the report we can't validate that column  because it doesn't exist. Therefore we warn the user about these cases. I don't understand what change you would want here, keeping in mind my above explanation about what we warn on

ShifaSZ (approved):

PR #3243, opened 2023-02-23
Separate RNAseq sample models by tissue types

ShifaSZ:
This PR separates sample models per tissue type. RNAseq outlier data don't have tissue types, so their sample models' tissue types are `None` and won't share sample models with TPM data.


seqr/views/utils/dataset_utils.py
```diff
@@ -421,8 +421,7 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
         samples = validate_samples(samples, sample_id_to_tissue_type, warnings)
 
     # Delete old data
-    individual_db_ids = {s.individual_id for s in samples}
-    to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
+    to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
```
ShifaSZ, line 425: Please also review the logic for the code below using `prev_loaded_individual_ids`.

hanars: this is not the behavior we want- we want to remove all data previously loaded for this individual and tissue type. If this data is from a new file it will have a new sample model and then these won't be deleted anymore even if its the same tissue type

ShifaSZ: The `samples` meet the per `individual and tissue type` conditions. So the new code removes all data previously loaded for the individual and tissue type.

hanars: can you add unit tests for this behavior? We should test that data for an individual is not removed when its a new tissue type, but is removed for an existing tissue type when reloaded from a new file

ShifaSZ: Yes, I can add some tests. And I also need to make some changes to allow TPM data to have more than one tissue type. The existing code doesn't create new sample models for new tissues but validates the tissue and skips loading the unmatched tissue data.

ShifaSZ: After adding a sample of `fibroblasts` tissue and RnaSeqTpm data for the sample, the test for loaded data failed for both Outlier and Tpm.
The response for the Outlier is:
```
{'info': ['Parsed 1 RNA-seq samples', 'Attempted data loading for 1 RNA-seq samples in the following 1 projects: 1kg project nåme with uniçøde'],
'sampleGuids': ['S000152_na19675_d2'],
'warnings': ['Skipped loading for 1 samples already loaded from this file'],
'fileName': 'rna_sample_data__outlier__2020-04-15T00:00:00.json.gz' }
```
For TPM it is:
```
{'info': ['Parsed 1 RNA-seq samples', 'Attempted data loading for 0 RNA-seq samples in the following 0 projects: '], 
'sampleGuids': [],
'warnings': ['Skipped data loading for the following 1 samples due to mismatched tissue type: NA19675_D2 (muscle to fibroblasts)', 'Skipped loading for 1 samples already loaded from this file']',
fileName': 'rna_sample_data__tpm__2020-04-15T00:00:00.json.gz'}
```
While the expected response is:
```
{'info': ['Parsed 1 RNA-seq samples', 'Attempted data loading for 0 RNA-seq samples in the following 0 projects: '], 'warnings': ['Skipped loading for 1 samples already loaded from this file'],
'sampleGuids': [], 'fileName': <ANY>}
```

hanars: Make sure you are using a different file name for the new tissue type in your tests. If you are still seeing an issue, that either means you wrote the test case wrong or it means you did not correctly update the application behavior

ShifaSZ: The new sample has a new tissue type and data source, as shown follows.
```
        "tissue_type": "F",
        "data_source": "fibroblasts_samples.tsv.gz",
```

ShifaSZ: The Outlier test fails because the function `_find_matching_sample_records` return more samples than expected. It returns all the samples with the same `sample_id` and different tissue types. It results in duplicating data for each tissue type.
The TMP test doesn't load the loaded data which is correct because it filters out the extra samples by matching the tissue type. The Outlier can't do the filtering because Outlier data don't have a tissue type.
The TMP tests fails for it has an extra warning message while filtering out the unwanted samples. And it also deletes the existing TMP data unexpectedly if we don't change the code for the `to_delete`.

To solve all these issues, I'm going to add tissue type filtering for both Outlier and TPM samples and remove the warnings if samples are filtered out.
But there is still an issue: The Outlier samples must have a `None` tissue type. It could have backward compatibility issue. There are quite a lot of Outlier data share the sample models with TPM data. These sample models' tissue type is not `None`. When loading new Outlier data the existing Outlier data won't be deleted automatically.

hanars: > The Outlier samples must have a None tissue type.

They don't have to have a None tossue type but they can. If we load Outlier data first and don't have a tissue type, and then later add TPM data and do have a tissue type, we want to set the tissue type on the existing sample model, not create a new model with no tissue type

> When loading new Outlier data the existing Outlier data won't be deleted automatically.

You will have to add special handling for if you are loading new data with no known tissue type, you want to delete any data previously loaded on any tissue type in that case

hanars: Alternatively, we could stop sharing sample models between different calling types. So there would no longer be the case where Outlier and TPM models share the same sample, each would have their own sample. That would also solve a bunch of the issue you outline here

seqr/fixtures/1kg_project.json
```diff
@@ -1237,6 +1236,44 @@
         "loaded_date": "2017-02-05T06:42:55.397Z"
     }
 },
+{
+    "model": "seqr.sample",
+    "pk": 152,
+    "fields": {
```
ShifaSZ, line 1242: Add this sample for TPM data after separating the Outlier and TPM data samples.

seqr/fixtures/1kg_project.json
```diff
@@ -1237,6 +1236,44 @@
...
+        "tissue_type": "M",
+        "data_source": "muscle_samples.tsv.gz",
+        "loaded_date": "2017-02-05T06:42:55.397Z"
+    }
+},
+{
+    "model": "seqr.sample",
+    "pk": 153,
+    "fields": {
+        "guid": "S000153_na19675_d2",
```
ShifaSZ, line 1262: Add this sample and a TPM entry afterward to test if the TPM data is deleted when adding data for a different tissue.

seqr/views/utils/dataset_utils.py
```diff
@@ -97,7 +97,7 @@ def load_mapping_file_content(file_content):
 def match_sample_ids_to_sample_records(
         projects,
         user,
-        sample_ids,
+        sample_id_to_tissue_type,
```
ShifaSZ, line 100: Pass both sample ids and the mapping from sample ids to tissue types to the function.

seqr/views/utils/dataset_utils.py
```diff
@@ -133,9 +133,11 @@ def match_sample_ids_to_sample_records(
             [2] array: ids of Individuals with exact-matching existing samples
     """
 
+    sample_ids = sample_id_to_tissue_type.keys()
     samples = _find_matching_sample_records(
         projects, sample_ids, sample_type, dataset_type, elasticsearch_index,
     )
+    samples = _check_invalid_tissues(samples, sample_id_to_tissue_type)
```
ShifaSZ, line 140: Filter out the samples for different tissue types.

seqr/views/utils/dataset_utils.py
```diff
@@ -320,43 +323,29 @@ def _parse_tsv_row(row):
...
+        tissue_type = sample_id_to_tissue_type[sample.sample_id]
+        if sample.tissue_type != tissue_type:
             invalid_tissues[sample] = tissue_type
 
-    if invalid_tissues:
-        mismatch = ', '.join([
-            f'{sample.sample_id} ({REVERSE_TISSUE_TYPE[expected_tissue]} to {REVERSE_TISSUE_TYPE[sample.tissue_type]})'
-            for sample, expected_tissue in invalid_tissues.items()])
-        message = f'Skipped data loading for the following {len(invalid_tissues)} samples due to mismatched tissue type: {mismatch}'
-        warnings.append(message)
```
ShifaSZ, line 358: The `invalid` tissue samples will be filtered out. So no warning messages are required.

seqr/views/utils/dataset_utils.py
```diff
@@ -320,43 +323,29 @@ def _parse_tsv_row(row):
...
+def _parse_tpm_row(row):
     sample_id = row[SAMPLE_ID_COL]
     if row[TPM_COL] != '0.0' and not sample_id.startswith('GTEX'):
-        prev_tissue = sample_id_to_tissue_type.get(sample_id)
         tissue = row[TISSUE_COL]
         if not tissue:
             raise ValueError(f'Sample {sample_id} has no tissue type')
-        if prev_tissue and prev_tissue != tissue:
-            raise ValueError(f'Mismatched tissue types for sample {sample_id}: {prev_tissue}, {tissue}')
-        sample_id_to_tissue_type[sample_id] = tissue
```
ShifaSZ, line 335: Move the creation of sample id to tissue type mapping to the primary parsing process because all RNA data types will need it.

seqr/views/utils/dataset_utils.py
```diff
@@ -385,9 +371,18 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
     for line in tqdm(f, unit=' rows'):
         row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in parse_row(row, sample_id_to_tissue_type=sample_id_to_tissue_type):
+        for sample_id, row_dict in parse_row(row):
+            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None), None)
+            if sample_id in sample_id_to_tissue_type:
+                prev_tissue_type = sample_id_to_tissue_type[sample_id]
+                if tissue_type != prev_tissue_type:
+                    raise ValueError(f'Mismatched tissue types for sample {sample_id}:'
+                                     f' {REVERSE_TISSUE_TYPE[prev_tissue_type]}, {REVERSE_TISSUE_TYPE[tissue_type]}')
```
ShifaSZ, line 383: We allow only one tissue type per sample in each input data file.

seqr/views/utils/dataset_utils.py
```diff
@@ -432,10 +426,11 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
         sample: samples_by_id[sample.sample_id] for sample in samples if sample.id not in loaded_sample_ids
     }
 
+    filters = {'filter': ~Q(family__individual__id__in=prev_loaded_individual_ids)} if prev_loaded_individual_ids else {}
     sample_projects = Project.objects.filter(family__individual__sample__in=samples_to_load.keys()).values(
         'guid', 'name', new_sample_ids=ArrayAgg(
             'family__individual__sample__sample_id', distinct=True, ordering='family__individual__sample__sample_id',
-            filter=~Q(family__individual__id__in=prev_loaded_individual_ids)
+            **filters,
```
ShifaSZ, line 433: The reason for the change is that it fails when `prev_loaded_individual_ids` is empty.

seqr/views/apis/individual_api.py
```diff
@@ -754,7 +754,7 @@ def save_individuals_metadata_table_handler(request, project_guid, upload_file_i
 def get_individual_rna_seq_data(request, individual_guid):
     individual = Individual.objects.get(guid=individual_guid)
     check_project_permissions(individual.family.project, request.user)
-    sample = Sample.objects.get(individual=individual, is_active=True, sample_type=Sample.SAMPLE_TYPE_RNA)
+    sample = Sample.objects.get(individual=individual, is_active=True, sample_type=Sample.SAMPLE_TYPE_RNA, tissue_type=None)
```
hanars, line 757: this is not a stable way of determining if a sample has outrider data or not. There is a chance in the future we would add other RNA data without a tissue type, or would start adding tissue type to outlier data. It would be better to skip the query for the sample altogether and just directly query the `RnaSeqOutlier` models instead - `RnaSeqOutlier.objects.filter(sample__individual=individual, sample__is_active=True, sample__sample_type=Sample.SAMPLE_TYPE_RNA)`

seqr/views/utils/dataset_utils.py
hanars: if so, then theres no need to keep a dictionary of a sample to tissue type mapping. It would be much neater to keep a single `tissue_type` variable, validate all the samples in the file have that tissue type, and then pass that tissue type directly

seqr/views/utils/dataset_utils.py
hanars: "not in empty set" should just return True for all the queries so I don't see why this change is necessary

hanars (changes requested):
I have a high level question - if we have splice junction outlier data and tpm data in the same tissue type, would those share a sample model or no? In other words, is it one sample per tissue per RNA data type or one sample per tissue?

seqr/views/utils/dataset_utils.py
ShifaSZ: I thought that different samples could have different tissue types. The example data file that Ben provided has a single tissue type. Maybe we should conform with Ben if it is always the case.

hanars: Your comment here is "We allow only one tissue type per sample in each input data file." You are currently failing the file if any sample has a different tissue type. So either you should clean up the code if this is the case, or you need to completely change this validation

seqr/views/utils/dataset_utils.py
ShifaSZ: When `prev_loaded_individual_ids` is empty and `samples_to_load` is non-empty, there will be an error with the message below.
```
"message": "syntax error at or near \")\"\nLINE 1: ...ORDER BY \"seqr_sample\".\"sample_id\") FILTER (WHERE ) AS \"new_...
"user": "test_data_manager@test.com",
"traceback": "Traceback (most recent call last):
  File \"/Users/shifa/opt/anaconda2/envs/mito/lib/python3.7/site-packages/django/db/backends/utils.py\", line 84, in _execute
    return self.cursor.execute(sql, params)\npsycopg2.errors.SyntaxError: syntax error at or near \")\"
LINE 1: ...ORDER BY \"seqr_sample\".\"sample_id\") FILTER (WHERE ) AS \"new_...

The above exception was the direct cause of the following exception: ...
```

ShifaSZ: I've just added a test for `loading new data without existing data`. If we don't change the code, the new test will fail.

seqr/views/utils/dataset_utils.py
hanars: I realize I misread this - we are allowing one tissue type per sample, not one tissue type per file

seqr/views/utils/dataset_utils.py
hanars: but we need to continue to warn the user that these are filtered out, that behavior can't change

seqr/views/utils/dataset_utils.py
```diff
@@ -385,9 +374,18 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+    sample_id_to_tissue_type = {}
     for line in tqdm(f, unit=' rows'):
         row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in parse_row(row, sample_id_to_tissue_type=sample_id_to_tissue_type):
+        for sample_id, row_dict in parse_row(row):
+            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None), None)
+            if sample_id in sample_id_to_tissue_type:
+                prev_tissue_type = sample_id_to_tissue_type[sample_id]
+                if tissue_type != prev_tissue_type:
+                    raise ValueError(f'Mismatched tissue types for sample {sample_id}:'
```
hanars, line 385: this needs to maintain previous behavior of skipping the sample, continuing to load the rest of the file, and showing the user a warning saying which samples were skipped

seqr/views/utils/dataset_utils.py
```diff
@@ -417,12 +416,11 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
     )
 
     warnings = []
-    if validate_samples:
-        samples = validate_samples(samples, sample_id_to_tissue_type, warnings)
 
     # Delete old data
     individual_db_ids = {s.individual_id for s in samples}
-    to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
+    filters = {'sample__individual_id__in': individual_db_ids} if model_cls == RnaSeqOutlier else {'sample__in': samples}
```
hanars, line 422: why would the behavior be different for different classes?

seqr/views/utils/dataset_utils.py
```diff
@@ -125,6 +126,7 @@ def match_sample_ids_to_sample_records(
         loaded_date (object): datetime object
         raise_no_match_error (bool): whether to raise an exception if no sample matches are found
         raise_unmatched_error_template (string): optional template to use to raise an exception if samples are unmatched, will not raise if not provided
+        sample_id_to_tissue_type (object): a mapping from sample ids to tissue types
```
hanars, line 129: I think it would actually be better (and more in keeping with this function) to do one call to this function per tissue type, and explicitly pass in one tissue type here that will be passed to `_find_matching_sample_records` and sample creation the same way as all the other arguments.

Technically this is less performant becuase its one database query per tissue type instead of one total, but it will make the ongoing maintenance of this function and buggyness of htis code much better, and the performance will still be acceptable

hanars (changes requested):

seqr/views/utils/dataset_utils.py
ShifaSZ: Oh, yes. I should have added comments for this. I was thinking about the existing outlier data loaded to the sample models with tissue types. If we handle them like TPM data, these existing outlier data will be retained when loading new data.
We could manually delete the existing data or make the code backward compatible. What do you think?

seqr/views/utils/dataset_utils.py
ShifaSZ: It doesn't filter out any data. It filters out sample models that don't match the tissues in the input file. New sample models will be created for all the data with mismatched tissue types.

seqr/views/apis/individual_api.py
```diff
@@ -789,10 +789,10 @@ def save_individuals_metadata_table_handler(request, project_guid, upload_file_i
 def get_individual_rna_seq_data(request, individual_guid):
     individual = Individual.objects.get(guid=individual_guid)
     check_project_permissions(individual.family.project, request.user)
-    sample = Sample.objects.get(individual=individual, is_active=True, sample_type=Sample.SAMPLE_TYPE_RNA)
+    outlier_data = RnaSeqOutlier.objects.filter(sample__individual=individual, sample__is_active=True)
```
ShifaSZ, line 792: The change is because the `Sample.objects.get()` fails for multiple matched objects. We could add tissue-type conditions, but it is not general as the new code.

seqr/views/utils/dataset_utils.py
```diff
@@ -285,15 +289,16 @@ def _get_unmatched_error(sample_keys):
...
         raise_no_match_error=False,
         get_unmatched_error=_get_unmatched_error if raise_unmatched_error_template else None,
         create_active=True,
         get_individual_sample_lookup=lambda inds: {(i.individual_id, i.family.project.name):
                                                        i for i in inds.select_related('family__project')},
+        sample_id_to_tissue_type=sample_id_to_tissue_type,
+        data_source=data_source,
+        sample_type=Sample.SAMPLE_TYPE_RNA,
+        dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS,
+        loaded_date=timezone.now(),
```
ShifaSZ, line 301: Re-order for the `kwargs`.

seqr/views/utils/dataset_utils.py
```diff
@@ -321,60 +326,34 @@ def _parse_tsv_row(row):
 
 REVERSE_TISSUE_TYPE = {v: k for k, v in TISSUE_TYPE_MAP.items()}
 
-def _parse_outlier_row(row, **kwargs):
+def _parse_outlier_row(row):
     yield row['sampleID'], {mapped_key: row[key] for key, mapped_key in RNA_OUTLIER_COLUMNS.items()}
 
-def _parse_tpm_row(row, sample_id_to_tissue_type=None):
+def _parse_tpm_row(row):
```
ShifaSZ, line 332: All RNAseq data will need a tissue-type field in the parsed output. Therefore, we remove the `kwargs` arguments for the parser functions.

seqr/views/utils/dataset_utils.py
ShifaSZ: We no longer have the warning message because we will create a new sample model and load the data in the case of a mismatched tissue type.

seqr/views/utils/dataset_utils.py
```diff
@@ -321,60 +326,34 @@ def _parse_tsv_row(row):
...
     expected_columns = ['sampleID'] + list(RNA_OUTLIER_COLUMNS.keys())
     return _load_rna_seq(
         RnaSeqOutlier, file_path, user, mapping_file, ignore_extra_samples, _parse_outlier_row, expected_columns,
     )
 
 def load_rna_seq_tpm(file_path, user=None, mapping_file=None, ignore_extra_samples=False):
-    sample_id_to_tissue_type = {}
     return _load_rna_seq(
         RnaSeqTpm, file_path, user, mapping_file, ignore_extra_samples, _parse_tpm_row, TPM_HEADER_COLS,
-        sample_id_to_tissue_type=sample_id_to_tissue_type, validate_samples=_check_invalid_tissues,
```
ShifaSZ, line 373: Move all tissue type-specific arguments into the `_load_rna_seq()` to be shared for all RNAseq data.

seqr/views/utils/dataset_utils.py
ShifaSZ: Sounds good.

seqr/views/utils/dataset_utils.py
```diff
@@ -413,13 +404,10 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+        sample_id_to_tissue_type=sample_id_to_tissue_type,
     )
 
-    warnings = []
-    if validate_samples:
-        samples = validate_samples(samples, sample_id_to_tissue_type, warnings)
-
     # Delete old data
     to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
```
ShifaSZ, line 413: It is still a problem for the `data_source` attribute of the sample. We don't update the attribute for the existing sample.

seqr/views/utils/dataset_utils.py
```diff
@@ -155,6 +156,7 @@ def _find_or_create_missing_sample_records(
                 individual=individual,
                 created_date=timezone.now(),
                 is_active=create_active,
+                tissue_type=sample_id_to_tissue_type.get(sample_key[0], None) if sample_id_to_tissue_type else None,
```
hanars, line 159: this needs to be by sample key not by sample ID, as otherwise we will get incorrect behavior for the same sample ID in multiple projects

seqr/views/utils/dataset_utils.py
```diff
@@ -276,7 +279,8 @@ def _get_unmatched_error(sample_keys):
         sample_id__in={sample_id for sample_id, _ in sample_project_tuples},
     )
 
-    samples = [s for s in samples if (s.sample_id, s.individual.family.project.name) in sample_project_tuples]
+    samples = [s for s in samples if (s.sample_id, s.individual.family.project.name) in sample_project_tuples and
+               sample_id_to_tissue_type[s.sample_id] == s.tissue_type]
```
hanars, line 283: its not necessary, but you could also add an additional filter above for `tissue_type__in= sample_id_to_tissue_type.values()` to get back fewer incorrect samples

seqr/views/utils/dataset_utils.py
hanars: I thought we still are only supporting one tissue type per sample per file. What if there are multiple tissue types for the same sample in the file?

seqr/views/utils/dataset_utils.py
hanars: I don't 100% follow - are you saying that if we had previously loaded data for a specific tissue type we are not updating the data source when we reload it? Because if so that seems like a pretty straightforward change to make

hanars (changes requested):

seqr/views/utils/dataset_utils.py
ShifaSZ: Suppose we have previously loaded samples from a data source in the DB when we reload data for the samples from a different data source. In that case, we delete the existing data from the previous data source before loading the new data. But we don't update the data source of the sample model to the latest data source. Do you think it is an issue?

seqr/views/utils/dataset_utils.py
ShifaSZ: Can we assume the samples with the same sample ID for different projects are identical in the real world? It seems reasonable because the samples with the same sample ID look like the same sample but have been loaded and analyzed by multiple projects.

seqr/views/utils/dataset_utils.py
ShifaSZ: I can add the tissue type condition to the filter to reduce incorrect samples. But we may still need the condition here because the filter above can get extra samples with different combinations of sample IDs and tissue types.

seqr/views/utils/dataset_utils.py
hanars: yes I think we should do a bulk update of the sample models to the latest data source

seqr/views/utils/dataset_utils.py
hanars: No this is not acceptable

seqr/views/utils/dataset_utils.py
hanars: we will still need the condition here, correct

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -637,46 +640,54 @@ def test_kibana_proxy(self):
...
             'message_data_type': 'Expression',
             'header': ['sample_id', 'project', 'gene_id', 'individual_id', 'tissue', 'TPM'],
             'optional_headers': ['individual_id'],
-            'loaded_data_row': ['NA19675_D2', '1kg project nåme with uniçøde', 'ENSG00000135953', 'NA19675_D3', 'muscle', 1.34],
+            'loaded_data_row': ['NA19675_D2', '1kg project nåme with uniçøde', 'ENSG00000135953', '', 'muscle', 1.34],
             'no_existing_data': ['NA19678', '1kg project nåme with uniçøde', 'ENSG00000233750', 'NA19678', 'fibroblasts', 0.064],
-            'reused_indiv_id_data': ['NA20870', 'Test Reprocessed Project', 'ENSG00000233750', 'NA20870', 'fibroblasts', 0.064],
+            'duplicated_indiv_id_data': [
+                ['NA20870', 'Test Reprocessed Project', 'ENSG00000233750', 'NA20870', 'fibroblasts', 0.01],
+                ['NA20870', '1kg project nåme with uniçøde', 'ENSG00000233750', 'NA20870', 'muscle', 1.34],
```
ShifaSZ, line 655: Test duplicated sample/individual IDs and different tissues in 2 projects.

seqr/views/utils/dataset_utils.py
ShifaSZ: We don't allow multiple tissue types for the same sample in a project. It makes sense because a sample usually is not possible from various tissues.

seqr/views/utils/dataset_utils.py
```diff
@@ -386,11 +367,24 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None), None)
             project = row_dict.pop(PROJECT_COL)
+            if (sample_id, project) in sample_id_to_tissue_type:
+                prev_tissue_type = sample_id_to_tissue_type[(sample_id, project)]
+                if tissue_type != prev_tissue_type:
+                    warnings.append(f'Skipped loading row with mismatched tissue types for sample {sample_id}:'
+                                    f' {REVERSE_TISSUE_TYPE[prev_tissue_type]}, {REVERSE_TISSUE_TYPE[tissue_type]}')
+                    continue
+
+            sample_id_to_tissue_type[(sample_id, project)] = tissue_type
```
ShifaSZ, line 385: Use sample id and project tuples for the keys of the tissue type mapping.

seqr/views/utils/dataset_utils.py
ShifaSZ: This issue is still open for discussion. Can a sample ID have multiple tissue types?

seqr/views/apis/dataset_api_tests.py
```diff
@@ -65,7 +65,8 @@ def test_add_variants_dataset(self, mock_send_email, mock_send_slack, mock_rando
         self.assertEqual(existing_sample.elasticsearch_index, INDEX_NAME)
         self.assertFalse(existing_sample.is_active)
         existing_sample_guid = existing_sample.guid
-        existing_rna_seq_sample_guid = Sample.objects.get(sample_id='NA19675_D2', sample_type='RNA').guid
+        existing_rna_seq_samples = Sample.objects.filter(sample_id='NA19675_D2', sample_type='RNA')
+        existing_rna_seq_sample_guids = {sample.guid for sample in existing_rna_seq_samples}
```
hanars, line 69: would be cleaner as 
```
existing_rna_seq_sample_guids = Sample.objects.filter(
sample_id='NA19675_D2', sample_type='RNA').values_list('guid', flat=True)
```

seqr/views/apis/dataset_api_tests.py
```diff
@@ -231,7 +233,8 @@ def test_add_variants_dataset(self, mock_send_email, mock_send_slack, mock_rando
         self.assertListEqual(list(response_json['individualsByGuid'].keys()), ['I000001_na19675'])
         self.assertListEqual(list(response_json['individualsByGuid']['I000001_na19675'].keys()), ['sampleGuids'])
         self.assertSetEqual(set(response_json['individualsByGuid']['I000001_na19675']['sampleGuids']),
-                            {sv_sample_guid, existing_index_sample_guid, new_sample_type_sample_guid, existing_rna_seq_sample_guid})
+                            {sv_sample_guid, existing_index_sample_guid, new_sample_type_sample_guid} |
+                            existing_rna_seq_sample_guids)
```
hanars, line 237: I think we do want to be more careful about testing if some of the new guids we care about are also in `existing_rna_seq_sample_guids`

seqr/views/utils/dataset_utils.py
```diff
@@ -265,18 +266,23 @@ def _get_unmatched_error(remaining_sample_ids):
...
 
-    samples = Sample.objects.select_related('individual__family__project').filter(
+    query = Q(
         individual__family__project__in=projects,
         sample_type=Sample.SAMPLE_TYPE_RNA,
         dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS,
         sample_id__in={sample_id for sample_id, _ in sample_project_tuples},
     )
+    tissues = set(sample_id_to_tissue_type.values())
+    query = query & Q(tissue_type__isnull=True) if None in tissues else query & Q(tissue_type__in=tissues)
```
hanars, line 281: This is neater as ` query &= Q(tissue_type__isnull=True) if None in tissues else Q(tissue_type__in=tissues)`

Also, I think it would be safer if the conditional was `tissues == {None}` to make sure you aren't in a weird edge case where one of the values is None

hanars (changes requested):

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -77,8 +73,8 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
         self.assertEqual(new_sample.tissue_type, 'WB')
 
         models = RnaSeqTpm.objects.all()
-        self.assertEqual(models.count(), 4)
-        self.assertSetEqual({model.sample for model in models}, {existing_sample, existing_sample2, new_sample})
+        self.assertEqual(models.count(), 6)
+        self.assertSetEqual({model.sample for model in models}, {*list(existing_rna_samples), new_sample})
```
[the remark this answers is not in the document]
ShifaSZ: Oh, yes, `existing_rna_samples` include `new_sample`. The test should still have value. I will update and keep it.

seqr/views/apis/dataset_api_tests.py
ShifaSZ: No new guids can be in `existing_rna_seq_sample_guids`, correct? If yes, I can add a test for it.

seqr/views/utils/dataset_utils.py
```diff
@@ -386,11 +367,24 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
 
+    warnings = []
+
+    sample_id_to_tissue_type = {}
     for line in tqdm(f, unit=' rows'):
         row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in parse_row(row, sample_id_to_tissue_type=sample_id_to_tissue_type):
-            gene_id = row_dict['gene_id']
+        for sample_id, row_dict in parse_row(row):
+            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None), None)
```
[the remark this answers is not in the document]
ShifaSZ: Okay, I was confused by the differences between `.get()` and `.pop()`.

seqr/views/utils/dataset_utils.py
```diff
@@ -386,11 +367,24 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
             project = row_dict.pop(PROJECT_COL)
+            if (sample_id, project) in sample_id_to_tissue_type:
+                prev_tissue_type = sample_id_to_tissue_type[(sample_id, project)]
+                if tissue_type != prev_tissue_type:
+                    warnings.append(f'Skipped loading row with mismatched tissue types for sample {sample_id}:'
+                                    f' {REVERSE_TISSUE_TYPE[prev_tissue_type]}, {REVERSE_TISSUE_TYPE[tissue_type]}')
+                    continue
```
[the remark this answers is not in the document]
ShifaSZ: If we want to load none of the data with conflict tissues, we must remove all data for the samples with conflict tissues. Is it correct?

seqr/views/utils/dataset_utils.py
```diff
@@ -413,19 +407,20 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
-
     # Delete old data
     to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
     if to_delete:
         model_cls.bulk_delete(user, to_delete)
 
+    samples_to_update = [s.id for s in samples if s.data_source != data_source]
+    if samples_to_update:
+        Sample.bulk_update(user, {'data_source': data_source}, id__in=samples_to_update)
```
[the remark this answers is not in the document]
ShifaSZ: Can some of the existing samples have the same data source while loading the same data file multiple times while fixing errors in the file?

seqr/views/utils/dataset_utils.py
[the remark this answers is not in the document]
hanars: correct

seqr/views/utils/dataset_utils.py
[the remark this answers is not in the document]
hanars: its possible but its a pretty rare edge case and in that case doing an extra bulk update on those samples wouldn't hurt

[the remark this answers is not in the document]
ShifaSZ: The `samples` for the existing samples is not a `queryset` because we remove project and tissue mismatched models from the queryset. The easiest way to create a queryset from the `samples` is to filter again with the `id__in={s.id for s in samples}` condition.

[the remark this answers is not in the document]
hanars: ah I forgot about that, good catch

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -637,40 +645,57 @@ def test_kibana_proxy(self):
...
                 ['NA19675_D3', '1kg project nåme with uniçøde', 'ENSG00000233750', 'NA19675_D3', 'fibroblasts', 0.064],
+                # skip GTEX samples
                 ['GTEX_001', '1kg project nåme with uniçøde', 'ENSG00000233750', 'NA19675_D3', 'whole_blood', 1.95],
+                # a different project sample NA20888
                 ['NA20888', 'Test Reprocessed Project', 'ENSG00000240361', 'NA20888', 'fibroblasts', 0.112],
+                # a project mismatched sample NA20878
                 ['NA20878', 'Test Reprocessed Project', 'ENSG00000233750', 'NA20878', 'fibroblasts', 0.064],
+                # conflict tissue types samples
+                ['NA19678', '1kg project nåme with uniçøde', 'ENSG00000233750', 'NA19678', 'muscle', 1.34],
+                ['NA19678', '1kg project nåme with uniçøde', 'ENSG00000135954', 'NA19678', 'fibroblasts', 0.05],
```
ShifaSZ, line 679: All conflict tissue types samples won't be loaded.

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -77,21 +73,26 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
         self.assertEqual(new_sample.tissue_type, 'WB')
 
         models = RnaSeqTpm.objects.all()
-        self.assertEqual(models.count(), 4)
-        self.assertSetEqual({model.sample for model in models}, {existing_sample, existing_sample2, new_sample})
-        self.assertEqual(models.get(sample=existing_sample, gene_id='ENSG00000240361').tpm, 12.6)
+        self.assertEqual(models.count(), 5)
+        self.assertSetEqual({model.sample for model in models}, set(existing_rna_samples))
+        self.assertEqual(len(models.filter(sample=existing_sample, gene_id='ENSG00000240361')), 0)
```
hanars, line 78: nitpicky, but `models.filter().count()` is a bit better than `len(model.filters())`. It mean django executes the more efficient COUNT query on the database and never actually pulls out the records

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -77,21 +73,26 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
...
-        # Test fails on mismatched tissue
-        mock_gzip_file.__iter__.return_value.pop()
-        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t6.04\tfibroblasts\n'
+        # Test a new sample created for a mismatched tissue
+        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\tNA19678\tENSG00000233750\t6.55\tfibroblasts\n'
         call_command('load_rna_seq_tpm', 'new_file.tsv.gz', '--ignore-extra-samples')
-        mock_utils_logger.warning.assert_any_call(EXPECTED_MISMATCHED_TISSUE_WARNING, None)
+        models = RnaSeqTpm.objects.filter(sample__sample_id='NA19678_D1')
+        self.assertEqual(models.count(), 2)
+        self.assertEqual(models.get(gene_id='ENSG00000233750', sample__tissue_type='F').tpm, 6.55)
```
hanars, line 94: lets also add a test here that the 2 samples have the 2 different tissue types we expect

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -763,76 +788,85 @@ def _set_file_iter_stdout(rows):
...
                 mock_open.return_value.__enter__.return_value.write.side_effect = mock_write
                 body.update({'ignoreExtraSamples': True, 'mappingFile': {'uploadedFileId': 'map.tsv'}, 'file': RNA_FILE_ID})
                 response_json, new_sample_guid = _test_basic_data_loading(
-                    params['new_data'], params["num_parsed_samples"], 2,
-                    '1kg project nåme with uniçøde, Test Reprocessed Project', 2, 16, 1)
-                self.assertTrue(RNA_SAMPLE_GUID in response_json['sampleGuids'])
-                warnings = [f'Skipped loading for the following {len(params["skipped_samples"].split(","))} '
-                            f'unmatched samples: {params["skipped_samples"]}']
-                if params.get('extra_warnings'):
-                    warnings = params['extra_warnings'] + warnings
```
hanars, line 814: it seems like the previous behvaior of specifying the skipped samples and the extra_warnings would still work. Is there a reason you changed this?

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: The `extra_warnings` are generated when the data mismatch tissue types with the existing sample. We want to create new samples for that case; no data will be skipped, so no warning is required.

hanars: that makes sense. Why change `skipped_samples` to `earnings` though?

ShifaSZ: We could use `skipped_samples` to generate the warning messages. But when there are different numbers of warning messages, it is getting harder to generate them. Moving the warning messages into the `params` works better than generating them.

hanars: I really just don't understand why you made this change at all though. Before, Outlier had 1 warning about skipped data generated from skipped_samples and TPM had 2 warnings, one generated by skipped_samples and 1 in extra_warnings. We still have that exact same pattern of warnings, why not just change the value of extra_warnings for the warning that changed instead of completely redoing this part of the code?

seqr/views/utils/dataset_utils.py
```diff
@@ -386,11 +368,26 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
         row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in parse_row(row, sample_id_to_tissue_type=sample_id_to_tissue_type):
-            gene_id = row_dict['gene_id']
+        for sample_id, row_dict in parse_row(row):
+            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None))
             project = row_dict.pop(PROJECT_COL)
+            if (sample_id, project) in sample_id_to_tissue_type:
+                prev_tissue_type = sample_id_to_tissue_type[(sample_id, project)]
+                if tissue_type != prev_tissue_type:
+                    warnings.append(f'Skipped loading row with mismatched tissue types for sample {sample_id}:'
```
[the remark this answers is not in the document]
hanars: we should not use the word "row" here, we are skipping tons of rows. We should have 1 warning that is formatted liek the existing warning for skipping unmatched samples:
```
f'Skipped loading for the following 2 unmatched samples: NA19677 (muscle, fibroblast), NA19678(fibroblast, whole blood)'
```

ShifaSZ:
Bulk updating of the data source is still to-be-done.

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -77,21 +71,28 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
...
-        # Test fails on mismatched tissue
-        mock_gzip_file.__iter__.return_value.pop()
-        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t6.04\tfibroblasts\n'
+        # Test a new sample created for a mismatched tissue
+        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\tNA19678\tENSG00000233750\t6.55\tfibroblasts\n'
         call_command('load_rna_seq_tpm', 'new_file.tsv.gz', '--ignore-extra-samples')
-        mock_utils_logger.warning.assert_any_call(EXPECTED_MISMATCHED_TISSUE_WARNING, None)
+        models = RnaSeqTpm.objects.select_related('sample').filter(sample__sample_id='NA19678_D1')
+        self.assertEqual(models.count(), 2)
+        self.assertSetEqual({v[0] for v in models.values_list('sample__tissue_type').distinct()}, {'F', 'WB'})
```
hanars, line 93: would be neater as `set(models.values_list('sample__tissue_type', flat=True))`

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -851,11 +891,16 @@ def mock_write(content):
...
 
                 # Test loading data when where are duplicated individual ids in different projects.
-                data = [params['reused_indiv_id_data']]
-                _test_basic_data_loading(data, 1, 1, 'Test Reprocessed Project', 1, 20, 0)
+                data = params['duplicated_indiv_id_data']
+                mock_writes = []
+                _test_basic_data_loading(data, 2, 2, 20, body,
+                                         project_names='1kg project nåme with uniçøde, Test Reprocessed Project',
+                                         num_created_samples=2)
+                self.assertSetEqual({re.sub(r'^S[0-9]*', 'S', s) for s in mock_writes}, params['write_data'])
```
hanars, line 902: would be neater to mock sample guid creation instead, but this is okay

seqr/views/utils/dataset_utils.py
```diff
@@ -401,30 +395,41 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
+    for project, sample in samples_with_conflict_tissues.items():
+        messages = []
+        for sample_id, tissue_types in sample.items():
+            sample_id_to_tissue_type.pop((sample_id, project))
+            samples_by_id.pop((sample_id, project))
+            messages.append(f'{sample_id} ({", ".join(sorted([REVERSE_TISSUE_TYPE[tissue_type] for tissue_type in tissue_types]))})')
+        conflict_messages.append(f'{", ".join(messages)} in the project {project}')
+    num_samples = sum([len(v) for v in samples_with_conflict_tissues.values()])
+    warnings = [f'Skipped loading for the following {num_samples} tissue type unmatched sample(s) in '
+                f'{len(samples_with_conflict_tissues)} project(s): {", ".join(conflict_messages)}'] if num_samples else []
```
hanars, line 408: grouping samples by project is much less readable to the end user than a clear list of samples, like we curretly have in our existing warnings. In the event a sample ID is included multiple times, the user will be smart enough to figure out which one is the issue. Just make this warning look like the warning we show for unmatched samples, per my previous review

seqr/views/utils/dataset_utils.py
```diff
@@ -401,30 +395,41 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
-    if validate_samples:
-        samples = validate_samples(samples, sample_id_to_tissue_type, warnings)
-
     # Delete old data
     to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
     if to_delete:
         model_cls.bulk_delete(user, to_delete)
 
+    Sample.bulk_update(user, {'data_source': data_source}, id__in={s.id for s in existing_samples})
```
hanars, line 431: I thought we discussed filtering this so it would not update samples that already had a matched data source?

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: It must be doable.

seqr/models.py
```diff
@@ -135,7 +135,9 @@ def bulk_create(cls, user, new_models):
     def bulk_update(cls, user, update_json, queryset=None, **filter_kwargs):
         """Helper bulk update method that logs the update"""
         if queryset is None:
-            queryset = cls.objects.filter(**filter_kwargs)
+            queryset = cls.objects.filter(**filter_kwargs).exclude(**update_json)
+            if not queryset:
+                return []
```
ShifaSZ, line 140: Stop updating the model if `update_json` doesn't change any data.

seqr/views/utils/dataset_utils.py
ShifaSZ: I updated the `bulk_update` function in the `models.py`. See this [comment](https://github.com/broadinstitute/seqr/pull/3243/files#r1154823791).

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -642,27 +651,42 @@ def test_kibana_proxy(self):
...
             ],
             'skipped_samples': 'NA19675_D3, NA20878',
+            'exist_sample_tissue_type': 'M',
             'created_sample_tissue_type': 'F',
-            'num_parsed_samples': 5,
-            'initial_model_count': 3,
-            'deleted_count': 2,
             'extra_warnings': [
-                'Skipped data loading for the following 1 samples due to mismatched tissue type: NA20889 (fibroblasts to muscle)'],
+                'Skipped loading for the following 2 tissue-unmatched sample(s): NA19678 (fibroblasts, muscle)',
```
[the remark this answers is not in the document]
ShifaSZ: Yes, it is a bug. It counts the tissues rather than the samples. I've fixed it.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -774,15 +798,18 @@ def _set_file_iter_stdout(rows):
...
-                def _test_basic_data_loading(data, parsed_samples, loaded_samples, project_names, projects,
-                                             individual_id, sample_guid_idx, warnings=None, additional_logs=None):
+                def _test_basic_data_loading(data, num_parsed_samples, num_loaded_samples, individual_id, body,
+                                             project_names=None, num_created_samples=1, warnings=None, additional_logs=None):
                     self.reset_logs()
                     _set_file_iter_stdout([header] + data)
                     response = self.client.post(url, content_type='application/json', data=json.dumps(body))
                     self.assertEqual(response.status_code, 200)
+                    if not project_names:
+                        project_names = Individual.objects.filter(id=individual_id).values('family__project__name')[0]['family__project__name']
```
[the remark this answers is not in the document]
ShifaSZ: Okay

hanars (approved):

PR #3245, opened 2023-02-24
fix family individual guids

hanars:


seqr/views/apis/project_api_tests.py
```diff
@@ -334,6 +334,8 @@ def test_project_families(self):
         family_fields.update(FAMILY_FIELDS)
         self.assertSetEqual(set(family_1.keys()), family_fields)
 
+        self.assertEqual(len(family_1['individualGuids']), 3)
+        self.assertEqual(len(family_1['individualGuids']), 3)
```
ShifaSZ, line 338: Duplicated?

ShifaSZ (changes requested):

seqr/views/apis/project_api_tests.py
hanars: yes one of those should be testing family 3

ShifaSZ (approved):

PR #3251, opened 2023-02-28
Unsolved case search

hanars:


ui/pages/Search/components/filters/ProjectFamiliesFilter.jsx
```diff
@@ -27,15 +38,28 @@ class ProjectFamiliesFilterInput extends React.PureComponent {
...
+    }
+    return null
+  }
+
+  unsolvedFamilyGuids = () => {
+    const { familyOptions } = this.props
+    return familyOptions.filter(
+      ({ analysisStatus }) => !SOLVED_FAMILY_STATUS_OPTIONS.has(analysisStatus),
+    ).map((({ value }) => value))
   }
```
ShifaSZ, line 57: It is neater for the function takes in the `familyOptions` parameter rather than get it from `this.props`. And it can also return a sorted and comma-joined string.

ShifaSZ (changes requested):

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -517,6 +513,19 @@ def _get_variants(results_model, **kwargs):
...
 
+        result_model.delete()
+        expected_searched_families.remove('F000007_7')
+        expected_searched_families.remove('F000010_10')
+        body['unsolvedFamiliesOnly'] = True
+        response = self.client.post(url, content_type='application/json', data=json.dumps(body))
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        self.assertSetEqual(set(response_json.keys()), set(EXPECTED_SEARCH_RESPONSE.keys()))
+        self.assertDictEqual(response_json, EXPECTED_SEARCH_RESPONSE)
```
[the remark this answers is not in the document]
hanars: this is copy-pasted from everywhere else in this test so I am leaving it as-is for clarity/consistency. The expected response has a `mock.ANY` for the families so the way this test is structured is you test that the response has the overall expected structure and then in each case you explicitly update and test the expected families

ui/pages/Search/components/filters/ProjectFamiliesFilter.jsx
hanars: That threw an error with our linter, I tried that. And it can't return a string, the value of the field that is sent to the server is a list

ShifaSZ (approved):

PR #3254, opened 2023-02-28
BCH aggregator link

hanars:


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -279,6 +287,7 @@ class BaseSearchLinks extends React.PureComponent {
 
 const mapStateToProps = state => ({
   genesById: getGenesById(state),
+  user: getUser(state),
```
ShifaSZ, line 290: Is it better to pass in `isAnalystUser` rather than the whole `user` object?

hanars: In this particular instance maybe but this is much more future proof, as it means that we can make any link easily showable/hideable based on any user field. The user in the state is a single object that is immutable so from a re-rendering perspective theres no case where the object would change

ShifaSZ (approved):

PR #3256, opened 2023-02-28
External analysis - fix 504

hanars:
For 20,000+ row file, speeds up response time from 80 seconds to 10 seconds

### DIFF
--- a/seqr/views/apis/summary_data_api.py
+++ b/seqr/views/apis/summary_data_api.py
@@ -113,23 +113,28 @@ def bulk_update_family_analysed_by(request):
         return create_json_response({'error': 'Project and Family columns are required'}, status=400)
     families_data = [dict(zip(header, row)) for row in family_upload_data[1:]]
 
-    family_qs = [Q(family_id=row['family'], project__name=row['project']) for row in families_data]
-    family_filter_q = family_qs[0]
-    for f_q in family_qs[1:]:
-        family_filter_q |= f_q
-    families = Family.objects.filter(family_filter_q)
+    projects = set()
+    families = set()
+    requested_families = set()
+    for row in families_data:
+        projects.add(row['project'])
+        families.add(row['family'])
+        requested_families.add((row['project'], row['family']))
+
+    family_db_id_lookup = {
+        (f['project__name'], f['family_id']): f['id'] for f in Family.objects.filter(
+            family_id__in=families, project__name__in=projects).values('id', 'family_id', 'project__name')
+    }
 
     warnings = []
-    if len(families) < len(families_data):
-        prefetch_related_objects(families, 'project')
-        family_models_set = {(f.family_id, f.project.name) for f in families}
-        requested_family_set = {(row['family'], row['project']) for row in families_data}
-        missing_families = ', '.join([f'{fam[0]} ({fam[1]})' for fam in sorted(requested_family_set - family_models_set)])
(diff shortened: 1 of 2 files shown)


seqr/views/apis/summary_data_api.py
```diff
@@ -113,29 +113,34 @@ def bulk_update_family_analysed_by(request):
...
+    families = set()
+    requested_families = set()
+    for row in families_data:
+        projects.add(row['project'])
+        families.add(row['family'])
+        requested_families.add((row['project'], row['family']))
+
+    family_db_id_lookup = {
+        (f['project__name'], f['family_id']): f['id'] for f in Family.objects.filter(
+            family_id__in=families, project__name__in=projects).values('id', 'family_id', 'project__name')
```
ShifaSZ, line 126: How about using `Concat` to match the `project name` and `family id` pairs?

hanars: I'd rather explicitly use tuples though, what is the reason why formatted string would be better?

seqr/views/apis/summary_data_api.py
```diff
@@ -113,29 +113,34 @@ def bulk_update_family_analysed_by(request):
...
+    missing_from_db = requested_families - set(family_db_id_lookup.keys())
+    if missing_from_db:
+        missing_families = ', '.join([f'{family} ({project})' for project, family in sorted(missing_from_db)])
         warnings.append(f'No match found for the following families: {missing_families}')
 
     analysed_by_models = [
-        FamilyAnalysedBy(family=family, data_type=data_type, last_modified_date=datetime.now())
-        for family in families
+        FamilyAnalysedBy(family_id=family_db_id_lookup[family_key], data_type=data_type, last_modified_date=datetime.now())
+        for family_key in requested_families if family_key in family_db_id_lookup
```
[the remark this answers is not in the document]
hanars: it does though? `family_key` is the project-family_id tuple

seqr/views/apis/summary_data_api.py
ShifaSZ: I mean using the Django `Concat` to connect two fields and filter out extra families. It makes the following code cleaner.

hanars: ohh you mean in the filter input not the values output. Yeah I like that!

seqr/views/apis/summary_data_api.py
```diff
@@ -113,29 +114,36 @@ def bulk_update_family_analysed_by(request):
...
-    for f_q in family_qs[1:]:
-        family_filter_q |= f_q
-    families = Family.objects.filter(family_filter_q)
+    projects = set()
+    families = set()
+    requested_families = set()
+    for row in families_data:
+        projects.add(row['project'])
+        families.add(row['family'])
+        requested_families.add((row['project'], row['family']))
```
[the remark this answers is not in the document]
hanars: No. Its more accurate to use tuples than formatted strings, the set substraction of the tuples is a much neater way of determining whats missing, for the missing message thats not how they are formatted so we would have to split and reformat the string, and also there is no disadvantage to the tuples

seqr/views/apis/summary_data_api.py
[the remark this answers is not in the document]
hanars: And of course there can be missing families, we literally have a missing handler right above this

ShifaSZ (approved):

Issue #3264, opened 2023-03-02
RNA-seq outlier data file


bw2:
Unable to load RNA-seq outlier data file:
gs://tgg-rnaseq/gagneur/outrider/results/latest_samples/fibroblasts__126_samples_7462C2E26C_without_GTEX__ods__q22_all_results.tsv

Getting this error:
<img width="804" alt="image" src="https://user-images.githubusercontent.com/6240170/222522265-a34581d3-c4d4-458a-a3d4-838c6476179e.png">

I'm still able to load this older data file with the same format though: 
gs://tgg-rnaseq/gagneur/outrider/results/latest_samples/whole_blood__110_samples_B25088DB1C_with_GTEX__ods__q20_all_results.tsv
So something about specific about the 1st file is causing an issue.

hanars:
Associated cloud reporting error: https://console.cloud.google.com/errors/detail/CPrJ8vj1veD55AE;time=P30D?project=seqr-project&utm_source=error-reporting-notification&utm_medium=email&utm_content=new-error

hanars:
This issue is fixed, although I get a readable "missing required column" error for that file now. Probably best to hold off on outlier loading until https://github.com/broadinstitute/seqr/issues/3265 can go live

PR #3269, opened 2023-03-03
Fix RNAseq data loading SQL syntax error.

ShifaSZ:
A fix for bug #3264.

### DIFF
--- a/seqr/views/apis/data_manager_api_tests.py
+++ b/seqr/views/apis/data_manager_api_tests.py
@@ -817,12 +817,32 @@ class DataManagerAPITest(AuthenticationTestCase):
                 self.assertListEqual(mock_writes, [row.replace(PLACEHOLDER_GUID, new_sample_guid) for row in params['parsed_file_data']])
 
                 # test loading new data without existing data
-                data = [params['new_data'][0]]
+                mock_logger.reset_mock()
+                data = [params['new_data'][3]]
                 data[0][0] = 'NA19678'  # load data for a new individual
                 _set_file_iter_stdout([header] + data)
                 body.pop('mappingFile')
                 response = self.client.post(url, content_type='application/json', data=json.dumps(body))
                 self.assertEqual(response.status_code, 200)
+                info = [
+                    f'Parsed 1 RNA-seq samples',
+                    'Attempted data loading for 1 RNA-seq samples in the following 1 projects: 1kg project nåme with uniçøde'
+                ]
+                response_json = response.json()
+                self.assertDictEqual(response_json, {'info': info, 'warnings': [], 'sampleGuids': mock.ANY, 'fileName': file_name})
+                new_sample_guid = self._check_rna_sample_model(
+                    individual_id=2, data_source='new_muscle_samples.tsv.gz', tissue_type=params.get('created_sample_tissue_type'),
+                )
+                self.assertListEqual(response_json['sampleGuids'], [new_sample_guid])
+                info_log_calls = [mock.call(info_log, self.data_manager_user) for info_log in info]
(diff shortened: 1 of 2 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -435,7 +435,7 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
     sample_projects = Project.objects.filter(family__individual__sample__in=samples_to_load.keys()).values(
         'guid', 'name', new_sample_ids=ArrayAgg(
             'family__individual__sample__sample_id', distinct=True, ordering='family__individual__sample__sample_id',
-            filter=~Q(family__individual__id__in=prev_loaded_individual_ids)
+            filter=~Q(family__individual__id__in=prev_loaded_individual_ids) if prev_loaded_individual_ids else None
```
ShifaSZ, line 438: When `prev_loaded_individual_ids` is empty, the SQL query generated by Django is incorrect (with an empty condition in the `where` clause).

hanars:
this needs a unit test that would have caught the error, so we don;t accidentally reintroduce this issue in the future

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -816,6 +816,14 @@ def mock_write(content):
                 mock_open.assert_called_with(file_name, 'wt')
                 self.assertListEqual(mock_writes, [row.replace(PLACEHOLDER_GUID, new_sample_guid) for row in params['parsed_file_data']])
 
+                # test loading new data without existing data
+                data = [params['new_data'][0]]
+                data[0][0] = 'NA19678'  # load data for a new individual
+                _set_file_iter_stdout([header] + data)
+                body.pop('mappingFile')
+                response = self.client.post(url, content_type='application/json', data=json.dumps(body))
+                self.assertEqual(response.status_code, 200)
```
hanars, line 825: You need to add tests that this is actually doing what we want, not just that it has a successful status code. Add a test for the `response_json` being what we expect and use `_check_rna_sample_model` to test the database mode is correct

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: When the bug happens, there is an uncaught exception. So the `response.status_code` will be 500. For testing the bug itself, it is enough to test the status code. But I will add the model check to ensure the change is correct.

hanars: tests are to test functionality for all changes going forward, not just one bug this one time. What if someone changed fixture data and this no longer was doing what we wanted it to do and was instead returning a 200 response with a warning saying everything in the file was skipped? How would we know from this test that we had lost coverage?

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -816,6 +816,34 @@ def mock_write(content):
...
+                # test loading new data without existing data
+                mock_logger.reset_mock()
+                data = [params['new_data'][3]]
+                data[0][0] = 'NA19678'  # load data for a new individual
+                _set_file_iter_stdout([header] + data)
+                body.pop('mappingFile')
+                response = self.client.post(url, content_type='application/json', data=json.dumps(body))
+                self.assertEqual(response.status_code, 200)
+                info = [
+                    f'Parsed 1 RNA-seq samples',
```
hanars, line 828: this does not need to be an f string, its failing codacy

hanars (changes requested):

hanars (approved):

PR #3270, opened 2023-03-03
Split sample matching functionality

hanars:
No functionality change, but splits apart the sample matching functionality, so we can update the RNA functionality without further diverging from the searching functionality

### DIFF
--- a/seqr/views/utils/dataset_utils.py
+++ b/seqr/views/utils/dataset_utils.py
@@ -216,7 +216,7 @@ def match_and_update_search_samples(
         elasticsearch_index=elasticsearch_index,
     )
     loaded_date = timezone.now()
-    samples, matched_individual_ids, remaining_sample_ids = _find_or_create_missing_sample_records(
+    samples, matched_individual_ids, _ = _find_or_create_missing_sample_records(
         samples=samples,
         projects=[project],
         user=user,
@@ -246,7 +246,7 @@ def match_and_update_search_samples(
     # refresh sample models to get updated values
     samples = Sample.objects.filter(id__in=[s.id for s in samples])
 
-    return samples, matched_individual_ids, activated_sample_guids, inactivated_sample_guids, family_guids_to_update, included_families
+    return samples, matched_individual_ids, activated_sample_guids, inactivated_sample_guids, family_guids_to_update
 
 
 def _match_and_update_rna_samples(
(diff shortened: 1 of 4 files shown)


ShifaSZ (changes requested):

seqr/views/utils/dataset_utils.py
```diff
@@ -264,29 +204,38 @@ def update_variant_samples(samples, user, elasticsearch_index, data_source=None,
...
     )
 
-    activated_sample_guids, inactivated_sample_guids = update_variant_samples(
-        samples, user, elasticsearch_index, data_source, loaded_date, dataset_type, sample_type)
+    prefetch_related_objects(samples, 'individual__family')
+    included_families = {sample.individual.family for sample in samples}
+    _validate_samples_families(samples, included_families, sample_type, dataset_type, expected_families=expected_families)
+
+    activated_sample_guids, inactivated_sample_guids = _update_variant_samples(
+        samples, user, elasticsearch_index, loaded_date=loaded_date, dataset_type=dataset_type, sample_type=sample_type)
```
ShifaSZ, line 238: It is okay to call with positional arguments rather than keyword arguments.

ShifaSZ (changes requested):

seqr/views/utils/dataset_utils.py
```diff
@@ -264,29 +204,38 @@ def update_variant_samples(samples, user, elasticsearch_index, data_source=None,
...
+        individual__family__project=project,
+        sample_type=sample_type,
+        dataset_type=dataset_type,
+        sample_id__in=sample_ids,
+        elasticsearch_index=elasticsearch_index,
+    )
     loaded_date = timezone.now()
-    samples, included_families, matched_individual_ids, remaining_sample_ids = match_sample_ids_to_sample_records(
-        projects=projects,
+    samples, matched_individual_ids, remaining_sample_ids = _find_or_create_missing_sample_records(
```
ShifaSZ, line 219: The `remaining_sample_ids` is never used, which causes the Codacy error.

ShifaSZ (changes requested):

seqr/views/utils/dataset_utils.py
hanars: yes

ShifaSZ (approved):

PR #3276, opened 2023-03-07
Fix RNAseq tpm load project

ShifaSZ:
Add the `project` column for the TPM input file and filter the individual lookups with project names. So that data won't be loaded into an unexpected project.

Keep the Outlier behaviors unchanged. No `project` field is added to the Outlier data.

### DIFF
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -47,7 +47,7 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Mismatched tissue types for sample NA19675_D2: fibroblasts, muscle')
 
-        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t1kg project\t\tENSG00000240361\t12.6\tmuscle\n'
+        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\tmuscle\n'
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
@@ -57,7 +57,7 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             call_command('load_rna_seq_tpm', RNA_FILE_ID, '--mapping-file', MAPPING_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677')
 
-        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
+        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
         call_command('load_rna_seq_tpm', RNA_FILE_ID, '--ignore-extra-samples')
 
         # Existing outlier data should be unchanged
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -31,6 +31,7 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000233750\t1.04\tmuscle\n',
             'NA19677\t1kg project nåme with uniçøde\t\tENSG00000233750\t5.31\tmuscle\n',
             'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000233750\t7.8\tmuscle\n',
+            'NA19678_D1\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n',
         ]
 
         with self.assertRaises(ValueError) as e:
@@ -52,11 +53,11 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
 
-        mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19678_D1\tNA19678']
-        mock_gzip_file.__iter__.return_value.append('NA19678_D1\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n')
+        mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19678_D1\tNA19678', 'NA19677\tNA20877']
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID, '--mapping-file', MAPPING_FILE_ID)
-        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
+        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19678_D1')
+        mock_open.assert_called_with(MAPPING_FILE_ID)
 
         mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
         call_command('load_rna_seq_tpm', RNA_FILE_ID, '--ignore-extra-samples')
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -23,31 +23,31 @@ class LoadRnaSeqTest(AuthenticationTestCase):
         mock_gzip_file = mock_gzip_open.return_value.__enter__.return_value
         mock_gzip_file.__iter__.return_value = [
             '',
-            'NA19675_D2\t\tENSG00000240361\t12.6\t\n',
-            'NA19678_D1\t\tENSG00000233750\t 6.04\twhole_blood\n',
-            'NA19678_D1\t\tENSG00000240361\t0.0\twhole_blood\n',
-            'NA19677\t\tENSG00000240361\t0.0\tinvalid\n',
-            'GTEX-001\t\tENSG00000240361\t3.1\tinvalid\n',
-            'NA19675_D2\t\tENSG00000233750\t1.04\tmuscle\n',
-            'NA19677\t\tENSG00000233750\t5.31\tmuscle\n',
-            'GTEX-001\t\tENSG00000233750\t7.8\tmuscle\n',
+            'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\t\n',
+            'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t 6.04\twhole_blood\n',
+            'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\twhole_blood\n',
+            'NA19677\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\tinvalid\n',
+            'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000240361\t3.1\tinvalid\n',
+            'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000233750\t1.04\tmuscle\n',
+            'NA19677\t1kg project nåme with uniçøde\t\tENSG00000233750\t5.31\tmuscle\n',
+            'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000233750\t7.8\tmuscle\n',
         ]
 
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
-        self.assertEqual(str(e.exception), 'Invalid file: missing column(s) TPM, gene_id, sample_id, tissue')
+        self.assertEqual(str(e.exception), 'Invalid file: missing column(s) TPM, gene_id, project, sample_id, tissue')
 
-        mock_gzip_file.__iter__.return_value[0] = 'sample_id\tindividual_id\tgene_id\tTPM\ttissue\n'
+        mock_gzip_file.__iter__.return_value[0] = 'sample_id\tproject\tindividual_id\tgene_id\tTPM\ttissue\n'
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Sample NA19675_D2 has no tissue type')
 
-        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t\tENSG00000240361\t12.6\tfibroblasts\n'
+        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\tproject\t\tENSG00000240361\t12.6\tfibroblasts\n'
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Mismatched tissue types for sample NA19675_D2: fibroblasts, muscle')
 
-        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t\tENSG00000240361\t12.6\tmuscle\n'
+        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t1kg project\t\tENSG00000240361\t12.6\tmuscle\n'
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
@@ -57,7 +57,7 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             call_command('load_rna_seq_tpm', RNA_FILE_ID, '--mapping-file', MAPPING_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677')
 
-        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
+        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
         call_command('load_rna_seq_tpm', RNA_FILE_ID, '--ignore-extra-samples')
 
         # Existing outlier data should be unchanged
@@ -95,6 +95,6 @@ class LoadRnaSeqTest(AuthenticationTestCase):
         ])
 
         # Test fails on mismatched tissue
-        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t\tENSG00000233750\t6.04\tfibroblasts\n'
+        mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t6.04\tfibroblasts\n'
         call_command('load_rna_seq_tpm', 'new_file.tsv.gz', '--ignore-extra-samples')
         mock_utils_logger.warning.assert_any_call(EXPECTED_MISMATCHED_TISSUE_WARNING, None)
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -53,9 +53,10 @@ class LoadRnaSeqTest(AuthenticationTestCase):
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
 
         mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19678_D1\tNA19678']
+        mock_gzip_file.__iter__.return_value.append('NA19678_D1\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n')
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID, '--mapping-file', MAPPING_FILE_ID)
-        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677')
+        self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
 
         mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\tNA19678\tENSG00000233750\t 6.04\twhole_blood\n'
         call_command('load_rna_seq_tpm', RNA_FILE_ID, '--ignore-extra-samples')
@@ -91,10 +92,11 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             mock.call('create 1 RnaSeqTpm for NA19678_D1'),
         ])
         mock_utils_logger.warning.assert_has_calls([
-            mock.call('Skipped loading for the following 1 unmatched samples: NA19677', None),
+            mock.call('Skipped loading for the following 2 unmatched samples: NA19677, NA19678_D1', None),
         ])
 
         # Test fails on mismatched tissue
+        mock_gzip_file.__iter__.return_value.pop()
         mock_gzip_file.__iter__.return_value[2] = 'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t6.04\tfibroblasts\n'
         call_command('load_rna_seq_tpm', 'new_file.tsv.gz', '--ignore-extra-samples')
         mock_utils_logger.warning.assert_any_call(EXPECTED_MISMATCHED_TISSUE_WARNING, None)
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -31,7 +31,7 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000233750\t1.04\tmuscle\n',
             'NA19677\t1kg project nåme with uniçøde\t\tENSG00000233750\t5.31\tmuscle\n',
             'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000233750\t7.8\tmuscle\n',
-            'NA19678_D1\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n',
+            'NA19678\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n',
         ]
 
         with self.assertRaises(ValueError) as e:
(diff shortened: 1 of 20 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -374,17 +386,22 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
     logger.info(message, user)
 
     data_source = file_path.split('/')[-1].split('_-_')[-1]
+    projects = Project.objects.filter(name__in={project for _, project in sample_id_project_tuples})
     samples, remaining_sample_ids = _match_and_update_rna_samples(
-        projects=get_internal_projects(),
+        projects=projects if len(projects) > 0 else get_internal_projects(),
         user=user,
         sample_ids=samples_by_id.keys(),
+        sample_id_project_names={f'{sample_id}--{project}' for sample_id, project in sample_id_project_tuples},
```
hanars, line 404: As discussed in the ticket and in the PR I linked there, doing filtration on a concatenated string is not sufficient. You need to add an explicit step that iterates through and confirms that every sample model has the correct project-sample. I recommend something like the set subtraction I added in https://github.com/broadinstitute/seqr/pull/3256/files#diff-a65acc9f2abbf6cf2a54fab4272e559e7bf7f79483a5f66b70e51219417dc84dR125

Also, for this many models I think the `Concat` you added is going to be way too slow for almost no gain. As I explicitly said on the ticket:
> Also, annotating that many samples with the concat field will almost certainly be slower than simply running a query that includes a few duplicates (the query we run now) and then making the sample lookup explicitly include the project

seqr/views/utils/dataset_utils.py
```diff
@@ -374,17 +386,22 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
 
     message = f'Parsed {len(samples_by_id)} RNA-seq samples'
     info = [message]
     logger.info(message, user)
 
     data_source = file_path.split('/')[-1].split('_-_')[-1]
+    projects = Project.objects.filter(name__in={project for _, project in sample_id_project_tuples})
     samples, remaining_sample_ids = _match_and_update_rna_samples(
-        projects=get_internal_projects(),
+        projects=projects if len(projects) > 0 else get_internal_projects(),
```
hanars, line 401: This is too fragile/confusing. The project filtration is different for outlier and tpm models - that means you need to explicitly pass different projects in from those 2 methods

hanars (changes requested):

hanars:
> There is no change to the new sample creation, which we should make some changes to.

Why? The only changes we need are to the lookup and mapping for existing Sample and Individual models to make sure they have the right project associated with them. Sample creation should not change

seqr/views/utils/dataset_utils.py
ShifaSZ: Okay, I'll do a sample lookup after the querying.

ShifaSZ:
> > There is no change to the new sample creation, which we should make some changes to.
> 
> Why? The only changes we need are to the lookup and mapping for existing Sample and Individual models to make sure they have the right project associated with them. Sample creation should not change
If we have two individuals with the same individual id and we are creating a new sample for one of the individuals, then it is possible we create the sample for the un-wanted individual.

hanars:
> If we have two individuals with the same individual id and we are creating a new sample for one of the individuals, then it is possible we create the sample for the un-wanted individual.

Not if you explicitly pass in the individual model, which we are

ShifaSZ:
The code for getting the individual models is as the following.
https://github.com/broadinstitute/seqr/blob/309d79f42c5fe313a8f35ac4c23db5ee4155eab2/seqr/views/utils/dataset_utils.py#L112-L114
The code uses the individual id as the key. So, if two individuals have the same individual id, only one will remain and it may not be the one we want.

hanars:
You either need to change that lookup to be a tuple lookup with the project as well as the ID, or you need to call this function once per project instead of once for all the samples

hanars (changes requested):

seqr/views/utils/dataset_utils.py
```diff
@@ -285,7 +298,8 @@ def _parse_tsv_row(row):
 TPM_COL = 'TPM'
 TISSUE_COL = 'tissue'
 INDIV_ID_COL = 'individual_id'
-TPM_HEADER_COLS = [SAMPLE_ID_COL, GENE_ID_COL, TISSUE_COL, TPM_COL]
+PROJECT_COL = 'project'
```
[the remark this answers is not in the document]
ShifaSZ: I double-checked the project column name. It is `project`.

seqr/views/utils/dataset_utils.py
ShifaSZ: I could pass in `projects` for outliers, but the `projects` for the TPMs will be available after the data is read from the file. It's more consistent to get the projects here. What do you think about that?

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -23,31 +23,31 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
...
-            'NA19677\t\tENSG00000233750\t5.31\tmuscle\n',
-            'GTEX-001\t\tENSG00000233750\t7.8\tmuscle\n',
+            'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\t\n',
+            'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t 6.04\twhole_blood\n',
+            'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\twhole_blood\n',
+            'NA19677\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\tinvalid\n',
+            'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000240361\t3.1\tinvalid\n',
+            'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000233750\t1.04\tmuscle\n',
+            'NA19677\t1kg project nåme with uniçøde\t\tENSG00000233750\t5.31\tmuscle\n',
+            'GTEX-001\t1kg project nåme with uniçøde\t\tENSG00000233750\t7.8\tmuscle\n',
```
hanars, line 33: Add a row with a sample_id that is present in our fixture data but specify a different project for it. For example, maybe include another row for  `NA19678_D1` but with `Test Reprocessed Project` as the project column, and then confirm that there is a warning returned to the user about it and that no data was created from that row

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -725,7 +725,7 @@ def _set_file_iter_stdout(rows):
                 self.assertEqual(response.status_code, 400)
                 self.assertDictEqual(response.json(), {'error': mock.ANY})
                 self.assertTrue(response.json()['error'].startswith(
-                    f'Error in NA19675_D2 data for {mismatch_row[1]}: mismatched entries '))
+                    f'Error in NA19675_D2 data for {mismatch_row[2 if data_type == "tpm" else 1]}: mismatched entries '))
```
hanars, line 728: I would rather you explicity set mismatch_row_index in the params

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -756,36 +756,53 @@ def _set_file_iter_stdout(rows):
                 self.assertEqual(model_cls.objects.count(), params['initial_model_count'])
                 mock_send_slack.assert_not_called()
 
+                def _test_basic_data_loading(data, samples, projects, project_names, individual_id):
```
hanars, line 759: did anything in this block change, or is it just copy-pasted? It might be better to do the refactor as a separate PR with this branched off of it to make it really clear what is changing in the tests

seqr/views/utils/dataset_utils.py
```diff
@@ -374,6 +392,8 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
             if indiv_id and sample_id not in sample_id_to_individual_id_mapping:
                 sample_id_to_individual_id_mapping[sample_id] = indiv_id
 
+            if PROJECT_COL in row_dict:
```
hanars, line 395: it would be better to check `if PROJECT_COL in expected_columns `

seqr/views/utils/dataset_utils.py
```diff
@@ -382,12 +402,14 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
 
     data_source = file_path.split('/')[-1].split('_-_')[-1]
     samples, remaining_sample_ids = _match_and_update_rna_samples(
-        projects=get_internal_projects(),
+        projects=get_internal_projects() if not sample_id_to_project_name else Project.objects.filter(
```
hanars, line 405: this should also be conditional on `if PROJECT_COL in expected_columns`

hanars (changes requested):

seqr/views/apis/data_manager_api_tests.py
ShifaSZ: Yes, they are copy-pasted from the loading new data test. I can create a PR for it. The basic test function is necessary because the additional data loading tests can do a partial test requiring little effort.

seqr/management/tests/load_rna_seq_tpm_tests.py
ShifaSZ: Added `mock_gzip_file.__iter__.return_value.append('NA19678_D1\tTest Reprocessed Project\t\tENSG00000240361\t0.2\twhole_blood\n')` a few lines below.

seqr/views/utils/dataset_utils.py
```diff
@@ -111,39 +111,41 @@ def _find_or_create_missing_sample_records(
...
         projects,
         user,
-        sample_ids,
+        sample_project_tuples,
+        get_individual_sample_key,
         raise_no_match_error=False,
         raise_unmatched_error_template=None,
         create_active=False,
         get_individual_sample_lookup=_get_individual_sample_lookup,
-        get_individual_sample_key=None,
```
ShifaSZ, line 119: Change the `get_individual_sample_key` from an optional argument to a required one.

seqr/views/utils/dataset_utils.py
```diff
@@ -111,39 +111,41 @@ def _find_or_create_missing_sample_records(
         samples,
         projects,
         user,
-        sample_ids,
+        sample_project_tuples,
```
ShifaSZ, line 114: Change the `sample_ids` to a sample-project tuple.

seqr/views/utils/dataset_utils.py
```diff
@@ -111,39 +111,41 @@ def _find_or_create_missing_sample_records(
...
         raise_no_match_error=False,
         raise_unmatched_error_template=None,
         create_active=False,
         get_individual_sample_lookup=_get_individual_sample_lookup,
-        get_individual_sample_key=None,
         **kwargs
 ):
     samples = list(samples)
-    remaining_sample_ids = set(sample_ids) - {sample.sample_id for sample in samples}
+    remaining_sample_keys = set(sample_project_tuples) - {(sample.sample_id, sample.project_name) for sample in samples}
```
ShifaSZ, line 123: We annotate a project_name field for the existing samples. It looks cleaner than the `sample.individual.family.project.name`, and it embedded the logic of using `None` for non-RNA and no-project info data.

seqr/views/utils/dataset_utils.py
```diff
@@ -111,39 +111,41 @@ def _find_or_create_missing_sample_records(
...
         remaining_individuals_dict = get_individual_sample_lookup(
             Individual.objects.filter(family__project__in=projects).exclude(id__in=matched_individual_ids)
         )
 
         # find Individual records with exactly-matching individual_ids
         sample_id_to_individual_record = {}
-        for sample_id in remaining_sample_ids:
-            individual_key = get_individual_sample_key(sample_id) if get_individual_sample_key else sample_id
+        for sample_key in remaining_sample_keys:
+            individual_key = (get_individual_sample_key(sample_key[0]), sample_key[1])
```
ShifaSZ, line 134: Since `get_individual_sample_key` is not optional, we don't have to handle the case when it is None.

seqr/views/utils/dataset_utils.py
```diff
@@ -261,28 +263,41 @@ def match_and_update_search_samples(
...
 
 def _match_and_update_rna_samples(
-    projects, user, sample_ids, data_source, sample_id_to_individual_id_mapping, raise_unmatched_error_template,
+    projects, user, sample_project_tuples, data_source, sample_id_to_individual_id_mapping, raise_unmatched_error_template,
 ):
-    samples = Sample.objects.select_related('individual').filter(
+    sample_project_lookup = defaultdict(set)
+    for sample_id, project_name in sample_project_tuples:
+        sample_project_lookup[sample_id].add(project_name)
+    no_project_key = next(iter(list(sample_project_tuples)), ('', None))[1] is None
```
ShifaSZ, line 271: After updating Outlier data to take in project names, we can eliminate `no_project_key`.

seqr/views/utils/dataset_utils.py
```diff
@@ -261,28 +263,41 @@ def match_and_update_search_samples(
...
+    samples = Sample.objects.select_related('individual__family__project').filter(
         individual__family__project__in=projects,
         sample_type=Sample.SAMPLE_TYPE_RNA,
         dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS,
-        sample_id__in=sample_ids,
-    )
+        sample_id__in=sample_project_lookup.keys(),
+    ).annotate(project_name=Value(None, output_field=TextField()) if no_project_key else F('individual__family__project__name'))
```
ShifaSZ, line 277: After updating Outlier, it can be simplified to `annotate(project_name=F('individual__family__project__name'))`.

seqr/views/utils/dataset_utils.py
```diff
@@ -377,7 +387,8 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
         row = dict(zip(header, _parse_tsv_row(line)))
         for sample_id, row_dict in parse_row(row, sample_id_to_tissue_type=sample_id_to_tissue_type):
             gene_id = row_dict['gene_id']
-            existing_data = samples_by_id[sample_id].get(gene_id)
+            project = row_dict.pop(PROJECT_COL, None)
```
ShifaSZ, line 390: We will remove `None` after all datasets have been upgraded to providing a project column.

seqr/views/utils/dataset_utils.py
```diff
@@ -396,26 +407,27 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
-    individual_db_ids = {s.individual_id for s in samples}
-    to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
+    to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
     if to_delete:
         model_cls.bulk_delete(user, to_delete)
 
     loaded_sample_ids = set(model_cls.objects.filter(sample__in=samples).values_list('sample_id', flat=True).distinct())
+    samples = Sample.objects.filter(id__in={s.id for s in samples} - loaded_sample_ids).annotate(
+        project_name=F('individual__family__project__name') if PROJECT_COL in expected_columns else Value(None, output_field=TextField()))
```
ShifaSZ, line 428: We should be able to check `rnaseqtpm_set` or `rnaseqoutlier_set` or similar fields for samples requiring a loading without `loaded_sample_ids`. But I need to learn how to make it work.

seqr/views/utils/dataset_utils.py
hanars: we should not add clunky code we plan to immediately remove. Just add outlier as part of this PR, or do outlier branched off this PR so it can be reviewed separately but then merge it here before this goes to dev

hanars (changes requested):

seqr/views/utils/dataset_utils.py
ShifaSZ: I have created a [PR](https://github.com/broadinstitute/seqr/pull/3287) for Outliers.

seqr/management/tests/load_rna_seq_tpm_tests.py
ShifaSZ: It should also be possible. But adding it below can reduce possible churns with the existing tests.

seqr/views/utils/dataset_utils.py
ShifaSZ: Adding a helper function argument for getting the remaining sample keys is a great idea. I will do that.

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -831,10 +839,20 @@ def mock_write(content):
...
-                data = [params['new_data'][3]]
+                data = [params['new_data'][4]]
                 data[0][0] = 'NA19678'  # load data for a new individual
                 body.pop('mappingFile')
-                _test_basic_data_loading(data, 1, 1, '1kg project nåme with uniçøde', 2, 0)
+                _test_basic_data_loading(data, 1, 1, '1kg project nåme with uniçøde', 1, 2, 0)
+
+                # Test loading data when where are duplicated individual ids in different projects.
+                # Add an individual with the individual_is identical to an existing individual
+                indiv = Individual(id=20, family_id=12, individual_id='NA20881', sex='M', affected='A')
```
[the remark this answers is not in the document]
ShifaSZ: Sounds good. I will add the individual that I added permanently to the fixture data. It belongs to project 3 and has the same ID as an individual in project 1.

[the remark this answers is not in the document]
ShifaSZ: The individual IDs in project 3 are heavily used. We can't change them easily. The most simple way is to add a new individual to the project. After adding the individual, a couple of other tests need some updates but are not hard to manage.

seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -23,41 +23,43 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
...
         self.assertEqual(str(e.exception), 'Mismatched tissue types for sample NA19675_D2: fibroblasts, muscle')
 
-        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t\tENSG00000240361\t12.6\tmuscle\n'
+        mock_gzip_file.__iter__.return_value[1] = 'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\tmuscle\n'
         with self.assertRaises(ValueError) as e:
             call_command('load_rna_seq_tpm', RNA_FILE_ID)
         self.assertEqual(str(e.exception), 'Unable to find matches for the following samples: NA19677, NA19678_D1')
 
-        mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19678_D1\tNA19678']
+        mock_open.return_value.__enter__.return_value.__iter__.return_value = ['NA19678_D1\tNA19678', 'NA19677\tNA20877']
```
hanars, line 56: how come you are adding a mapping for the new fixture data instead of just adding fixture data that works without mapping?

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -697,11 +697,13 @@ def test_add_workspace_data(self):
             ' HG00731, HG00732, HG00733, NA19675_1, NA20870, NA20874')
 
         # Test a valid operation
+        first_available_individual_id = Individual.objects.all().values_list('id', flat=True).order_by('-id').first() + 1
```
hanars, line 700: its better to mock guid creation

seqr/views/utils/dataset_utils.py
hanars: okay I will review this part of the code again once we have merged that to this branch

seqr/views/utils/dataset_utils.py
```diff
@@ -227,15 +229,17 @@ def match_and_update_search_samples(
     )
     loaded_date = timezone.now()
     get_individual_sample_key = _get_mapped_individual_lookup_key(sample_id_to_individual_id_mapping)
+    sample_project_tuples = [(sample_id, None) for sample_id in sample_ids]
```
hanars, line 232: for search samples we will not be using hacky null projects for tuples, there is no reason why this needs to change for search samples. Update the helper functions so that for search samples we continue to have a dictionary keyed just on sample ID and for RNA samples we are keying by tuples

hanars (changes requested):

seqr/management/tests/load_rna_seq_tpm_tests.py
ShifaSZ: The unmatched sample will be found in the previous test without the mapping. I can merge these two tests.

seqr/views/utils/dataset_utils.py
ShifaSZ: Okay. I will add a tuple type check when it needs to use the sample_id in the key.

seqr/views/apis/anvil_workspace_api_tests.py
ShifaSZ: Updated.

seqr/views/utils/dataset_utils.py
```diff
@@ -99,64 +99,70 @@ def _get_individual_sample_lookup(individuals):
...
+def _get_sample_id_from_dict_key(key):
+    return key[0] if isinstance(key, tuple) else key
 
 
 def _find_or_create_missing_sample_records(
         samples,
         projects,
         user,
-        sample_ids,
+        sample_count,
```
ShifaSZ, line 120: The `sample_ids` is replaced by the `sample_count` because we have moved the `remaining_sample_keys` up to the caller function.

seqr/views/utils/dataset_utils.py
```diff
@@ -407,15 +421,15 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
...
     # Delete old data
-    individual_db_ids = {s.individual_id for s in samples}
-    to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
+    to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
     prev_loaded_individual_ids = set(to_delete.values_list('sample__individual_id', flat=True))
     if to_delete:
         model_cls.bulk_delete(user, to_delete)
 
     loaded_sample_ids = set(model_cls.objects.filter(sample__in=samples).values_list('sample_id', flat=True).distinct())
+    samples = Sample.objects.select_related('individual__family__project').filter(id__in={s.id for s in samples} - loaded_sample_ids)
```
hanars, line 430: I don't think we want to do a full new query for these- you already call `select_related` above so theres no need to call it again

hanars (changes requested):

seqr/views/utils/dataset_utils.py
```diff
@@ -99,64 +99,70 @@ def _get_individual_sample_lookup(individuals):
...
-        remaining_sample_ids -= set(sample_id_to_individual_record.keys())
-        if raise_no_match_error and len(remaining_sample_ids) == len(sample_ids):
+        remaining_sample_keys -= set(sample_id_to_individual_record.keys())
+        if raise_no_match_error and len(remaining_sample_keys) == sample_count:
             raise ValueError(
-                'None of the individuals or samples in the project matched the {} expected sample id(s)'.format(
-                    len(sample_ids)
-                ))
+                'None of the individuals or samples in the project matched the {} expected sample id(s)'.format(sample_count))
+        remaining_sample_ids = {_get_sample_id_from_dict_key(sample_key) for sample_key in remaining_sample_keys}
```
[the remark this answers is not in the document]
ShifaSZ: I didn't notice that.

seqr/views/utils/dataset_utils.py
```diff
@@ -99,64 +99,70 @@ def _get_individual_sample_lookup(individuals):
...
         if raise_unmatched_error_template and remaining_sample_ids:
             raise ValueError(raise_unmatched_error_template.format(sample_ids=(', '.join(sorted(remaining_sample_ids)))))
 
         # create new Sample records for Individual records that matches
         new_samples = [
             Sample(
-                guid='S{}_{}'.format(random.randint(10**9, 10**10), sample_id)[:Sample.MAX_GUID_SIZE], # nosec
-                sample_id=sample_id,
+                guid='S{}_{}'.format(random.randint(10**9, 10**10), _get_sample_id_from_dict_key(sample_key))[:Sample.MAX_GUID_SIZE],  # nosec
+                sample_id=_get_sample_id_from_dict_key(sample_key),
```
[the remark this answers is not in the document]
ShifaSZ: When there is a `sample id` to `individual id` mapping, the `individual id` will differ from the `sample id`.

seqr/views/utils/dataset_utils.py
```diff
@@ -407,15 +421,15 @@ def _load_rna_seq(model_cls, file_path, user, mapping_file, ignore_extra_samples
         samples = validate_samples(samples, sample_id_to_tissue_type, warnings)
 
     # Delete old data
-    individual_db_ids = {s.individual_id for s in samples}
-    to_delete = model_cls.objects.filter(sample__individual_id__in=individual_db_ids).exclude(sample__data_source=data_source)
+    to_delete = model_cls.objects.filter(sample__in=samples).exclude(sample__data_source=data_source)
```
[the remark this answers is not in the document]
ShifaSZ: There can be different RNA samples of an individual. Do we want to delete the data of all the samples of the individual when loading new data to the individual?

seqr/views/utils/dataset_utils.py
ShifaSZ: The new samples don't have `select_related` called. But I'm unsure about the performance impact if we don't do that for the new samples.

seqr/views/utils/dataset_utils.py
```diff
@@ -99,64 +99,70 @@ def _get_individual_sample_lookup(individuals):
...
         raise_no_match_error=False,
         raise_unmatched_error_template=None,
         create_active=False,
         get_individual_sample_lookup=_get_individual_sample_lookup,
-        get_individual_sample_key=None,
         **kwargs
 ):
     samples = list(samples)
-    remaining_sample_ids = set(sample_ids) - {sample.sample_id for sample in samples}
+    remaining_sample_ids = set()
```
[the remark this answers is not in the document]
ShifaSZ: I will return the `remaining_sample_keys` directly and add special handling to RNA.

seqr/views/utils/dataset_utils.py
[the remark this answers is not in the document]
ShifaSZ: Why do we need to provide the `guid` here? The sample model will call _compute_guid() by default.

seqr/views/utils/dataset_utils.py
ShifaSZ: I'm going to eliminate the new query. If the performance is an issue for new samples, we may add `selete_related` just for new sample models.

seqr/views/utils/dataset_utils.py
```diff
@@ -111,40 +111,40 @@ def _find_or_create_missing_sample_records(
...
+        if raise_no_match_error and len(remaining_sample_keys) == sample_count:
             raise ValueError(
-                'None of the individuals or samples in the project matched the {} expected sample id(s)'.format(
-                    len(sample_ids)
-                ))
-        if raise_unmatched_error_template and remaining_sample_ids:
+                'None of the individuals or samples in the project matched the {} expected sample id(s)'.format(sample_count))
+        if raise_unmatched_error_template and remaining_sample_keys:
+            remaining_sample_ids = {sample_id for sample_id, _ in remaining_sample_keys}\
+                if isinstance(list(remaining_sample_keys)[0], tuple) else remaining_sample_keys
```
ShifaSZ, line 147: Still keep the `remaining_sample_ids` because the RNA has the `raise_unmatched_error_template` option. Move the `remaining_sample_ids` generation into the exception branch to avoid unnecessary computation.

hanars: Then we should change `raise_unmatched_error_template` to `get_unmatched_error` as a function that takes `sample_keys` as an argument and explicitly write the helper functions differently for search and rna samples

seqr/views/utils/dataset_utils.py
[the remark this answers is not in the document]
hanars: do we have a unit test for that?

hanars (changes requested):

seqr/views/utils/dataset_utils.py
```diff
@@ -155,12 +155,13 @@ def _find_or_create_missing_sample_records(
                 individual=individual,
                 created_date=timezone.now(),
                 is_active=create_active,
-                **kwargs,
-            ) for sample_id, individual in sample_id_to_individual_record.items()]
+                **kwargs
+            ) for sample_id, individual in [(sample_key[0] if isinstance(sample_key, tuple) else sample_key, individual)
```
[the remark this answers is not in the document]
ShifaSZ: We also need to use the `sample_id` for the `guid`. Can we use the individual id for the `guid`?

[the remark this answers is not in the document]
hanars: yes

hanars (approved):

PR #3280, opened 2023-03-10
helper functions for call-specific individual sample individual lookup

hanars:
This is the refactor I proposed in my last PR review. It would allow us to create an explicit mapping from samples to individual-project combinations.

### DIFF
--- a/seqr/views/utils/dataset_utils.py
+++ b/seqr/views/utils/dataset_utils.py
@@ -100,6 +100,8 @@ def _get_individual_sample_lookup(individuals):
 
 
 def _get_mapped_individual_lookup_key(sample_id_to_individual_id_mapping):
+    sample_id_to_individual_id_mapping = sample_id_to_individual_id_mapping or {}
+
     def _get_mapped_id(sample_id):
         return sample_id_to_individual_id_mapping.get(sample_id, sample_id)
     return _get_mapped_id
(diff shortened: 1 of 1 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -94,37 +94,45 @@ def load_mapping_file_content(file_content):
...
 
+
+def _get_individual_sample_lookup(individuals):
+    return {i.individual_id: i for i in individuals}
+
+
+def _get_mapped_individual_lookup_key(sample_id_to_individual_id_mapping):
+    def _get_mapped_id(sample_id):
+        return sample_id_to_individual_id_mapping.get(sample_id, sample_id)
+    return _get_mapped_id
```
ShifaSZ, line 105: I like your idea. I may want to do a little further by changing the `return` clause to the following.
```
return _get_mapped_id if sample_id_to_individual_id_mapping else lambda sample_id: sample_id
```
With the change, we can eliminate many `if-else` conditions afterward.

seqr/views/utils/dataset_utils.py
```diff
@@ -94,37 +94,45 @@ def load_mapping_file_content(file_content):
...
 def _find_or_create_missing_sample_records(
         samples,
         projects,
         user,
         sample_ids,
-        sample_id_to_individual_id_mapping=None,
         raise_no_match_error=False,
         raise_unmatched_error_template=None,
         create_active=False,
+        get_individual_sample_lookup=_get_individual_sample_lookup,
```
ShifaSZ, line 116: I see what you mean. It makes it ready for adding project name filters for RNAseq data.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #3288, opened 2023-03-20
update locus list access control

hanars:


seqr/views/utils/permissions_utils.py
```diff
@@ -216,24 +216,34 @@ def check_project_permissions(project, user, **kwargs):
...
+def _get_all_can_view_project_guids_set(user):
+    return set(get_project_guids_user_can_view(user, limit_data_manager=False))
+
+
 def check_projects_view_permission(projects, user):
-    can_view = set(get_project_guids_user_can_view(user, limit_data_manager=False))
-    no_access_projects = set(projects.values_list('guid', flat=True)) - can_view
+    no_access_projects = set(projects.values_list('guid', flat=True)) - _get_all_can_view_project_guids_set(user)
     if no_access_projects:
         raise PermissionDenied(f"{user} does not have sufficient permissions for {','.join(no_access_projects)}")
```
ShifaSZ, line 236: It should deny permission if `projects` is empty. How about checking if the project is empty first?

hanars: For this helper function if there are no projects than the user should have access, all this helper does is make sure the user has access to all the projects. The functionality for locus lists is really different from this function which is why why its no longer used for locus lists

PR #3289, opened 2023-03-21
abstract project selector component

hanars:
This moves existing behavior for selecting a project from the awesomebar and updating a component and makes it reusable, will be needed for a upcoming PR


ShifaSZ (approved):
LGTM

PR #3291, opened 2023-03-22
Fix counts for project variant notes

hanars:
x


seqr/views/apis/project_api.py
```diff
@@ -305,17 +305,14 @@ def project_mme_submisssions(request, project_guid):
...
-    num_tags = sum(count['count'] for count in note_counts_by_family)
     note_tag_type = {
         'variantTagTypeGuid': 'notes',
         'name': 'Has Notes',
         'category': 'Notes',
         'description': '',
         'color': 'grey',
         'order': 100,
-        'numTags': num_tags,
+        'numTags': SavedVariant.objects.filter(family__project=project, variantnote__isnull=False).distinct().count(),
```
ShifaSZ, line 315: How about adding a test to check `numTags` in the response changed?

ShifaSZ (changes requested):

seqr/views/apis/project_api.py
hanars: I did. I added test fixture data for an extra note on a variant that already had a note, the test started failing and reporting that there were 2 variants with notes instead of 1, I made this change, and the test started passing again

ShifaSZ: I see.

PR #3298, opened 2023-03-27
Anvil pedigree upload - require fields

hanars:


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -162,14 +162,15 @@ def _convert_fam_file_rows_to_json(column_map, rows, required_columns=None):
         json_record = {}
         for key, column in column_map.items():
             value = (row_dict.get(key) or '').strip()
+            if column in required_columns and not value:
+                errors.append(f'Missing {_to_title_case(_to_snake_case(column))} in row #{i + 1}')
```
ShifaSZ, line 166: You may want to add `continue` after appending to the errors.

seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -728,6 +730,12 @@ def _test_errors(self, url, fields, workspace_name):
         response_json = response.json()
         self.assertListEqual(response_json['errors'], ['Missing required columns: Affected, Sex'])
 
+        self.mock_load_file.return_value = LOAD_SAMPLE_DATA + MISSING_REQUIRED_SAMPLE_DATA
+        response = self.client.post(url, content_type='application/json', data=json.dumps(REQUEST_BODY))
+        self.assertEqual(response.status_code, 400)
+        response_json = response.json()
+        self.assertListEqual(response_json['errors'], ['Missing Sex in row #4', 'Missing Affected in row #4'])
```
ShifaSZ, line 737: It should test error messages like `Invalid value '' for ... in row #n` following the missing error messages if there is no `continue` after the missing column error.

hanars: those error messages don't exist, we have handing for blank values so they aren't generally invalid. We also wouldn't want there to be more errors here, this is exactly what we would want to tell users if they were missing the values

PR #3303, opened 2023-03-29
Bulk create logging error and better logging tests

hanars:
Based on the suggestion in the ticket I explored a few options for how to best test logging and ultimately decided to add a helper to our test function that redirects our log stream to a string buffer that we can access at any point in the. tests. This means we are able to test whatever the actual log output is. 

I then fixed the logging bug reported in the ticket, as well as a few other small bugs I caught once we actually had good testing for logs. I think eventually we will want to make these updates for everywhere in our tests that are mocking loggers, but that seems out of scope for this PR

### DIFF
--- a/seqr/views/utils/dataset_utils.py
+++ b/seqr/views/utils/dataset_utils.py
@@ -10,7 +10,7 @@ from seqr.models import Sample, Individual, Family, Project, RnaSeqOutlier, RnaS
 from seqr.utils.communication_utils import safe_post_to_slack
 from seqr.utils.elasticsearch.utils import get_es_client, get_index_metadata
 from seqr.utils.file_utils import file_iter
-from seqr.utils.logging_utils import log_model_bulk_update, SeqrLogger
+from seqr.utils.logging_utils import SeqrLogger
 from seqr.views.utils.file_utils import parse_file
 from seqr.views.utils.permissions_utils import get_internal_projects
 from seqr.views.utils.json_utils import _to_snake_case, _to_camel_case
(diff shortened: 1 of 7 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -158,7 +158,6 @@ def _find_or_create_missing_sample_records(
                 **kwargs
             ) for sample_key, individual in sample_id_to_individual_record.items()]
         samples += list(Sample.bulk_create(user, new_samples))
-        log_model_bulk_update(logger, new_samples, user, 'create')
```
hanars, line 161: `Sample.bulk_create`  actually calls `log_model_bulk_update` so we had a duplicate of this log

seqr/views/utils/test_utils.py
```diff
@@ -51,6 +53,9 @@ def setUp(self):
         self.mock_analyst_group.resolve_expression.return_value = 'analysts'
         self.addCleanup(patcher.stop)
 
+        self._log_stream = StringIO()
+        logging.getLogger().handlers[0].stream = self._log_stream
```
ShifaSZ, line 57: Good to know we can do that without mocking.

seqr/views/utils/test_utils.py
```diff
@@ -223,6 +228,18 @@ def check_no_analyst_no_access(self, url, get_response=None):
...
 
+    def reset_logs(self):
+        self._log_stream.truncate(0)
+        self._log_stream.seek(0)
+
+    def assert_json_logs(self, user, expected):
+        logs = self._log_stream.getvalue().split('\n')
+        for i, (message, extra) in enumerate(expected):
+            self.assertDictEqual(json.loads(logs[i]), {
+                'timestamp': mock.ANY, 'severity': 'INFO', 'user': user.email, 'message': message, **(extra or {}),
```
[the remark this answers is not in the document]
hanars: I think thats messier, I strongly prefer to have a single supported type per argument to a function. While python lest you have lists and function calls of mixed types its not really maintainable or readable

seqr/views/utils/test_utils.py
hanars: yeah took a while to figure this one out!

ShifaSZ (approved):

PR #3321, opened 2023-04-11
fix selected transcript for multi dataset comp het

hanars:
We are returning an invalid selected transcript ID in one particular edge case. We had a unit test for that case so our coverage was at 100% but we did not actually check that the outputted variants were formatted correctly so we missed the bug


ShifaSZ (approved):
LGTM

PR #3322, opened 2023-04-11
show SV new genotype

hanars:


ShifaSZ:
It hasn't implemented all the requested features. For example, it isn't displaying/filtering `new` variants. Right?

hanars:
>It hasn't implemented all the requested features. For example, it isn't displaying/filtering new variants. Right? 

It has, that was all implemented in the pipeline to work with the existing functionality we already built for gCNV, and I confirmed that it is working for this new SV data

seqr/utils/elasticsearch/constants.py
```diff
@@ -426,6 +426,7 @@ def get_prediction_response_key(key):
     'prev_call': {'format_value': bool},
     'prev_overlap': {'format_value': bool},
     'new_call': {'format_value': bool},
+    'prev_num_alt': {'format_value': lambda i: None if i is None else int(i)},
```
ShifaSZ, line 429: The type of `prev_num_alt` is an integer in the ES. Do we still need to format it?

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -291,6 +298,21 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
     previousCall = { content: 'Identical Call', hover: 'Identical call in previous callset', color: 'blue' }
   } else if (genotype.prevOverlap) {
     previousCall = { content: 'Overlapping Call', hover: 'Overlapping call in previous callset', color: 'teal' }
+  } else if (Number.isInteger(genotype.prevNumAlt)) {
+    const hasSameHemiXGenotype = isHemiX && (genotype.numAlt === 1 || genotype.numAlt === 2) &&
+      (genotype.prevNumAlt === 1 || genotype.prevNumAlt === 2)
```
ShifaSZ, line 303: I need clarification about the logic for the same hemiX genotype. Does it mean it is a hemiX with the same `numAlt` and `prevNumAlt` values?

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -291,6 +298,21 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    previousCall = {
+      content: 'New Genotype',
+      color: 'olive',
+      hover: (
+        <span>
+          Previous callset:
+          <AlleleContainer>{svNumAltGenotype(genotype.prevNumAlt, isHemiX)}</AlleleContainer>
+          {/* eslint-disable-next-line react/jsx-one-expression-per-line */}
+          {hasSameHemiXGenotype && <i>Hemi alt allele change: {genotype.prevNumAlt} to {genotype.numAlt}</i>}
```
ShifaSZ, line 312: The variable name `hasSameHemiXGenotype` looks like something changed, but the displaying text says `allele changed`.

seqr/utils/elasticsearch/constants.py
hanars: yes if format is omitted it defaults to string, which is wrong

ui/shared/components/panel/variants/VariantIndividuals.jsx
hanars: For hemiX num alt of 1 and 2 are displayed the exact same so to the user they would be unable to see that anything had changed, so this is checking for that case

ui/shared/components/panel/variants/VariantIndividuals.jsx
hanars: the number of called alt alleles has changed

ShifaSZ (approved):
LGTM.

PR #3339, opened 2023-04-21
Disable load data until loaded

hanars:


ui/pages/Project/components/ProjectOverview.jsx
```diff
@@ -332,7 +333,7 @@ const Dataset = React.memo(({ showLoadWorkspaceData, hasAnvil, samplesByType, us
     <DetailSection
       {...sectionProps}
       button={(datasetSections.length - 1 === i) ?
-        <EditDatasetsButton showLoadWorkspaceData={showLoadWorkspaceData} user={user} /> : null}
+        <EditDatasetsButton showLoadWorkspaceData={showLoadWorkspaceData && !noLoadedData} user={user} /> : null}
```
ShifaSZ, line 336: It could be more readable by adding `!noLoadedData` to the ternary condition and turning off the `EditDatasetsButton` button when no data is loaded.

hanars: no because for non-anvil users we do want to show the loading data option, like so we can actually add the data for the first time, so this only changes the behavior for the `showLoadWorkspaceData` case

ShifaSZ (approved):
LGTM

PR #3341, opened 2023-04-21
Show clinvar last updated date

hanars:


seqr/utils/elasticsearch/es_search.py
```diff
@@ -1206,6 +1207,12 @@ def _merge_duplicate_variants(cls, variant, duplicate_variant):
                 variant['genotypes'][guid] = genotype
         variant['familyGuids'] = sorted(set(variant['familyGuids'] + duplicate_variant['familyGuids']))
 
+        # Always show the most up-to-date clinvar
+        clinvar_version = variant[CLINVAR_KEY]['version']
+        dup_clinvar_version = duplicate_variant[CLINVAR_KEY]['version']
+        if dup_clinvar_version and (clinvar_version is None or dup_clinvar_version > clinvar_version):
```
ShifaSZ, line 1213: Would it be cleaner to assign the versions a default value of `0` using `or` like the below example? 
```
clinvar_version = variant[CLINVAR_KEY]['version'] or 0
dup_clinvar_version = duplicate_variant[CLINVAR_KEY]['version'] or 0
```
Then, the if condition can simply compare the two versions.

ShifaSZ (approved):

PR #3343, opened 2023-04-24
abstract component for loading data to state

hanars:
The diff is hard to read, but I basically moved the `LoadOptionsSelect` into a new component and made the response parsing a bit more generic so that it could be used for other components


ShifaSZ (approved):
LGTM

PR #3344, opened 2023-04-25
Hpo summary page

hanars:


ui/pages/Search/selectors.js
```diff
@@ -58,6 +58,21 @@ export const getProjectFamilies = (params, familiesByGuid, familiesByProjectGuid
   return null
 }
 
+export const getMultiProjectFamilies = createSelector(
+  (state, props) => props.match.params,
+  params => ({
```
ShifaSZ, line 63: It does nothing with the state. Is it helpful to be a selector?

ui/pages/Search/components/VariantSearchForm.jsx
```diff
@@ -33,3 +34,11 @@ const mapStateToProps = (state, ownProps) => ({
 })
 
 export default connect(mapStateToProps)(VariantSearchForm)
+
+const mapNoEditProjectStateToProps = (state, ownProps) => ({
+  noEditProjects: true,
```
ShifaSZ, line 39: The `noEditProjects` is always `true`. Are there any use cases that the `noEditProjects` are false?

ShifaSZ (changes requested):

ui/pages/Search/selectors.js
hanars: yes because it memoizes the function so its not recomputed every time

ui/pages/Search/components/VariantSearchForm.jsx
hanars: It is always true for the `NoEditProjectsVariantSearchForm` and always false for the regular `VariantSearchForm`

ShifaSZ (approved):

PR #3345, opened 2023-04-25
move es utils into nested folder

hanars:
This moves without changing the elasticsearch code into a nested folder, to support adding the new hail backend search

### DIFF
--- a/seqr/management/commands/lift_project_to_hg38.py
+++ b/seqr/management/commands/lift_project_to_hg38.py
@@ -11,7 +11,7 @@ from seqr.views.utils.dataset_utils import match_and_update_search_samples, \
 from seqr.views.utils.json_to_orm_utils import update_model_from_json
 from seqr.views.utils.orm_to_json_utils import get_json_for_saved_variants
 from seqr.views.utils.variant_utils import reset_cached_search_results
-from seqr.utils.elasticsearch.utils import get_es_variants_for_variant_tuples, get_single_es_variant
+from seqr.utils.search.utils import get_es_variants_for_variant_tuples, get_single_es_variant
 from seqr.utils.xpos_utils import get_xpos
 
 logger = logging.getLogger(__name__)
--- a/seqr/utils/search/elasticsearch/es_utils_tests.py
+++ b/seqr/utils/search/elasticsearch/es_utils_tests.py
@@ -11,9 +11,9 @@ from sys import maxsize
 from urllib3.exceptions import ReadTimeoutError
 
 from seqr.models import Family, Sample, VariantSearch, VariantSearchResults
-from seqr.utils.elasticsearch.utils import get_es_variants_for_variant_tuples, get_single_es_variant, get_es_variants, \
+from seqr.utils.search.utils import get_es_variants_for_variant_tuples, get_single_es_variant, get_es_variants, \
     get_es_variant_gene_counts, get_es_variants_for_variant_ids, InvalidIndexException, InvalidSearchException
-from seqr.utils.elasticsearch.es_search import _get_family_affected_status, _liftover_grch38_to_grch37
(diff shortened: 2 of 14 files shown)


seqr/management/commands/lift_project_to_hg38.py
```diff
@@ -11,7 +11,7 @@
 from seqr.views.utils.json_to_orm_utils import update_model_from_json
 from seqr.views.utils.orm_to_json_utils import get_json_for_saved_variants
 from seqr.views.utils.variant_utils import reset_cached_search_results
-from seqr.utils.elasticsearch.utils import get_es_variants_for_variant_tuples, get_single_es_variant
+from seqr.utils.search.utils import get_es_variants_for_variant_tuples, get_single_es_variant
```
ShifaSZ, line 14: I think all the `es` related utils should be under the `elasticsearch` sub-folder.

seqr/utils/search/elasticsearch/es_utils_tests.py
```diff
@@ -11,9 +11,9 @@
 from urllib3.exceptions import ReadTimeoutError
 
 from seqr.models import Family, Sample, VariantSearch, VariantSearchResults
-from seqr.utils.elasticsearch.utils import get_es_variants_for_variant_tuples, get_single_es_variant, get_es_variants, \
+from seqr.utils.search.utils import get_es_variants_for_variant_tuples, get_single_es_variant, get_es_variants, \
     get_es_variant_gene_counts, get_es_variants_for_variant_ids, InvalidIndexException, InvalidSearchException
-from seqr.utils.elasticsearch.es_search import _get_family_affected_status, _liftover_grch38_to_grch37
+from seqr.utils.search.elasticsearch.es_search import _get_family_affected_status, _liftover_grch38_to_grch37
```
ShifaSZ, line 16: I suggest removing the first underscore for the public function names.

seqr/views/apis/variant_search_api.py
```diff
@@ -12,8 +12,8 @@
 from reference_data.models import GENOME_VERSION_GRCh37
 from seqr.models import Project, Family, Individual, SavedVariant, VariantSearch, VariantSearchResults, ProjectCategory, \
     Sample
-from seqr.utils.elasticsearch.utils import get_es_variants, get_single_es_variant, get_es_variant_gene_counts
-from seqr.utils.elasticsearch.constants import XPOS_SORT_KEY, PATHOGENICTY_SORT_KEY, PATHOGENICTY_HGMD_SORT_KEY
+from seqr.utils.search.utils import get_es_variants, get_single_es_variant, get_es_variant_gene_counts
+from seqr.utils.search.elasticsearch.constants import XPOS_SORT_KEY, PATHOGENICTY_SORT_KEY, PATHOGENICTY_HGMD_SORT_KEY
```
ShifaSZ, line 16: Is it possible for the search API functions not to use the `es` functions directly? Let the search utility functions encapsulate the details of the search databases.

seqr/management/commands/lift_project_to_hg38.py
hanars: thats the plan, so  we would refactor `get_es_variants_for_variant_tuples` to `get_variants_for_variant_tuples` and then move the es-specific parts of the function into the nested `elasticsearch` folder. But to keep the diff on this sane, I started by just moving the files with no changes, and will have a leter follow up PR that splits up the logic

seqr/utils/search/elasticsearch/es_utils_tests.py
hanars: this PR is only moving files, not changing anything else. Subsequent PRs can clean up functions calls but the diff would not be readable if I start changing things in the files

seqr/views/apis/variant_search_api.py
hanars: see above

ShifaSZ (approved):
LGTM.

PR #3346, opened 2023-04-25
Move ES helpers

hanars:
Move all functions that directly interact with elasticsearch into search utils file and create shared constants file


seqr/utils/search/add_data_utils.py
```diff
@@ -0,0 +1,66 @@
+import elasticsearch_dsl
+
+from seqr.models import Sample
+from seqr.utils.search.constants import VCF_FILE_EXTENSIONS
+
+
+SAMPLE_FIELDS_LIST = ['samples', 'samples_num_alt_1']
+#  support .bgz instead of requiring .vcf.bgz due to issues with DSP delivery of large callsets
+DATASET_FILE_EXTENSIONS = VCF_FILE_EXTENSIONS[:-1] + ('.bgz', '.bed', '.mt')
+
```
hanars, line 10: everything in this file is copy-pasted from dataset_utils

seqr/utils/search/utils.py
```diff
@@ -32,6 +35,94 @@ def get_es_client(timeout=60, **kwargs):
     return elasticsearch.Elasticsearch(**client_kwargs, **kwargs)
 
 
+def ping_search_backend():
+    if not get_es_client(timeout=3, max_retries=0).ping():
+        raise ValueError('No response from elasticsearch ping')
+
+
+def get_elasticsearch_status():
```
hanars, line 43: copy-pasted from data_manager_api

ShifaSZ: The better file location for the non-search database management operations (e.g., status checking, loading, deleting, etc.) should be under the database-specific folder because they are not general to all types of databases. But here is the right place if we have a general DB status checking for all DBs.

seqr/utils/search/utils.py
```diff
@@ -32,6 +35,94 @@ def get_es_client(timeout=60, **kwargs):
...
+
+    return {
+        'indices': indices,
+        'diskStats': list(disk_status.values()),
+        'nodeStats': list(node_stats.values()),
+        'errors': errors,
+    }
+
+
+def delete_es_index(index):
```
ShifaSZ, line 72: Similarly, deleting an ES index should be under the `elasticsearch` folder. It is better to have a general DB deletion function under the `search` folder, which calls the `delete_es_index` from the Elasticsearch folder.

seqr/utils/search/add_data_utils.py
```diff
@@ -0,0 +1,67 @@
...
+from seqr.utils.search.constants import VCF_FILE_EXTENSIONS
+from seqr.utils.search.utils import get_es_client, get_index_metadata
+
+
+SAMPLE_FIELDS_LIST = ['samples', 'samples_num_alt_1']
+#  support .bgz instead of requiring .vcf.bgz due to issues with DSP delivery of large callsets
+DATASET_FILE_EXTENSIONS = VCF_FILE_EXTENSIONS[:-1] + ('.bgz', '.bed', '.mt')
+
+
+def validate_index_metadata_and_get_samples(elasticsearch_index, **kwargs):
```
ShifaSZ, line 13: This function is for Elasticsearch only. It is better to put it into the Elasticsearch folder. It must be okay for Some Elasticsearch-only APIs and management commands to call the functions defined in the ES folder.

seqr/utils/search/utils.py
hanars: Thats a fair point, I can give this a more generic name and then move all the ES specific logic into the elasticsearch helper function

seqr/utils/search/add_data_utils.py
hanars: It's not okay for anything outside the search folder to call any of the elasticsearch specific functios because it will make the switchover to a new backend and subsequent cleanup too hard. However, we can make all the function that are used outside the search folder much more generic and keep the elaticsearch specific login in the elasticsearch subfolder

PR #3347, opened 2023-04-25
search helpers for middleware errors

hanars:


seqr/utils/search/utils.py
```diff
@@ -18,6 +18,34 @@ class InvalidSearchException(Exception):
     pass
 
 
+SEARCH_EXCEPTION_ERROR_MAP = {
+    InvalidIndexException: 400,
+    InvalidSearchException: 400,
+    elasticsearch.exceptions.ConnectionError: 504,
+    elasticsearch.exceptions.TransportError: lambda e: int(e.status_code) if e.status_code != 'N/A' else 400,
+}
```
ShifaSZ, line 26: It may be better to define ES-specific things in the ES folder and use them here. For example, in the ES folder, we define `ES_SEARCH_EXCEPTION_ERROR_MAP`. And we define `SEARCH_EXCEPTION_ERROR_MAP = ES_SEARCH_EXCEPTION_ERROR_MAP` here. It is similar to the other constants below.

ShifaSZ (changes requested):

ShifaSZ (approved):
LGTM

PR #3353, opened 2023-04-26
Shared funtioanlity for add new search samples

hanars:
Shares functionality for adding search samples to minimize the number of public helper methods needed, and moves elasticsearch specific logic into the search utils

### DIFF
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -27,7 +27,7 @@ def add_new_search_samples(request_json, project, user, summary_template=False, 
     sample_id_to_individual_id_mapping = load_mapping_file(
         request_json['mappingFilePath'], user) if request_json.get('mappingFilePath') else {}
 
-    sample_ids, sample_type = validate_index_metadata_and_get_samples(
+    sample_ids, sample_type = _validate_index_metadata_and_get_samples(
         elasticsearch_index, project=project, dataset_type=dataset_type, genome_version=genome_version)
     if not sample_ids:
         raise ValueError('No samples found in the index. Make sure the specified caller type is correct')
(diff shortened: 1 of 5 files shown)


seqr/utils/search/add_data_utils.py
```diff
@@ -3,14 +3,73 @@
...
+        raise ValueError('No samples found in the index. Make sure the specified caller type is correct')
+
+    sample_data = {
+        'elasticsearch_index': elasticsearch_index,
+    }
+    num_samples, matched_individual_ids, activated_sample_guids, inactivated_sample_guids, updated_family_guids = match_and_update_search_samples(
+        project=project,
+        user=user,
+        sample_ids=sample_ids,
+        sample_data=sample_data,
```
ShifaSZ, line 42: Are there any benefits to replacing the `elasticsearch_index` with `sample_data`?

ShifaSZ (changes requested):

seqr/utils/search/add_data_utils.py
hanars: yeah we are gonna replace this with a different model field for the new backend and it will be good to be able to use the existing helper function but have all the backend-specific logic in one place

seqr/utils/search/elasticsearch/es_utils.py
```diff
@@ -92,7 +92,17 @@ def get_index_metadata(index_name, client, include_fields=False, use_cache=True)
     return index_metadata
 
 
-def validate_index_metadata_and_get_samples(elasticsearch_index, **kwargs):
+def validate_es_index_metadata_and_get_samples(request_json, project=None):
```
ShifaSZ, line 95: The `project` seems to be a required parameter.

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #3358, opened 2023-05-01
Rename es-specific search helpers

hanars:
This renames the generic search functions in the search utils file, although it does not change the logic. There will be a subsequent PR to move the es-specific code out of the generic helpers, but the diff was already pretty big for this given that these functions are tested and mocked by name so I wanted to keep the PR separate to keep the test changes simple


ShifaSZ (approved):
Looks good.

PR #3359, opened 2023-05-02
Split ES logic out of search utlities

hanars:
Moves the ES-specific parts of the search utility functions into ES-specific files/helper functions, and moves the more generally useful parts of those functions into the main search utils file


seqr/utils/search/utils.py
```diff
@@ -96,37 +115,57 @@ def query_variants(search_model, es_search_cls=EsSearch, sort=XPOS_SORT_KEY, ski
...
+
+    if len(previous_search_results.get('all_results', [])) == previous_search_results.get('total_results'):
+        return _get_gene_aggs_for_cached_variants(previous_search_results)
+
+    previously_loaded_results = process_es_previously_loaded_gene_aggs(previous_search_results)
+    if previously_loaded_results is not None:
+        return previously_loaded_results
+
+    gene_counts, _ = _query_variants(search_model, user, previous_search_results, gene_agg=True)
     return gene_counts
```
ShifaSZ, line 152: The rew return values are not `counts`. The name of the function could be more straightforward.

seqr/utils/search/elasticsearch/es_utils.py
```diff
@@ -248,3 +250,69 @@ def _get_es_indices(client):
...
+def process_es_previously_loaded_results(previous_search_results, start_index, end_index):
+    grouped_results = previous_search_results.get('grouped_results')
+    results = None
+    if grouped_results:
+        results = get_compound_het_page(grouped_results, start_index, end_index)
+
+    return results
+
+
+def process_es_previously_loaded_gene_aggs(previous_search_results):
```
ShifaSZ, line 295: This function does little with the `ES`. You may want to remove the `es` from its name and move it to the search util file.

ShifaSZ (changes requested):

seqr/utils/search/elasticsearch/es_utils.py
hanars: It is ES specific, only ES searches will generate this weird partial cache field, the hail backend will not, so I broke the caching apart into the caching that will be shared and the caching that will be ES only

ShifaSZ (approved):

Issue #3324, opened 2023-04-11
Show "RNA splice" tags on search result variants


ShifaSZ:
Show "RNA splice" tags on search result variants that overlap an outlier splice junction +/- 200bp

- hover over the tag should show info about the splice junction outlier (chrom/start/end/strand, z-score, p-value, maybe tissue and read count also)

- variant search results:

- OUTRIDER tag should be changed to: "RNA expression"
- double-check that splice junction outlier tags will work for SVs (needs interval intersection to work)

hanars:
this ticket is about showing a tag on the search results page and your comment is about the outlier table on the RNA page, so I do not think the question belongs on this issue. However, on the RNA page, IGV should show between the volcano plots and the table

ShifaSZ:
I mean the outlier table in the hover over as the figure below.
![image](https://user-images.githubusercontent.com/33550059/236555726-1675eb5f-81f3-490d-8f04-170ad7b6123e.png)

ShifaSZ:
We could show the IGV by clicking the tag or the outlier row. But the outlier row is more precise. We can get the locus of the splice junction outlier for the IGV.

Issue #3200, opened 2023-02-06
Splice junction outliers


bw2:
Each *splice junction* is represented by 4 values: chromosome, start (1-based) position, end position, strand (+ or -). 
Usually (but not always) the splice junction is associated with a gene. The gene may be missing if it's a novel junction outside of annotated genes. 

A *splice junction outlier* is a splice junction that also has a p-value (ie. 0.0002), a z-score (ie. -3.5), a tool name that generated these (ie. "FRASER"), and an integer read count representing the number of RNA-seq reads that span the junction. 

To add splice junction outliers to seqr, we should:
- show a "splice junction outliers " button on the family page under the individual that has RNA-seq data (similar to expression outliers)
- RNA-seq sample tissue type (eg. "fibroblasts" or "whole blood") should also be shown somewhere
- splice junction outlier page for that individual should show:
    - volcano plot (each point is a splice junction, x-axis is the z-score, y-axis is the p-value)
    - table with one row per gene or splice junction  (TBD) - with columns: rank by p-value, gene name (empty if gene is missing), splice junction chrom/start/end/strand (clicking here should jump to this interval in IGV.js),  read count, zscore, p-value
    - button to search for variants in this family with the region filter set to all splice junction intervals above some p-value threshold (TBD)
  
- variant search results should show "splice junction outlier" tags on variants that are near outlier junction intervals +/- 200bp 
	- hover over the tag should show info about the splice junction outlier (chrom/start/end/strand,  z-score, pvalue, maybe tissue and read count also)
	- there should be a way to show the splice junction region in IGV (+/- 200bp) and automatically show the splice junction track for that sample + the GTEx track for the same tissue as the sample

----
Input format for splice junction outliers. 
The data upload UI should make it possible to upload a table with these columns:

- seqr individual id 
- tissue type
- ENSG gene id 
- chrom
- start (1-based coordinates)
- end 
- type (string) - example values: psi5, psi3, theta
- strand  ("+" or "-")
- read count (integer number of RNA-seq reads that span the splice junction)
- z-score
- p-value
- deltaPsi (float) 
- rareDiseaseSamplesWithJunction - integer - this is the number of rare disease samples that have support for that junction
- rareDiseaseSamplesTotal - integer - number of rare disease samples total (for that tissue)
- dotSize (float) - values are between 0 and 1 and represent the size of the circle to draw in the volcano plot.

hanars:
[...]

PR #3366, opened 2023-05-10
Shifa/variant rnaseq label

ShifaSZ:
See ticket #3324


ui/pages/Project/components/RnaSeqJunctionOutliersTable.jsx
```diff
@@ -11,25 +10,6 @@ import ShowGeneModal from 'shared/components/buttons/ShowGeneModal'
...
-    {
-      name,
-      content: camelcaseToTitlecase(name).replace(' ', '-'),
-      format: row => row[name].toPrecision(3),
-    }
-  )),
-  ...RNA_SEQ_SPLICE_DETAIL_FIELDS.map(name => (
-    {
-      name,
-      content: camelcaseToTitlecase(name).replace(' ', '-'),
```
ShifaSZ, line 28: Move into utils for sharing between the RNAseq result page and the search result variants.

ShifaSZ:
After moving the `RNA Splice` tag to the annotation, the UI is like the below figure.
image

hanars:
> After moving the `RNA Splice` tag to the annotation, the UI is like the below figure.

Some notes on the UI:

- The tag should be pink, the same as we use for the expression outliers
- We do not need the header for the popup
- Do not show the gene in the hover
- The junction position should not be a link, and the popup should not open IGV. 
- The last two column headers are too long for the column content (this is also a problem on the main RNA page). Change them to `RD Junctions` and `RD Total` and add an info icon in the header where the user can hover to see the full header title
- It would be helpful to have a screenshot without the popup being open, in order to see how the label looks in general


Also, re: IGV, per the orginal ticket discussion it does not open from the new label, it opens from the existing "Show RNA-Seq "button". https://github.com/broadinstitute/seqr/issues/3200#issuecomment-1419571806
> In the control panel for IGV we allow users to toggle the "Range" show between "Variant" and "Gene". For variants in an outlier junction, we can add "Splice Junction" as an option here

ShifaSZ:
> In the control panel for IGV we allow users to toggle the "Range" show between "Variant" and "Gene". For variants in an outlier junction, we can add "Splice Junction" as an option here

I see what you mean by adding a `Splice Junction` IGV display range. My question is that the variant can have multiple overlapping junctions. How can we decide which junction's range to be used?

hanars:
The updates to IGV should be a separate PR, but in answer to your question, if there are multiple junctions show all of them (so the minimum start position and the maximum end position)

ShifaSZ:
> if there are multiple junctions show all of them (so the minimum start position and the maximum end position)

Okay

ShifaSZ:
The `RNA expression`  and `RNA splice` labels without a popup:

ShifaSZ:
With the popup info:

ui/shared/components/panel/variants/SpliceOutlierLabel.jsx
```diff
@@ -0,0 +1,71 @@
...
+  ), []).filter((outlier) => {
+    const { start: junctionStart, end: junctionEnd } = outlier
+    if ((pos >= junctionStart - RNASEQ_JUNCTION_PADDING) && (pos <= junctionEnd + RNASEQ_JUNCTION_PADDING)) {
+      return true
+    }
+    if (end && !endChrom) {
+      return (end >= junctionStart - RNASEQ_JUNCTION_PADDING) && (end <= junctionEnd + RNASEQ_JUNCTION_PADDING)
+    }
+    return false
+  })
```
ShifaSZ, line 39: Should we use a selector to get the overlapped outliers of the variant?

seqr/views/utils/variant_utils.py
```diff
@@ -131,15 +132,22 @@ def _add_locus_lists(projects, genes, add_list_detail=False, user=None):
...
+    for outlier, outlier_cls, max_sign in [('outliers', RnaSeqOutlier, None),
+                                           ('spliceOutliers', RnaSeqSpliceOutlier, MAX_SIGNIFICANT_OUTLIER_NUM)]:
+        outlier_data = get_json_for_rna_seq_outliers(
+            outlier_cls.objects.filter(
+                gene_id__in=gene_ids,
+                sample__individual__family__guid__in=family_guids,
+                sample__is_active=True,
+            ).order_by('p_value'),
+            nested_fields=[{'fields': ('sample', 'individual', 'guid'), 'key': 'individualGuid'},
+                           {'fields': ('sample', 'tissue_type'), 'key': 'tissueType'}],
```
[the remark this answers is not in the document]
hanars: Isn't that a lot of duplicated data though? Like I would expect there to only be a couple sample models returned, but this way you would be returning tissue type hundereds of times

ui/redux/selectors.js
```diff
@@ -433,6 +433,7 @@ export const getRnaSeqSignificantJunctionData = createSelector(
           junctionLocus: `${row.chrom}:${row.start}-${row.end} ${row.strand}`,
           idField: getSpliceId(row),
           familyGuid: individualsByGuid[individualGuid].familyGuid,
+          individualName: individualsByGuid[individualGuid].displayName || individualGuid,
```
hanars, line 436: what is the reason for this change?

hanars (changes requested):

seqr/views/utils/variant_utils.py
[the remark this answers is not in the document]
ShifaSZ: If we return `sampleGuid`, the `sampleGuid` will be duplicated hundreds of times, just like the `tissueType`. Returning `tissueType` saves steps to get the `tissueType` from the sample data.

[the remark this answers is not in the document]
hanars: But if we return `sampleGuid` only `sampleGuid` will be duplicated, whereas now we are duplicating `tissueType` and `individualGuid`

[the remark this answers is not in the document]
hanars: Also, in the UI getting the possible tissue types will be much faster iterating over a couple samples than hundreds of rnaseq models

ui/redux/selectors.js
hanars: but why add the fallback value? If we are not showing the data results page theres no need to add this, it can just be null

ui/shared/components/panel/variants/SpliceOutlierLabel.jsx
hanars: I think you should abstract the logic for getting overlapped junctions into a helper function, but that should be done in the IGV PR. The pre-grouping in the selector by family and chromosome (similar to what the  `VariantLocusListLabels` uses) will make that pretty efficient

ui/shared/components/table/RnaSeqJunctionOutliersTable.jsx
```diff
@@ -110,12 +110,12 @@ const RNA_SEQ_SPLICE_COLUMNS = [
...
 
 const RnaSeqJunctionOutliersTable = React.memo(
-  ({ variant, reads, showReads, updateReads, data, dispatch, ...props }) => (
+  ({ variant, reads, showReads, updateReads, data, columns, dispatch, ...props }) => (
     <div>
       {reads}
       <DataTable
         idField="idField"
-        columns={RNA_SEQ_SPLICE_COLUMNS}
+        columns={columns || RNA_SEQ_SPLICE_COLUMNS}
```
hanars, line 118: I think it would be better to have a boolean field passed in and then select which set of columns to use based on that

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -764,7 +764,7 @@ def get_json_for_rna_seq_outliers(models, max_significant_num_per_tissue=None, *
     if max_significant_num_per_tissue:
         models = models.annotate(place=Window(
             expression=DenseRank(),
-            partition_by=[F("sample__tissue_type")],
+            partition_by=[F('sample__tissue_type'), F('sample__individual_id')],
```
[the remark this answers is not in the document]
ShifaSZ: It is good for the previous PR because there is always a single individual in it. This PR has to fix it because the variant API needs to return the RnaSeq data on a family basis.

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -293,26 +294,14 @@ const BaseVariantLocusListLabels = React.memo(({ locusListIntervalsByProject, fa
...
-      return true
-    }
-    if (end && !variant.endChrom) {
-      const variantPosEnd = variantPos + (end - pos)
-      return (variantPosEnd >= interval.start) && (variantPosEnd <= interval.end)
-    }
-    return false
-  }).map(({ locusListGuid }) => locusListGuid)
+  const locusListGuids = locusListIntervals.filter(variantIntervalOverlap(variant))
+    .map(({ locusListGuid }) => locusListGuid)
```
ShifaSZ, line 304: It looks no simple way to make the `VariantLocusListLabels` reusable for `SpliceOutlierLabel`. So I just created a reusable variant and `Interval` overlapping checking function.

seqr/views/utils/variant_utils.py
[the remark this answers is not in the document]
ShifaSZ: The `rnaSeqData` in the response is organized on a per-individual basis. So the `individualGuid` won't be duplicated. If we return the tissue type, no iteration is required in the UI. It just uses the tissue type in the data. The only thing is that we need to map the single or two letters tissue types (e.g., 'M', 'F', etc.) to displayable strings. But the costs of lookup the mapping table are smaller than looking up the sample table.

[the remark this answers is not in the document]
hanars: If you have a nested field for `individualGuid`, as you do above, it will be added for every object

We define basically all the enumerations in constants.js and do the mapping on the client side, and return single letter codes on the server side, so that is probably the right approach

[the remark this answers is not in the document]
ShifaSZ: We need the nested field `individualGuid`. The field will be popped and used as a key in the response. This differs from the individual API, which has the `individualGuid` before fetching the outlier models and no need for a nested `individualGuid` field.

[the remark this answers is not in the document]
hanars: I think we are talking past each other. I think there are some high level questions that need to be aligned on before figuring out the best approach here. 

1. Why do we need `tissueType` at all here? In the new UI you are showing tissue type is not shown
2. Why are we customizing `get_json_for_rna_seq_outliers` so much here? In general, we do not want our json helpers to be overly customizable, we want to be really consistent everywhere they are used. Ideally we would have the same json for these everywhere, but in the event that we can't we need to be explicit about what is being added. In particular, a) why would the `max_significant_num_per_tissue` be different here than in other places, that should be a property of the model class? And b) why are the nested fields required here different? If the nested fields do need to be different `get_json_for_rna_seq_outliers` should accept a flag like `add_individual_guid_field` or something specific like that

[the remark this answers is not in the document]
ShifaSZ: 1. I should have added a `tissue type` column according to https://github.com/broadinstitute/seqr/issues/3200#:~:text=hover%20over%20the%20tag%20should%20show%20info%20about%20the%20splice%20junction%20outlier%20(chrom/start/end/strand%2C%20z%2Dscore%2C%20pvalue%2C%20maybe%20tissue%20and%20read%20count%20also) "hover over the tag should show info about the splice junction outlier (chrom/start/end/strand, z-score, pvalue, `maybe tissue` and read count also)"

2. We can avoid overly customizing `get_json_for_rna_seq_outliers` by creating a new function `get_json_for_splice_outliers`. a) there is only one value for `max_significant_num_per_tissue` when it is not `None`. If we separate a function for splice junction outliers, it is a constant. b) the variant search API gets the outliers per family, while the individual API gets them per individual. This difference makes the code for the two APIs different, but they share the same `get_json...` function. And the variant search API needs the nested individual GUID field, and the individual API doesn't.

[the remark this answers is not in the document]
hanars: The ticket says "maybe tissue" and we will be leaving tissue out for MVP, so you do not need to add it.

I still don't why the functionality here needs to be refactored so much. There is already a function to get the data for the volcano plot page, which by definition will contain all the info needed for the the junction table. Why can't we reuse that ?

seqr/views/apis/dataset_api_tests.py
```diff
@@ -103,13 +103,17 @@ def test_add_variants_dataset(self, mock_send_email, mock_send_slack, mock_rando
...
             'I000013_na20878': {'sampleGuids': [new_sample_guid]},
         })
         self.assertSetEqual(
             set(response_json['individualsByGuid']['I000002_na19678']['sampleGuids']),
             {replaced_sample_guid, existing_old_index_sample_guid}
         )
+        self.assertSetEqual(
+            set(response_json['individualsByGuid']['I000003_na19679']['sampleGuids']),
+            {'S000153_na19679', existing_sample_guid}
+        )
```
ShifaSZ, line 116: Change to avoid order churn.

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -12,8 +12,9 @@ import { ButtonLink, HelpIcon } from '../../StyledComponents'
 import { getOtherGeneNames } from '../genes/GeneDetail'
 import Transcripts from './Transcripts'
 import VariantGenes, { LocusListLabels } from './VariantGene'
-import { getLocus, has37Coords, Sequence, ProteinSequence, TranscriptLink } from './VariantUtils'
+import { getLocus, has37Coords, Sequence, ProteinSequence, TranscriptLink, variantIntervalOverlap } from './VariantUtils'
 import { GENOME_VERSION_37, GENOME_VERSION_38, getVariantMainTranscript, SVTYPE_LOOKUP, SVTYPE_DETAILS, SCREEN_LABELS } from '../../../utils/constants'
+import SpliceOutlierLabel from './SpliceOutlierLabel'
```
hanars, line 17: It would be more in keeping with the style of this folder to just define this component in this file directly

ui/shared/components/panel/variants/SpliceOutlierLabel.jsx
```diff
@@ -0,0 +1,49 @@
...
+import { getSpliceOutliersByChromFamily } from 'redux/selectors'
+import { RNASEQ_JUNCTION_PADDING } from 'shared/utils/constants'
+import RnaSeqJunctionOutliersTable from 'shared/components/table/RnaSeqJunctionOutliersTable'
+import { variantIntervalOverlap } from './VariantUtils'
+
+const HOVER_DATA_TABLE_PROPS = { basic: 'very', compact: 'very', singleLine: true }
+
+const BaseSpliceOutlierLabel = React.memo(({ variant, spliceOutliersByFamily }) => {
+  const { pos, end, familyGuids, endChrom } = variant
+  const outliers = familyGuids.reduce((acc, fGuid) => (
```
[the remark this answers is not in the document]
ShifaSZ: It is a little tricky because one is per family basis the other is per project basis. But it's doable.

ui/shared/components/panel/variants/selectors.js
```diff
@@ -30,7 +30,7 @@ export const getIndividualGeneDataByFamilyGene = createSelector(
       if (rnaSeqData) {
         acc[familyGuid] = acc[familyGuid] || {}
         acc[familyGuid].rnaSeqData = Object.entries(rnaSeqData).reduce(
-          (acc2, [geneId, data]) => (data.isSignificant ? {
+          (acc2, [geneId, [data]]) => (data.isSignificant ? {
```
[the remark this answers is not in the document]
ShifaSZ: The list for the expression outliers always has one element.

ui/shared/components/panel/variants/VariantUtils.jsx
```diff
@@ -60,3 +60,29 @@ export const ProteinSequence = React.memo(({ hgvs }) => <Sequence color="black"
 ProteinSequence.propTypes = {
   hgvs: PropTypes.string.isRequired,
 }
+
+const variantIntervalOverlap = (variant, padding, noGenomeVersion) => (interval) => {
+  const { pos, end, genomeVersion, liftedOverPos } = variant
+  const variantPos = (genomeVersion === interval.genomeVersion) || noGenomeVersion ? pos : liftedOverPos
```
ShifaSZ, line 66: For RNAseq outliers, no genome version is provided in the data. Should we always use `pos` rather than `leftedOverPos`?

hanars: yes

ui/shared/components/panel/variants/selectors.js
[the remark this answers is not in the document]
ShifaSZ: I keep expression outliers data unchanged, removing the list under each geneId.

ui/redux/selectors.js
```diff
@@ -360,6 +360,20 @@ export const getSearchGeneBreakdownValues = createSelector(
   })),
 )
 
+const groupDataNestedByChrom = (accInit, groupedData, nestedKey) => groupedData.reduce(
```
hanars, line 363: `accInit` is a confusing variable, `initialData` might be clearer

ui/redux/selectors.js
```diff
@@ -428,9 +433,22 @@ export const getRnaSeqSignificantJunctionData = createSelector(
...
         })),
     } : acc), {},
   ),
 )
+
+export const getSpliceOutliersByChromFamily = createSelector(
+  getRnaSeqSignificantJunctionData,
+  spliceDataByIndiv => Object.values(spliceDataByIndiv).reduce(
+    (acc, spliceData) => {
+      if (!spliceData || spliceData.length < 1) {
```
hanars, line 448: It would be better if `getRnaSeqSignificantJunctionData` had a filter in the logic to prevent empty values in the first place

ui/shared/components/panel/variants/VariantUtils.jsx
hanars: I'm not sure the `noGenomeVersion` is really a necessary argument here - couldn't you just do this condition based on whether or not `interval.genomeVersion` is defined?

ui/shared/components/panel/variants/VariantUtils.jsx
```diff
@@ -60,3 +60,29 @@ export const ProteinSequence = React.memo(({ hgvs }) => <Sequence color="black"
...
+  }
+  return false
+}
+
+export const getOverlappedIntervals = (
+  variant, intervals, familiesByGuid = null, padding = 0, noGenomeVersion = false,
+) => {
+  const { familyGuids = [] } = variant
+  const familyIntervals = familyGuids.reduce((acc, fGuid) => (
+    [...acc, ...(intervals[familiesByGuid ? familiesByGuid[fGuid].projectId : fGuid] || [])]), [])
```
hanars, line 85: It feels a little magical that if `familiesByGuid` is defined it should use `projectId` but otherwise not. I think it would be better to be more explicit and instead of `familiesByGuid` as the parameter, have `getIntervalGroupId` as a a parameter and then here it could be `intervals[getIntervalGroupId(fGuid)]`

hanars (changes requested):

ui/shared/components/panel/variants/VariantUtils.jsx
ShifaSZ: The existing code allows `interval.genomeVersion` to be undefined. In this case, it will use the `liftedOverPos` rather than `pos`. Is it a desired behavior? If not, we could change its behavior to use `pos` when `interval.genomeVersion` is undefined.

hanars: its probably better to swap the logic around to check if `variant.liftedOverGenomeVersion === interval.genomeVersion` and use that to determine if `liftedOverPos` should be used

seqr/views/utils/variant_utils.py
```diff
@@ -250,7 +240,8 @@ def get_variants_response(request, saved_variants, response_variants=None, add_a
         )
 
     if include_individual_gene_scores:
-        response['rnaSeqData'] = _get_rna_seq_outliers(genes.keys(), loaded_family_guids)
+        filters = {'gene_id__in': genes.keys(), 'sample__individual__family__guid__in': loaded_family_guids}
+        response['rnaSeqData'] = get_json_for_rna_seq_outliers(filters)
```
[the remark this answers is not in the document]
ShifaSZ: We could get rid of the helper function.

ui/pages/Project/reducers.js
```diff
@@ -303,7 +303,9 @@ export const searchMmeMatches = submissionGuid => (dispatch) => {
 export const loadRnaSeqData = individualGuid => (dispatch, getState) => {
   const data = getState().rnaSeqDataByIndividual[individualGuid]
   // If variants were loaded for the individual, the significant gene data will be loaded but not all the needed data
-  if (!data?.outliers || Object.values(data.outliers).every(({ isSignificant }) => isSignificant)) {
+  if (!data?.outliers || (Object.values(data.outliers).every(({ isSignificant }) => isSignificant) &&
+    Object.values(data.spliceOutliers).flat().every(({ isSignificant }) => isSignificant)
```
hanars, line 307: shouldn't you also be checking for `!data?.spliceOutliers`

ui/shared/components/panel/variants/VariantUtils.jsx
```diff
@@ -60,3 +60,28 @@ export const ProteinSequence = React.memo(({ hgvs }) => <Sequence color="black"
...
+  if (end && !variant.endChrom) {
+    const variantPosEnd = variantPos + (end - pos)
+    return (variantPosEnd >= interval.start - padding) && (variantPosEnd <= interval.end + padding)
+  }
+  return false
+}
+
+export const getOverlappedIntervals = (variant, intervals, getIntervalGroup, padding = 0) => {
+  const { familyGuids = [] } = variant
+  const familyIntervals = familyGuids.reduce((acc, fGuid) => (
```
hanars, line 83: it would be neater to do this with the filter in the reduce, instead of in 2 passes:
```
[...acc, ...(intervals[getIntervalGroup(fGuid)] || []).filter(variantIntervalOverlap(variant, padding)]
```

ui/pages/Project/reducers.js
ShifaSZ: I was thinking about the case that no spiceOutlierss data. We could have no `outliers` data for an individual as well. So, should we reload data when the `data` is null or undefined? This means not checking both `!data?.outliers` and `!data?.spliceOutliers` but checking `!data`.

hanars: So we should load an empty list for both `outliers` and `spliceOutliers` if data has been fetched but there is no data. So if its undefined it means its never been fetched and we should fetch and an empty list means it has been fetched and theres no need to re-fetch

ShifaSZ: So, is it safe to check `!data?.outliers` and not check `!data?.spliceOutliers`?

hanars: no, what if the `outliers` are loaded but not `spliceOutliers`

hanars (approved):

PR #3369, opened 2023-05-12
Add a new fixture sample preparing for splice outlier data fetching API tests

ShifaSZ:


seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -55,7 +55,7 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
 
         # Test database models
         existing_sample = Sample.objects.get(individual_id=1, sample_type='RNA', tissue_type='M')
-        existing_rna_samples = Sample.objects.filter(sample_type='RNA')
+        existing_rna_samples = Sample.objects.filter(sample_type='RNA').exclude(id=153)
```
hanars, line 58: Maybe be more descriptive of what sample you are omitting, like adding `rnaseqspliceoutlier__isnull=True` to the filter instead of excluding it by id

hanars (changes requested):

seqr/management/tests/load_rna_seq_tpm_tests.py
ShifaSZ: I add `rnaseqtpm__isnull=False` to the filter.

hanars (approved):

PR #3373, opened 2023-05-15
Fix read many files from GS

hanars:
For many files, we are running into a documented issue when using `wait` with stdout piping. Switching to the recommended `communicate` function fixes the issue: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -13,6 +13,9 @@ def run_command(command, user=None):
 
 
 def _run_gsutil_command(command, gs_path, gunzip=False, user=None):
+    if not is_google_bucket_file_path(gs_path):
+        raise Exception('A Google Storage path is expected.')
+
     #  Anvil buckets are requester-pays and we bill them to the anvil project
     google_project = get_google_project(gs_path)
     project_arg = '-u {} '.format(google_project) if google_project else ''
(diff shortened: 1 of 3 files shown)


seqr/utils/file_utils.py
```diff
@@ -87,22 +90,27 @@ def get_gs_file_list(gs_path, user=None, check_subfolders=True):
...
-        return [line.decode('utf-8').rstrip('\n') for line in process.stdout]
     return process
+
+
+def _run_gsutil_with_stdout(command, gs_path, user=None):
+    process = _run_gsutil_command(command, gs_path, user=user)
+    output, errs = process.communicate()
+    if errs:
+        errors = errs.decode('utf-8').strip().replace('\n', ' ')
+        raise Exception(f'Run command failed: {errors}')
```
ShifaSZ, line 115: In the case of the path not existing or insufficient permissions, the `errs` is still None. The error message is in the output because we set `stderr=subprocess.STDOUT`. We have to use other ways to detect errors, and I don't know how yet.

ShifaSZ (changes requested):

seqr/utils/file_utils.py
hanars: oh good cacth

seqr/utils/file_utils_tests.py
```diff
@@ -39,25 +39,23 @@ def test_get_gs_file_list(self, mock_logger, mock_subproc):
         self.assertEqual(str(ee.exception),  'A Google Storage path is expected.')
 
         process = mock_subproc.Popen.return_value
-        process.wait.return_value = -1
-        process.stdout = [b'-bash: gsutil: command not found.\n', b'Please check the path.\n']
+        process.communicate.return_value = None, b'-bash: gsutil: command not found.\nPlease check the path.\n'
```
ShifaSZ, line 42: It is better to change `None` to `b''` because that is the value in the real world.

seqr/utils/file_utils_tests.py
```diff
@@ -39,25 +39,23 @@ def test_get_gs_file_list(self, mock_logger, mock_subproc):
...
 
         mock_subproc.reset_mock()
         mock_logger.reset_mock()
-        process.wait.return_value = 0
-        process.stdout = [b'\n', b'\n', b'Updates are available for some Cloud SDK components.  To install them,\n',
-                          b'please run:\n', b'  $ gcloud components update\n',
-                          b'gs://bucket/target_path/id_file.txt\n', b'gs://bucket/target_path/data.vcf.gz\n']
+        process.communicate.return_value = b'\n\nUpdates are available for some Cloud SDK components.  To install them,\n' \
+                                           b'please run:\n  $ gcloud components update\ngs://bucket/target_path/id_file.txt\n' \
+                                           b'gs://bucket/target_path/data.vcf.gz\n', None
```
ShifaSZ, line 55: Same as above, `None` is better to be `b''`.

ShifaSZ (approved):

PR #3397, opened 2023-05-23
Update latest gencode

hanars:


reference_data/management/commands/update_gencode_latest.py
```diff
@@ -0,0 +1,62 @@
...
+from reference_data.management.commands.update_refseq import RefseqReferenceDataHandler
+from reference_data.models import GeneInfo, TranscriptInfo
+
+logger = logging.getLogger(__name__)
+
+BATCH_SIZE = 5000
+
+
+class Command(BaseCommand):
+    help = 'Loads genes and transcripts from the latest supported Gencode release, updating previously loaded gencode data'
```
ShifaSZ, line 17: I see the differences between `update_gencode` and `update_gencode_latest`. But it can be confusing for the users. How about add a scenario description about when to use which command?

ShifaSZ (changes requested):

ShifaSZ (approved):

PR #3401, opened 2023-05-25
Fix error formatting for IGV loading 

hanars:
`individualsByGuid` is no longer initially loaded with the project page state so formatting error messages on IGV loading is failing. This change more explicitly maps the user-provided ID to the error message shown to the user


ShifaSZ (approved):
LGTM

PR #3410, opened 2023-05-30
use xpos for X-linked filtering

hanars:
We no longer index the `contig` field and can not search against it in new indices


ShifaSZ (approved):
LGTM

PR #3414, opened 2023-05-31
Anvil gregor report fixes

hanars:


seqr/views/apis/report_api.py
```diff
@@ -686,7 +686,9 @@ def _get_sample_airtable_metadata(sample_ids, user, include_collaborator=False):
     'aligned_dna_short_read_file', 'aligned_dna_short_read_index_file', 'md5sum', 'reference_assembly',
     'alignment_software', 'mean_coverage', 'analysis_details',
 ]
-READ_TABLE_COLUMNS = ['aligned_dna_short_read_id', 'experiment_dna_short_read_id'] + READ_TABLE_AIRTABLE_FIELDS
+READ_TABLE_COLUMNS = ['aligned_dna_short_read_id', 'experiment_dna_short_read_id'] + READ_TABLE_AIRTABLE_FIELDS + ['quality_issues']
+READ_TABLE_COLUMNS.insert(6, 'reference_assembly_details')
+READ_TABLE_COLUMNS.insert(6, 'reference_assembly_uri')
```
ShifaSZ, line 691: I'm curious about why we have to add these fields. They aren't mentioned in the requirement ticket.

hanars: because the ticket also says to use the latest schema file for validation, and these columns are in that schema and we were throwing an error without them

ShifaSZ (approved):

PR #3415, opened 2023-05-31
Disable large search download

hanars:


seqr/utils/search/utils.py
```diff
@@ -125,12 +125,18 @@ def _get_cached_search_results(search_model, sort=None):
...
+
+
 def query_variants(search_model, sort=XPOS_SORT_KEY, skip_genotype_filter=False, load_all=False, user=None, page=1, num_results=100):
     previous_search_results = _get_cached_search_results(search_model, sort=sort)
     total_results = previous_search_results.get('total_results')
 
     if load_all:
-        num_results = total_results or MAX_VARIANTS
+        num_results = total_results or MAX_EXPORT_VARIANTS
+        _validate_export_variant_count(num_results)
```
ShifaSZ, line 139: Do we need to validate the number of results? We limit the downloading but shouldn't restrict displaying.

hanars: If you say "I want to download these 2000 variants" and then we return a file with 1000 variants, thats not successful. In the event that we don't know how many variants there will be before we trigger the search we kick it off with this reasonable threshold, but if it turns out that there were more variants in the search than we are willing to return then we need to tell the user they can't download everything and not give them this truncated result set

hanars: This is conditional on `load_all` which is only used for exports

ShifaSZ (approved):

PR #3420, opened 2023-06-01
Remove hail

hanars:
The hail import in seqr has not been working as we do not have java installed/configured etc. With the new hail backend running as a separate service it makes sense to remove hail from the main seqr repo entirely, and anywhere that really needs hail in the future will be run in that backend. The one place that currently uses hail is just for checking if files exist which is faster in hail but fine without it


requirements.txt
```diff
@@ -1,79 +1,29 @@
 #
-# This file is autogenerated by pip-compile with Python 3.9
-# by the following command:
+# This file is autogenerated by pip-compile with python 3.9
+# To update, run:
 #
 #    pip-compile requirements.in
 #
```
hanars, line 6: @sjahl adding you as reviewer mostly for a sanity check if anything here looks concerning/ if you think there is anything we might need to change in the docker image related to this

sjahl: Nope, I don't think anything with regards to the image needs to change. This looks reasonable to me -- we had quite a lot of transitive dependencies from hail!

sjahl (approved):

seqr/management/commands/check_bam_cram_paths.py
```diff
@@ -31,13 +31,13 @@ def handle(self, *args, **options):
             individual__family__project__name__in=args
         ) if args else IgvSample.objects.all()).filter(
             file_path__startswith='gs://'
-        ).prefetch_related('individual', 'individual__family__project')
+        ).order_by('id').prefetch_related('individual', 'individual__family__project')
```
ShifaSZ, line 34: Are there any reasons to add `order_by`?

hanars: mostly test stability

ShifaSZ: The test: https://github.com/broadinstitute/seqr/blob/a3cff8bc472d5b3efa6861e51e227b28f92f1453/seqr/management/tests/check_bam_cram_paths_tests.py#L43-L46
It uses a `sorted`. Does it stabilize the order?

ShifaSZ (approved):

PR #3422, opened 2023-06-02
Add hail backend status check

hanars:


seqr/views/status_tests.py
```diff
@@ -2,61 +2,86 @@
...
-
-        with mock.patch('seqr.utils.search.elasticsearch.es_utils.ELASTICSEARCH_SERVICE_HOSTNAME', ''):
-            self._test_status_error(url, mock_logger, es_error='Elasticsearch backend is disabled')
+        self._test_status_error(url, mock_logger)
 
         mock_db_connections.__getitem__.return_value.cursor.side_effect = None
         mock_redis.return_value.ping.side_effect = None
+        responses.add(responses.GET, 'http://test-hail:5000/status', status=200)
         urllib3_responses.add(urllib3_responses.HEAD, '/', status=200)
         urllib3_responses.add(urllib3_responses.HEAD, '/status', status=500)
```
ShifaSZ, line 47: Can the responses setting be moved to the child classes? So that we don't have to mix the setting in the shared base class.

PR #3425, opened 2023-06-05
Hail backend - add additional calls

hanars:
This implements all the remaining requests to the hail backend, or explicitly sets exceptions for requests we do not plan to support

### DIFF
--- a/seqr/utils/search/elasticsearch/es_utils.py
+++ b/seqr/utils/search/elasticsearch/es_utils.py
@@ -262,7 +262,7 @@ def _get_es_indices(client):
     return indices, seqr_index_projects
 
 
-def get_es_variants_for_variant_ids(samples, genome_version, variants_by_id, user=None, return_all_queried_families=False):
+def get_es_variants_for_variant_ids(samples, genome_version, variants_by_id, user, return_all_queried_families=False):
     variants = EsSearch(
         samples, genome_version, user=user, return_all_queried_families=return_all_queried_families,
     ).filter_by_variant_ids(list(variants_by_id.keys()))
--- a/seqr/utils/search/elasticsearch/es_utils.py
+++ b/seqr/utils/search/elasticsearch/es_utils.py
@@ -262,13 +262,11 @@ def _get_es_indices(client):
     return indices, seqr_index_projects
 
 
-def get_es_variants_for_variant_ids(samples, genome_version, variant_ids, user, dataset_type=None, return_all_queried_families=False):
+def get_es_variants_for_variant_ids(samples, genome_version, variants_by_id, user=None, return_all_queried_families=False):
     variants = EsSearch(
         samples, genome_version, user=user, return_all_queried_families=return_all_queried_families,
-    ).filter_by_variant_ids(variant_ids)
-    if dataset_type:
-        variants = variants.update_dataset_type(dataset_type)
-    return variants.search(num_results=len(variant_ids))
+    ).filter_by_variant_ids(list(variants_by_id.keys()))
+    return variants.search(num_results=len(variants_by_id))
 
 
 def get_es_variants(samples, search, user, previous_search_results, genome_version, sort=None, page=None, num_results=None,
--- a/seqr/utils/search/hail_search_utils.py
+++ b/seqr/utils/search/hail_search_utils.py
@@ -4,7 +4,7 @@ from django.db.models import F, Min
 import requests
 from reference_data.models import Omim, GeneConstraint, GENOME_VERSION_LOOKUP
 from seqr.models import Sample, PhenotypePrioritization
-from seqr.utils.search.constants import PRIORITIZED_GENE_SORT, DATASET_TYPES_LOOKUP
+from seqr.utils.search.constants import PRIORITIZED_GENE_SORT
 from seqr.utils.xpos_utils import MIN_POS, MAX_POS
 from settings import HAIL_BACKEND_SERVICE_HOSTNAME, HAIL_BACKEND_SERVICE_PORT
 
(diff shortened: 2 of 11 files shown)


seqr/utils/search/utils.py
```diff
@@ -103,21 +105,45 @@ def delete_search_backend_data(data_id):
...
+        try:
+            parsed_variant_ids[variant_id] = _parse_variant_id(variant_id)
+        except (KeyError, ValueError):
+            parsed_variant_ids[variant_id] = None
+
+    if dataset_type:
+        def is_valid(v_id):
+            if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS:
+                return bool(v_id)
+            return not v_id
```
ShifaSZ, line 138: Will it be more neat to write these lines:
 `return bool(v_id) if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS else not v_id`

ShifaSZ (commented):
It mostly looks excellent. Only a few nitpicking.

seqr/utils/search/elasticsearch/es_utils.py
```diff
@@ -262,13 +262,11 @@ def _get_es_indices(client):
...
-def get_es_variants_for_variant_ids(samples, genome_version, variant_ids, user, dataset_type=None, return_all_queried_families=False):
+def get_es_variants_for_variant_ids(samples, genome_version, variants_by_id, user, return_all_queried_families=False):
     variants = EsSearch(
         samples, genome_version, user=user, return_all_queried_families=return_all_queried_families,
-    ).filter_by_variant_ids(variant_ids)
-    if dataset_type:
-        variants = variants.update_dataset_type(dataset_type)
-    return variants.search(num_results=len(variant_ids))
+    ).filter_by_variant_ids(list(variants_by_id.keys()))
+    return variants.search(num_results=len(variants_by_id))
```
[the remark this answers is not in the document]
hanars: The ES function and the hail backend function need to have the same signature to make calling them straightforward, and the hail backend function needs this and it doesn't hurt the ES function to get it

seqr/utils/search/hail_search_utils.py
```diff
@@ -129,3 +152,20 @@ def _format_interval(chrom=None, start=None, end=None, offset=None, **kwargs):
...
+        missing_families = expected_families - set(result['familyGuids'])
+        if missing_families:
+            invalid_family_variants.append((result['variantId'], missing_families))
+
+    if invalid_family_variants:
+        from seqr.utils.search.utils import InvalidSearchException
+        missing = ', '.join([
+            f'{variant_id} ({"; ".join(sorted(families))})' for variant_id, families in invalid_family_variants
+        ])
+        raise InvalidSearchException(f'Unable to return all families for the following variants: {missing}')
```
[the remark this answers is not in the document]
hanars: this is only if you pass the specific `return_all_queried_families` argument to the function, which is only used in our manage scripts for lifting data between builds. In that case, a call might have dropped out of a callset but we still want the variant data returned. Its an unlikely edge case, which is why throwing an error here is an acceptable way of handling it

seqr/utils/search/utils.py
```diff
@@ -132,11 +132,11 @@ def _get_variants_for_variant_ids(families, variant_ids, user, dataset_type=None
...
     if dataset_type:
-        def is_valid(v_id):
-            if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS:
-                return bool(v_id)
-            return not v_id
-        parsed_variant_ids = {k: v for k, v in parsed_variant_ids.items() if is_valid(v)}
+        parsed_variant_ids = {
+            k: v for k, v in parsed_variant_ids.items()
+            if (dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS and v) or
+               (dataset_type == Sample.DATASET_TYPE_SV_CALLS and not v)
```
ShifaSZ, line 138: I am concerned about the precedence of `==` and the `and` operators. It is better to add parentheses around the comparisons. And what happens when dataset_type is not `Sample.DATASET_TYPE_VARIANT_CALLS` and `Sample.DATASET_TYPE_SV_CALLS`?

ShifaSZ (changes requested):

seqr/utils/search/utils.py
hanars: The precedence works as expected here, this is a very straightforward boolean in python:
```
s = 'A'
s == 'A' and False
False
s == 'A' and True
True
s = 'B'
s == 'A' and False
False
s == 'A' and True
False
```

That case can't happen, there are only 2 possible dataset types here. I thought it was more readable to explicitly do `dataset_type == Sample.DATASET_TYPE_SV_CALLS` but `dataset_type != Sample.DATASET_TYPE_VARIANT_CALLS` would do the same thing and not leave that edge case open

seqr/utils/search/utils.py
```diff
@@ -135,7 +135,7 @@ def _get_variants_for_variant_ids(families, variant_ids, user, dataset_type=None
         parsed_variant_ids = {
             k: v for k, v in parsed_variant_ids.items()
             if (dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS and v) or
-               (dataset_type == Sample.DATASET_TYPE_SV_CALLS and not v)
+               (dataset_type != Sample.DATASET_TYPE_VARIANT_CALLS and not v)
```
ShifaSZ: Removing the open edge case is good. And the condition can be rewritten as `if (v if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS else not v)`

hanars: this is more readable and explicit. I am leaving it as is

PR #3431, opened 2023-06-08
Shifa/update splice outlier model

ShifaSZ:
Add a rank field to the RnaSeqSpliceOutlier model to filter the significant outliers better.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # _seqr_ Changes
 
 ## dev
+* Add a 'rank' field to the RnaSeqSpliceOutlier model
```
hanars, line 4:
```suggestion
* Add a 'rank' field to the RnaSeqSpliceOutlier model  (REQUIRES DB MIGRATION)
```

seqr/models.py
```diff
@@ -1111,6 +1111,7 @@ class RnaSeqSpliceOutlier(DeletableSampleMetadataModel):
         ('*', 'Any direction'),
     )
 
+    rank = models.IntegerField(null=True)
```
hanars, line 1114: this should never be null

hanars (changes requested):

seqr/models.py
ShifaSZ: Okay, I thought we should support existing data in the database before the migration.

hanars: There shouldn't be any of this data in the database yet, and if there was it would be broken without this field so we would want the migration to fail

seqr/migrations/0052_rnaseqspliceoutlier_rank.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('seqr', '0051_auto_20230426_1615'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rnaseqspliceoutlier',
+            name='rank',
+            field=models.IntegerField(null=True),
```
hanars, line 16: it looks like you changed the model definition but did not recreate the actual migration file

ShifaSZ: Oh, you are right.

seqr/migrations/0052_rnaseqspliceoutlier_rank.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('seqr', '0051_auto_20230426_1615'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rnaseqspliceoutlier',
+            name='rank',
+            field=models.IntegerField(default=0),
```
hanars, line 16: We need this to fail if there is data loaded - setting this default will break behavior for displaying these models, we are relying on this field to be accurate. This needs to be non-nullable with no default - you will need to drop all your data locally to run this

hanars (changes requested):

seqr/migrations/0052_rnaseqspliceoutlier_rank.py
ShifaSZ: It still requires a default value after dropping all local data. And I did a little investigation and found there is a [trick ](https://stackoverflow.com/questions/30885466/how-can-i-resolve-the-django-error-add-a-non-nullable-field) to adding a non-nullable with no default field by two steps: 1) adding a nullable field first, 2) changing it to a non-nullable (by selecting manually handling null values). There will be two migration files. Do you think this approach is acceptable?

hanars: I think you are misunderstanding how the second option works. If you select "manually handle" it will generate a migration file for you with no default and  a non-nullable field, which is exactly what we want. You then have the ability to manually add your own handling for that,  but you should just not add anything and we will be all set

ShifaSZ: I know "manually handle" is what we want. But we can't simply add a new field to achieve that. We have to add a new field with "null=True" first and do "makemigrations", then remove "null=True" and do "makemigrations" again. We can only select "manually handle" when we do "makemigrations" for the second time.

ShifaSZ: I pushed an update that has two migrations. If there is any RnaSeqSpliceOutlier data, the migration will fail because the data is non-nullable. But the migrations will be successful if there are no existing RnaSeqSpliceOutlier data.

hanars: Why would there be 2 migrations? This needs to be one migration that fails if there is data, and since there is not any data we are fine

ShifaSZ: I know how to reverse the migrations. Doing it in one migration is only possible by providing a default value. The two migrations approach is the trick I mentioned above.

ShifaSZ: It also works by merging the two migration files manually.

PR #3434, opened 2023-06-12
add bulk upload validation

hanars:


seqr/views/utils/pedigree_info_utils_tests.py
```diff
@@ -55,13 +55,15 @@ def test_parse_pedigree_table(self):
             parse_pedigree_table(
                 [['family_id', 'individual_id', 'sex', 'affected', 'father', 'mother', 'proband_relation'],
                  ['fam1', 'ind1', 'male', 'aff.', 'ind3', 'ind2', 'mother'],
-                 ['fam2', 'ind2', 'male', 'unknown', 'ind2', '.', '']],
+                 ['fam2', 'ind1', 'male', 'unknown', 'ind2', '.', ''],
+                 ['fam2', 'ind2', 'male', 'unknown', 'ind2', '.', ''],],
```
ShifaSZ, line 59: It is neater not to add a comma before the last `]`.

hanars: No its better to add it so when we add a row in the future the diff is cleaner

PR #3437, opened 2023-06-14
Hail backend - change callset population

hanars:


seqr/utils/search/hail_search_utils.py
```diff
@@ -28,10 +28,14 @@ def get_hail_variants(samples, search, user, previous_search_results, genome_ver
     end_offset = num_results * page
     search_body = _format_search_body(samples, genome_version, user, end_offset, search)
 
+    frequencies = search_body.pop('freqs', None)
+    if frequencies and frequencies.get('callset'):
+        frequencies['seqr'] = frequencies.pop('callset')
```
ShifaSZ, line 33: It seems to me that `seqr` here means `seqr callset`. Is it better to use `seqrCallset` or `seqrPop`?

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -157,11 +157,15 @@ const sectionTitle = ({ fieldTitle, section }) => (
...
 const HET_SECTION = 'Heteroplasmy'
 
+const SV_CALLSET_POP = { field: 'sv_callset', fieldTitle: 'This Callset', acDisplay: 'AC', helpMessage: SV_CALLSET_CRITERIA_MESSAGE }
 const CALLSET_POP = { field: 'callset', fieldTitle: 'This Callset', acDisplay: 'AC' }
+const SEQR_POP = { ...CALLSET_POP, field: 'seqr', fieldTitle: 'seqr' }
 
 const POPULATIONS = [
-  { field: 'sv_callset', fieldTitle: 'This Callset', acDisplay: 'AC', helpMessage: SV_CALLSET_CRITERIA_MESSAGE },
+  SV_CALLSET_POP,
+  { ...SV_CALLSET_POP, field: 'sv_seqr', fieldTitle: 'seqr' },
```
ShifaSZ, line 166: Same as above for the `fieldTitle` field.

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -157,11 +157,15 @@ const sectionTitle = ({ fieldTitle, section }) => (
 const HOM_SECTION = 'Homoplasmy'
 const HET_SECTION = 'Heteroplasmy'
 
+const SV_CALLSET_POP = { field: 'sv_callset', fieldTitle: 'This Callset', acDisplay: 'AC', helpMessage: SV_CALLSET_CRITERIA_MESSAGE }
 const CALLSET_POP = { field: 'callset', fieldTitle: 'This Callset', acDisplay: 'AC' }
+const SEQR_POP = { ...CALLSET_POP, field: 'seqr', fieldTitle: 'seqr' }
```
ShifaSZ, line 162: Should the value for the `fieldTitle` be `seqr Callset`?

seqr/utils/search/hail_search_utils.py
hanars: No, that would be an anti-pattern.  This is nested under a `freqs` dictionary in the body, so its very explicit that its rteferring to a frequency, and we don't  use `gnomad_genomes_pop` or the like for the other populations

ui/shared/components/panel/variants/Frequencies.jsx
hanars: Its very clear in the UI that this is referring to a population frequency. Adding the word "Callset" is really inaccurate or misleading - it implies all the samples in seqr have been called together, which they haven't. And adding the words "population" or "frequency" clutters to UI without adding any information to the user

ui/shared/components/panel/variants/Frequencies.jsx
hanars: see above, but no

ui/shared/components/panel/variants/Frequencies.jsx
ShifaSZ: I see. But the statement `all the samples in seqr have been called together` confuses me. What does the word `called` mean exactly?

ShifaSZ (approved):

ui/shared/components/panel/variants/Frequencies.jsx
hanars: It means that they were run through a joint calling pipeline, like GATK, which is how a joint called VCF is generated

PR #3439, opened 2023-06-14
better test stablity for gregor report

hanars:
There are going to be substantial changes to which fields we query from airtable, so I would like the order of these fields to be  clear in the test so its easy to see what changed


ShifaSZ (approved):
LGTM

PR #3444, opened 2023-06-16
Better pedigree utils tests

hanars:
When investigating https://github.com/broadinstitute/seqr-private/issues/1246 I discovered that a pretty significant bug had slipped through due to a gap in our unit tests. We had extensive tests for different pedigree file types in pedigree_info_utils_tests, and then tests in individual_api_tests only included one basic pedigree file type. This meant that even though our coverage was incredibly high, we were not really testing the end-to-end behavior for some of the pedigree file types. In general, we have been trying to move away from directly testing utility functions to avoid this very sort of issue. This PR does not change behavior, it just moves all the tests that were directly calling the utility function into the individual_api_tests file and fully tests their behavior


seqr/views/apis/individual_api_tests.py
```diff
@@ -463,6 +511,310 @@ def test_individuals_table_handler(self):
         response = self.client.post(save_url)
         self.assertEqual(response.status_code, 200)
 
+    @mock.patch('seqr.views.utils.permissions_utils.PM_USER_GROUP', 'project-managers')
+    @mock.patch('seqr.views.utils.pedigree_info_utils.NO_VALIDATE_MANIFEST_PROJECT_CATEGORIES')
+    @mock.patch('seqr.utils.communication_utils.EmailMultiAlternatives')
+    def test_individuals_sample_manifest_table_handler(self, mock_email, mock_no_validate_categories):
```
hanars, line 517: this is broadly speaking all copy pasted from the other test file, just reformatted to calling the endpoint instead of the helper function directly

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -411,11 +411,11 @@ def _set_proband_relationship(json_records):
 
         maternal_siblings = {
             r[JsonConstants.INDIVIDUAL_ID_COLUMN] for r in records
-            if affected[JsonConstants.MATERNAL_ID_COLUMN] == r[JsonConstants.MATERNAL_ID_COLUMN]
+            if affected[JsonConstants.MATERNAL_ID_COLUMN] and affected[JsonConstants.MATERNAL_ID_COLUMN] == r[JsonConstants.MATERNAL_ID_COLUMN]
         }
         paternal_siblings = {
             r[JsonConstants.INDIVIDUAL_ID_COLUMN] for r in records
-            if affected[JsonConstants.PATERNAL_ID_COLUMN] == r[JsonConstants.PATERNAL_ID_COLUMN]
+            if affected[JsonConstants.PATERNAL_ID_COLUMN] and affected[JsonConstants.PATERNAL_ID_COLUMN] == r[JsonConstants.PATERNAL_ID_COLUMN]
```
hanars, line 418: I did fix a small bug I caught here when updating the test, which is that previously if 2 individuals were both missing one parent ID we were counting them as half siblings because they had the "same" parent (which is "None")

seqr/views/apis/individual_api_tests.py
```diff
@@ -371,6 +372,20 @@ def test_individuals_table_handler(self):
...
+        self.assertEqual(response.status_code, 400)
+        self.assertDictEqual(response.json(), {'warnings': None, 'errors': [
+            'Missing Family Id in row #1', 'Missing Individual Id in row #1',
+        ]})
+
+        response = self.client.post(individuals_url, {'f': SimpleUploadedFile(
+            'test.tsv',  '#Some comments\n#Family ID	#Individual ID	Previous Individual ID\n"1"	"NA19675_1"""'.encode('utf-8'))})
+        self.assertEqual(response.status_code, 400)
+        self.assertDictEqual(response.json(), {'warnings': [], 'errors': [
+            'Error while parsing file: test.tsv. Row 1 contains 2 columns: 1, NA19675_1, while header contains 3: Family ID, Individual ID, Previous Individual ID',
```
ShifaSZ, line 386: The test data `#Some comments\n#Family ID	#Individual ID	Previous Individual ID\n` is weird. The header line starts with a `#` and looks like a comment line.

seqr/views/apis/individual_api_tests.py
```diff
@@ -463,6 +513,310 @@ def test_individuals_table_handler(self):
...
+        response = _send_request_data(data)
+        self.assertEqual(response.status_code, 400)
+        self.assertDictEqual(response.json(), {
+            'errors': ['Error while parsing file: sample_manifest.tsv. Unsupported file format'], 'warnings': [],
+        })
+
+        data[1] = header_2[:5] + header_2[7:10] + header_2[14:17] + ['Coded Phenotype'] + header_2[19:]
+        response = _send_request_data(data)
+        self.assertDictEqual(response.json(), {
+            'errors': ['Error while parsing file: sample_manifest.tsv. Unsupported file format'], 'warnings': [],
```
ShifaSZ, line 551: You can add comments about the two tests' different purposes. Such as `test incomplete headers`.

seqr/views/apis/individual_api_tests.py
```diff
@@ -463,6 +513,310 @@ def test_individuals_table_handler(self):
...
+    @mock.patch('seqr.views.utils.permissions_utils.PM_USER_GROUP', 'project-managers')
+    @mock.patch('seqr.views.utils.pedigree_info_utils.NO_VALIDATE_MANIFEST_PROJECT_CATEGORIES')
+    @mock.patch('seqr.utils.communication_utils.EmailMultiAlternatives')
+    def test_individuals_sample_manifest_table_handler(self, mock_email, mock_no_validate_categories):
+        receive_url = reverse(receive_individuals_table_handler, args=[PROJECT_GUID])
+        self.check_manager_login(receive_url)
+
+        def _send_request_data(data):
+            return self.client.post(receive_url, {'f': SimpleUploadedFile(
+                'sample_manifest.tsv', '\n'.join(['\t'.join(row) for row in data]).encode('utf-8')),
```
ShifaSZ, line 525: It can be a class function and shared with other tests.

ShifaSZ (changes requested):

seqr/views/apis/individual_api_tests.py
hanars: This is maintaining existing behavior and moves an [existing test](https://github.com/broadinstitute/seqr/blob/master/seqr/views/utils/pedigree_info_utils_tests.py#L71), as we do allow `#` characters to be ignored in file headers. . Changing supported behavior is out of scope for this PR

seqr/views/apis/individual_api_tests.py
hanars: Thats out of scope for this PR, the goal here is to make our tests able to catch an existing bug and any similar bugs going forward. Refactoring the other tests in this file is out of scope

ShifaSZ (approved):
LGTM

PR #3445, opened 2023-06-16
split parse pedigree functionality

hanars:
while implementing https://github.com/broadinstitute/seqr/pull/3444 there was an edge case in the code that was nearly impossible to test, essentially because `parse_pedigree_table` was being called in 2 different places with 2 pretty different sets of arguments, but the code had error handling for all the possible argument combinations. This splits the functionality out, creating a `parse_basic_pedigree_table` helper that only works on standard pedigree formats and does not allow all the extra arguments for the sample manifest

### DIFF
--- a/seqr/views/apis/dataset_api_tests.py
+++ b/seqr/views/apis/dataset_api_tests.py
@@ -103,12 +103,16 @@ class DatasetAPITest(object):
         )
         self.assertDictEqual(response_json['individualsByGuid'], {
             'I000002_na19678': {'sampleGuids': mock.ANY},
-            'I000003_na19679': {'sampleGuids': ['S000153_na19679', existing_sample_guid]},
+            'I000003_na19679': {'sampleGuids': mock.ANY},
             'I000013_na20878': {'sampleGuids': [new_sample_guid]},
         })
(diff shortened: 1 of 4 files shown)


ShifaSZ (changes requested):

seqr/views/apis/dataset_api_tests.py
```diff
@@ -103,13 +103,17 @@ def test_add_variants_dataset(self, mock_send_email, mock_send_slack, mock_rando
...
+            'I000003_na19679': {'sampleGuids': mock.ANY},
             'I000013_na20878': {'sampleGuids': [new_sample_guid]},
         })
         self.assertSetEqual(
             set(response_json['individualsByGuid']['I000002_na19678']['sampleGuids']),
             {replaced_sample_guid, existing_old_index_sample_guid}
         )
+        self.assertSetEqual(
+            set(response_json['individualsByGuid']['I000003_na19679']['sampleGuids']),
+            {'S000153_na19679', existing_sample_guid}
```
ShifaSZ, line 115: I made the exact change in my branch. But it must work.

ShifaSZ (approved):

PR #3464, opened 2023-07-07
Avoid `None` as a key in the response JSON.

ShifaSZ:
The `create_json_response` function calls the `JsonResponse` function with `sort_keys` as the JSON dump params. In this case, `None` as a key in the JSON is not allowed. Converting `None` to a string `'null'` aligns the `json.dumps` behavior without the `sort_keys` param.


hanars:
How does this actually show to the user? My gut would be that this will look wrong and we are better off filtering these entirely, but depending on how they look it might be okay?

ShifaSZ:
I got a screenshot for the UI of a `None` gene_id transcript in the popup transcript modal, as shown below.
image

hanars:
What does the variant itself look like, these are SVs so they should be showing all the genes without clicking into the modal

But "Gene Id: null" is certainly not acceptable to show the user

ShifaSZ:
It is an SV, and it has many transcripts/genes.
image

hanars:
so what does in look like when you expand the genes?

ShifaSZ:
It shows thousands of genes. The screenshot below shows the last few of them. You can see the transcript data on the right side of the screenshot.
image

hanars:
I really need to be able to see how these new transcripts will be shown to the user. You either need to find an example with fewer genes, provide a screenshot that actually shows the "null" genes, or schedule time with me to meet. Showing just the end of the screenshot is not sufficient for me to figure out what changes are necessary

ShifaSZ:
In the VariantGenes component, there is a statement as the following. Since genesById['null'] is undefined, the transcripts with the `null` gene id will be filtered out.
https://github.com/broadinstitute/seqr/blob/c200fdc3eee153efde298027946099be92cde59c/ui/shared/components/panel/variants/VariantGene.jsx#L699

ShifaSZ:
The issue with the first option is that we don't have gene details.  We can only display the information in the `transcripts` field of the variant. So I also prefer the second option.

seqr/utils/search/elasticsearch/es_utils_tests.py
```diff
@@ -138,7 +138,12 @@ def call_request_json(self, index=-1):
               'consequence_terms': [
                   'frameshift_variant'
               ]
-            }
+            },
+            {
```
hanars, line 142: it would be better/ more accurate to add this to the mock SV variant, instead of here

hanars (approved):

PR #3477, opened 2023-07-11
handle no intervals

hanars:


ui/shared/components/panel/variants/VariantUtils.jsx
```diff
@@ -81,7 +81,10 @@ const variantIntervalOverlap = (variant, padding) => (interval) => {
 export const getOverlappedIntervals = (variant, intervals, getIntervalGroup, padding = 0) => {
   const { familyGuids = [] } = variant
   return familyGuids.reduce((acc, fGuid) => (
-    [...acc, ...(intervals[getIntervalGroup(fGuid)] || []).filter(variantIntervalOverlap(variant, padding))]), [])
+    intervals ? [
+      ...acc, ...(intervals[getIntervalGroup(fGuid)] || []).filter(variantIntervalOverlap(variant, padding)),
+    ] : []
+  ), [])
```
ShifaSZ, line 87: Adding a statement before the return as the following is neater?
```
if (!intervals) {
  return []
}
```
or write in as:
```
return intervals ? familyGuids.reduce((acc, fGuid) => (
    [...acc, ...(intervals[getIntervalGroup(fGuid)] || []).filter(variantIntervalOverlap(variant, padding))]), [])
 : []
```

hanars: this is fine. In an urgent release situation we do not nitpick style

ShifaSZ (approved):

PR #3480, opened 2023-07-11
Add hail backend - intitial status check

hanars:
This adds the directory and basic scaffolding for adding the hail search app to seqr. Currently, all this adds is a status endpoint and unit tests. The hail search app uses aiohttp, as this library comes built in on the hail docker image we are using, and is therefore the most lightweight and best supported approach. 

Future work will be needed to actually release docker images for the hail search backend, but I wanted to wait until there was a bit more functionality integrated in before adding that


hail_search/__main__.py
```diff
@@ -0,0 +1,16 @@
+from aiohttp import web
+
+from hail_search.web_app import init_web_app
+
+
+def run():
+    app = init_web_app()
+    web.run_app(
+        app,
+        host='0.0.0.0',  # nosec
```
ShifaSZ, line 10: It looks like a backend service web app without authentication support. It should not be accessible from external.

hanars: Correct. The hail backend service will only be accessible via an internal, unauthenticated kubernetes service, so that only seqr can access it. Actually setting up the deployment is out of scope of this PR, but you can see the kubernetes yaml that I used to get this working in a test environment here: https://github.com/broadinstitute/seqr/blob/62084009dd60a35eef1c85608ad0c8a3e2a90cf8/deploy/kubectl_helpers/utils/hail-search.yaml

ShifaSZ (approved):

sjahl (approved):

PR #3481, opened 2023-07-11
handle invalid version directories for dag trigger

hanars:
Triggering the dag through the new UI fails if the directory has any non-version files or folders

### DIFF
--- a/seqr/views/apis/data_manager_api.py
+++ b/seqr/views/apis/data_manager_api.py
@@ -449,7 +449,8 @@ def load_data(request):
 
     version_path_prefix = f'{SEQR_DATSETS_GS_PATH}/GRCh38/{dag_name}'
     version_paths = get_gs_file_list(version_path_prefix, user=request.user, allow_missing=True, check_subfolders=False)
-    curr_version = max([int(re.findall(f'{version_path_prefix}/v(\d\d)/', p)[0]) for p in version_paths] + [0])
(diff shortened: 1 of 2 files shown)


seqr/views/apis/data_manager_api.py
```diff
@@ -449,7 +449,8 @@ def load_data(request):
 
     version_path_prefix = f'{SEQR_DATSETS_GS_PATH}/GRCh38/{dag_name}'
     version_paths = get_gs_file_list(version_path_prefix, user=request.user, allow_missing=True, check_subfolders=False)
-    curr_version = max([int(re.findall(f'{version_path_prefix}/v(\d\d)/', p)[0]) for p in version_paths] + [0])
+    versions = [re.findall(f'{version_path_prefix}/v(\d\d)/', p) for p in version_paths]
+    curr_version = max([int(v[0]) for v in versions if v] + [0])
```
ShifaSZ, line 453: Should we also provide a warning message about skipping version paths? The skipping can be found by comparing the lengths of `version_paths` and `versions`.

hanars: No. If there is no previous version this will set the version to 0, which is expected behavior. This is already unit tested. The only change here is that if there are other things in the bucket, which is allowed to happen, we don't try to extract a version number from them. There is nothing a user cares about happening here

PR #3482, opened 2023-07-12
Allow TPM value to be zero.

ShifaSZ:
See the connected issue.

### DIFF
--- a/seqr/management/tests/load_rna_seq_tpm_tests.py
+++ b/seqr/management/tests/load_rna_seq_tpm_tests.py
@@ -23,8 +23,6 @@ class LoadRnaSeqTest(AuthenticationTestCase):
             '',
             'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\t\n',
             'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t 6.04\twhole_blood\n',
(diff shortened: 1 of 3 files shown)


seqr/management/tests/load_rna_seq_tpm_tests.py
```diff
@@ -23,8 +23,6 @@ def test_command(self, mock_open, mock_logger, mock_utils_logger, mock_gzip_open
             '',
             'NA19675_D2\t1kg project nåme with uniçøde\t\tENSG00000240361\t12.6\t\n',
             'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000233750\t 6.04\twhole_blood\n',
-            'NA19678_D1\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\twhole_blood\n',
-            'NA19677\t1kg project nåme with uniçøde\t\tENSG00000240361\t0.0\tinvalid\n',
```
hanars, line 27: instead of removing these and then changing a value below, why not just keep one of them?

hanars (changes requested):

seqr/management/tests/load_rna_seq_tpm_tests.py
ShifaSZ: Simply keeping one of the two rows won't work because they have either invalid tissue or conflict data with other rows. Removing them and changing another success-loaded row to 0 TPM is much simpler.

hanars (approved):

PR #3489, opened 2023-07-13
add family guid to write pedigree

hanars:


ShifaSZ:
Isn't the `Family_ID` a unique ID already?

hanars:
> Isn't the `Family_ID` a unique ID already?

within a project yes but not across seqr, which means it is not acceptable for naming permanent files that contain family data for search in the v3 backend. Also, family names are editable by users, GUIDs are not

ShifaSZ (approved):

PR #3492, opened 2023-07-14
Support per project family-table-filters.

ShifaSZ:
See the connected issue.

### DIFF
--- a/ui/pages/Project/reducers.js
+++ b/ui/pages/Project/reducers.js
@@ -10,6 +10,7 @@ import {
 import { SHOW_ALL, SORT_BY_FAMILY_GUID, NOTE_TAG_NAME } from 'shared/utils/constants'
 import { HttpRequestHelper } from 'shared/utils/httpRequestHelper'
 import { SHOW_IN_REVIEW, SORT_BY_FAMILY_NAME, SORT_BY_FAMILY_ADDED_DATE, CASE_REVIEW_TABLE_NAME } from './constants'
+import { getFamiliesTableFiltersByProject } from './selectors'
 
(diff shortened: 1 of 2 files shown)


ui/pages/Project/selectors.js
```diff
@@ -432,6 +432,12 @@ const analysedByFilters = (filter, analysedByOptions) => {
   return filterGroups
 }
 
+export const getFamiliesTableFilters = createSelector(
+  getFamiliesTableFiltersByProject,
+  getCurrentProject,
```
hanars, line 437: it would be better just to use the `getProjectGuid` selector

ui/pages/Project/reducers.js
```diff
@@ -370,7 +371,15 @@ export const updateFamiliesTable = (updates, tableName) => (
   { type: tableName === CASE_REVIEW_TABLE_NAME ? UPDATE_CASE_REVIEW_TABLE_STATE : UPDATE_FAMILY_TABLE_STATE, updates }
 )
 
-export const updateFamiliesTableFilters = updates => ({ type: UPDATE_FAMILY_TABLE_FILTER_STATE, updates })
+export const updateFamiliesTableFilters = updates => (dispatch, getState) => {
```
hanars, line 374: Because the data structure is changing, you need to change what kind of reducer `UPDATE_FAMILY_TABLE_FILTER_STATE` is, so its one that can be easily partially updated. If you change it to `createObjectsByIdReducer` then the built in update functionality will handle merging old and new state for you, and you wouldn't need `getFamiliesTableFiltersByProject` at all here

hanars (changes requested):

hanars (approved):

PR #3498, opened 2023-07-18
Fathmm mkl

hanars:


seqr/utils/search/elasticsearch/es_search.py
```diff
@@ -1533,7 +1555,7 @@ def _parse_es_sort(sort, sort_config):
 
 def _get_field_values(hit, field_configs, format_response_key=_to_camel_case, get_addl_fields=None, lookup_field_prefix='', existing_fields=None, skip_fields=None):
     return {
-        field_config.get('response_key', format_response_key(field)): _value_if_has_key(
+        field_config.get('response_key') or format_response_key(field): _value_if_has_key(
```
ShifaSZ, line 1558: It is unlikely that the `response_key` is configured to `None` or an empty string. It should be safe to keep it unchanged.

seqr/views/utils/test_utils.py
```diff
@@ -1056,7 +1056,8 @@ def _get_expected_dag_variables(self, omit_project=None, **kwargs):
                                    'hom': None, 'id': None, 'max_hl': None},
         },
         'pos': 248367227,
-        'predictions': {'splice_ai': 0.75, 'eigen': None, 'revel': None, 'mut_taster': None, 'fathmm': None,
+        'predictions': {'splice_ai': 0.75, 'eigen': None, 'revel': None, 'mut_taster': None, 'fathmm': 'D',
+                        'vest': None, 'mut_pred': None,
```
ShifaSZ, line 1060: Should we also test for non-None values for `vest` and `mut_pred`?

seqr/utils/search/elasticsearch/es_search.py
hanars: Its not about handling Nones, its about skipping an unnecessary computation. If `response_key` is provided, theres no reason `format_response_key` needs to be a valid function for that particular field. With these updates, theres a case where `format_response_key` throws an error but theres no reason to evaluate it at all since theres a `response_key`, so this changes causes the evaluation to be skipped

ShifaSZ (approved):

PR #3502, opened 2023-07-20
Multiple variant tags filtering

ShifaSZ:
Now, it updates the saved-variant filtering with multiple tags. No behavior changes to the project saved-variants.


seqr/views/apis/summary_data_api.py
```diff
@@ -87,8 +87,11 @@ def saved_variants_page(request, tag):
     if tag == 'ALL':
         saved_variant_models = SavedVariant.objects.exclude(varianttag=None)
     else:
-        tag_type = VariantTagType.objects.get(name=tag, project__isnull=True)
-        saved_variant_models = SavedVariant.objects.filter(varianttag__variant_tag_type=tag_type)
+        tags = tag.split(';')
+        saved_variant_models = SavedVariant.objects.filter(
+            varianttag__variant_tag_type__name__in=tags,
+            varianttag__variant_tag_type__project__isnull=True,
```
ShifaSZ, line 93: It returns more saved variants than we need. But the front end will filter out the proper variants to display.

ui/pages/SummaryData/reducers.js
```diff
@@ -43,22 +43,30 @@ export const loadSuccessStory = successStoryTypes => (dispatch) => {
...
+        if (tags[0] === SHOW_ALL) {
+          tags = SUMMARY_PAGE_SAVED_VARIANT_TAGS
+        }
         dispatch({
           type: RECEIVE_SAVED_VARIANT_TAGS,
-          updates: { [tag]: true },
+          updates: tags.reduce((acc, t) => ({ ...acc, [t]: true }), {}),
         })
+      } else if (gene) {
+        dispatch({ type: RECEIVE_SAVED_VARIANT_TAGS, updates: {} })
```
ShifaSZ, line 69: It is better to detect gene changes. But I don't know if it's worth doing.

seqr/views/apis/summary_data_api.py
hanars: we should be doing all the filtering on the backend for the summary data page. Why not keep the 2 distinct queries we had before so that we actually only retunr the correct variants:
```
tag_types = VariantTagType.objects.filter(name__in=tags, project__isnull=True)
saved_variant_models = SavedVariant.objects.filter(varianttag__variant_tag_type__in=tag_types)
```

ui/shared/utils/constants.js
```diff
@@ -1041,6 +1041,29 @@ export const REVIEW_TAG_NAME = 'Review'
 export const KNOWN_GENE_FOR_PHENOTYPE_TAG_NAME = 'Known gene for phenotype'
 export const DISCOVERY_CATEGORY_NAME = 'CMG Discovery Tags'
 export const MME_TAG_NAME = 'MME Submission'
+export const SUMMARY_PAGE_SAVED_VARIANT_TAGS = [
```
hanars, line 1044: it is confusing to have a page-specific constant in the global constants list, exspecially sicnce its used in other forms elsewhere. It is better to have a single `VARIANT_TAGS` constant and then if there need to be different tag options on different pages you can filter out the ones you don;t want for that specific page

ui/pages/SummaryData/reducers.js
```diff
@@ -43,21 +43,32 @@ export const loadSuccessStory = successStoryTypes => (dispatch) => {
...
   } else if (!gene) {
     return
   }
 
   dispatch({ type: REQUEST_SAVED_VARIANTS })
-  new HttpRequestHelper(`/api/summary_data/saved_variants/${tag}`,
+  new HttpRequestHelper(`/api/summary_data/saved_variants/${tags.join(';')}`,
     (responseJson) => {
       if (tag && !gene) {
+        if (tags[0] === SHOW_ALL) {
```
hanars, line 61: this case is impossible - you can't return all variants with no gene filter applied

ui/pages/SummaryData/reducers.js
```diff
@@ -43,21 +43,32 @@ export const loadSuccessStory = successStoryTypes => (dispatch) => {
...
+        }
         dispatch({
           type: RECEIVE_SAVED_VARIANT_TAGS,
-          updates: { [tag]: true },
+          updates: tags.reduce((acc, t) => ({ ...acc, [t]: true }), {}),
+        })
+      } else if (gene) {
+        dispatch({
+          type: RECEIVE_SAVED_VARIANT_TAGS,
+          updates: SUMMARY_PAGE_SAVED_VARIANT_TAGS.reduce((acc, t) => ({ ...acc, [t]: false }), {}),
```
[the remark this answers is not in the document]
hanars: why would we clear previously loaded data because we loaded more data?

seqr/views/apis/summary_data_api.py
ShifaSZ: > tag_types = VariantTagType.objects.filter(name__in=tags, project__isnull=True)
> saved_variant_models = SavedVariant.objects.filter(varianttag__variant_tag_type__in=tag_types)

I did some tests and found the result is the same. I can't find a Django filter that means our requirements. What I could do is add a Python filter afterward.

ui/pages/SummaryData/reducers.js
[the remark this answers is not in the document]
ShifaSZ: All the data loaded for the tags must be reloaded if the gene changes because the data differs for the genes with the same tags. I don't detect gene changes (I probably should). So clear tags loaded flag if the gene is not null to ensure correct results.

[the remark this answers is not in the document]
hanars: but if we previously loaded like all the variants for a different tag, why would clear that? We should only clear for the tag(s) that were loaded

seqr/views/apis/summary_data_api.py
ShifaSZ: It works. Can it be a concern about having many DB queries?

hanars: django converts that to 1 db query. Chaining `filter` statements does not execute a query

ui/shared/utils/constants.js
ShifaSZ: The dropdown tag options for the summary page and project page are different.
On the summary page, they are as the following figure:

On the project pages, they are:

hanars: there should not be a page specific constant in the shared constants file. If things are different on different pages, there should be page specific constants in the page specific constants files. If this constant is used in multiple places, then it should not be named `SUMMARY_PAGE_SAVED_VARIANT_TAGS` it should just be `SAVED_VARIANT_TAGS`. If it is just used in one place for one page, then it should be in a different file

ShifaSZ: Okay.

ShifaSZ: Move the constant into the SummaryData folder.

ui/pages/SummaryData/constants.js
```diff
@@ -1,9 +1,7 @@
-/* eslint-disable import/prefer-default-export */
-
```
ShifaSZ, line 2: There is a warning that the `eslint-disable` doesn't change anything because there are no `import/prefer-default-export` violations.

ui/pages/Project/components/SavedVariants.jsx
```diff
@@ -142,12 +143,15 @@ class BaseProjectSavedVariants extends React.PureComponent {
       project.variantTagTypes.map(type => type.category).filter(category => category),
     )]
 
-    const isCategory = categoryOptions.includes(newTag)
-    updateTableField('categoryFilter')(isCategory ? newTag : null)
+    const lastNewTag = newTag.length > 0 ? newTag[newTag.length - 1] : null
+    const isCategory = categoryOptions.includes(lastNewTag)
+    const [firstTag, ...otherTag] = newTag
+    const updatedTag = firstTag === ALL_FILTER || categoryOptions.includes(firstTag) ? otherTag : newTag
+    updateTableField('categoryFilter')(isCategory ? lastNewTag : null)
```
ShifaSZ, line 150: The behaviors include: if the latest selected option is a category, remove all selected options; otherwise if the previous (or the first) selected option is a category, remove the category option.
The issue is that when a category is selected, the category disappears from the option list.
No category is selected:


After a category (`Collaboration`) is selected:

ui/pages/SummaryData/reducers.js
```diff
@@ -43,8 +43,9 @@ export const loadSuccessStory = successStoryTypes => (dispatch) => {
 
 export const loadSavedVariants = ({ tag, gene = '' }) => (dispatch, getState) => {
   // Do not load if already loaded
+  const stateKey = tag && `${tag.split(TAG_URL_DELIMITER).sort().join(TAG_URL_DELIMITER)}${gene}`
   if (tag) {
-    if (getState().savedVariantTags[tag]) {
+    if (getState().savedVariantTags[stateKey]) {
```
ShifaSZ, line 48: Add the gene to the key for the state so that no states have to be removed when gene changed.

hanars: this seems to be adding a lof of complexity  that will be hard to maintain and have almost no performance improvement - we don't anticipate users toggling back and forth to the same gene, so refetching a gene is not an issue. Please explain why this was not needed before but is now needed because we are adding support for multiple tags

ui/pages/Project/components/SavedVariants.jsx
hanars: if a category is selected, all the tags in the category should be shown as selected and should be added to the list, the category itself should remain in the dropdown and clicking it again will have no effect

ui/pages/SummaryData/reducers.js
ShifaSZ: It was a bug before when the gene changed. The UI still displayed the variant data for the previous gene.

ui/pages/Project/components/SavedVariants.jsx
ShifaSZ: > the category itself should remain in the dropdown and clicking it again will have no effect

How about the category also disappear from the dropdown list? It is much easier to implement and it doesn't make sense to click on a selected category.

ui/pages/SummaryData/reducers.js
hanars: then aither a) make a separate PR to fix the bug and branch from that. or b) make a ticket to fix that bug and fix it later and not as part of this PR. This PR is really hard to understand because there is a lot in here that is unrelated to the functionality change requested in the ticket, and I can't review it effectively

ui/pages/SummaryData/components/SavedVariants.jsx
```diff
@@ -29,29 +29,7 @@ const FILTER_FIELDS = [
   VARIANT_PER_PAGE_FIELD,
 ]
 
-const TAG_OPTIONS = [
```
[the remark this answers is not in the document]
ShifaSZ: Oh, you are right. I used the tag options for the reducer to update the tag-loaded status before my latest update. We don't need it now.

hanars (changes requested):
This PR has gotten unwieldy, in large part because the scope has been really blown up. The initial ticket requested only that behavior for the summary data page change. In an attempt to simplify the change we decided to also change the functionality on the project page, but this has gotten way too complicated in how categories will be handled etc. I am restricting the scope of this back to only the summary page, ad the project page will continue to be a single select

ui/pages/SummaryData/reducers.js
ShifaSZ: I can leave the reducer unchanged for this PR and create a bug ticket.

seqr/views/apis/summary_data_api.py
```diff
@@ -87,8 +87,11 @@ def saved_variants_page(request, tag):
     if tag == 'ALL':
         saved_variant_models = SavedVariant.objects.exclude(varianttag=None)
     else:
-        tag_type = VariantTagType.objects.get(name=tag, project__isnull=True)
-        saved_variant_models = SavedVariant.objects.filter(varianttag__variant_tag_type=tag_type)
+        tags = tag.split(';')
+        tag_type = VariantTagType.objects.filter(name__in=tags, project__isnull=True)
```
hanars, line 91: should be `tag_types` plural, as it is a collection of models not a single model

seqr/views/apis/summary_data_api_tests.py
```diff
@@ -134,6 +134,11 @@ def test_saved_variants_page(self):
         expected_variant_guids.add('SV0000002_1248367227_r0390_100')
         self.assertSetEqual(set(response.json()['savedVariantsByGuid'].keys()), expected_variant_guids)
 
+        multi_tag_url = reverse(saved_variants_page, args=['Review;Tier 1 - Novel gene and phenotype'])
+        response = self.client.get('{}?gene=ENSG00000135953'.format(multi_tag_url))
+        self.assertEqual(response.status_code, 200)
+        self.assertSetEqual(set(response.json()['savedVariantsByGuid'].keys()), {'SV0000001_2103343353_r0390_100'})
```
hanars, line 140: use a constant for this guid

seqr/views/apis/summary_data_api_tests.py
```diff
@@ -231,14 +236,18 @@ class LocalSummaryDataAPITest(AuthenticationTestCase, SummaryDataAPITest):
     MANAGER_VARIANT_GUID = 'SV0000006_1248367227_r0004_non'
 
 
-def assert_has_expected_calls(self, users, skip_group_call_idxs=None):
+def assert_has_expected_calls(self, users, skip_group_call_idxs=None, has_ws_access_level_call=False):
```
hanars, line 239: this function is only used twice so theres no reused behavior for `mock_get_ws_access_level` at all anymore - it wold be better to take that check out of the helper and explicitly add the correct check for it after calling this helper

ui/shared/components/panel/variants/SavedVariants.jsx
```diff
@@ -76,14 +78,17 @@ class SavedVariants extends React.PureComponent {
...
+      variantExportConfig, loadVariants, error, multiple,
     } = this.props
     const { showAllFilters } = this.state
     const { familyGuid, variantGuid, tag } = match.params
 
-    const appliedTagCategoryFilter = tag || (variantGuid ? null : (tableState.categoryFilter || ALL_FILTER))
+    const tags = tag ? tag.split(TAG_URL_DELIMITER) : tag
+    const tagCategoryFilters = tags || (variantGuid ? [] : [(tableState.categoryFilter || ALL_FILTER)])
+    const firstAppliedTagCategoryFilter = tagCategoryFilters.length > 0 ? tagCategoryFilters[0] : null
+    const appliedTagCategoryFilter = multiple ? tagCategoryFilters : firstAppliedTagCategoryFilter
```
hanars, line 89: this functionality should not be shared, its really page specific and hard to understand whats happening here. Having a `dropdownValue` prop explicitly passed through here would make the page specific logic much clearer

ui/shared/components/panel/variants/SavedVariants.jsx
```diff
@@ -76,14 +78,17 @@ class SavedVariants extends React.PureComponent {
...
 
-    let shownFilters = (discoveryFilters && appliedTagCategoryFilter === DISCOVERY_CATEGORY_NAME) ?
+    let shownFilters = (discoveryFilters && firstAppliedTagCategoryFilter === DISCOVERY_CATEGORY_NAME) ?
```
hanars, line 91: similar to above, `discoveryFilters` is page specific so its confusing to have this logic here, it would be better to move this logic into the page-specific component and conditionally pass the correct `filters` prop

ui/shared/components/panel/variants/selectors.js
```diff
@@ -135,9 +136,19 @@ export const getPairedSelectedSavedVariants = createSelector(
     } else if (tag === MME_TAG_NAME) {
       pairedVariants = matchingVariants(pairedVariants, ({ mmeSubmissions = [] }) => mmeSubmissions.length)
     } else if (tag && tag !== SHOW_ALL) {
-      pairedVariants = matchingVariants(
-        pairedVariants, ({ tagGuids }) => tagGuids.some(tagGuid => tagsByGuid[tagGuid].name === tag),
-      )
+      const tags = tag.split(TAG_URL_DELIMITER)
```
hanars, line 139: Doesn't doing the split here means all the above logic for MME/notes would not work?

hanars (changes requested):

ui/shared/components/panel/variants/selectors.js
ShifaSZ: `MME Submission/Has Notes` won't work for multiple-tag filtering. They are not required since they are not on the summary data options list. But we should have a more systemic solution for handling tag/tags here.

seqr/views/apis/summary_data_api_tests.py
ShifaSZ: It is in a test assertion. Are you sure a constant is better than a string?

hanars: I was thinking since its reused, because its in the single tag filtering and the multi tag filtering, its better to share a constant. But if you prefer for consistency with the test file, we can leave it

ui/shared/components/panel/variants/selectors.js
ShifaSZ: Since we filter the tags in the backend and don't support the project saved-variant multiple-tag filtering, these changes are unnecessary.

ui/shared/components/panel/variants/SavedVariants.jsx
```diff
@@ -74,17 +72,15 @@ class SavedVariants extends React.PureComponent {
...
       tableSummaryComponent, loading, filteredVariantsCount, tagOptions, additionalFilter, updateTableField,
-      variantExportConfig, loadVariants, error,
+      variantExportConfig, loadVariants, error, multiple, getSelectedTag,
     } = this.props
     const { showAllFilters } = this.state
-    const { familyGuid, variantGuid, tag } = match.params
-
-    const appliedTagCategoryFilter = tag || (variantGuid ? null : (tableState.categoryFilter || ALL_FILTER))
+    const { familyGuid, variantGuid } = match.params
+    const optionValue = getSelectedTag(variantGuid ? null : (tableState.categoryFilter || ALL_FILTER))
```
ShifaSZ, line 81: Use a function prop to get the selected option value because the `tableState` is only available here.

ui/shared/components/panel/variants/selectors.js
```diff
@@ -130,7 +131,15 @@ export const getPairedSelectedSavedVariants = createSelector(
       return acc
     }, [])
 
-    if (tag === NOTE_TAG_NAME) {
+    const tags = (tag || '').split(TAG_URL_DELIMITER)
+    if (tags.length > 1) {
+      pairedVariants = matchingVariants(
+        pairedVariants, ({ tagGuids }) => {
+          const tagNames = tagGuids.map(tagGuid => tagsByGuid[tagGuid].name)
+          return tags.every(tagName => tagNames.includes(tagName))
```
ShifaSZ, line 139: In the end, we still need to add filtering for multiple tags.

ShifaSZ:
Each time when the user changes a selected tag, the summary data API is called twice. I think it is caused by the `ownProp.match.params` and `newParams` always being identical.
https://github.com/broadinstitute/seqr/blob/6b7a039380dbae91ff728d70774952188dad62d8/ui/pages/SummaryData/components/SavedVariants.jsx#L133-L143

ui/shared/components/panel/variants/selectors.js
hanars: Why? The combined DB filter you had implemented in the last round of review worked?

hanars (changes requested):

ui/shared/components/panel/variants/selectors.js
ShifaSZ: We have the state for the tag loaded but don't organize the variants per tag. This means all loaded variants are still available when the tag is changed. So we must filter the variants to display according to the current tag.

hanars: so don't set the "tag loaded" state when multiple tags are queried, and always refetch for multiple tags. We are going to run into untenable performance problems if we use AND instead of OR in the database query, so adding a tiny frontend optimization to prevent a couple extra backend requests is not worth the tradeoff of having actual needed queries time out because they return twice as much data instead of half

ShifaSZ: We don't refetch for every tag change because we have the tag-loaded state. But the loaded variants won't be removed when more variants are loaded. We need some code to filter out the variants to display in the front end. We could add/update the reducer to store the variants per tag(s). But we can also use the selectors to organize the variants. That's pretty much the current code.

hanars: for multiple tags, always load variants and then use the results of that load as the variants to show. Do not use any of the other variants that have been previously loaded. Put the results of a multi-tag search into a different field in the state and then use that if its present and don;t use the other state fields for the variant list

hanars: This should bevery similar to how `searchedVariants` are handled for variant searches

ShifaSZ: Then, we may be unable to use the `SavedVariants` shared component.

hanars: They already use different code to load saved variants, which means the code for where in the state to put the results after loading is already separate. Changing the selector to look first and see if theres data in the "use this if populated" part of the state, and then falling back to the current logic if not would be very straightforward to add

hanars: theres also already a `projectGuid` prop in the selector for what variants to show, so you could also make it conditional on that

ui/shared/components/panel/variants/selectors.test.js
```diff
@@ -67,7 +79,7 @@ test('getPairedFilteredSavedVariants', () => {
 
 test('getVisibleSortedSavedVariants', () => {
   const savedVariants = getVisibleSortedSavedVariants(
-    STATE_WITH_2_FAMILIES, { match: { params:  {} } }
+    STATE_WITH_2_FAMILIES, { project: { projectGuid: 'R0237_1000_genomes_demo' }, match: { params:  {} } },
```
hanars, line 82: we should also be testing without the project and with the new state field to check that that works correctly too

ui/pages/SummaryData/components/SavedVariants.jsx
```diff
@@ -67,11 +68,17 @@ TAG_OPTIONS.push({
 
 const PAGE_URL = '/summary_data/saved_variants'
 
-const getUpdateTagUrl =
-  (selectedTag, match) => `${PAGE_URL}/${selectedTag}${match.params.gene ? `/${match.params.gene}` : ''}`
+const getUpdateTagUrl = (selectedTag, match) => {
```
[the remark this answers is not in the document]
hanars: show all should not act like a multi-option - there should be no way to select "show all" and also select other tags. Is there a reason users should even have the show all option anymore on the summary page? Presumably, if they send a request with no tags we will return results for all tags, so just have "All" be something thats shown by default when the tag list is empty should be suufficient

ui/pages/SummaryData/reducers.js
```diff
@@ -54,13 +53,10 @@ export const loadSavedVariants = ({ tag, gene = '' }) => (dispatch, getState) =>
...
-      if (tag && !gene) {
-        dispatch({
-          type: RECEIVE_SAVED_VARIANT_TAGS,
-          updates: { [tag]: true },
-        })
-      }
-      dispatch({ type: RECEIVE_DATA, updatesById: responseJson })
+      dispatch({
+        type: RECEIVE_DATA,
+        updatesById: { ...responseJson, multiTagVariants: { [tag]: Object.keys(responseJson.savedVariantsByGuid) } },
```
hanars, line 58: `if (tag && !gene) ` still seems like an important conditional, why was it removed?

hanars: I proposed naming this `multiTagVariants` because I thought this would be special-cased only for multi variant tags, but this implementation is nicer. However, I would rename the key `savedVariantsByTag`

ui/shared/components/panel/variants/SavedVariants.jsx
```diff
@@ -74,17 +71,15 @@ class SavedVariants extends React.PureComponent {
...
       tableSummaryComponent, loading, filteredVariantsCount, tagOptions, additionalFilter, updateTableField,
-      variantExportConfig, loadVariants, error,
+      variantExportConfig, loadVariants, error, multiple, getSelectedTag,
     } = this.props
     const { showAllFilters } = this.state
-    const { familyGuid, variantGuid, tag } = match.params
-
-    const appliedTagCategoryFilter = tag || (variantGuid ? null : (tableState.categoryFilter || ALL_FILTER))
+    const { familyGuid, variantGuid } = match.params
+    const optionValue = getSelectedTag(tableState.categoryFilter)
```
hanars, line 80: notpick, but theres no need to assign this to a variable, it would be more readable if you directly assigned it to the prop

ui/shared/components/panel/variants/selectors.js
```diff
@@ -73,8 +73,13 @@ export const getPairedSelectedSavedVariants = createSelector(
   (state, props) => (props.project || {}).projectGuid,
   getVariantTagsByGuid,
   getVariantNotesByGuid,
+  state => state.savedVariantsByTag,
   (savedVariants, { tag, gene, familyGuid, analysisGroupGuid, variantGuid }, familiesByGuid, analysisGroupsByGuid,
-    projectGuid, tagsByGuid, notesByGuid) => {
+    projectGuid, tagsByGuid, notesByGuid, savedVariantsByTags) => {
+    if (!projectGuid) {
```
hanars, line 79: For the project page, why not use this if its available? You could skip the `else if (tag && tag !== SHOW_ALL)` block below if you start with this. Also, bailing out early like this totally skips the "paired" part of this "getPairedSelectedSavedVariants" helper for the summary page, which is not an okay change. If you just make the following change and then take out the additional tag filtering below, everything should work fine
```
let variants = savedVariantsByTags[tag] ? savedVariantsByTags[tag].map(guid => savedVariants[guid]) : Object.values(savedVariants)
```

hanars (changes requested):

ui/shared/components/panel/variants/selectors.js
ShifaSZ: Hmm, the project page could also use the cache. But we need to update the saved variant fetching API response handling first. It could fall back to using all loaded variants if the cache is unavailable, but we must keep the summary page from falling back when the cache is missing.

ShifaSZ: BTW. I don't understand what the `paired` means. It looks like pairing the variants with tags and notes. But for what and how?

ui/pages/SummaryData/reducers.js
ShifaSZ: > `if (tag && !gene)`  still seems like an important conditional, why was it removed?

Not do caching when there is a gene? It makes sense. I could add the condition to the value of the `savedVariantsByTag` generation. But it still has corner cases for switching from no gene to having a gene.

ui/shared/components/panel/variants/selectors.js
hanars: > But we need to update the saved variant fetching API response handling first. It could fall back to using all loaded variants if the cache is unavailable, but we must keep the summary page from falling back when the cache is missing.

It shouldn't be too hard to add a conditional check if that states populated, and then make requiring it conditional on the project

```
let variants = savedVariantsByTags[tag] ? savedVariantsByTags[tag].map(guid => savedVariants[guid]) : Object.values(savedVariants)
if (tag) {
  if (savedVariantsByTags[tag]) {
     variants = savedVariantsByTags[tag].map(guid => savedVariants[guid])
  } else if (!project) {
    return []
  }
}
```

> I don't understand what the paired means

We are displaying compound hets as pairs. If variants are tagged together, they are a comp het pair

ui/pages/SummaryData/reducers.js
hanars: How is that a corner case? The filtering in the selector will work fine for that

ShifaSZ: I want to add the `gene` to the key to solve the issues. So the `savedVariantsByTag` will become `savedVariantsByTagGene`.

hanars: the amount we are caching is overkill. This is too complex and it will save basically no time for the user. Do not add this additional level of caching. Just do this like search results, where theres a single state key with a list (for search its `searchedVariants`, here it will be something like `summaryDataVariants`) that just gets overwritten every time new summary variants are loaded, and then the selector checks that first and uses that if its there

bpblanken (approved):

hanars (approved):

Issue #3485, opened 2023-07-13
Interval local filter error when search on SV indices


ShifaSZ:
**Describe the bug**
When a variant search with an interval location filter searches on a structural variant index, the interval filter does not properly handle breakend and multi-chromosome events. 
For example, if the location filter is `chr20:45824623-45827445`, ~the search results include variants from all chromosomes other than chr20.~ the search includes breakend events whose start is before chromosome 20 (i.e chromosome 1) and whose end is after chromosome 20 (i.e. chromosome 21)

**Link to page(s) where bug is occurring**
This is an example search: https://seqr.broadinstitute.org/variant_search/results/3fd78f3581752a065fd9af0bb07f6b3a?page=1&sort=pathogenicity

**Scope of the bug**
All projects with structural variant indices.

**Screenshots**
For example, the first search result variant in the following picture is on chromosome 1, but the search interval filter is on chromosome 20.
![image](https://github.com/broadinstitute/seqr/assets/33550059/9161b5e0-b66d-455c-8529-450f64a858ad)

hanars:
this is now fixed

PR #3503, opened 2023-07-20
Fix SV localtion interval filter bug.

ShifaSZ:
See issue #3485.

### DIFF
--- a/seqr/utils/search/elasticsearch/es_search.py
+++ b/seqr/utils/search/elasticsearch/es_search.py
@@ -1413,8 +1413,9 @@ def _location_filter(genes, intervals, exclude_locations):
                     }
                 } for key in ['xpos', 'xstop']]
                 interval_q = _build_or_filter('range', range_filters)
-                interval_q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop}) &
-                              Q('range', xpos={'gte': get_xpos(interval['chrom'], 0)}) & Q('range', xstop={'lte': get_xpos(interval['chrom'], MAX_POS)})
+                interval_q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop}) & \
+                              Q('range', xpos={'gte': get_xpos(interval['chrom'], MIN_POS)}) & \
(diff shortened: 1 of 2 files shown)


seqr/utils/search/elasticsearch/es_search.py
```diff
@@ -1413,7 +1413,8 @@ def _location_filter(genes, intervals, exclude_locations):
                     }
                 } for key in ['xpos', 'xstop']]
                 interval_q = _build_or_filter('range', range_filters)
-                interval_q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop})
+                interval_q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop}) &\
+                              Q('script', script="doc['xpos'].value / 1000000000 == doc['xstop'].value / 1000000000")
```
hanars, line 1417: script queries are really inefficient, so in general we don't want to use them. I think a better approach would be jusy to add additional checks here that `xstart` and `xstop` are both on the  same chromosome as `interval['chrom']`. You can do this by just adding the following to the current query:
```suggestion
                interval_q |= Q('range', xpos={'lte': xstart}) & Q('range', xstop={'gte': xstop}) &
                              Q('range', xpos={'gte': get_xpos(interval['chrom'], 0)}) & Q('range', xstop={'lte': get_xpos(interval['chrom'], MAX_POS)})
```

hanars (changes requested):

seqr/utils/search/elasticsearch/es_search.py
ShifaSZ: Sounds good. BTW. Do we need to handle the case that the `end` is on a different chromosome and the interval locates after the `xstart` on the same chromosome? Or the interval is before the `xstop` on the same chromosome of `xstop`?

hanars: No we don't need any special handling for that. If this is a breakend event and either xpos or xstop falls directly inside a filtered interval, then the `range_filters` above will capture that and those variants will be returned, which is the desired behavior. And otherwise if they are on different chromosomes we do not want them returned, and this change will handle that

hanars (approved):

Issue #3506, opened 2023-07-21
Update and harmonize outlier and gene search parameters 


lynnpais:
**Is your feature request related to a problem? Please describe.**
We would like to update and harmonize outlier gene search parameters across seqr. 

**Describe the solution you'd like**
Use the following search parameters :
Inheritance from ‘All’ to ‘Any affected’
Frequency of 0.03 in gnomAD and 0.03 for this callset 
Call quality GQ of 40 and AB of 10. No VQSR.

The proposed changes pertain to the following:
1. RNASeq page - Search for variants in [expression] outlier genes
2. RNAseq results page - Search for variants in [splice junction] outlier regions
3. RNAseq results page - Junction
4. Saved variants page - Gene search (change text in hover over to say <3%. It is currently 10%). 

**Describe alternatives you've considered**
NA

**Additional context**
Discussed in the analyst team meeting.

hanars:
"No VQSR" translates to "Pass Variants Only" under the quality filter

This involves changing the gene search everywhere is appears in seqr, so behavior can be shared

lynnpais:
By "No VQSR" we mean Show all variants, NOT pass only.
<img width="301" alt="Screenshot 2023-07-21 at 12 58 40 PM" src="https://github.com/broadinstitute/seqr/assets/37872186/4a3ef5d1-4978-4a54-a14b-e5de9a8f6f15">

ShifaSZ:
> Inheritance from ‘All’ to ‘Any affected’

The initial search for the gene search link has an `any_affected` inheritance mode. But it is converted to `all` mode because it doesn't have an inheritance filter field. What should the initial inheritance filter be? @hanars If we select the `Any Affected` on the search page, the inheritance filter is `{A: "has_alt"}`. Should we use the filter in the initial search?

ShifaSZ:
Okay. I'll update [these codes](https://github.com/broadinstitute/seqr/blob/df88da531228363944381504b65944b29e57d9e5/ui/pages/Search/components/VariantSearchFormContent.jsx#L67C1-L72C1) to accept a mode with an inheritance filter field.

ShifaSZ:
> the search link should be posting a search body directly to the backend.

The search link for the search button opens the search page and doesn't post to the backend. Only when the users click on the `Search` button on the search page the search request is sent to the backend.

hanars:
[...]

PR #3509, opened 2023-07-24
Update search params for outlier and gene searches.

ShifaSZ:
See issue #3506.


ShifaSZ:
The UI displayed:
image

ui/shared/components/buttons/SearchResultsLink.jsx
```diff
@@ -48,15 +48,21 @@ const ConnectedSearchResultsLink = connect(null, mapDispatchToProps)(SearchResul
...
 
-const getGeneSearchProps = af => ({
-  inheritance: { mode: ANY_AFFECTED },
-  freqs: FREQUENCIES.filter(({ name }) => name !== THIS_CALLSET_FREQUENCY && name !== SV_CALLSET_FREQUENCY).reduce(
-    (acc, { name }) => ({ ...acc, [name]: { af } }), {},
-  ),
+const getGeneSearchProps = (af, af2) => ({
+  inheritance: { mode: ANY_AFFECTED, filter: {} },
+  freqs: {
+    ...FREQUENCIES.filter(({ name }) => name !== THIS_CALLSET_FREQUENCY && name !== SV_CALLSET_FREQUENCY).reduce(
```
hanars, line 54: I didn't catch it when the ticket was made, but the intention here was not to have different frequency handling for different popualtions. Lynn has now clarified on the ticket that we should not filter on topmed at all, and everything else in `FREQUENCIES` is either a gnomad or a callset frequency and should have the new 0.03 value. It would be cleaner therefore to filter for "not topmed" 

Also, I'm still waiting for the cnfirmation form Lynn, but I think we can get rid of the ` RareGeneSearchLink` entirely (which is only used in `PhenotypePrioritizedGenes`) and have consistant gene search behavior across seqr

hanars (changes requested):

ShifaSZ:
New initial search frequency filters for those links:
image

ui/shared/components/buttons/SearchResultsLink.jsx
```diff
@@ -48,15 +48,14 @@ const ConnectedSearchResultsLink = connect(null, mapDispatchToProps)(SearchResul
...
-    (acc, { name }) => ({ ...acc, [name]: { af } }), {},
-  ),
-})
-
-const INITIAL_GENE_SEARCH = getGeneSearchProps(0.1)
-const RARE_GENE_SEARCH = getGeneSearchProps(0.01)
+const INITIAL_GENE_SEARCH = {
+  inheritance: { mode: ANY_AFFECTED, filter: {} },
+  freqs: {
+    ...FREQUENCIES.filter(({ name }) => name.startsWith('gnomad_') || name === THIS_CALLSET_FREQUENCY).reduce(
```
hanars, line 54: per my review, we should be filtering out topmed, not filtering for literally everything else

hanars (changes requested):

ui/shared/components/buttons/SearchResultsLink.jsx
ShifaSZ: Do you mean all freq filters other than topmed have a threshold of 0.03?

hanars: yes

ui/shared/components/buttons/SearchResultsLink.jsx
```diff
@@ -48,15 +48,14 @@ const ConnectedSearchResultsLink = connect(null, mapDispatchToProps)(SearchResul
...
-    (acc, { name }) => ({ ...acc, [name]: { af } }), {},
-  ),
-})
-
-const INITIAL_GENE_SEARCH = getGeneSearchProps(0.1)
-const RARE_GENE_SEARCH = getGeneSearchProps(0.01)
+const INITIAL_GENE_SEARCH = {
+  inheritance: { mode: ANY_AFFECTED, filter: {} },
+  freqs: {
+    ...FREQUENCIES.filter(({ name }) => name !== 'topmed').reduce(
```
hanars, line 54: all strings that are used more than once need to be constants. Similar to `THIS_CALLSET_FREQUENCY`, the constant needs to be defined in the constants file, used in the definition for the `FREQUENCIES`, and then imported here

hanars (changes requested):

ui/pages/Search/components/VariantSearchFormContent.jsx
```diff
@@ -57,7 +58,7 @@ const SAVED_SEARCH_FIELD = {
 }
 
 const INHERITANCE_PANEL = {
-  name: 'inheritance',
+  name: INHERITANCE_NAME,
```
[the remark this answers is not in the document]
ShifaSZ: The `inheritance` is tricky. How about `freqs`? I believe it has been updated completely.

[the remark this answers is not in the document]
hanars: leave this file alone except for the changes needed for this feature request. The changes needed for this feature request are not to filter on topmed. Therefore, make a constant for topmed, remove that from the filters to apply, and leave the rest of this alone

hanars (approved):

PR #3510, opened 2023-07-24
Saved variant timeout

hanars:
The issue appears to be that the increase in the number of TPM models now that we loaded TPM==0 has increased the size of the response and we are actually losing most of the time serializing and deserializing the response. This updates the behavior so we only return TPM genes for a family if the family has a variant in that gene, instead of if any family has a variant in that gene. For RGP this redoes the response size by over 100Mb

### DIFF
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -229,7 +229,7 @@ def get_variants_response(request, saved_variants, response_variants=None, add_a
     rna_tpm = None
     if include_individual_gene_scores:
         response['rnaSeqData'] = _get_rna_seq_outliers(genes.keys(), family_genes.keys())
-        rna_tpm = _get_family_has_rna_tpm(family_genes, response)
+        rna_tpm = _get_family_has_rna_tpm(family_genes, genes.keys())
         response['phenotypeGeneScores'] = get_phenotype_prioritization(family_genes.keys(), gene_ids=genes.keys())
 
     if add_all_context or request.GET.get(LOAD_PROJECT_TAG_TYPES_CONTEXT_PARAM) == 'true':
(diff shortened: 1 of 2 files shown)


seqr/views/utils/variant_utils.py
```diff
@@ -91,6 +92,8 @@ def _saved_variant_genes_transcripts(variants):
             for gene_id, transcripts in var.get('transcripts', {}).items():
                 gene_ids.add(gene_id)
                 transcript_ids.update([t['transcriptId'] for t in transcripts if t.get('transcriptId')])
+            for family_guid in var['familyGuids']:
+                family_genes[family_guid].update(var.get('transcripts', {}).keys())
```
ShifaSZ, line 96: Do we need to include the families who have no transcripts?

seqr/views/utils/variant_utils.py
```diff
@@ -152,12 +155,15 @@ def get_phenotype_prioritization(family_guids, gene_ids=None):
...
     tpm_family_genes = RnaSeqTpm.objects.filter(
-        sample__individual__family__guid__in=families_by_guid.keys(), gene_id__in=gene_ids,
+        sample__individual__family__guid__in=family_genes.keys(), gene_id__in=gene_ids,
     ).values('sample__individual__family__guid').annotate(genes=ArrayAgg('gene_id', distinct=True))
-    for agg in tpm_family_genes:
-        families_by_guid[agg['sample__individual__family__guid']]['tpmGenes'] = agg['genes']
+    return {
+        agg['sample__individual__family__guid']: {'tpmGenes': [
+            gene for gene in agg['genes'] if gene in family_genes[agg['sample__individual__family__guid']]
+        ]} for agg in tpm_family_genes
```
ShifaSZ, line 165: Should we filter out the families with no genes in the `family_genes`?

ShifaSZ (changes requested):

seqr/views/utils/variant_utils.py
hanars: yes overall, as this is used for the the list of families for filtering which projects to return etc. However, I can add an extra filter before we get the family list for rna data, as theres no reason to include rna data for families with no genes

hanars: added this below

ShifaSZ (approved):

PR #3512, opened 2023-07-25
shared utilities for hail backend tests

hanars:
moves constants and helper method for hail search tests into the hail search folder


ShifaSZ:
The folder location of the `test_utils.py` is weird to me. Will it be shared among the hail search implementation (under seqr/utils/search) and the hail search backend service (under hail_search)?

hanars:
> Will it be shared among the hail search implementation (under seqr/utils/search) and the hail search backend service (under hail_search)?

Yes, and due to how django apps are configured and the various tests are set up, it has to live where it does

PR #3513, opened 2023-07-26
Hail backend: variant response

hanars:
Adds the initial hail search backend code which reads the relevant hail tables and returns properly formatted variants. Does not implement most sorts or any filtering


hail_search/constants.py
```diff
@@ -0,0 +1,13 @@
+GENOME_VERSION_GRCh38_DISPLAY = 'GRCh38'
```
ShifaSZ, line 1: Is it possible to share the same constants with the ES search? Same for the rest of the constants?

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,322 @@
...
+            enum_array = hl.array(field_enum)
+            if is_array:
+                annotations[f'{field}s'] = value[value_field].map(lambda v: enum_array[v])
+            else:
+                annotations[field] = enum_array[value[value_field]]
+
+        value = value.annotate(**annotations)
+        if annotate:
+            annotations = annotate(value, enum, ht_globals)
+            value = value.annotate(**annotations)
```
ShifaSZ, line 119: The term `annotation` is used for both variant search and hail. It is confusing. Should we use `variant_annotation` or another name for the search filter?

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def __init__(self, data_type, sample_data, genome_version, sort=XPOS_SORT_KEY, num_results=100, **kwargs):
+        self._genome_version = genome_version
+        self._sort = sort
+        self._num_results = num_results
+
+        self._load_filtered_table(data_type, sample_data, **kwargs)
+
+    def _load_filtered_table(self, data_type, sample_data, **kwargs):
+        self._ht = self.import_filtered_table(data_type, sample_data, **kwargs)
```
ShifaSZ, line 135: We may not need a function if it is used just once in the `init`.

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,322 @@
...
+    def import_filtered_table(self, data_type, sample_data, **kwargs):
+        tables_path = f'{DATASETS_DIR}/{self._genome_version}/{data_type}'
+
+        family_samples = defaultdict(list)
+        project_samples = defaultdict(list)
+        for s in sample_data:
+            family_samples[s['family_guid']].append(s)
+            project_samples[s['project_guid']].append(s)
+
+        logger.info(f'Loading {data_type} data for {len(family_samples)} families in {len(project_samples)} projects')
```
ShifaSZ, line 146: The `project_samples` is solely for stats. A per-project counter should be enough.

hail_search/search.py
```diff
@@ -0,0 +1,15 @@
...
+
+def search_hail_backend(request):
+    sample_data = request.pop('sample_data', {})
+
+    data_types = list(sample_data.keys())
+    single_data_type = data_types[0] if len(data_types) == 1 else None
+
+    sample_data = sample_data[single_data_type]
+    data_type = single_data_type
+    query_cls = QUERY_CLASS_MAP[single_data_type]
```
ShifaSZ, line 12: It will be a key error when `single_data_type` is None. Should we report an error if the length of `data_types` is not 1?

hail_search/test_search.py
```diff
@@ -1,7 +1,286 @@
 from aiohttp.test_utils import AioHTTPTestCase
 
+from hail_search.test_utils import get_hail_search_body, FAMILY_2_VARIANT_SAMPLE_DATA, FAMILY_2_MISSING_SAMPLE_DATA
 from hail_search.web_app import init_web_app
 
+VARIANT_ID_1 = '1-10439-AC-A'
+VARIANT1 = {
```
ShifaSZ, line 7: Should we re-use the test data constants directly from the ES search? The pros include less typing and fewer mistakes. One of the main cons is coupling the tests with the ES searches.

hail_search/test_search.py
```diff
@@ -13,3 +292,23 @@ async def test_status(self):
             self.assertEqual(resp.status, 200)
             resp_json = await resp.json()
         self.assertDictEqual(resp_json, {'success': True})
+
+    async def test_search(self):
+        search_body = get_hail_search_body(sample_data=FAMILY_2_VARIANT_SAMPLE_DATA)
+        async with self.client.request('POST', '/search', json=search_body) as resp:
+            self.assertEqual(resp.status, 200)
+            resp_json = await resp.json()
```
ShifaSZ, line 300: Can we test the search class directly without going through the web server? So that no synchronization is needed for the search tests.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,322 @@
...
+    GNOMAD_GENOMES_FIELD, XPOS_SORT_KEY, GENOME_VERSION_GRCh38_DISPLAY
+
+DATASETS_DIR = os.environ.get('DATASETS_DIR', '/hail_datasets')
+
+logger = logging.getLogger(__name__)
+
+
+def _to_camel_case(snake_case_str):
+    converted = snake_case_str.replace('_', ' ').title().replace(' ', '')
+    return converted[0].lower() + converted[1:]
```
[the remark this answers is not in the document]
hanars: no. The docker image for the hail backend o=will only contain this folder, so all functionality needs to be implemented here

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def _format_transcript_args(self):
+        return {
+            'format_value': lambda value: value.rename({k: _to_camel_case(k) for k in value.keys()}),
+        }
+
+    @staticmethod
+    def _enum_field(value, enum, ht_globals=None, annotate=None, format_value=None, drop_fields=None, **kwargs):
+        annotations = {}
+        drop = [] + (drop_fields or [])
```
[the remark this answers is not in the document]
hanars: no we do not want to modify the `drop` variable, this is the same as writing `drop = deepcopy(drop_fields) or []`, but I though this was neater

hail_search/hail_search_query.py
hanars: per my comment in standup, this is a paired down version of what I already wrote, which includes support for multiple data types. This function is much longer in the final implementation and overridden by subclasses

hail_search/hail_search_query.py
hanars: I do not understand. How is creating a tracker better than logging as we go so you can check the logs and see what is happeneing?

hail_search/search.py
hanars: No, this is simply functionality thats not implemented yet but will be before we go live. It seemed unnecessary and too much work to add a bunch of error handling that we would immediately remove in a subsequent PR, when the end result would still be that search wouldn't work if you tried it. As I said in standup, this is only a partial implementation. The goal is to get stuff in in manageable chunks. If we are unwilling to merge code that does not have full functionality, we would have to review thousands of lines of code at once, which would result in worse review and worse code

hail_search/test_search.py
hanars: But that would be a substantially worse test suite. We don't care at all how the search class behaves, we only care about how the actual API requests that seqr will be using behave. In the main seqr repo we have been slowly phasing out testing utility functions directly in favor of testing them through API endpoints for that exact reason, and in fact we have had bugs in the past couple months that slipped through because we tested the function directly and not through the API. The async behavior here is part of the well supported `AioHTTPTestCase` and adds almost no time to the test suite as the fake responses are pretty quick

hail_search/search.py
ShifaSZ: Can you add a comment about something to do before going live?

hanars: I would rather not - there would be about 200 comments. Nothing works right now - none of the filters, multi family search, multi data type search, multi project search. The whole file would be comments. This is not a practical request

hanars: if you are worried about missing stuff, at the very end I will have moved an entire branch over and I can check that there is nothing left on that branch thats different from dev, in order to confirm all fucntionality was transferred. But putting in hundereds of comments would be less accurate than that approach and a huge amount of my time

ShifaSZ (approved):
LGTM

bpblanken (commented):
Just some questions, in general everything here looks solid!

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,323 @@
...
+        ht = families_ht.annotate(**annotation_ht_query_result)
+        # Add globals
+        ht = ht.join(hl.read_table(annotations_ht_path).head(0).select().select_globals(*self.GLOBALS), how='left')
+
+        ht = ht.transmute(
+            genotypes=ht.family_entries.flatmap(lambda x: x).filter(
+                lambda gt: hl.is_defined(gt.individualGuid)
+            ).map(lambda gt: gt.select(
+                'sampleId', 'individualGuid', 'familyGuid',
+                numAlt=hl.if_else(hl.is_defined(gt.GT), gt.GT.n_alt_alleles(), -1),
```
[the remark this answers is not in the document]
hanars: I'd rather not. This calculation is pretty fast, we have to iterate the entries anyways so its not like we would really save much time, we still want the `GT` field for querying so we'd be storing duplicate data, and having a larger table and larger data slows down our read time for the whole table which I think might be more costly than annotating a couple hundred rows.

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,323 @@
...
+            project_samples[s['project_guid']].append(s)
+
+        logger.info(f'Loading {data_type} data for {len(family_samples)} families in {len(project_samples)} projects')
+        if len(family_samples) == 1:
+            family_guid, family_sample_data = list(family_samples.items())[0]
+            family_ht = hl.read_table(f'{tables_path}/families/{family_guid}.ht')
+            families_ht = self._add_entry_sample_families(family_ht, family_sample_data)
+            families_ht = families_ht.select_globals()
+        else:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
hanars: Per the discussion here, this was just a placehloder for functionality that will be implemented before go live: https://github.com/broadinstitute/seqr/pull/3513/files#r1277635482

It seems adding error handling or not adding error handling are equally confusing, but I'll just take this out until we actually add the functionality

hail_search/hail_search_query.py
```diff
@@ -0,0 +1,323 @@
...
+        'ref': lambda r: r.alleles[0],
+        'alt': lambda r: r.alleles[1],
+        'genotypeFilters': lambda r: hl.str(' ,').join(r.filters),
+        'mainTranscriptId': lambda r: r.sorted_transcript_consequences.first().transcript_id,
+    }
+    BASE_ANNOTATION_FIELDS.update(BaseHailTableQuery.BASE_ANNOTATION_FIELDS)
+    ENUM_ANNOTATION_FIELDS = {
+        'clinvar': {'annotate_value': lambda value, enum, ht_globals: {
+            'conflictingPathogenicities': value.conflictingPathogenicities.map(
+                lambda p: p.annotate(pathogenicity=hl.array(enum['pathogenicity'])[p.pathogenicity_id]).drop(
```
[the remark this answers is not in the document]
hanars: I can actually just use the `_enum_field` helper method, which will be clearer

bpblanken (approved):

PR #3518, opened 2023-07-27
Hail backend - multiple projects

hanars:


hail_search/test_search.py
```diff
@@ -16,22 +88,44 @@ async def test_status(self):
...
         self.assertSetEqual(set(resp_json.keys()), {'results', 'total'})
-        self.assertEqual(resp_json['total'], 4)
-        results = [VARIANT1, VARIANT2, VARIANT3, VARIANT4]
+        self.assertEqual(resp_json['total'], len(results))
         self.assertListEqual(
-            [v['variantId'] for v in resp_json['results']], [v['variantId'] for v in results]
+            [v['variantId'] for v in resp_json['results']], [v['variantId'] for v in results],
+        )
+        for i, result in enumerate(resp_json['results']):
+            self.assertDictEqual(result, results[i])
```
ShifaSZ, line 102: The ` assertListEqual(resp_json['results], results)` should work.

hail_search/hail_search_query.py
```diff
@@ -146,8 +146,32 @@ def import_filtered_table(self, data_type, sample_data, **kwargs):
...
+            families_ht = filtered_project_hts[0]
+            default_entries = hl.empty_array(families_ht.family_entries.dtype.element_type)
+            for project_ht in filtered_project_hts[1:]:
+                families_ht = families_ht.join(project_ht, how='outer')
+                families_ht = families_ht.select(
+                    filters=families_ht.filters.union(families_ht.filters_1),
+                    family_entries=hl.bind(
+                        lambda a1, a2: a1.extend(a2),
+                        hl.or_else(families_ht.family_entries, default_entries),
+                        hl.or_else(families_ht.family_entries_1, default_entries),
```
ShifaSZ, line 172: Where are the `filters_1` and `family_entries_1` fields from?

ShifaSZ (changes requested):

hail_search/test_search.py
hanars: yeah but its impossible to understand the results of the error for a list of dicts - its not clear whats actually different. This is barely less efficient in terms of runtime, but when there is an issue the error is actually helpful

hail_search/hail_search_query.py
hanars: its what happens when you join tables in hail that have the same field names, instead of trying to combine them hail moves them to a new field. Theres no way to avoid it

hail_search/hail_search_query.py
```diff
@@ -168,6 +192,15 @@ def import_filtered_table(self, data_type, sample_data, **kwargs):
 
         return ht
 
+    def _filter_entries_table(self, ht, sample_data):
+        ht = self._add_entry_sample_families(ht, sample_data)
+
+        ht = ht.annotate(family_entries=ht.family_entries.map(
+            lambda entries: hl.or_missing(entries.any(lambda x: x.GT.is_non_ref()), entries))
```
bpblanken, line 199: I think this filtration is unneeded now that the pipeline is doing this right?

bpblanken (commented):
logic looks right to me!

hail_search/hail_search_query.py
hanars: no, its still needed because there could be many families in a project but a user has only searched for some of them

ShifaSZ (approved):

PR #3519, opened 2023-07-27
Hail backend filter inheritance

hanars:


hail_search/test_search.py
```diff
@@ -117,6 +128,29 @@ async def test_multi_project_search(self):
             sample_data=MULTI_PROJECT_SAMPLE_DATA,
         )
 
+    async def test_inheritance_filter(self):
+        await self._assert_expected_search(
+            [VARIANT1, VARIANT2, MULTI_FAMILY_VARIANT, VARIANT4], inheritance_mode='any_affected', omit_sample_type='SV_WES',
```
ShifaSZ, line 133: You may want to add a test for both inheritance_filter and inheritance_mode are undefined.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
```diff
@@ -22,6 +23,15 @@ def _to_camel_case(snake_case_str):
 
 class BaseHailTableQuery(object):
 
+    GENOTYPE_QUERY_MAP = {
+        REF_REF: lambda gt: gt.is_hom_ref(),
+        REF_ALT: lambda gt: gt.is_het(),
```
bpblanken, line 28: Would you mind giving some context about why both `REF_ALT` and `COMP_HET_ALT` are both `gt.is_het()`?  I might have expected one `is_het_non_ref` and one `is_het_ref`.

hail_search/hail_search_query.py
```diff
@@ -243,13 +255,72 @@ def _add_entry_sample_families(cls, ht, sample_data):
...
+            lambda entries: hl.or_missing(entries.any(any_valid_entry), entries))
+        )
+
+        if not (inheritance_filter or inheritance_mode):
+            return ht
+
+        if inheritance_mode == X_LINKED_RECESSIVE:
+            x_chrom_interval = hl.parse_locus_interval(
+                hl.get_reference(self._genome_version).x_contigs[0], reference_genome=self._genome_version)
+            ht = ht.filter(self.get_x_chrom_filter(ht, x_chrom_interval))
```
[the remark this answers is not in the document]
hanars: the filtering works differently for SVs, which is why this is abstracted out into a helper function. I'm not sure how "contains" is clearer as a function name - that implies that a set or list contains something, not that an interval overlaps a locus. For SVs, `contains` will not be used at all in the filter itself. The function returns the filter, so `get_x_chrom_filter` seems like a pretty accurate name to me, but if you have a better suggestion I would be open to it

hail_search/test_search.py
hanars: thats literally every other test in the file

hail_search/hail_search_query.py
hanars: for gCNV the behavior is different for `COMP_HET_ALT` and its overridden for that class. But if its confusing I can take this out for now and add it back when I actually add gCNV support

ShifaSZ (approved):
LGTM

PR #3520, opened 2023-07-28
Hail backend filter quality

hanars:


hail_search/hail_search_query.py
```diff
@@ -318,6 +331,39 @@ def _valid_genotype_family_entries(cls, entries, gentoype_entry_indices, genotyp
...
         return hl.or_missing(is_valid, entries)
 
+    @classmethod
+    def _get_genotype_passes_quality_filter(cls, quality_filter):
+        affected_only = quality_filter.get('affected_only')
+        passes_quality = None
+        for filter_k, value in quality_filter.items():
+            field = cls.GENOTYPE_FIELDS.get(filter_k.replace('min_', ''))
+            if field and value:
+                passes_quality = cls._get_genotype_passes_quality_field(field, value, affected_only, passes_quality)
```
ShifaSZ, line 341: Would it be more straightforward to use an array of functions as the following?
```
passes_qualities = []
for ...:
    ...
    passes_qualities.append(cls. _get_genotype_passes_quality_field(...))

passes_quelity = None if not passes_qualities else lambda gt: passes_qualities.all(lambda: field_filter: field_filter(gt))
```
And we don't need the cascading filter_func parameter for cls._get_genotype_passes_quality_field function below.

hail_search/hail_search_query.py
```diff
@@ -318,6 +331,39 @@ def _valid_genotype_family_entries(cls, entries, gentoype_entry_indices, genotyp
...
+
+        return passes_quality
+
+    @classmethod
+    def _get_genotype_passes_quality_field(cls, field, value, affected_only, filter_func):
+        field_config = cls.QUALITY_FILTER_FORMAT.get(field) or {}
+        if field_config.get('scale'):
+            value = value / field_config['scale']
```
ShifaSZ, line 349: The `cls.QUALITY_FILTER_FORMAT` fields, such as `scale` and `override`, are used multiple times. It may be better to define constants for them.

hail_search/hail_search_query.py
```diff
@@ -203,13 +204,25 @@ def import_filtered_table(self, data_type, sample_data, **kwargs):
...
+    def _filter_entries_table(self, ht, sample_data, inheritance_mode=None, inheritance_filter=None, quality_filter=None,
+                              **kwargs):
         ht, sample_id_family_index_map = self._add_entry_sample_families(ht, sample_data)
 
         ht = self._filter_inheritance(
             ht, inheritance_mode, inheritance_filter, sample_data, sample_id_family_index_map,
         )
 
+        quality_filter = quality_filter or {}
+        if quality_filter.get('vcf_filter'):
```
ShifaSZ, line 216: It seems to me that the `get` function can be `pop`. Otherwise, the `vcf_filter` field will be an unnecessary round of iterating in `_get_genotype_passes_quality_filter`.

hail_search/hail_search_query.py
```diff
@@ -318,6 +331,39 @@ def _valid_genotype_family_entries(cls, entries, gentoype_entry_indices, genotyp
...
+
+    @classmethod
+    def _get_genotype_passes_quality_field(cls, field, value, affected_only, filter_func):
+        field_config = cls.QUALITY_FILTER_FORMAT.get(field) or {}
+        if field_config.get('scale'):
+            value = value / field_config['scale']
+
+        def passes_quality_field(gt):
+            is_valid = (gt[field] >= value) | hl.is_missing(gt[field])
+            if field_config.get('override'):
```
ShifaSZ, line 353: Use `'override' in field_config` instead of the `get` function?

hail_search/test_search.py
```diff
@@ -428,6 +428,33 @@ async def test_inheritance_filter(self):
         await self._assert_expected_search(
             [VARIANT2, VARIANT3], inheritance_filter=gt_inheritance_filter, sample_data=FAMILY_2_VARIANT_SAMPLE_DATA)
 
+    async def test_quality_filter(self):
+        await self._assert_expected_search(
+            [VARIANT1, VARIANT2, MULTI_FAMILY_VARIANT], quality_filter={'vcf_filter': 'pass'}, omit_sample_type='SV_WES',
+        )
+
+        await self._assert_expected_search(
+            [VARIANT2, MULTI_FAMILY_VARIANT], quality_filter={'min_gq': 40}, omit_sample_type='SV_WES',
```
ShifaSZ, line 437: We may need another test for combining the `min_x` filter and `vcf_filter`.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: the additional iterating is necessary regardless, because there may be quality filters that are data type specific to a different data type. So for instance, if there is an SV-specific quality filter we do not want to apply it here. And then since we are iterating anyways, theres no need to modify the input unnecessarily

hail_search/hail_search_query.py
```diff
@@ -318,6 +331,39 @@ def _valid_genotype_family_entries(cls, entries, gentoype_entry_indices, genotyp
         )
         return hl.or_missing(is_valid, entries)
 
+    @classmethod
+    def _get_genotype_passes_quality_filter(cls, quality_filter):
+        affected_only = quality_filter.get('affected_only')
```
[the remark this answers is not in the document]
hanars: same response as above

hail_search/hail_search_query.py
hanars: thats not really any more readable, and would add a weird edge case if `override` has null. Not sure what the advantage is?

hail_search/hail_search_query.py
```diff
@@ -317,6 +331,46 @@ def _valid_genotype_family_entries(cls, entries, gentoype_entry_indices, genotyp
...
+        passes_quality_filters = []
+        for filter_k, value in quality_filter.items():
+            field = cls.GENOTYPE_FIELDS.get(filter_k.replace('min_', ''))
```
bpblanken, line 339: Is there any reason to format these with `min` if the only applicable filter is `>=`?  We'd never want variants "under" a certain quality right?

bpblanken: Unless this is just how the parameters are passed from the frontend, which would make sense.

hanars: its how they are passed form the frontend, yeah

hail_search/hail_search_query.py
ShifaSZ: I thought that the `override` was a configuration field. Configuring it to `null` or `false` doesn't make sense because it needs to be a function. The `'key' in dict` checks the key while the `get` function uses both key and value. It saves a bit of performance.
Now, using the `namedtuple` class is all good.

PR #3522, opened 2023-07-28
Hail backend frequency filter

hanars:


hail_search/hail_search_query.py
```diff
@@ -250,6 +250,40 @@ def _missing_entry(entry):
...
+            if freqs.get('hh') is not None:
+                hom_field = pop_config['hom']
+                hemi_field = pop_config['hemi']
+                if hom_field:
+                    pop_filters.append(pop_expr[hom_field] <= freqs['hh'])
+                if hemi_field:
+                    pop_filters.append(pop_expr[hemi_field] <= freqs['hh'])
+
+            if pop_filters is not None:
+                pop_filter = pop_filters[0]
```
ShifaSZ, line 282: The condition for the if statement should be `len(pop_filters) > 0`. Otherwise, `pop_filters[0]` can be out of range error.

hail_search/test_search.py
```diff
@@ -117,6 +117,36 @@ async def test_multi_project_search(self):
             sample_data=MULTI_PROJECT_SAMPLE_DATA,
         )
 
+    async def test_frequency_filter(self):
+        await self._assert_expected_search(
+            [VARIANT1, VARIANT4], frequencies={'seqr': {'af': 0.2}}, omit_sample_type='SV_WES',
```
ShifaSZ, line 122: It is better to have a test for a population that has no frequencies or filters on a frequency other than 'af', 'ac', and 'hh'. It is the corner case that the `pop_filters` above can be empty.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: `len(None)` throws an error though. But yes, this could be cleaned up

hail_search/test_search.py
hanars: every other test in the file tests what happens when theres no frequency filter

hanars: oh you mean has values that are all None. Yes, I can add a test

ShifaSZ (approved):

PR #3529, opened 2023-07-31
fix on success function order

hanars:
I'm not sure how these arguments got flipped, but they are supposed to be the other way around and everything works fine once this is changed


ui/shared/components/panel/LoadWorkspaceDataForm.jsx
```diff
@@ -148,7 +148,7 @@ const onProjectCreateSuccess = (responseJson) => {
 
 const formatAddDataUrl = ({ projectGuid }) => (`/api/project/${projectGuid}/add_workspace_data`)
 
-const onAddDataFromWorkspace = dispatch => (responseJson) => {
+const onAddDataFromWorkspace = responseJson => (dispatch) => {
```
ShifaSZ, line 151: The existing code won't work. It couldn't be caught because we usually load data for newly created projects and never add data for current projects.

ShifaSZ (approved):

PR #3530, opened 2023-07-31
Hail backend - In-silico filter

hanars:


hail_search/hail_search_query.py
```diff
@@ -411,6 +419,44 @@ def _filter_by_frequency(self, frequencies):
                     pop_filter &= pf
                 self._ht = self._ht.filter(hl.is_missing(pop_expr) | pop_filter)
 
+    def _filter_by_in_silico(self, in_silico_filters):
+        in_silico_filters = in_silico_filters or {}
+        require_score = in_silico_filters.get('requireScore', False)
```
ShifaSZ, line 424: Using `includes_missing_score` for the `requireScore` is more understandable.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: but this is passed from the ui so its not something we can change. I think its clearer to name the variable what its called in the UI code and what its passed in the json as rather than to rename it to something else that means the same thing

hail_search/hail_search_query.py
```diff
@@ -355,6 +370,44 @@ def _filter_by_frequency(self, frequencies):
...
+        in_silico_filters = {
+            k: v for k, v in in_silico_filters.items()
+            if k in self.PREDICTION_FIELDS_CONFIG and v is not None and len(v) != 0
```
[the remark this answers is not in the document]
hanars: yeah, this was copy-pasted from elasticsearch code, but it does not seem to be neccessary

ShifaSZ (approved):

hail_search/hail_search_query.py
```diff
@@ -484,6 +492,41 @@ def _filter_by_frequency(self, frequencies):
...
+            if not require_score:
+                missing_qs.append(hl.is_missing(ht_value))
+
+        if missing_qs:
+            missing_q = missing_qs[0]
+            for q in missing_qs[1:]:
+                missing_q &= q
+            in_silico_qs.append(missing_q)
+
+        in_silico_q = in_silico_qs[0]
```
bpblanken, line 524: Can this block be replaced by `hl.any(in_silico_qs)`?

bpblanken: Much cleaner 🎉!

bpblanken (approved):

hail_search/hail_search_query.py
hanars: I feel like I was frustrated that there was no easy way to do this in hail, and now I'm like "oh duh, of course there is"

PR #3537, opened 2023-08-03
Hail backend - filter annotations

hanars:


hail_search/hail_search_query.py
```diff
@@ -625,6 +654,63 @@ def _format_transcript_args(self):
...
+            )
+            annotation_exprs['allowed_transcripts'] = allowed_transcripts
+            annotation_filters = annotation_filters + [hl.is_defined(allowed_transcripts.first())]
+
+        consequence_filter = self._or_filter(annotation_filters)
+        if consequence_filter is not None:
+            annotation_exprs['has_allowed_annotation'] = consequence_filter
+
+        if annotation_exprs:
+            self._ht = self._ht.annotate(**annotation_exprs)
```
hanars, line 679: these are added as annotations instead of directly filtering because they are used again later for annotating the selected transcript and for filtering compound hets

hail_search/hail_search_query.py
```diff
@@ -521,11 +516,41 @@ def _filter_by_in_silico(self, in_silico_filters):
...
+            ht_value = self._ht[score_path.source][score_path.field]
+            score_filter = ht_value >= float(value)
+
+        return score_filter, ht_value
+
+    def _filter_by_annotations(self, pathogenicity, annotations):
+        annotation_override_filters = self._get_annotation_override_filters(pathogenicity, annotations)
+
+        self._annotate_allowed_consequences(annotations, annotation_override_filters)
+        if 'has_allowed_annotation' in self._ht.row:
```
ShifaSZ, line 537: Using `hasattr(self._ht, 'has_allowed_annotation')` is more straightforward.

hail_search/hail_search_query.py
```diff
@@ -625,6 +654,63 @@ def _format_transcript_args(self):
...
+
+        for key, *args in self.PATHOGENICITY_FILTERS:
+            path_terms = (pathogenicity or {}).get(key)
+            if path_terms:
+                annotation_filters.append(self._has_terms_range_expr(path_terms, key, *args))
+        if (annotations or {}).get(SCREEN_KEY):
+            screen_enum = self._get_enum_lookup(SCREEN_KEY.lower(), 'region_type')
+            allowed_consequences = hl.set({screen_enum[c] for c in annotations[SCREEN_KEY]})
+            annotation_filters.append(allowed_consequences.contains(self._ht.screen.region_type_ids.first()))
+        if (annotations or {}).get(SPLICE_AI_FIELD):
```
ShifaSZ, line 692: It may be cleaner to update the `annotations` with `annotations = annotations or {}` before using it.

hail_search/hail_search_query.py
```diff
@@ -625,6 +654,63 @@ def _format_transcript_args(self):
...
+            annotation_filters = annotation_filters + [hl.is_defined(allowed_transcripts.first())]
+
+        consequence_filter = self._or_filter(annotation_filters)
+        if consequence_filter is not None:
+            annotation_exprs['has_allowed_annotation'] = consequence_filter
+
+        if annotation_exprs:
+            self._ht = self._ht.annotate(**annotation_exprs)
+
+    def _get_annotation_override_filters(self, pathogenicity, annotations):
```
ShifaSZ, line 681: I'm curious about the `override` in the function's name. The function gets annotation filters other than the allowed transcripts. Why is it a kind of `overriding`?

hail_search/hail_search_query.py
```diff
@@ -625,6 +654,63 @@ def _format_transcript_args(self):
...
+
+    def _has_terms_range_expr(self, terms, field, subfield, range_configs):
+        enum_lookup = self._get_enum_lookup(field, subfield)
+
+        ranges = [[None, None]]
+        for path_filter, start, end in range_configs:
+            if path_filter in terms:
+                ranges[-1][1] = len(enum_lookup) if end is None else enum_lookup[end]
+                if ranges[-1][0] is None:
+                    ranges[-1][0] = enum_lookup[start]
```
ShifaSZ, line 706: Would it be cleaner by writing it as `ranges[-1][0] = ranges[-1][0] or enum_lookup[start]`?

hail_search/test_search.py
```diff
@@ -245,6 +245,29 @@ async def test_frequency_filter(self):
             omit_sample_type='SV_WES',
         )
 
+    async def test_annotations_filter(self):
```
ShifaSZ, line 248: It is unclear to me whether the tests cover the cases of allowed transcript filters.

hail_search/test_search.py
```diff
@@ -245,6 +245,29 @@ async def test_frequency_filter(self):
...
+            [VARIANT2, VARIANT4], pathogenicity=pathogenicity, annotations={'SCREEN': ['CTCF-only', 'DNase-only']},
+            omit_sample_type='SV_WES',
+        )
+
+        annotations = {'missense': ['missense_variant'], 'in_frame': ['inframe_insertion', 'inframe_deletion'], 'frameshift': None}
+        await self._assert_expected_search(
+            [VARIANT2, VARIANT4], pathogenicity=pathogenicity, annotations=annotations, omit_sample_type='SV_WES',
+        )
+
+        await self._assert_expected_search([VARIANT2, VARIANT4], annotations=annotations, omit_sample_type='SV_WES')
```
ShifaSZ, line 264: Are there special cases covered with and without the `pathogenicity` in these two tests?

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: I prefer it this way, as it is specifically requiring it to be a row annotation, and not a table function or a global

hail_search/hail_search_query.py
hanars: These are the filters that override the annotation filter, thus the name `annotation_override_filters`. Certain filters  such as pathogenicity and splice AI override the annotation filter, so the variant is returned even if it does not pass the annotation filter

hail_search/hail_search_query.py
hanars: no because that overwrites it when the value is 0, which is incorrect. `ranges[-1][0] = enum_lookup[start] if ranges[-1][0] is None else  ranges[-1][0] ` is the correct one liner version for this logic, but I think thats less readable

hail_search/test_search.py
hanars: It does? Thats what the final 3 assertions in this test are testing. What case are you worried about that you don't see here?

hail_search/test_search.py
hanars: There are 6 different assertions in this one new test, so I don't know which 2 tests you are referring to, but this test has multiple cases with pathogenicity only, annotations only, and both annotations and pathogenicity. What is the actual case you are concerned about?

hail_search/test_search.py
ShifaSZ: My concern was if any variants were filtered out by the transcript consequences with [the code](https://github.com/broadinstitute/seqr/blob/12c49bc8c2a326a76791543fce9a8df473d2a411/hail_search/hail_search_query.py#L663C24-L668). But I think the assertion with `missense` annotations has done the test.

hail_search/test_search.py
ShifaSZ: The `pathogenicity` could override the `annotations`. But the search results are the same. So the `pathogenicity` and `annotations` are overlapped. The concern is the test efficiency.

ShifaSZ (approved):

hail_search/constants.py
```diff
@@ -10,6 +10,16 @@
 
 VARIANT_KEY_FIELD = 'variantId'
 GNOMAD_GENOMES_FIELD = 'gnomad_genomes'
+SPLICE_AI_FIELD = 'splice_ai'
+NEW_SV_FIELD = 'new_structural_variants'
+SCREEN_KEY = 'SCREEN'
```
bpblanken, line 15: Does `SCREEN` need to be all caps?

hanars: yeah it was a weird choice we made in the UI. If it really bothers you we could transform it in the server side proxy

hail_search/hail_search_query.py
```diff
@@ -618,6 +630,62 @@ def _format_transcript_args(self):
...
+
+        return annotation_filters
+
+    def _has_terms_range_expr(self, terms, field, subfield, range_configs):
+        enum_lookup = self._get_enum_lookup(field, subfield)
+
+        ranges = [[None, None]]
+        for path_filter, start, end in range_configs:
+            if path_filter in terms:
+                ranges[-1][1] = len(enum_lookup) if end is None else enum_lookup[end]
```
bpblanken, line 679: This is maybe out of the scope of this pr, but would it make sense to do the enum lookups of the statically configured ranges at startup?

hanars: The lookup is pretty fast and we need to evaluate the enum for formatting results anyways, so it doesn't really add any time relative to the code before this PR. Theres maybe a bigger question of should we read all the enums and globals at application startup instead of on a per search basis, which might be faster in the long run. But I think thats out of scope of this PR. If you think thats a better approach, I can make a follow up for that

bpblanken: Yeah, I think doing the enums and globals at application startup would make more sense than at query time.  And yeah, follow up make more sense than on this pr.

bpblanken (approved):

PR #3538, opened 2023-08-04
Hail backend selected transcript

hanars:


hail_search/hail_search_query.py
```diff
@@ -618,6 +625,27 @@ class VariantHailTableQuery(BaseHailTableQuery):
...
+
+        main_transcript = ht.sorted_transcript_consequences.first()
+        if has_gene_transcripts:
+            matched_transcript = ht.gene_transcripts.first()
+            if has_allowed_transcripts:
+                allowed_transcript_ids = hl.set(ht.allowed_transcripts.map(lambda t: t.transcript_id))
+                matched_transcript = hl.or_else(
+                    ht.gene_transcripts.find(lambda t: allowed_transcript_ids.contains(t.transcript_id)),
+                    matched_transcript,
+                )
```
ShifaSZ, line 641: Moving `matched_transcript = ht.gene_transcripts.first()` into an `else` branch is cleaner.

hail_search/hail_search_query.py
```diff
@@ -618,6 +625,27 @@ class VariantHailTableQuery(BaseHailTableQuery):
         },
     }
 
+    @staticmethod
+    def _selected_main_transcript_expr(ht):
+        has_gene_transcripts = 'gene_transcripts' in ht.row
```
ShifaSZ, line 630: The `gene_transcripts` is always annotated with either null or non-null values. So, is the `'gene_transcripts' in ht.row`  always `True`?

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: I don't see how? moving it into an ekse block would mean defining it twice, we need it here and in the else statement.  leaving it where it is means we can conditionally reuse it the way we are now

hail_search/hail_search_query.py
hanars: its not always annotated, the annotation is in a method thats only conditionally called

ShifaSZ (approved):

bpblanken (approved):

PR #3540, opened 2023-08-07
Hail backend - clinvar overrides

hanars:


hail_search/hail_search_query.py
```diff
@@ -658,6 +677,32 @@ def _format_transcript_args(self):
         })
         return args
 
+    def _get_family_passes_quality_filter(self, quality_filter, ht=None, pathogenicity=None, **kwargs):
+        passes_quality = super(VariantHailTableQuery, self)._get_family_passes_quality_filter(quality_filter)
+        clinvar_path_ht = False if passes_quality is None else self._get_clinvar_filter_ht(pathogenicity)
+        if not clinvar_path_ht:
+            return passes_quality
+
+        return lambda entries: hl.is_defined(clinvar_path_ht[ht.key]) | passes_quality(entries)
```
ShifaSZ, line 686: The `ht` here can be `None`. So `ht.key` can fail.

hail_search/hail_search_query.py
```diff
@@ -658,6 +677,32 @@ def _format_transcript_args(self):
...
+
+    def _get_clinvar_filter_ht(self, pathogenicity):
+        if self._filter_hts.get(CLINVAR_KEY) is not None:
+            return self._filter_hts[CLINVAR_KEY]
+
+        clinvar_path_filters = self._get_clinvar_path_filters(pathogenicity)
+        if not clinvar_path_filters:
+            self._filter_hts[CLINVAR_KEY] = False
+            return False
```
ShifaSZ, line 695: Is it possible that `self._filter_hts[CLINVAR_KEY]` is assigned a value `False` during previous calls to this function and `self._filter_hts.get(CLINVAR_KEY) is not None` becomes `True` in the subsequent calls?

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: it can never be None, its just a function of how overriding keyword args works

hail_search/hail_search_query.py
hanars: no? theres only one possible return value from this function, this is just a way of caching the result

hail_search/hail_search_query.py
ShifaSZ: Then you may want to remove the default value for `ht` from the params.

hail_search/hail_search_query.py
ShifaSZ: Okay, I see. `False` also needs to be cached, representing Clinvar is unavailable.

hail_search/hail_search_query.py
```diff
@@ -142,42 +143,46 @@ def __init__(self, data_type, sample_data, genome_version, sort=XPOS, num_result
...
-            variant_ids=variant_ids, **kwargs)
+            sample_data, excluded_intervals=excluded_intervals, variant_ids=variant_ids, **kwargs)
 
-    def import_filtered_table(self, data_type, sample_data, intervals=None, **kwargs):
-        tables_path = f'{DATASETS_DIR}/{self._genome_version}/{data_type}'
-        load_table_kwargs = {'_intervals': intervals, '_filter_intervals': bool(intervals)}
+    def _get_table_path(self, path):
+        return f'{DATASETS_DIR}/{self._genome_version}/{self._data_type}/{path}'
 
+    def _read_table(self, path):
```
bpblanken, line 166: If we want a cache over our table reads, maybe https://cachetools.readthedocs.io/en/latest/#cachetools.cachedmethod?

hanars: its not really caching, its assigning it to an instance variable. This is essentially the same thing we are doing when we say `self._ht = ...` but I tried to make it a little more generic to make it extensible for some other known cases. I'll refactor this to make it a clearer case of reuse though, since this seems to be really confusing

hanars: also these tables live in memory, so from a performance standpoint I'm not sure caching helps. This is juts a helper function to make reading multiple tables with the same arguments easier, but its a different table for each call

hanars: the first above comment was because I thought you were commenting on a different part of the code. The second comment I think actually responds to your comment

hail_search/hail_search_query.py
bpblanken: I had a comment here not get posted that might make the other one make more sense:

"Is this function actually called multiple times to justify the `self._filter_hts` cache?"

bpblanken: And a second question, is `clinvar_path_variants.ht` a static reference table?

hail_search/hail_search_query.py
hanars: oh so then my comment above does basically work - its not a cache its an instance variable, I just know theres gonna be at least 1 more so I assigned it to a dictionary instead of its own variable

hail_search/hail_search_query.py
hanars: This function is called every time `_filter_entries_table` is called, which is once per project. So yes, it can be reused. And `clinvar_path_variants.ht` is a static reference table, its just the path and likely path variants without the other annotations. I realize we probably should have talked offline about the couple of pre-filter tables seqr will need, its for performance optimization so it wasn't on my radar initially

bpblanken (approved):

PR #3546, opened 2023-08-10
Saved variant timeout fix

hanars:
Having a gene_id specific db index was not enough to improve performance when querying RNA data for both sample and gene. This adds a compound index for sample and gene together, and updates the queries to directly use the underlying sample foreign key field instead of doing a costly and unindexed join


ShifaSZ:
You may want to update the change log.

ShifaSZ (approved):
LGTM

PR #3547, opened 2023-08-10
cache globals for hail backend

hanars:


hail_search/hail_search_query.py
```diff
@@ -59,6 +62,15 @@ class BaseHailTableQuery(object):
         XPOS: lambda r: [r.xpos],
     }
 
+    @classmethod
+    def load_globals(cls):
+        loaded_globals = {}
+        for genome_version in cls.GENOME_VERSIONS:
+            ht_path = cls._get_generic_table_path(genome_version, 'annotations.ht')
+            globals_ht = hl.read_table(ht_path).head(0).select()
+            loaded_globals[genome_version] = {k: hl.eval(globals_ht[k]) for k in cls.GLOBALS}
```
bpblanken, line 71: Will `cls.GLOBALS` ever have more than a single element?  If so maybe faster to call `hl.eval()` only once (though it gets real jank):

```
dict(hl.eval(hl.Struct(**{k:globals_ht[x] for k in cls.GLOBALS})))
```

hail_search/hail_search_query.py
```diff
@@ -59,6 +62,15 @@ class BaseHailTableQuery(object):
         XPOS: lambda r: [r.xpos],
     }
 
+    @classmethod
+    def load_globals(cls):
+        loaded_globals = {}
+        for genome_version in cls.GENOME_VERSIONS:
+            ht_path = cls._get_generic_table_path(genome_version, 'annotations.ht')
+            globals_ht = hl.read_table(ht_path).head(0).select()
```
bpblanken, line 70: can we just access the globals themselves w/ `globals_struct = hl.read_table(ht_path).globals`?

hail_search/hail_search_query.py
hanars: it can, because we need the clinvar version which is in the version globals. I can make this one eval statement

hail_search/hail_search_query.py
hanars: 🤦 yes

hail_search/web_app.py
```diff
@@ -29,4 +31,5 @@ def init_web_app():
         web.get('/status', status),
         web.post('/search', search),
     ])
+    app[APP_GLOBALS] = load_globals()
```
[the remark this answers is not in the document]
hanars: I will try and see

bpblanken (approved):
lgtm!

hail_search/hail_search_query.py
```diff
@@ -59,6 +63,14 @@ class BaseHailTableQuery(object):
         XPOS: lambda r: [r.xpos],
     }
 
+    @classmethod
+    def load_globals(cls):
+        cls.LOADED_GLOBALS = {}
+        for genome_version in cls.GENOME_VERSIONS:
+            ht_path = cls._get_generic_table_path(genome_version, 'annotations.ht')
+            ht_globals = hl.eval(hl.read_table(ht_path).globals.select(*cls.GLOBALS))
+            cls.LOADED_GLOBALS[genome_version] = {k: ht_globals[k] for k in cls.GLOBALS}
```
ShifaSZ, line 72: Does it necessary to load multiple genome versions? It looks that only the version of `self._genome_version` is useful.

hail_search/search.py
```diff
@@ -3,13 +3,18 @@
...
     query_cls = QUERY_CLASS_MAP[single_data_type]
 
-    query = query_cls(data_type, sample_data=sample_data, **request)
+    query = query_cls(sample_data, genome_version, **request)
     return query.search()
+
+
+def load_globals():
+    for cls in QUERY_CLASS_MAP.values():
+        cls.load_globals()
```
ShifaSZ, line 20: I need clarification on why loading once per data type is sufficient.

hail_search/hail_search_query.py
hanars: yes because we have data on both builds so we need to support both builds, so when users do searches on either build they work. we different reference data sources for different data builds, so the enums could easily be different

hail_search/search.py
hanars: why would it not be? We only have one annotation table per genome version per data type, so we load the globals for each data type for each genome version, which means we have loaded all the possible globals

ShifaSZ: I got it.

ShifaSZ (approved):

PR #3550, opened 2023-08-11
Hail backend sort

hanars:


hail_search/hail_search_query.py
```diff
@@ -683,7 +689,34 @@ def _sort_order(self, ht):
         return sort_expressions
 
     def _get_sort_expressions(self, ht, sort):
-        return self.SORTS[sort](ht)
+        sort_expressions = []
+        if sort in self.SORTS:
+            sort_expressions = self.SORTS[sort](ht)
+
+        elif sort in POPULATION_SORTS:
```
bpblanken, line 696: Should `POPULATION_SORTS` be present as a class var in some capacity?  Maybe by combining with `self.POPULATIONS`?

hail_search/hail_search_query.py
```diff
@@ -683,7 +690,34 @@ def _sort_order(self, ht):
...
+            prediction_path = self.PREDICTION_FIELDS_CONFIG[sort]
+            sort_expressions = [-hl.float64(ht[prediction_path.source][prediction_path.field])]
+
+        elif sort == 'in_omim':
+            sort_expressions = self._omim_sort(ht, hl.set(set(self._sort_metadata)))
+
+        elif self._sort_metadata:
+            sort_expressions = self._gene_rank_sort(ht, hl.dict(self._sort_metadata))
+
+        else:
```
ShifaSZ, line 707: We don't usually add a blank link between the `if` and the `else` lines. Do we?

hanars: because all the elifs are on their own lines I thought this was more readable

ShifaSZ (approved):
LGTM

hail_search/hail_search_query.py
```diff
@@ -695,7 +701,34 @@ def _sort_order(self, ht):
         return sort_expressions
 
     def _get_sort_expressions(self, ht, sort):
-        return self.SORTS[sort](ht)
+        sort_expressions = []
+        if sort in self.SORTS:
```
bpblanken, line 705: Can every assigment to `sort_expressions` here just be a `return`?

hail_search/hail_search_query.py
```diff
@@ -695,7 +701,34 @@ def _sort_order(self, ht):
...
+            sort_expressions = [-hl.float64(ht[prediction_path.source][prediction_path.field])]
+
+        elif sort == 'in_omim':
+            sort_expressions = self._omim_sort(ht, hl.set(set(self._sort_metadata)))
+
+        elif self._sort_metadata:
+            sort_expressions = self._gene_rank_sort(ht, hl.dict(self._sort_metadata))
+
+        else:
+            sort_field = next((field for field, config in self.POPULATIONS.items() if config.get('sort') == sort), None)
```
bpblanken, line 719: maybe personal preference but I think just making this a list and accessing the first element with `[0]` is easier to read.  We expect `(field for field, config in self.POPULATIONS.items() if config.get('sort') == sort)` to have at most one element right?

hanars: I prefer `next` because it stops iteration when it finds the match

bpblanken: what about statically defining the `{config['sort']: field}` mapping so the lookup is `O(1)`?

hanars: I had that originally, and you asked to change it so it would be part of the main `POPULATIONS` configs

hanars: theres are max 5 populations, I think we are overthinking the optimization here

bpblanken: Sure.. I think I was originally looking for something like:

```
POPULATIONS = {
   ...
 }
POPULATION_SORTS = {config['sort'] :field for field, config in POPULATIONS if config.get('sort')}
```
I agree that the optimization bit is unnecessary, so going to go ahead and approve!

bpblanken (approved):

PR #3552, opened 2023-08-11
Hail backend - AF prefilter

hanars:
Adds a frequency prefilter to filter out high AF variants from the entries table to reduce the number of reads to the annotations table


hail_search/hail_search_query.py
```diff
@@ -677,29 +681,59 @@ def _format_transcript_args(self):
...
+            return 'is_pathogenic'
         elif CLINVAR_PATH_FILTER not in clinvar_path_filters:
-            clinvar_path_ht = clinvar_path_ht.filter(clinvar_path_ht.is_likely_pathogenic)
-        self._filter_hts[CLINVAR_KEY] = clinvar_path_ht
+            return 'is_likely_pathogenic'
+        return True
+
+    def _prefilter_entries_table(self, ht, **kwargs):
+        af_ht = self._get_loaded_filter_ht(
+            GNOMAD_GENOMES_FIELD, 'high_af_variants.ht', self._get_gnomad_af_prefilter, **kwargs)
```
ShifaSZ, line 720: The function `_prefilter_entries_table` looks prefiltering for `gnomAD`. How about adding `gnomad` to the function name?

hail_search/test_search.py
```diff
@@ -243,15 +243,15 @@ async def test_frequency_filter(self):
         )
 
         await self._assert_expected_search(
-            [VARIANT1, VARIANT2, VARIANT4], frequencies={'gnomad_genomes': {'af': 0.41}}, omit_sample_type='SV_WES',
+            [VARIANT1, VARIANT2, VARIANT4], frequencies={'gnomad_genomes': {'af': 0.05}}, omit_sample_type='SV_WES',
```
ShifaSZ, line 246: Need a test for `af` smaller than 0.01.

ShifaSZ (changes requested):

hail_search/hail_search_query.py
hanars: this is a subclass specific override of a base method. For the variants table it filters for gnomad, for other data types it will have different filtering

bpblanken (approved):

ShifaSZ (approved):

PR #3553, opened 2023-08-11
Hail backend gene counts

hanars:


hail_search/search.py
```diff
@@ -1,7 +1,7 @@
 from hail_search.hail_search_query import QUERY_CLASS_MAP
 
 
-def search_hail_backend(request):
+def search_hail_backend(request, gene_counts=False):
```
bpblanken, line 4: Small style suggestion here.  The linter I have over in the pipeline calls this a [bool trap](https://adamj.eu/tech/2021/07/10/python-type-hints-how-to-avoid-the-boolean-trap/) and recommends against it.  

It may be easier to understand the switchey-ness here by having this method just `return query` and calling `gene_counts()` or `search()` upstream.

bpblanken (approved):
One style nit/thought but otherwise this is a cool feature!

PR #3555, opened 2023-08-15
handle dup individual ids for sample metadata

hanars:
Discovered a bug in the sample metadata report if individual IDs are not unique across projects

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -205,9 +205,9 @@ def sample_metadata_export(request, project_guid):
     )
     family_rows_by_id = {row['family_id']: row for row in family_rows}
 
-    rows_by_subject_id = {row['subject_id']: row for row in subject_rows}
+    rows_by_subject_family_id = {(row['subject_id'], row['family_guid']): row for row in subject_rows}
     for row in sample_rows:
-        rows_by_subject_id[row['subject_id']].update(row)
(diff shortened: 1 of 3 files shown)


seqr/views/apis/report_api.py
```diff
@@ -205,9 +205,9 @@ def sample_metadata_export(request, project_guid):
     )
     family_rows_by_id = {row['family_id']: row for row in family_rows}
 
-    rows_by_subject_id = {row['subject_id']: row for row in subject_rows}
+    rows_by_subject_family_id = {(row['subject_id'], row['family_guid']): row for row in subject_rows}
     for row in sample_rows:
-        rows_by_subject_id[row['subject_id']].update(row)
+        rows_by_subject_family_id[(row['subject_id'], row['family_guid'])].update(row)
```
ShifaSZ, line 210: Can we use the `family_id` field instead of the newly added `family_guid` field?

hanars: no because family guids are unique across projects and family ids are not

ShifaSZ: Can it be an issue for the `family_rows_by_id`, which uses the `family_id` as the key? The `family_rows` generation in the function `_get_loaded_before_date_project_individual_samples` is similar, which relies on the `family_id`.

hanars: this is out of scope

ShifaSZ (approved):

PR #3558, opened 2023-08-16
New in-silico thresholds

ShifaSZ:
It is ready for review.

### DIFF
--- a/ui/shared/components/panel/variants/Predictions.jsx
+++ b/ui/shared/components/panel/variants/Predictions.jsx
@@ -5,10 +5,10 @@ import { connect } from 'react-redux'
 import { Icon, Transition, Popup } from 'semantic-ui-react'
 
 import { getGenesById } from 'redux/selectors'
-import { PREDICTOR_FIELDS, getPredictColor, getVariantMainGeneId } from 'shared/utils/constants'
+import { PREDICTOR_FIELDS, PRED_COLOR_MAP, getVariantMainGeneId } from 'shared/utils/constants'
 import { snakecaseToTitlecase } from 'shared/utils/stringUtils'
 import { HorizontalSpacer } from '../../Spacers'
-import { ButtonLink } from '../../StyledComponents'
+import { ButtonLink, ColoredIcon } from '../../StyledComponents'
 
 const PredictionValue = styled.span`
   margin-left: 5px;
@@ -19,8 +19,27 @@ const PredictionValue = styled.span`
 
 const NUM_TO_SHOW_ABOVE_THE_FOLD = 6 // how many predictors to show immediately
 
+const comparePathScores = (value, i, thresholds) => {
+  if (i < 2) { // Benign thresholds
+    if (i === 0) {
+      return value <= thresholds[0]
+    }
+    return (thresholds[0] === undefined || value > thresholds[0]) && value <= thresholds[1]
+  }
+
+  if (i === 2) { // Grey area
+    return (thresholds[1] === undefined || value > thresholds[1]) && value < thresholds[2]
+  }
+
+  // Pathogenic thresholds
+  if (i === 5) {
+    return true
+  }
+  return value >= thresholds[i - 1] && (thresholds[i] === undefined || value < thresholds[i])
+}
+
 const predictionFieldValue = (
-  predictions, { field, pathHigher, thresholds, indicatorMap, infoField, infoTitle },
+  predictions, { field, thresholds, indicatorMap, infoField, infoTitle },
 ) => {
   let value = predictions[field]
   if (value === null || value === undefined) {
@@ -31,17 +50,17 @@ const predictionFieldValue = (
 
   if (thresholds) {
     value = parseFloat(value).toPrecision(3)
-    const color = getPredictColor(value, pathHigher, thresholds)
-    return { value, color, infoValue, infoTitle, pathHigher, thresholds }
+    const color = PRED_COLOR_MAP.find((clr, i) => comparePathScores(value, i, thresholds))
+    return { value, color, infoValue, infoTitle, thresholds }
   }
 
   return indicatorMap[value[0]] || indicatorMap[value]
 }
 
-const PATHOGENIC_COLORS = ['green', 'light green', 'grey', 'yellow', 'red', 'dark red']
+const coloredIcon = color => <ColoredIcon name="circle" size="small" color={color} />
 
 const Prediction = (
-  { field, fieldTitle, value, color, infoValue, infoTitle, pathHigher, thresholds, href },
+  { field, fieldTitle, value, color, infoValue, infoTitle, thresholds, href },
 ) => {
   const indicator = infoValue ? (
     <Popup
@@ -49,23 +68,32 @@ const Prediction = (
       content={infoValue}
       trigger={<Icon name="question circle" size="small" color={color} />}
     />
-  ) : <Icon name="circle" size="small" color={color} />
+  ) : coloredIcon(color)
   const fieldName = fieldTitle || snakecaseToTitlecase(field)
   const fieldDisplay = thresholds ? (
     <Popup
       header={`${fieldName} Color Ranges`}
       content={
         <div>
-          {thresholds.map((th, i) => {
-            if (!th) {
-              return null
-            }
-            const t = pathHigher ? th : -1 * th
-            if (i < 3) {
-              return <div>{`${PATHOGENIC_COLORS[i]} ${pathHigher ? '<' : '>'}= ${t}`}</div>
-            }
-            return <div>{`${PATHOGENIC_COLORS[i]} ${pathHigher ? '>' : '<'}= ${t}`}</div>
-          }).filter(e => !!e)}
+          {[0, 1].map(i => thresholds[i] !== undefined && (
+            <div>
+              {coloredIcon(PRED_COLOR_MAP[i])}
+              {i > 0 && thresholds[i - 1] !== undefined && ` > ${thresholds[i - 1]} and`}
+              {` <= ${thresholds[i]}`}
+            </div>
+          ))}
+          <div>
+            {coloredIcon(PRED_COLOR_MAP[2])}
+            {thresholds[1] === undefined ? '' : ` > ${thresholds[1]} and`}
+            {` < ${thresholds[2]}`}
+          </div>
+          {[2, 3, 4].map(i => thresholds[i] !== undefined && (
+            <div>
+              {coloredIcon(PRED_COLOR_MAP[i + 1])}
+              {` >= ${thresholds[i]}`}
+              {i < 4 && thresholds[i + 1] !== undefined && ` and < ${thresholds[i + 1]}`}
+            </div>
+          ))}
         </div>
       }
       trigger={<span>{fieldName}</span>}
@@ -90,7 +118,6 @@ Prediction.propTypes = {
   infoTitle: PropTypes.string,
   fieldTitle: PropTypes.string,
   color: PropTypes.string,
-  pathHigher: PropTypes.bool,
   thresholds: PropTypes.arrayOf(PropTypes.number),
   href: PropTypes.string,
 }
@@ -121,8 +148,8 @@ class Predictions extends React.PureComponent {
     if (gene && gene.primateAi) {
       genePredictors.primate_ai = {
         field: 'primate_ai',
-        pathHigher: gene.primateAi.percentile75 >= gene.primateAi.percentile25,
-        thresholds: [null, null, gene.primateAi.percentile25, gene.primateAi.percentile75, null],
+        thresholds: [undefined, undefined, gene.primateAi.percentile25.toPrecision(3),
+          gene.primateAi.percentile75.toPrecision(3), undefined],
       }
     }
 
--- a/ui/shared/components/panel/variants/Predictions.jsx
+++ b/ui/shared/components/panel/variants/Predictions.jsx
@@ -19,25 +19,6 @@ const PredictionValue = styled.span`
 
 const NUM_TO_SHOW_ABOVE_THE_FOLD = 6 // how many predictors to show immediately
 
-const comparePathScores = (value, i, thresholds) => {
-  if (i < 2) { // Benign thresholds
-    if (i === 0) {
-      return value <= thresholds[0]
-    }
-    return (thresholds[0] === undefined || value > thresholds[0]) && value <= thresholds[1]
-  }
-
-  if (i === 2) { // Grey area
-    return (thresholds[1] === undefined || value > thresholds[1]) && value < thresholds[2]
-  }
-
-  // Pathogenic thresholds
-  if (i === 5) {
-    return true
-  }
-  return value >= thresholds[i - 1] && (thresholds[i] === undefined || value < thresholds[i])
-}
-
 const predictionFieldValue = (
   predictions, { field, thresholds, indicatorMap, infoField, infoTitle },
 ) => {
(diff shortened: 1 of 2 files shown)


ui/shared/utils/constants.js
```diff
@@ -1316,38 +1316,70 @@ export const SV_IN_SILICO_GROUP = 'Structural'
...
+  } else {
+    colorIndex = 5
+    if (v < thresholds[2]) {
+      colorIndex = 2
+    } else if (v < thresholds[3]) {
+      colorIndex = 3
+    } else if (v < thresholds[4]) {
+      colorIndex = 4
+    }
+  }
```
ShifaSZ, line 1337: This implementation uses the leftmost color if benign thresholds are missing and the right-most color if pathogenic points are missing.

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -54,13 +51,21 @@ const Prediction = (
...
-          {warningThreshold < dangerThreshold && <div>{`Yellow > ${warningThreshold}`}</div>}
+          {thresholds.map((th, i) => {
+            if (!th) {
+              return null
+            }
+            const t = pathHigher ? th : -1 * th
+            if (i < 3) {
+              return <div>{`${PATHOGENIC_COLORS[i]} ${pathHigher ? '<' : '>'}= ${t}`}</div>
+            }
+            return <div>{`${PATHOGENIC_COLORS[i]} ${pathHigher ? '>' : '<'}= ${t}`}</div>
```
ShifaSZ, line 67: It simply shows the threshold and doesn't show the color ranges. Is it okay? And it still needs to align with the color display because I'm still determining what color we should use for both ends if thresholds are missing.

ui/shared/utils/constants.js
```diff
@@ -1316,38 +1316,70 @@ export const SV_IN_SILICO_GROUP = 'Structural'
...
+  { field: 'strvctvre', group: SV_IN_SILICO_GROUP, thresholds: [null, null, 0.5, 0.75, null] },
+  { field: 'polyphen', group: MISSENSE_IN_SILICO_GROUP, thresholds: [null, 0.113, 0.978, 0.999, null], indicatorMap: POLYPHEN_MAP },
+  { field: 'sift', group: MISSENSE_IN_SILICO_GROUP, thresholds: [null, 0.08, 0.001, 0, null], indicatorMap: INDICATOR_MAP },
   { field: 'mut_taster', group: MISSENSE_IN_SILICO_GROUP, indicatorMap: MUTTASTER_MAP },
-  { field: 'fathmm', group: MISSENSE_IN_SILICO_GROUP, indicatorMap: FATHMM_MAP },
-  { field: 'vest', warningThreshold: 0.5, dangerThreshold: 0.764 },
-  { field: 'mut_pred', warningThreshold: 0.392, dangerThreshold: 0.737 },
-  { field: 'apogee', warningThreshold: 0.5, dangerThreshold: 0.5 },
-  { field: 'gnomad_noncoding', fieldTitle: 'gnomAD Constraint', displayOnly: true, warningThreshold: 2.18, dangerThreshold: 4 },
+  { field: 'fathmm', group: MISSENSE_IN_SILICO_GROUP, thresholds: [null, 3.32, -4.14, -5.04, null], indicatorMap: FATHMM_MAP },
```
hanars, line 1363: if there is an indicator map, there is no threshhold. By definition, if we define an indicator map it means the value coming back from search is a text representation of the pathogenoicity and we do not even have a numeric score in our search data. The ticket was to update the numeric thresholds for numeric scores only. Any socre that was in the image Lynn shared that is absent from seqr or not a numeric score in seqr should not be changed as part of this ticket

ui/shared/utils/constants.js
```diff
@@ -1316,38 +1316,70 @@ export const SV_IN_SILICO_GROUP = 'Structural'
 export const NO_SV_IN_SILICO_GROUPS = [MISSENSE_IN_SILICO_GROUP, CODING_IN_SILICO_GROUP]
 export const SPLICE_AI_FIELD = 'splice_ai'
 
+const PRED_COLOR_MAP = ['green', '#90ee90', 'grey', 'yellow', 'red', '#8b0000']
+
+export const getPredictColor = (value, pathHigher, thresholds) => {
+  let colorIndex = 0
+  const v = pathHigher ? value : -1 * value
```
hanars, line 1323: all of the `pathHigher` logic seems to have been added for scores that we have as text scores in the data in seqr, not as numeric scores. Once you take out all the thresholds for scores that did not used to have thresholds and can't have them, this logic should be unneeded

hanars (changes requested):

ui/shared/utils/constants.js
ShifaSZ: You are right. SIFT and FATHMM have lower pathogenic scores. They both have non-numeric scores.

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -19,8 +19,27 @@ const PredictionValue = styled.span`
...
+
+  if (i === 2) { // Grey area
+    return (thresholds[1] === undefined || value > thresholds[1]) && value < thresholds[2]
+  }
+
+  // Pathogenic thresholds
+  if (i === 5) {
+    return true
+  }
+  return value >= thresholds[i - 1] && (thresholds[i] === undefined || value < thresholds[i])
```
ShifaSZ, line 38: I changed the null thresholds to `undefined` because any numbers compared with `undefined` get a `false`. The null is similar to zero. A number compared with null can get `true` or `false` depending on whether it is negative or positive.

We still have to handle the benign and pathogen threshold differently because the sides of the close/open ends differ.

ui/shared/utils/constants.js
```diff
@@ -1316,37 +1316,43 @@ export const SV_IN_SILICO_GROUP = 'Structural'
 export const NO_SV_IN_SILICO_GROUPS = [MISSENSE_IN_SILICO_GROUP, CODING_IN_SILICO_GROUP]
 export const SPLICE_AI_FIELD = 'splice_ai'
 
+export const PRED_COLOR_MAP = ['green', '#90ee90', 'grey', 'yellow', 'red', '#8b0000']
```
ShifaSZ, line 1319: I changed to use the `ColoredIcon` component, which doesn't support the `olive` color. So I keep the light green color encoded.

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -29,38 +48,52 @@ const predictionFieldValue = (
...
+            {thresholds[1] === undefined ? '' : ` > ${thresholds[1]} and`}
+            {` < ${thresholds[2]}`}
+          </div>
+          {[2, 3, 4].map(i => thresholds[i] !== undefined && (
+            <div>
+              {coloredIcon(PRED_COLOR_MAP[i + 1])}
+              {` >= ${thresholds[i]}`}
+              {i < 4 && thresholds[i + 1] !== undefined && ` and < ${thresholds[i + 1]}`}
+            </div>
+          ))}
```
ShifaSZ, line 96: The UI looks like the following picture.
image

ui/shared/utils/constants.js
hanars: So we do not want to change to that, that will change all the main colors to be css colors and not semantic UI colors, and as is discussed extensively in the ticket, we are trying to keep using the sematic ui colors wherever possible. You need to add conditional logic for which icon component to use, and only use the custom color for dark red

ui/shared/components/panel/variants/Predictions.jsx
hanars: Just show "color > threshold" for everyhting except the first score which shows <=. DO not show both boundry numbers for each color, just the lower boundry

ShifaSZ: How about we do it like the following?
```
<colored dot> <= 0.0016
<colored dot> <= 0.29
<colored dot> < 0.644
<colored dot> < 0.773
<colored dot> < 0.932
<colored dot> >= 0.932
```

hanars: I'm concerned about the mixing of `<=` and `<`, we need to be consistent in setting our thresholds so we always use the same comparator. Otherwise this is fine, having the final one be `>=` works

ShifaSZ: The lower color ranges open on the left and close on the right. The upper ones close at the left and open at the right. That's why we have both `<=` and `<`.

hanars: And what I am saying is we should define all the thresholds so they have a consistant cutoff. Wither everything should be strict less than, or <=, you define the threshhol so you can change it if you need to to get this consistancy

hanars: In the example you gave, this is the functional equivalent:
```
<colored dot> < 0.0017
<colored dot> < 0.2901
<colored dot> < 0.644
<colored dot> < 0.773
<colored dot> < 0.932
<colored dot> >= 0.932
```

ShifaSZ: The code will be much easier. But the thresholds are slightly different from the provided numbers. Is it okay for the analysts?

hanars: You should also be doing this in a single list iteration. The logic is not 100% correct here but something like this will work:
```
thresholds.map((threshold, i) => {
  if (threshold === undefined && thresholds[i+1] === undefined) {
    return null
  }
  return (
      <div>
          {coloredIcon(PRED_COLOR_MAP[i])}
          {thresholds[i+1] !== undefined ? ` < ${thresholds[i+1]}` : ` >= ${threshold}` }
       </div>
  )
})
```

ShifaSZ:
Current popup thresholds are like the following pictures.
image
image

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -29,39 +29,48 @@ const predictionFieldValue = (
...
-    if (value >= dangerThreshold) {
-      color = 'red'
-    } else if (value >= warningThreshold) {
-      color = 'yellow'
-    }
-    return { value, color, infoValue, infoTitle, dangerThreshold, warningThreshold }
+  if (thresholds) {
+    value = parseFloat(value).toPrecision(3)
+    const color = PRED_COLOR_MAP.find(
+      (clr, i) => (thresholds[i - 1] || thresholds[i - 1]) &&
```
hanars, line 35: isn't `(thresholds[i - 1] || thresholds[i - 1])` the same as `thresholds[i - 1]`

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -29,39 +29,48 @@ const predictionFieldValue = (
...
+        (thresholds[i - 1] === undefined || value >= thresholds[i - 1]) &&
+        (thresholds[i] === undefined || value < thresholds[i]),
+    )
+    return { value, color, infoValue, infoTitle, thresholds }
   }
 
   return indicatorMap[value[0]] || indicatorMap[value]
 }
 
+const coloredIcon = color => (color === 'darkred' ? <ColoredIcon name="circle" size="small" color="#8b0000" /> :
```
hanars, line 45: use the actual html hex color #8b0000 in the list, and use a conditional to check if the color starts with #

hanars (changes requested):

ui/shared/components/panel/variants/Predictions.jsx
ShifaSZ: My fault, it needs to be `(thresholds[i - 1] || thresholds[i ])`

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -29,39 +29,48 @@ const predictionFieldValue = (
...
-    } else if (value >= warningThreshold) {
-      color = 'yellow'
-    }
-    return { value, color, infoValue, infoTitle, dangerThreshold, warningThreshold }
+  if (thresholds) {
+    value = parseFloat(value).toPrecision(3)
+    const color = PRED_COLOR_MAP.find(
+      (clr, i) => (thresholds[i - 1] || thresholds[i - 1]) &&
+        (thresholds[i - 1] === undefined || value >= thresholds[i - 1]) &&
+        (thresholds[i] === undefined || value < thresholds[i]),
```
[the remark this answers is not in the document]
ShifaSZ: After changing the first term to `(thresholds[i - 1] || thresholds[i])`, the first entry will return `false` if it is `undefined`. It is because `thresholds[-1]` is `undefined` so `thresholds[-1] || thresholds[0]` is `false`.

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -116,8 +123,8 @@ class Predictions extends React.PureComponent {
     if (gene && gene.primateAi) {
       genePredictors.primate_ai = {
         field: 'primate_ai',
-        warningThreshold: gene.primateAi.percentile25,
-        dangerThreshold: gene.primateAi.percentile75,
+        thresholds: [undefined, undefined, gene.primateAi.percentile25.toPrecision(3),
+          gene.primateAi.percentile75.toPrecision(3), undefined],
```
hanars, line 127: I think this spacing should cause a linting error

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -116,8 +123,8 @@ class Predictions extends React.PureComponent {
     if (gene && gene.primateAi) {
       genePredictors.primate_ai = {
         field: 'primate_ai',
-        warningThreshold: gene.primateAi.percentile25,
-        dangerThreshold: gene.primateAi.percentile75,
+        thresholds: [undefined, undefined, gene.primateAi.percentile25.toPrecision(3),
```
hanars, line 126: why add the `toPrecision` here?

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -29,39 +29,48 @@ const predictionFieldValue = (
...
     <Popup
       header={`${fieldName} Color Ranges`}
       content={
-        <div>
-          <div>{`Red > ${dangerThreshold}`}</div>
-          {warningThreshold < dangerThreshold && <div>{`Yellow > ${warningThreshold}`}</div>}
-        </div>
+        PRED_COLOR_MAP.map((c, i) => {
+          if (thresholds[i] === undefined && thresholds[i - 1] === undefined) {
+            return null
```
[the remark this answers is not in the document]
ShifaSZ: This works even when the first threshold is `undefined` because `threshold[-1]` is undefined. We won't show the color for any two consecutive `undefined`.

hanars (changes requested):

ui/shared/components/panel/variants/Predictions.jsx
ShifaSZ: I added it because they display oddly (having many digits) on the popup compared to other in silico scores, which are constants.

ui/shared/components/panel/variants/Predictions.jsx
hanars: all should be on their own lines, or all should be on the same line

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -19,8 +19,10 @@ const PredictionValue = styled.span`
 
 const NUM_TO_SHOW_ABOVE_THE_FOLD = 6 // how many predictors to show immediately
 
+export const PRED_COLOR_MAP = ['green', 'olive', 'grey', 'yellow', 'red', '#8b0000']
```
hanars, line 22: should not be exported

PR #3565, opened 2023-08-18
Add "wait_for_routes" helper script for CronJobs.

bpblanken:


deploy/docker/seqr/wait_for_routes
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+###
+# Waits for network endpoints.  Intended useage is within Kubernetes CronJobs to wait for sidecar availability.
```
sjahl, line 4: don't want to be that guy, but `s/useage/usage` 😄

sjahl (approved):
the Bash all looks good to me here, just the one spelling mistake

deploy/docker/seqr/wait_for_routes
bpblanken: ❤️ that the typo is in the comment and not the bash

hanars:
all seqr PRs should be against dev, not master, unless there is an overwhelmingly compelling reason why we want to release it ahead of everything else on dev. If that is the case, then you need to fill out the [release checklist](https://github.com/broadinstitute/seqr-private/blob/master/release_pr_checklist.md) and make sure to a do a deployment as soon as it is merged

bpblanken:
@hanars I've pointed at `dev`.

hanars (approved):

PR #3575, opened 2023-08-23
Hail backend - SV WES

hanars:
Getting this up early for review, still needs unit tests


ShifaSZ (commented):
Mostly, it looks good to me.

hail_search/hail_search_query.py
```diff
@@ -1158,4 +1154,79 @@ def _additional_annotation_fields(self):
         }
 
 
-QUERY_CLASS_MAP = {cls.DATA_TYPE: cls for cls in [VariantHailTableQuery, SvHailTableQuery]}
+class GcnvHailTableQuery(SvHailTableQuery):
```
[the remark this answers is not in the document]
hanars: yeah, Ben suggested the same thing on the other PR. I agree, but to keep the diff sane I will do a follow up PR to move everything into its own file

hail_search/hail_search_query.py
```diff
@@ -1171,4 +1171,91 @@ def _additional_annotation_fields(self):
...
+            get_default(r), agg(entries.map(lambda g: g[sample_field]))
+        )
+
+    def _format_results(self, ht, annotation_fields):
+        ht = ht.annotate(**{k: self._get_genotype_override_field(ht, k) for k in self.GENOTYPE_OVERRIDE_FIELDS})
+        return super()._format_results(ht, annotation_fields)
+
+    def get_allowed_sv_type_ids(self, sv_types):
+        return super().get_allowed_sv_type_ids([
+            type.replace('gCNV_', '') for type in sv_types if type.startswith('gCNV_')
```
bpblanken, line 1246: Is there a case for doing this replacement in the pipeline?  So that instead of passing the enum ids for `gCNV_DEL` and `gCNV_DUP` we instead pass the ids for `DEL` and `DUP`?  Or are `gCNV_DEL` and `gCNV_DUP` used elsewhere?

hail_search/web_app.py
```diff
@@ -8,6 +8,8 @@
 def _hl_json_default(o):
     if isinstance(o, hl.Struct) or isinstance(o, hl.utils.frozendict):
         return dict(o)
+    elif isinstance(o, set):
```
bpblanken, line 11: Should this be `hl.set`?

hail_search/hail_search_query.py
bpblanken: Yeah, I agree the string replace in python is better.  I guess, if we'd like to remove `gCNV_DUP` and `gCNV_DEL` from the enum it's no problem on my end.

hail_search/web_app.py
hanars: so it was an issue with a regular python set, sets are not actually json serializable and this solved the particular issue. But I can also add a case for a hail set

hanars: actually, since we are only json serializing collected results we do not run into this case, as hail will have already converted all hail setexpressions to python sets. The only cases where we end up with hail-specific data types are those that hail provides which can exist as raw dat atypes and not expressions, which is just Interval, Struct, and frozendict - https://hail.is/docs/0.2/utils/index.html#utils

bpblanken: yep, I think that's right.

bpblanken (approved):

Issue #3539, opened 2023-08-04
Updated LoF constraint tag incorporating new Shet values


lynnpais:
**Is your feature request related to a problem? Please describe.**
We would like to merge the LoF + HI tags for genes in seqr and add one more metric (Shet) for calling LoF constraint. 

**Describe the solution you'd like**
_Variant page:_
Display only one 'LoF constr' tag for a gene if it meets criteria for any one LoF, HI or Shet (new) data. Shet constrained genes are here on sheet 1 - https://docs.google.com/spreadsheets/d/1enxGBWCAFBHdrRlqCj_ueleiDo9K9GWn/edit#gid=1642044433

In the hover over, show the rank for LoF (currently the first line in the LoF hover over), Shet value (column C in sheet 1 above), and HI score (currently last line in HI hover over) for each gene. Include the citation for each, LoF - Karczewski (2020), SHet - Zeng (2023), and HI - Collins (2022). Remove all additional text currently displayed.
Note for Shet - The S is capitalized and het is subscript. Paper here - https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10245655/

_Gene page:_
Add a row for Shet (after LoF but before Haploinsufficient) and show the score for a gene. Scores for all genes are in column C on sheet 2 - https://docs.google.com/spreadsheets/d/1enxGBWCAFBHdrRlqCj_ueleiDo9K9GWn/edit#gid=1146995171

Text to display with the RPS15A gene as an example- Shet: 0.61.
Note: This score was developed by the Pritchard lab [Zeng et al 2023] to predict gene constraint based on functional and evolutionary information. Scores >0.1 are considered to have high likelihood to be under extreme selection.

**Describe alternatives you've considered**
NA

**Additional context**
As discussed

ShifaSZ:
The task includes a backend database loading, looking up, and frontend display. Is that right? @hanars

hanars:
Yes

ShifaSZ:
Separated DB tables for the two sheets may also work because the number of rows differs.

hanars:
we will have one DB table for Shet. We will not accept loading from google sheets. You need to save this as a single standardized file (tsv, csv) and put it in GCP. If there is one sheet that works for loading, just put that in GCP. If not, merge sheets together into one big sheet, and put that in GCP. It can go in the gs:///seqr-reference-data/gene_constraint bucket, which is where we currently load gene constraint from

ShifaSZ:
[...]

PR #3576, opened 2023-08-24
Create a Gene Shet reference data update cmd.

ShifaSZ:
This is one of the PRs for #3539.


reference_data/management/commands/update_gene_shet.py
```diff
@@ -0,0 +1,23 @@
...
+from reference_data.management.commands.utils.update_utils import GeneCommand, ReferenceDataHandler
+from reference_data.models import GeneShet
+
+logger = logging.getLogger(__name__)
+
+
+class ShetReferenceDataHandler(ReferenceDataHandler):
+
+    model_cls = GeneShet
+    url = 'https://storage.googleapis.com/seqr-reference-data/Shet/Shet_Zeng_2023.tsv'
```
hanars, line 11: this belongs in the gene_constraint subfolder, as specified in the ticket

hanars, line 11: add a comment in this file explaining how that file was generated/ where it came from

reference_data/management/commands/update_gene_shet.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @staticmethod
+    def parse_record(record):
+        yield {
+            'gene_id': record['ensg'],
+            'shet': float(record['post_mean_shet']),
+            'shet_constrained': bool(int(record['shet_constrained'])),
```
hanars, line 18: I'm not sure we need this field - we should load the Shet score for any gene we have data for, and we will define a cutoff in seqr for  whether or not to flag a gene based on the score cutoff, not a database field. What in the ticket do you anticipate will require this column?

hanars (changes requested):

reference_data/management/commands/update_gene_shet.py
ShifaSZ: Hmm, the current cutoff looks like 0.1. The requirement ticket doesn't tell the cutoff explicitly. The new `'LoF constr' tag` will be displayed when anyone of LoF, HI, or Shet (this column is true).

hanars: the cutoff is something that we use for display purposes. It should not live in the database in any way, so we should not be loading this

hanars (changes requested):

reference_data/management/commands/update_gene_shet.py
```diff
@@ -0,0 +1,24 @@
...
+    model_cls = GeneShet
+    # The .tsv file is generated from the Google Doc at https://docs.google.com/spreadsheets/d/1enxGBWCAFBHdrRlqCj_ueleiDo9K9GWn/edit#gid=1146995171
+    # by downloading with a tsv format.
+    url = 'https://storage.googleapis.com/seqr-reference-data/gene_constraint/shet_Zeng(2023).xlsx%20-%20All%20scores-for%20gene%20page.tsv'
+
+    @staticmethod
+    def parse_record(record):
+        yield {
+            'gene_id': record['ensg'],
+            'shet': float(record['post_mean (Shet)']),
```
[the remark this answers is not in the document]
ShifaSZ: It is the column name of the spreadsheet. I don't know what name is more suitable. Maybe we should ask Lynn.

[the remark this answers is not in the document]
hanars: the column name is `post_mean (Shet)`. The name of this method is Shet. Therefore, this column is representing the post_mean score for the Shet method. Since the name of the table is Shet, we do not need to capture Shet in the name of the score in the table. Therefore, the name of the score in the table should be post_mean

bpblanken (approved):

hanars (approved):

PR #3586, opened 2023-09-05
allow missing gene id in rna splice outlier data

hanars:


seqr/views/utils/dataset_utils.py
```diff
@@ -378,7 +379,11 @@ def _load_rna_seq_file(file_path, user, mapping_file, parse_row, expected_column
 
             sample_id_to_tissue_type[(sample_id, project)] = tissue_type
 
-            gene_ids.add(row_dict[GENE_ID_COL])
+            gene_id = row_dict[GENE_ID_COL]
```
bpblanken, line 382: Can we now use the `gene_id` variable on line 391 below?

hanars: no because the `gene_or_unique_id` below might be a unique id and not a gene id, and we need the gene id here

bpblanken (approved):

PR #3587, opened 2023-09-05
allow empty text field updates

hanars:


ui/shared/components/panel/view-fields/TextFieldView.jsx
```diff
@@ -27,7 +27,11 @@ const MAPPED_HTML_COMPONENTS = { code: 'div', pre: 'p' }
 
 const LazyRichTextEditor = props => <React.Suspense fallback={<Loader />}><RichTextEditor {...props} /></React.Suspense>
 
-const RICH_TEXT_FIELD = { component: LazyRichTextEditor }
+const RICH_TEXT_FIELD = {
+  component: LazyRichTextEditor,
+  // Override default behavior for undefined values, which is to exclude them from form values
+  parse: val => val || null,
```
bpblanken, line 33: Super pedantic here... but do we want `0` and `''` to also coerce to `null`?

hanars: 0 here would always be the string "0" which is truthy, and the empty string is fine to coerce to null , I think we are doing that in the backend in django anyways

bpblanken (approved):

PR #3589, opened 2023-09-06
split query classes into multiple files

hanars:
splits hail query code into one file per class. Other than tweaking imports, does not change any code, just copy-paste


bpblanken (approved):
lgtm assuming this is just moving code around.

Is now a reasonable time to do the `variants` -> `snv_indel` switch?

hanars:
> lgtm assuming this is just moving code around.

yes. Per the PR description: "Other than tweaking imports, does not change any code, just copy-paste"

> Is now a reasonable time to do the `variants` -> `snv_indel` switch?

No, thats also needs to change a bunch of how seqr formats the requests to the backend and possibly a database migration and I want to do that as its own thing

Issue #3595, opened 2023-09-08
Standard search page not loading for families in RGP project


Ikeoluwa-O-O:
**Describe the bug**
I am not able to load a page for any of the standard searches in [RGP_2660](https://seqr.broadinstitute.org/project/R0594_rare_genomes_project_gen/family_page/F043609_rgp_2660). I tried this for another family in the same project ([Rare Genomes Project_Genomes_GRU](https://seqr.broadinstitute.org/project/R0594_rare_genomes_project_gen/project_page))  and got the same result (blank page). Another analyst is experiencing this issue as well. 

**Link to page(s) where bug is occurring**
https://seqr.broadinstitute.org/variant_search/results/ad6e291c1423503700a50480e171379d?page=1&sort=splice_ai

**Scope of the bug**
No, I am not experiencing this issue in all projects. The searches work fine in the [CMG_Manton_Exomes](https://seqr.broadinstitute.org/project/R0303_cmg_manton_exomes/project_page) project. 

**Screenshots**
<img width="1406" alt="Screenshot 2023-09-08 at 2 06 34 PM" src="https://github.com/broadinstitute/seqr/assets/84587859/9d6ffcdc-a10d-499e-9123-2366d1d96b01">

hanars:
This is now causing backend errors as well: https://console.cloud.google.com/errors/detail/CIih2vS-3MzLDQ;time=P30D?project=seqr-project&utm_source=error-reporting-notification&utm_medium=email&utm_content=resolved-error

hanars:
this is now fixed

PR #3596, opened 2023-09-08
properly format vest

hanars:
recent changes to the UI cause an unhandled JS error on an invalid search results value which was previously being hidden from the user. The fix is to properly format this value so it is displayed properly


bpblanken (approved):

bpblanken:
This is resolving https://github.com/broadinstitute/seqr/issues/3595 right?

PR #3603, opened 2023-09-13
Hail backend - multi data type compound hets

hanars:
Adds functionality for querying compound hets split across data types. Requires some updates to the fixture data to test some edge cases


hail_search/queries/base.py
```diff
@@ -192,14 +192,18 @@ def _enum_field(value, enum, ht_globals=None, annotate_value=None, format_value=
...
+    def __init__(self, sample_data, genome_version, sort=XPOS, sort_metadata=None, num_results=100, inheritance_mode=None,
+                 override_comp_het_alt=False, **kwargs):
         self._genome_version = genome_version
         self._sort = sort
         self._sort_metadata = sort_metadata
         self._num_results = num_results
+        self._override_comp_het_alt = override_comp_het_alt
         self._ht = None
         self._comp_het_ht = None
+        self.unfiltered_comp_het_ht = None
```
bpblanken, line 204: It stands out that this isn't denoted as "private".  Is there a reason?

hail_search/queries/base.py
```diff
@@ -273,6 +277,9 @@ def import_filtered_table(self, sample_data, intervals=None, **kwargs):
             entry_type = families_ht.family_entries.dtype.element_type
             for project_ht, num_project_families in filtered_project_hts[1:]:
                 families_ht = families_ht.join(project_ht, how='outer')
+                families_ht = families_ht.annotate_globals(
```
bpblanken, line 280: Is this using the `family_guids` global that's on the families table?  I think that it's actually something I was using internally (to pass the list of loadable families between tasks) and don't think it's reliable.

hail_search/queries/base.py
hanars: It's accessed directly by the multi data type query class, not as like a `super` call but by actually reading the property from an instantiated class, so I thought it was better to be explicit about the fact that it is intended to be used outside the class - https://github.com/broadinstitute/seqr/pull/3603/files#diff-e646625d305f6373c7ae24a18d8b0857c2bb1eceaa924076023103695d9f67b2R45

hail_search/queries/base.py
bpblanken: 👍 I'm going to remove it from the pipeline output then.

hail_search/queries/base.py
bpblanken: 👍 silly, but I'd maybe move the var declaration to be above or below all the privates.

bpblanken (approved):

bpblanken (approved):

PR #3606, opened 2023-09-14
Hail backend - noncoding exon canonical transcript

hanars:
Adds new search behavior into hail backend

### DIFF
--- a/hail_search/queries/mito.py
+++ b/hail_search/queries/mito.py
@@ -181,10 +181,28 @@ class MitoHailTableQuery(BaseHailTableQuery):
             ht = hl.filter_intervals(ht, parsed_intervals, keep=False)
         return ht
 
-    def _get_consequence_filter(self, allowed_consequence_ids, annotation_exprs):
+    def _get_consequence_filter(self, allowed_consequence_ids, allowed_consequences, annotation_exprs):
+        canonical_consequences = {
+            c.replace('__canonical', '') for c in allowed_consequences if c.endswith('__canonical')
+        }
+        canonical_consequences -= allowed_consequences
+
+        all_consequence_ids = None
+        if canonical_consequences:
+            all_consequence_ids = hl.set({
+                *self._get_allowed_consequence_ids(canonical_consequences), *allowed_consequence_ids,
+            })
+        elif not allowed_consequence_ids:
+            return None
+
+        allowed_consequence_ids = hl.set(allowed_consequence_ids) if allowed_consequence_ids else hl.empty_set(hl.tint)
         allowed_transcripts = self._ht.sorted_transcript_consequences.filter(
-            lambda tc: tc.consequence_term_ids.any(allowed_consequence_ids.contains)
+            lambda tc: tc.consequence_term_ids.any(
+                (hl.if_else(hl.is_defined(tc.canonical), all_consequence_ids, allowed_consequence_ids)
+                 if canonical_consequences else allowed_consequence_ids
+            ).contains)
         )
+
         annotation_exprs['allowed_transcripts'] = allowed_transcripts
         return hl.is_defined(allowed_transcripts.first())
 
--- a/hail_search/queries/base.py
+++ b/hail_search/queries/base.py
@@ -644,14 +644,12 @@ class BaseHailTableQuery(object):
             ann for field, anns in annotations.items()
             if anns and (field not in ANNOTATION_OVERRIDE_FIELDS) for ann in anns
         }
-        allowed_consequence_ids = self._get_enum_terms_ids(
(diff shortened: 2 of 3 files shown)


hail_search/queries/mito.py
```diff
@@ -181,10 +181,28 @@ def _prefilter_entries_table(self, ht, parsed_intervals=None, exclude_intervals=
             ht = hl.filter_intervals(ht, parsed_intervals, keep=False)
         return ht
 
-    def _get_consequence_filter(self, allowed_consequence_ids, annotation_exprs):
+    def _get_consequence_filter(self, allowed_consequence_ids, allowed_consequences, annotation_exprs):
+        canonical_consequences = {
+            c.replace('__canonical', '') for c in allowed_consequences if c.endswith('__canonical')
+        }
+        canonical_consequences -= allowed_consequences
```
bpblanken, line 188: Is this line needed if we have the `c.endswith('__canonical')` in the set expression above?

hail_search/queries/mito.py
```diff
@@ -181,10 +181,28 @@ def _prefilter_entries_table(self, ht, parsed_intervals=None, exclude_intervals=
...
+
+        all_consequence_ids = None
+        if canonical_consequences:
+            all_consequence_ids = hl.set({
+                *self._get_allowed_consequence_ids(canonical_consequences), *allowed_consequence_ids,
+            })
+        elif not allowed_consequence_ids:
```
bpblanken, line 195: I think this line, [line 198](https://github.com/broadinstitute/seqr/pull/3606/files#diff-71932dea9e59d8d1ce70d88444658acffce2f3865c9afb11b57f93daf6610f66R198) and [line 202](https://github.com/broadinstitute/seqr/pull/3606/files#diff-71932dea9e59d8d1ce70d88444658acffce2f3865c9afb11b57f93daf6610f66R202) can maybe be consolidated in some way.

Maybe letting `all_consequence_ids` default to `hl.set(allowed_consequence_ids)` and checking the existence of `allowed_consequence_ids` at the top of the function?

bpblanken (commented):
so nice that search can handle this reasonably without needing pipeline changes!

hail_search/queries/mito.py
hanars: yeah because if a user selected both the canonical filter and the regular filter for that same consequence, by definition all the canonicals will also pass the regular so theres no need to do any special canonical handling for that consequence

hail_search/queries/mito.py
hanars: I tried that but the issue is that we need to define the expression inside the lambda in order to have the consequence in scope to check for canonical. I could not come up with a cleaner way to combine these. I am open to suggestions, but I would need a bit more detailed guidance, as based on what you suggest here I can't figure out how to get it to work

bpblanken: would something like this work?

```
allowed_consequence_ids = hl.set(allowed_consequence_ids) if allowed_consequence_ids else hl.empty_set(hl.tint32)
canonical_consequence_ids = hl.set(self._get_allowed_consequence_ids(canonical_consequences)) if canonical_consequences else hl.empty_set(hl.tint32)
allowed_transcripts = self._ht.sorted_transcript_consequences.filter(
    lambda tc: ((tc.canonical & tc.consequence_term_ids.any(canonical_consequence_ids.contains)) | tc.consequence_term_ids.any(allowed_consequence_ids.contains))
)
annotation_exprs['allowed_transcripts'] = allowed_transcripts
return hl.is_defined(allowed_transcripts.first())
```

hanars: This si less efficient though. The way I have it for each transcript it checks for the presence of the consequence in 1 set, here it potentially checks 2. I'm not sure this is substantively more readable than what I have, so the performance hit seems not worth it

bpblanken: 👍 Yep! there is two set checks.  Performance-wise though, I'm not sure it matters... since set creation itself will overwhelm set lookup:

```
In [25]: lookup = set(range(10))

In [26]: %timeit 'd' in lookup
16.5 ns ± 0.0281 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [27]: %timeit 'd' in set(range(10))
240 ns ± 2.52 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

if you don't think it's more readable though, then we can agree to disagree!

hanars: no thats a good point about set creation also being slow. It is slightly more readable the way you propose, so if thees no performance advantage to the way it is, I can change it

hanars: wait, except the way I have it written now the sets are created out of the loop, so that only happens at most twice, so then I do think the duplicate lookups is not worth it. I think this is a good agree to disagree situation

bpblanken: 👍

PR #3607, opened 2023-09-15
Change dataset_type VARIANTS to SNV_INDEL

hanars:


ui/shared/utils/constants.js
```diff
@@ -111,7 +111,7 @@ export const MATCHMAKER_CONTACT_URL_FIELD = {
 
 // SAMPLES
 
-export const DATASET_TYPE_VARIANT_CALLS = 'VARIANTS'
+export const DATASET_TYPE_VARIANT_CALLS = 'SNV_INDEL'
```
bpblanken, line 114: Is it painful to also rename the constant `DATASET_TYPE_SNV_INDEL_CALLS`?

seqr/migrations/0054_alter_sample_dataset_type.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 3.2.18 on 2023-09-15 14:54
+
+from django.db import migrations, models
+
+OLD_DT = 'SNV_INDEL'
```
bpblanken, line 5: Are `OLD_DT` and `NEW_DT` meant to be the same?

ui/shared/utils/constants.js
hanars: nope! thats a good idea

seqr/migrations/0054_alter_sample_dataset_type.py
hanars: 🤦 no, there was an overzealous refactor that I failed to fully undo. I tested it locally with them being different, and then later this got changed, I will change it back

bpblanken (approved):

PR #3623, opened 2023-09-21
Dockerfile for hail search.

bpblanken:


deploy/docker/hail_search/start_server.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+gunicorn hail_search.web_app:init_web_app -c /gunicorn_config.py
```
bpblanken, line 2: I opted for `gunicorn` here for consistency with the Django app and because it's more "best-practicey".  I'm not sure it's necessary though, and it requires a pip install inside the docker container.

[Docs here](https://docs.aiohttp.org/en/stable/deployment.html#start-gunicorn)

.github/workflows/hail-search-unit-tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# Run the test suite on pushes (incl. merges) to master and dev
+# Run the test suite when a PR is opened, pushed to, or reopened
+on:
+  push:
+    branches:
+      - master
+      - dev
+    paths:
+      - 'hail_search/**'
```
bpblanken, line 11: Split the `hail_search` unit tests out into their own GitHub action so that we can trigger the docker build based on this `paths_switch`.

deploy/docker/hail_search/readiness_probe
```diff
@@ -0,0 +1 @@
+curl http://localhost:${HAIL_SEARCH_SERVICE_PORT} | grep -i "html" >& /readiness_probe.log
```
bpblanken, line 1: k8s conceptual question here.  I'm using an environment variable for the service port here, to mirror what the seqr app is doing.  But I'm actually wondering if we should have this be hard-coded to 8000, which is hard-coded inside the container and also as the `containerPort` [here](https://github.com/broadinstitute/seqr-helm/blob/main/charts/seqr/templates/deployment.yaml#L38).  From my understanding, "service port" is the port the service itself is listening, and that what we really want is the 'containerPort' or the service's 'targetPort' for the readiness probe.

hail_search/web_app.py
```diff
@@ -29,7 +29,7 @@ async def status(request: web.Request) -> web.Response:
     return web.json_response({'success': True})
 
 
-def init_web_app():
+async def init_web_app():
```
bpblanken, line 32: I didn't change this yet, but `load_globals()` uses `hail` functions so we should add `hl.init()` to this function.

.dockerignore
```diff
@@ -5,7 +5,8 @@ static/*
 !static/images/*
 deploy/*
 !deploy/docker/seqr/*
-hail_search/*
+!deploy/docker/hail_search/*
+hail_search/fixtures/*
```
hanars, line 9: one thing to note is this is including the hail search code in the main seqr docker image, which it previously wasn't. Not sure its a huge deal, but I liked having them be totally distinct

deploy/docker/hail_search/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM hailgenetics/hail:0.2.115
```
hanars, line 1: for the protoype, we used a much simpler dockerfile - https://github.com/broadinstitute/seqr/blob/d2c4b3c7022166c66fdd50113eb8e16551551482/deploy/docker/hail_search/Dockerfile

Is there a reason to add all this here? I really liked having a slim image with no additional deployment requirements, and I feel like gunicorn is really overkill here

deploy/docker/hail_search/start_server.sh
hanars: I don't love it, and I don't really understand what advantage we are getting by having it. Aiohttp seems to work just fine on its own

deploy/docker/hail_search/readiness_probe
hanars: FWIW I don't really see any point in mirroring the seqr app, its not like its a particularly great example of using docker and kubernetes. I would go with what seems logical, not whats consistent with seqr

hail_search/web_app.py
hanars: I think that messes up unit tests, so I would be careful and do it as its own PR

hanars (changes requested):

deploy/docker/hail_search/Dockerfile
bpblanken: It's all gunicorn, except for the readiness_probe, which probably need.  It's possible to configure the readiness probe entirely in k8s though, rather than having a command, which we can opt for.

.dockerignore
bpblanken: What would be the list of directories that we'd want in the `seqr` Dockerfile?

```
admin
matchmaker
panelapp
reference_data
seqr
static
ui
```?  

I think being explicit is probably the best option at this point.

Comment (hanars) [reply .dockerignore:9]: this looks right to me

File: deploy/docker/hail_search/start_server.sh (hunk 1 above)
Comment (bpblanken) [reply deploy/docker/hail_search/start_server.sh:2]: yeah, I don't love it either.

I think my gut was to go for it because `seqr` is using it, and I've consistently seen the `nginx -> gunicorn -> python web framework` stack, with `gunicorn` doing the process management (how many workers, what happens when they die etc, routing concurrent requests to different workers) as well as the gateway interface stuff.  I _think_ the gateway stuff is unnecessary because aiohttp has support for it (in ways that Django and Flask do not).

Regarding concurrency, I _think_ we only ever want to run a single worker here right?  So that a single request blocks every subsequent request while it's running?  With concurrency being managed at the k8s level (I want 4 containers running, with load distributed by the nginx-ingress)?  In that case, we should be fine to remove `gunicorn`.  

@sjahl if you had any thoughts.

Comment (sjahl) [reply deploy/docker/hail_search/start_server.sh:2]: historically, i've always had gunicorn on python web app deployments for the reasons Ben mentioned... the app servers from flask and django are usually not suitable for production without gunicorn and a webserver in front of them.

I can see the process lifecycle stuff being useful here, for restarting processes that crash? if the app containers spawn quickly and kubernetes can determine that dead containers need replacing, maybe it's less critical. But, we'd also want to keep in mind that the k8s health check would need to be fairly conservative to avoid killing pods unnecessarily... so you're probably looking at a several second health check interval, and a certain number of consecutive failures before taking action.

In any case, if this is only going to dev, i'd be pro experimentation and seeing how well it handles things without the gunicorn layer.

Comment (bpblanken) [reply deploy/docker/hail_search/start_server.sh:2]: +1

I think I'm generally in favor of putting the concurrency management in the kubernetes here and only having a single process, that if it fails causes the Docker container to exit, run as a server.  This means that our health checks will be completely inaccessible while a hail query or any other request is running, but allows us to prevent multiple hail queries running concurrently within the same container.  Our health check will need to be very very conservative.

File: hail_search/web_app.py (hunk 4 above)
Comment (bpblanken) [reply hail_search/web_app.py:32]: I added `hl.init(idempotent=True)`, which I think is equivalent to what's happening under the hood now.

Comment (bpblanken) [reply hail_search/web_app.py:32]: Tests are ok!

File: deploy/docker/hail_search/start_server.sh (hunk 1 above)
Comment (bpblanken) [reply deploy/docker/hail_search/start_server.sh]: I found [a question](https://github.com/aio-libs/aiohttp/discussions/6805) which is unanswered, but at least points to somebody else having similar thoughts.

Comment (bpblanken) [reply deploy/docker/hail_search/start_server.sh]: And a [decent discussion](https://github.com/aio-libs/aiohttp/issues/552) about deployment

File: deploy/docker/hail_search/sync_datasets [hunk 7]
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
bpblanken, line 1: I've pushed the `initContainers` logic into this script to help with the environment variable looping piece.  To me, this is _right_ on the line of needing to be written in python.

deploy/docker/hail_search/sync_datasets
```diff
@@ -0,0 +1,23 @@
...
+)
+
+function to_gcs_path () {
+    # e.g. GRCh38_SNV_INDEL_VERSION -> GRCh38/SNV_INDEL
+    echo $1 | sed "s|GRCh38_|GRCh38/|g" | sed "s|GRCh37_|GRCh37/|g" | sed "s|_VERSION||g"
+}
+
+for suffix in "${suffixes[@]}"; do
+    path=$(to_gcs_path $suffix)
+    gsutil -m rsync -r $SOURCE_DATASETS_DIR/$path/runs/${!suffix} $DATASETS_DIR/$path
```
bpblanken, line 22: A combination of env vars like:
```
SOURCE_DATASETS_DIR=gs://seqr-datasets/v03
DATASETS_DIR=/datasets
GRCh38_SNV_INDEL_VERSION=manual__2023-07-07T18-43-59.702744+00-00
```
doe the trick.

deploy/docker/seqr/Dockerfile
```diff
@@ -80,7 +80,13 @@ RUN echo "deb http://packages.cloud.google.com/apt cloud-sdk-bullseye main" > /e
...
 
 WORKDIR /seqr
-COPY . .
+COPY admin/ .
+COPY matchmaker/ .
+COPY panelapp/ .
+COPY reference_data/ .
+COPY seqr/ .
+COPY static/ .
+COPY ui/ .
```
hanars, line 89: actually, I'm not sure we need ui here because we copy the build ui below, but I would double check that

.github/workflows/hail-search-unit-tests.yaml
hanars: If we update the docker image but not the search code it looks like this won't get run. Could we also add the path to the dockerfile here?

deploy/docker/hail_search/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+LABEL maintainer="Broad TGG"
+
+WORKDIR /hail_search
+
+# Infrastucture Code
+COPY deploy/docker/hail_search/sync_datasets /
```
hanars, line 8: might be a little neater to have a `bin` directory or another scripts directory with `sync_datasets `in it and copy that whole directory, so if we ever need to add another script to our deployment we don't have to keep updating a list here

deploy/docker/hail_search/sync_datasets
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -e
+
+suffixes=(
+    GRCh37_SNV_INDEL_VERSION
+    GRCh37_MITO_VERSION
+    GRCh37_SV_VERSION
+    GRCh37_GCNV_VERSION
```
hanars, line 8: we are not going to have GRCh37 SV, GCNV, or MITO data (we will never have internal data for it, so it would only be possible once we make these available to anvil users, which will involve lots of code changes) so we may want to take those out

hail_search/__main__.py
```diff
@@ -7,12 +7,11 @@
 
 def run():
     logging.basicConfig(level=logging.INFO)
-    hl.init()
     app = init_web_app()
     web.run_app(
         app,
         host='0.0.0.0',  # nosec
-        port=5000,
+        port=8000,
```
hanars, line 14: When running seqr locally, seqr is running on port 8000 , so this change makes it impossible to run seqr and search locally at the same time. Is there a reason for this change?

hail_search/web_app.py
```diff
@@ -29,12 +29,13 @@ async def status(request: web.Request) -> web.Response:
...
+    hl.init(idempotent=True)
+    load_globals()
```
hanars, line 34: not  an issue, but curious why its better to load globals before app initialization?

.github/workflows/hail-search-unit-tests.yaml
bpblanken: yeah, we can add no problem, but do we want to add something in here that actually uses/builds the docker image?

deploy/docker/hail_search/sync_datasets
bpblanken: Yep, I think just removing for now makes sense.

hail_search/web_app.py
bpblanken: I moved the line so that "hail" code was being executed together, but other than that no reason 🤷

.github/workflows/hail-search-unit-tests.yaml
hanars: I guess not, I don't think we actually need to run the unit tests if the docker image changes, but I guess I was trying to solve the issue that the docker build is only triggered based on this job. If theres a different way to trigger building a docker image when the dockerfile is updated, that seems better to me

hail_search/__main__.py
bpblanken: I think I was actually looking for consistency _with_ the seqr app, because the service port itself doesn't matter if we're running a separate k8s service, but that doesn't make sense at all with local development.  I'll revert.

.github/workflows/hail-search-unit-tests.yaml
bpblanken: Maybe splitting out the `Dockerfile Linting` job into one-per service also makes sense?

deploy/docker/hail_search/bin/sync_datasets
```diff
@@ -0,0 +1,20 @@
...
+)
+
+function to_gcs_path () {
+    # e.g. GRCh38_SNV_INDEL_VERSION -> GRCh38/SNV_INDEL
+    echo $1 | sed "s|GRCh38_|GRCh38/|g" | sed "s|GRCh37_|GRCh37/|g" | sed "s|_VERSION||g"
+}
+
+for suffix in "${suffixes[@]}"; do
+    path=$(to_gcs_path $suffix)
+    gsutil -m rsync -r $SOURCE_DATASETS_DIR/$path/runs/${!suffix} $DATASETS_DIR/$path
```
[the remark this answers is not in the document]
hanars: for the prototype, I didn't run the copying in the seqr from the hail search image at all, I ran it as an initContainer defined in the yaml file using a gcloud cli docker image: https://github.com/broadinstitute/seqr/compare/dev...hail-backend#diff-1170a92e32cc8b8d3707780992a5ce0046c05f6d2c7e3560c71694ee6c1e4c71R62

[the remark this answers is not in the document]
hanars: I feel like templating that out in helm would not be too bad actually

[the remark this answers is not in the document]
bpblanken: Yep, my initial pass was using that image... then I moved the code here because the bash got unwieldy in helm.  I'll give it another shot though, since I think installing `gcloud` into this image isn't great.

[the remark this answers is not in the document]
bpblanken: I [end up with this](https://github.com/broadinstitute/seqr-helm/pull/19/files#diff-5f718a9c818a4d6921db78158471daa121a855a05148515dc9938ebd28711955R35-R48) which I think is the most reasonable thing at this point.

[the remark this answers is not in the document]
hanars: we could also store the variables in some sort of dictionary, so the values would look like
```
versions:
  GRCh37/SNV_INDEL: manual__2023-..
  GRCh38/SNV_INDEL: manual__2023-..
  GRCh38/MITO: manual__2023-..
  GRCh38/GCNV: manual__2023-..
  GRCh38/SV: manual__2023-..
```

and then the helm template would look like
```
{{- range $path, $version := .Values.environment.versions }}
  gsutil -m rsync -r {{ .Values.environment.SOURCE_DATASETS_DIR }}/{{ $path }}/runs/{{ version }} {{ .Values.environment.DATASETS_DIR }}/{{ $path }};
{{- end }}
```

.github/workflows/hail-search-unit-tests.yaml
bpblanken: actually, what about just moving the `Dockerfile` to a `deploy` directory inside `hail_search`?

hanars: yes! I like this

deploy/docker/hail_search/bin/sync_datasets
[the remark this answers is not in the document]
bpblanken: yep! something like that works.

hanars (approved):

PR #3632, opened 2023-09-26
Do not delete search samples if hail backend enabled

hanars:


seqr/views/utils/individual_utils.py
```diff
@@ -195,6 +193,27 @@ def delete_individuals(project, individual_guids, user):
...
+
+
+def _validate_no_submissions(individuals_to_delete):
+    return _validate_delete_individuals(
+        individuals_to_delete, 'MME submission',
+        dict(matchmakersubmission__isnull=False, matchmakersubmission__deleted_date__isnull=True)
+    )
+
+
+def _validate_no_search_samples(individuals_to_delete):
```
bpblanken, line 211: would it be clearer to name this function `_validate_no_search_and_no_submissions_samples`, 
just so it's clear in [line 177] above that the hail backend isn't just validating search, but both search and submissions?

bpblanken (approved):

PR #3636, opened 2023-09-27
Local dev run-through changes

bpblanken:
I got local seqr running and hit a few small things.

### DIFF
--- a/deploy/LOCAL_DEVELOPMENT_INSTALL.md
+++ b/deploy/LOCAL_DEVELOPMENT_INSTALL.md
@@ -17,8 +17,7 @@ After installation, run `psql -l` and if there is no user named `postgres`, run 
 
 - [redis](https://redis.io/topics/quickstart)
 
-- [node/npm <14](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).  Note: more recent versions of `node` may or may not function correctly and
-are not officially supported.
+- [node/npm <14](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).  Note: more recent versions of `node` may not function are not officially supported.
 
 Optionally, if planning to use elasticsearch from docker-compose, install:
 - [docker](https://docs.docker.com/install/)
--- a/deploy/LOCAL_DEVELOPMENT_INSTALL.md
+++ b/deploy/LOCAL_DEVELOPMENT_INSTALL.md
@@ -17,7 +17,8 @@ After installation, run `psql -l` and if there is no user named `postgres`, run 
 
 - [redis](https://redis.io/topics/quickstart)
 
-- [node/npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+- [node/npm <14](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).  Note: more recent versions of `node` may or may not function correctly and
+are not officially supported.
 
 Optionally, if planning to use elasticsearch from docker-compose, install:
 - [docker](https://docs.docker.com/install/)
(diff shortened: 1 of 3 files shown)


deploy/LOCAL_DEVELOPMENT_INSTALL.md
```diff
@@ -17,7 +17,7 @@ After installation, run `psql -l` and if there is no user named `postgres`, run
 
 - [redis](https://redis.io/topics/quickstart)
 
-- [node/npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+- [node/npm <14](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm).  Note: more recent versions of `node` may not function are not officially supported.
```
bpblanken, line 20: `node 20` (brew install default) hits [this issue](https://stackoverflow.com/questions/75167770/error-error0308010cdigital-envelope-routinesunsupported-node-js-v19-4-0)

deploy/LOCAL_DEVELOPMENT_INSTALL.md
```diff
@@ -82,6 +82,7 @@ export ANALYST_USER_GROUP=TGG_Users
 export PM_USER_GROUP=TGG_PM
     
 # Set the client ID and secret for the seqr-local OAuth client (from GCP)
+# Note: do not use the values from `seqr-secrets` in secret manager, the local credentials are saved [here](https://console.cloud.google.com/apis/credentials?project=seqr-project).
```
bpblanken, line 85: I initially fetched these values in the same way the infrastructure does... which was wrong.  Took a bit of poking to figure out why.

settings.py
```diff
@@ -258,8 +258,8 @@
         'http://localhost:3000',
         'http://localhost:8000',
     )
-    # STATICFILES_DIRS.append(STATIC_ROOT)
-    # STATIC_ROOT = None
+    STATICFILES_DIRS.append(STATIC_ROOT)
```
bpblanken, line 261: my static assets didn't load with this commented out... was there maybe a reason?

hanars: were you running localhost:8000 or localhost:3000 ?

deploy/LOCAL_DEVELOPMENT_INSTALL.md
hanars: 😢  can you make a ticket that we should update to a more recent node version? Its low priority, but worth doing at some point, and for now this works

settings.py
hanars: I think the reason we left these commented out is if you want to use the built assets and localhost:8000 you have to uncomment them. But if you want to be testing using the live javascript assets on localhost:3000, which is how we basically always want to do locally since the built assets in the seqr repo are not necessarily current, then they need to be commented out

bpblanken: I'm running the js on `localhost:3000`, which is proxying to the Django server on 8000.  The dev server live updating seems to be working.  

The issue is isolated to the icons in `seqr/static` itself.   I don't have any actually built javascript in `ui/dist`.

deploy/LOCAL_DEVELOPMENT_INSTALL.md
bpblanken: Ticket created [here](https://github.com/broadinstitute/seqr/issues/3640)

hanars (approved):

settings.py
hanars: For future reference, this breaks prod deploys which is why we've always kept it commented out 🤦 
https://github.com/broadinstitute/seqr/actions/runs/6435806250/job/17477833506

PR #3637, opened 2023-09-27
Add helpText for SpliceAI

bpblanken:
Screenshot 2023-09-27 at 5 01 15 PM

### DIFF
--- a/ui/shared/components/panel/search/constants.js
+++ b/ui/shared/components/panel/search/constants.js
@@ -453,7 +453,7 @@ const REQUIRE_SCORE_FIELD = {
   labelHelp: 'Only return variants where at least one filtered predictor is present. By default, variants are returned if a predictor meets the filtered value or is missing entirely',
 }
 export const IN_SILICO_FIELDS = [REQUIRE_SCORE_FIELD, ...PREDICTOR_FIELDS.filter(({ displayOnly }) => !displayOnly).map(
-  ({ field, fieldTitle, indicatorMap, group, min, max }) => {
+  ({ field, fieldTitle, warningThreshold, dangerThreshold, indicatorMap, group, min, max }) => {
     const label = fieldTitle || snakecaseToTitlecase(field)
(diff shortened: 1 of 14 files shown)


hanars:
Sorry, I should have checked that this ticket was clearer before assigning it over to you. When the ticket says we should show it "similar to how we show it for the Pathogenicity override", it means we should show it directly in the panel above all the filters, not on a hover. The text we should use should be something like "Filter by reported annotation. Variants will be returned if they have ANY of the specified annotations, including if they have a Splice AI score above the threshold and no other annotations. This filter is overridden by the pathogenicity filter, so variants will be returned if they have the specified pathogenicity even if none of the annotation filters match"

 
Screenshot 2023-09-28 at 8 35 36 AM

bpblanken:
👍 Do we also want to mention "screen"?  It has the same behavior as splice_ai I believe?

hanars:
I feel like screen is covered by "ANY of the specified annotations", it pretty much looks to users like to other annotations in the filter so while in the database th efiltering is quite different, to the user I do not think it is

bpblanken:
Updated screenshot:
Screenshot 2023-09-28 at 2 54 59 PM

ui/shared/components/panel/search/constants.js
```diff
@@ -472,13 +472,6 @@ export const IN_SILICO_FIELDS = [REQUIRE_SCORE_FIELD, ...PREDICTOR_FIELDS.filter
     const labelHelp = (
       <div>
         {`Enter a numeric cutoff for ${label}`}
-        {dangerThreshold && (
```
bpblanken, line 475: I believe this is dead code after the recent changes... is that correct?

hanars: yeah, but actually we should be replacing this with updated logic from the config thresholds, not just removing it. I would leave it for now and make a ticket to properly update it (or you can give that a go if you want)

hanars (approved):

PR #3644, opened 2023-10-02
Require Sample tissue type - fix rna outlier duplicated data

hanars:
The lack of tissue type for rna outliers was causing the logic for reusing and replacing RNA data on reload to break, causing data to be loaded multiple times instead of overwriting previous data. This PR makes tissue type required at the database level to prevent leaving it out, and migrates the previously loaded data so that the duplicated data is cleaned up


seqr/views/apis/data_manager_api_tests.py
```diff
@@ -819,194 +828,194 @@ def _check_rna_sample_model(self, individual_id, data_source, tissue_type):
...
     @mock.patch('seqr.utils.file_utils.subprocess.Popen')
     @mock.patch('seqr.views.apis.data_manager_api.gzip.open')
-    def test_update_rna_seq(self, mock_open, mock_subprocess, mock_load_uploaded_file,
+    def _test_update_rna_seq(self, data_type, mock_open, mock_subprocess, mock_load_uploaded_file,
                             mock_os, mock_datetime, mock_send_slack):
         url = reverse(update_rna_seq)
         self.check_data_manager_login(url)
 
-        for data_type, params in self.RNA_DATA_TYPE_PARAMS.items():
-            with self.subTest(data_type):
```
hanars, line 828: this needed to be broken into separate tests instead of a subTest because the different dat atypes share Sample models now and modify them so the tests have unexpected interactions, but if they are broken into explicit separate tests they each run on a clean database and work as expected

hanars: The only change from here down is indentation, nothing was actually changed

seqr/views/utils/dataset_utils.py
```diff
@@ -241,16 +241,17 @@ def _get_unmatched_error(sample_keys):
 def _parse_tsv_row(row):
     return [s.strip().strip('"') for s in row.rstrip('\n').split('\t')]
 
-PROJECT_COL = 'project'
-RNA_OUTLIER_COLUMNS = {'geneID': 'gene_id', 'pValue': 'p_value', 'padjust': 'p_adjust', 'zScore': 'z_score',
-                       PROJECT_COL: PROJECT_COL}
 
+PROJECT_COL = 'project'
```
hanars, line 245: These database/ outlier input changes allow much better shared behaviors across the RNA data types, so the code is refactored here to take advantage of that

seqr/views/utils/dataset_utils.py
```diff
@@ -356,55 +326,76 @@ def _load_rna_seq_file(file_path, user, mapping_file, parse_row, expected_column
...
     sample_id_to_tissue_type = {}
     samples_with_conflict_tissues = defaultdict(set)
     errors = []
+    missing_required_fields = defaultdict(list)
     gene_ids = set()
-    missing_gene_samples = set()
     for line in tqdm(f, unit=' rows'):
         row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in parse_row(row):
-            tissue_type = TISSUE_TYPE_MAP.get(row_dict.pop(TISSUE_COL, None))
```
hanars, line 371: a lof of this change is just unindenting, since theres no longer `parse_row` iterator with a for loop and instead just a one-to-one parsing of each row

seqr/migrations/0055_alter_sample_tissue_type.py
```diff
@@ -0,0 +1,93 @@
...
+    has_rna_data = {
+        field: set(dup_samples.filter(**{f'{field}__isnull': False}).values_list('id', flat=True))
+        for field in rna_data_classes
+    }
+    for new_s_id, old_sample_ids in tqdm(dup_id_map.items(), unit=' duplicates'):
+        for field, model_cls in rna_data_classes.items():
+            has_data_samples = has_rna_data[field]
+            for old_s_id in old_sample_ids:
+                if old_s_id in has_data_samples and new_s_id not in has_data_samples:
+                    model_cls.objects.using(db_alias).filter(sample_id=old_s_id).update(sample_id=new_s_id)
```
bpblanken, line 68: Is it correct that this reassignment of the duplicate sample id to the `new_s_id` can happen several times, once for each `rna_data_class`?  

Or is it that a sample may only have one non null `rna_data_class`, so that the `values()` stored in `has_rna_data` are disjoint.

seqr/migrations/0055_alter_sample_tissue_type.py
```diff
@@ -0,0 +1,93 @@
...
+        'rnaseqtpm': apps.get_model('seqr', 'RnaSeqTpm'),
+    }
+    db_alias = schema_editor.connection.alias
+
+    samples = Sample.objects.using(db_alias).filter(tissue_type__isnull=True)
+    if not samples:
+        return
+
+    rna_samples = samples.filter(sample_type='RNA', is_active=True)
+    non_rna_samples = samples.exclude(id__in=[s.id for s in rna_samples])
```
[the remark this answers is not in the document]
hanars: no because its AND vs OR logic. The equivalent would be `filter(~Q(sample_type='RNA') | Q(is_active=False))` which felt less readable but I can change to if you think its more explicit

seqr/migrations/0055_alter_sample_tissue_type.py
hanars: reassignment can happen multiple times for different data classes, you could have scenario where it needs to happen for, say, both tpm and outlier because the old sample has those and the new sample only has splice. But you also have a scenario where the old sample has outlier and tpm models and the new samples only has tpm, so we want to assign the old samples outlier modes to the new sample but not the old tpm modles

seqr/migrations/0055_alter_sample_tissue_type.py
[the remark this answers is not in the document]
bpblanken: cool, I wouldn't bother changing it since this is a migration.  That makes sense to me though.

seqr/migrations/0055_alter_sample_tissue_type.py
```diff
@@ -0,0 +1,93 @@
...
+    grouped_samples = defaultdict(set)
+    for s in rna_samples:
+        if len(individual_tissue_sample_map[s.individual_id]) == 1:
+            s.tissue_type = list(individual_tissue_sample_map[s.individual_id].keys())[0]
+        elif s.data_source.startswith('fibroblasts__'):
+            s.tissue_type = 'F'
+        elif s.data_source.startswith('muscle__'):
+            s.tissue_type = 'M'
+        else:
+            s.tissue_type = SAMPLE_TISSUE_MAP[s.sample_id]
```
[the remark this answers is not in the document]
bpblanken: ok, that seems reasonable.  Was curious about why it wasn't just `s.tissue_type = 'A'`.

PR #3646, opened 2023-10-04
Missing paths filter in GitHub actions.

bpblanken:
Opening this pr should also provide a test of itself!

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -9,10 +9,12 @@ on:
       - dev
     paths-ignore:
       - 'hail_search/**'
+      - '.github/workflows/hail-search-unit-tests.yaml'
   pull_request:
     types: [opened, synchronize, reopened]
     paths-ignore:
       - 'hail_search/**'
+      - '.github/workflows/hail-search-unit-tests.yaml'
 
 jobs:
   python:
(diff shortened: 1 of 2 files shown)


sjahl (approved):
this looks fine to me. Kind of wish paths-ignore was a workflow-wide thing, so we didn't need to put it in each of `{push,pull_request,etc}`.

One thing I might add, is an ignore in the seqr unit tests, so that updates to the hail-search-unit-tests.yaml file don't trigger the full seqr test suite.

.github/workflows/unit-tests.yml
```diff
@@ -9,8 +9,12 @@ on:
       - dev
     paths-ignore:
       - 'hail_search/**'
+      - '.github/workflows/hail-search-unit-tests.yaml'
   pull_request:
     types: [opened, synchronize, reopened]
+    paths-ignore:
+      - 'hail_search/**'
+      - '.github/workflows/hail-search-unit-tests.yaml'
```
bpblanken, line 17: I thought this wasn't working correctly because the seqr unit tests are getting triggered, but this file is `unit-tests.yml`, which should trigger the tests.  🤦

hanars (approved):

.github/workflows/unit-tests.yml
sjahl: been there, done that, got the T shirt 😂

PR #3665, opened 2023-10-10
Faq spanish

hanars:


ui/pages/Public/components/Faq.jsx
```diff
@@ -1,171 +1,503 @@
...
-    Alternatively, if you want to add data to the existing project, you will need to move the new joint called VCF to
-    the original workspace and request loading additional data from that project as described above.
-
-    <Header content="Q. Have ideas for seqr?" size="medium" />
-    We are excited to see seqr&apos;s features grow to support your and others analysis needs, and welcome your
-    suggestions or code contributions to the open source project. Please open a new  &nbsp;
-    <a href="https://github.com/broadinstitute/seqr/discussions" target="blank">Github discussion</a> to discuss your
-    proposed ideas, or email us at <a href="mailto:seqr@broadinstitute.org">seqr@broadinstitute.org</a>.
+const FAQS = [
+  {
```
bpblanken, line 12: Is it possible/reasonable for us to just change the `faq` route to redirect to `faq/en` or `faq/en-us` so that the object structure is consistent?

hanars: its possible, but we do not do it anywhere else and its a bit of a pain with the combination of server side and client side routing we use in seqr. That said, changing the object structure makes sense to me and would be more readable, but we can acheive that without changing the routing and just setting a default for the language when its not in the route

bpblanken: 👍 let's do that then!

bpblanken (approved):

PR #3667, opened 2023-10-11
update `hail-search` -> `hail_search`

bpblanken:


hanars:
Generally, we do not release directly to master without first going through dev as a matter of policy. If having this live in production is critical, you can do concurrent PRs to dev and master, and add the [release checklist](https://github.com/broadinstitute/seqr-private/blob/master/release_pr_checklist.md) to the master PR

bpblanken:
yeah, it should be into `dev`.  I branched correctly locally and didn't notice on the pr because `master` and `dev` are the same right now.

hanars (approved):

PR #3668, opened 2023-10-11
shared igv upload field

hanars:
Updates description for the "optional" gCNV column, and also combines the logic for the IGV upload field so future updates to this field can happen in one place instead of tracking it doen in multiple places

### DIFF
--- a/ui/shared/components/form/IGVUploadField.jsx
+++ b/ui/shared/components/form/IGVUploadField.jsx
@@ -1,0 +1,59 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { Table } from 'semantic-ui-react'
+
+import { NoBorderTable } from '../StyledComponents'
+import FileUploadField, { validateUploadedFile } from './XHRUploaderField'
+
+const IgvDropzoneLabel = ({ columns }) => (
+  <NoBorderTable basic="very" compact="very">
+    <Table.Body>
+      <Table.Row>
+        <Table.Cell colSpan={2}>
+          Upload a file that maps seqr Individuals to IGV file paths. Include one row per track.
+          <br />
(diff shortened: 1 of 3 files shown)


ui/shared/components/form/IGVUploadField.jsx
```diff
@@ -0,0 +1,59 @@
...
+  'IGV Track File Path',
+  'gCNV Sample ID, to identify the sample in the gCNV batch path. Not used for other track types',
+]
+
+// eslint-disable-next-line react-perf/jsx-no-new-array-as-prop
+const COLUMNS = ['Project', ...NO_PROJECT_COLUMNS]
+
+export const UPLOAD_IGV_FIELD = {
+  name: 'mappingFile',
+  validate: validateUploadedFile,
```
bpblanken, line 50: From what I can tell, this validator isn't able to check the column number/contents.  Should we also update [the backend](https://github.com/broadinstitute/seqr/blob/fb35df0b867504b71012d7d358cd4411c928faf9/seqr/views/apis/igv_api.py#L30) so that we throw an error if the wrong number of columns is supplied?

hanars: this field uses different URLs different that already properly handle their own validation

bpblanken: gotcha, do you mean `receive_igv_table_handler` and `receive_bulk_igv_table_handler`, or even different URLs?

hanars: even more different

bpblanken: ok.. I'm gonna ask about this in person cause I'm a bit confused but not a blocker here.

bpblanken (approved):

PR #3674, opened 2023-10-17
Hail backend - better error handling

hanars:


seqr/utils/search/hail_search_utils.py
```diff
@@ -15,7 +15,10 @@ def _hail_backend_url(path):
 
 def _execute_search(search_body, user, path='search'):
     response = requests.post(_hail_backend_url(path), json=search_body, headers={'From': user.email}, timeout=300)
-    response.raise_for_status()
+
+    if 400 <= response.status_code:
```
bpblanken, line 19: can we flip this to `response.status_code >= 400`?  This is a little [yoda-ey].(https://en.wikipedia.org/wiki/Yoda_conditions)

hail_search/web_app.py
```diff
@@ -1,9 +1,29 @@
...
 import hail as hl
+import logging
 
 from hail_search.search import search_hail_backend, load_globals
 
+logger = logging.getLogger(__name__)
+
+
+def _handle_exception(e, request):
+    if not e.text:
```
bpblanken, line 12: Does this line ever run?  It seems to me like only subclasses of `HTTPError` will enter here, and it seems like `aiohttp` manages the `reason` -> `text` propagation.

hanars: this line always runs, HTTPError leaves text blank by default

bpblanken: I think it [gets set for us](https://docs.aiohttp.org/en/v3.8.0/_modules/aiohttp/web_exceptions.html):
```
In [3]: try:
   ...:      raise HTTPError(reason='hello world')
   ...:   except HTTPError as e:
   ...:      print(e.text)
   ...:
-1: hello world
```

bpblanken (approved):
ping on the yoda condition if you want to fix it, but it's a nit anyways.

PR #3678, opened 2023-10-19
Hail backend post body size

hanars:
We ran into issues with the post body size for the all-projects searches. This removes some unnecessary data from the search bodies, and also changes the setting for the hail search server


hail_search/web_app.py
```diff
@@ -50,7 +50,7 @@ async def status(request: web.Request) -> web.Response:
 async def init_web_app():
     hl.init(idempotent=True)
     load_globals()
-    app = web.Application(middlewares=[error_middleware])
+    app = web.Application(middlewares=[error_middleware], client_max_size=(1024**2)*10)
```
bpblanken, line 53: out of curiosity, could we also [compress](https://stackoverflow.com/a/25284140) the post body?  I guess the pros are it would allow us to keep a tighter bound on the max size, and the cons are it's sorta unnecessary work?  🤷

bpblanken (approved):
inline question/thought exercise about compression but everything else is reasonable!

hail_search/web_app.py
hanars: I tried it, but the compression was not sufficient to get below the threshold so we would have to add all the work to handle compression and still need to increase the max size

PR #3679, opened 2023-10-20
Hail backend mito population fields

hanars:
Update hail backend code to work with the actual data schema for the MITO table


bpblanken (approved):
reminder that the `mito` clinvar field is going to change to `clinvar_mito` as soon as I re-run the pipeline.  Maybe a "blocked" ticket?

PR #3688, opened 2023-10-25
Comp het gene- hail backend

hanars:


hail_search/queries/base.py
```diff
@@ -679,18 +693,31 @@ def _filter_compound_hets(self):
...
+            k: ch_ht[k].filter(lambda t: t.gene_id == ch_ht.gene_ids)
+            for k in [ALLOWED_TRANSCRIPTS, ALLOWED_SECONDARY_TRANSCRIPTS] if k in ch_ht.row
+        }
+        if transcript_annotations:
+            ch_ht = ch_ht.annotate(**transcript_annotations)
+        primary_filters = self._get_annotation_filters(ch_ht)
+        secondary_filters = self._get_annotation_filters(ch_ht, is_secondary=True)
+
+        self.unfiltered_comp_het_ht = ch_ht.filter(hl.any(primary_filters + secondary_filters))
+        if self._has_secondary_annotations and not (primary_filters and secondary_filters):
```
bpblanken, line 708: This line is a little unclear to me.  I think right now `not (primary_filters and secondary_filters)` is `True` if either `primary_filters` or `secondary_filters` has length > 0.  Is that what we want?

hanars: In english, this is saying that if there are supposed to be secondary filters, but we do not have both primary and secondary filters, then there are no valid pars for this data type and we should break. So `(primary_filters and secondary_filters)` is only true if both have length > 0 , so the `not` means that if either have length 0 we meet the condition

PR #3692, opened 2023-10-30
Remove crons from app

bpblanken:
bam/crams job has finally succeeded!

### DIFF
--- a/deploy/docker/seqr/entrypoint.sh
+++ b/deploy/docker/seqr/entrypoint.sh
@@ -75,26 +75,5 @@ python -u manage.py check
 # launch django server in background
 /usr/local/bin/start_server.sh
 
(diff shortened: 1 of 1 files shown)


deploy/docker/seqr/entrypoint.sh
```diff
@@ -75,26 +75,5 @@ python -u manage.py check
 # launch django server in background
 /usr/local/bin/start_server.sh
 
-if [ "$RUN_CRON_JOBS" ]; then
```
bpblanken, line 78: `RUN_CRON_JOBS` has been off in production for a bit

hanars: we shouldk probably fully remove it from the helm charts as well after this is deployed

Issue #3642, opened 2023-09-29
Use updated logic from config thresholds on Variant Search page.


bpblanken:
**Is your feature request related to a problem? Please describe.**
Variant Search page does not display color thresholds:
<img width="297" alt="Screenshot 2023-09-29 at 1 35 28 PM" src="https://github.com/broadinstitute/seqr/assets/5325727/2757a2d7-02a7-404b-a4ae-17c8594f8b0f">

**Describe the solution you'd like**
Also show the color thresholds for each predictor:
<img width="197" alt="Screenshot 2023-09-29 at 1 33 58 PM" src="https://github.com/broadinstitute/seqr/assets/5325727/b0d8c20f-e1ac-4fb2-8e65-0504a9a1ff73">

**Describe alternatives you've considered**
NA

**Additional context**
Per discussion [here](https://github.com/broadinstitute/seqr/pull/3637#discussion_r1340685577)

hanars:
this is now live in production

PR #3693, opened 2023-10-30
Shared config threshold logic.

bpblanken:
Ticket: #3642 

Screenshot 2023-10-26 at 12 10 54 PM


ui/shared/components/panel/search/constants.js
```diff
@@ -472,17 +473,12 @@ export const IN_SILICO_FIELDS = [REQUIRE_SCORE_FIELD, ...PREDICTOR_FIELDS.filter
...
-            <div>{`Red > ${dangerThreshold}`}</div>
-            <div>{`Yellow > ${warningThreshold}`}</div>
-          </div>
-        )}
+        {thresholds && predictorColorRanges(thresholds)}
       </div>
     )
     return {
       labelHelp,
+      popupProps: thresholds ? { hoverable: true } : {},
```
bpblanken, line 481: `hoverable` makes the link clickable.

ui/shared/components/panel/search/constants.js
```diff
@@ -472,17 +473,12 @@ export const IN_SILICO_FIELDS = [REQUIRE_SCORE_FIELD, ...PREDICTOR_FIELDS.filter
     const labelHelp = (
       <div>
         {`Enter a numeric cutoff for ${label}`}
-        {dangerThreshold && (
```
hanars, line 475: 🤦

ui/shared/components/panel/search/constants.js
hanars: I think its fine to just always make the help label popup `hoverable`, I don't think theres any real advantage to it not being anywhere else

ui/shared/utils/constants.js
```diff
@@ -1343,6 +1344,36 @@ export const PREDICTOR_FIELDS = [
   { field: 'mitotip', indicatorMap: MITOTIP_MAP },
   { field: 'hmtvar', thresholds: [undefined, undefined, 0.35, 0.35, undefined] },
 ]
+export const PRED_COLOR_MAP = ['green', 'olive', 'grey', 'yellow', 'red', '#8b0000']
```
hanars, line 1347: the name of this constant is okay for something used internally in a file, but if we are going to make it an importable constant we should probably name it something more decriptive and accurate, maybe `PREDICTION_COLOR_LIST` or `PREDICTION_COLORS`? Also, maybe it makes more sense not to export it at all and instead move all the helper functions that use it into this file, but thats up to you

bpblanken: I moved the function using `PRED_COLOR_MAP` into here.  Seemed cleaner!

hanars (approved):

PR #3694, opened 2023-10-30
use query table for variant id query

hanars:
Changes how variant-id specific queries are executed, uses query_table instead of filtering to specific variant IDs to improve performance

### DIFF
--- a/hail_search/queries/mito.py
+++ b/hail_search/queries/mito.py
@@ -181,7 +181,7 @@ class MitoHailTableQuery(BaseHailTableQuery):
 
         return [
             hl.struct(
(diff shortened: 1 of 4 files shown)


hail_search/queries/mito.py
```diff
@@ -174,6 +175,18 @@ def _filter_variant_ids(self, ht, variant_ids):
             ])
         return ht.filter(variant_id_q)
 
+    def _parse_variant_keys(self, variant_ids=None, **kwargs):
+        if not variant_ids:
+            return variant_ids
+
+        should_add_chr_prefix = any(c.startswith('chr') for c in hl.get_reference(self._genome_version).contigs)
```
bpblanken, line 182: Could we just do: 

```
should_add_chr_prefix = self._genome_version == 'GRCh38'
```
on this one?

bpblanken (approved):

PR #3697, opened 2023-10-31
Remove dependency on service account key

bpblanken:
Service account is now enabled via workload identity (at least on dev!)


deploy/docker/seqr/entrypoint.sh
```diff
@@ -17,22 +17,6 @@ if [ "$GCLOUD_ZONE" ]; then
     gcloud config set compute/zone "$GCLOUD_ZONE"
 fi
 
-if [ -e "/.config/service-account-key.json" ]; then
```
hanars, line 20: 😍

hanars (approved):
we should probably do a follow up PR once this is live to clean up our helm templates/ configs

PR #3711, opened 2023-11-07
Update LOCAL_INSTALL.md

TheACKx:
To update reference data in seqr code section it writes manage.py and that causes errors need to write  ./manage.py  
image

Wrong one (in docs) : 
docker-compose exec seqr manage.py update_all_reference_data --use-cached-omim --skip-gencode
image

Fixed one : 
docker-compose exec seqr ./manage.py update_all_reference_data --use-cached-omim --skip-gencode


hanars (approved):
thanks for fixing!

PR #3717, opened 2023-11-08
Remove hail shuffle

hanars:
Updates key_by behavior to keep datasets sorted by locus-allele instead of switching to the non-sorted variant_id

### DIFF
--- a/hail_search/test_search.py
+++ b/hail_search/test_search.py
@@ -300,9 +300,9 @@ class HailSearchTestCase(AioHTTPTestCase):
             inheritance_mode=inheritance_mode, gene_counts={
                 'ENSG00000097046': {'total': 2, 'families': {'F000002_2': 2}},
                 'ENSG00000177000': {'total': 2, 'families': {'F000002_2': 2}},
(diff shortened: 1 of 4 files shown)


hail_search/test_search.py
```diff
@@ -300,9 +300,9 @@ async def test_inheritance_filter(self):
             inheritance_mode=inheritance_mode, gene_counts={
                 'ENSG00000097046': {'total': 2, 'families': {'F000002_2': 2}},
                 'ENSG00000177000': {'total': 2, 'families': {'F000002_2': 2}},
-                'ENSG00000275023': {'total': 2, 'families': {'F000002_2': 2}},
-                'ENSG00000277258': {'total': 2, 'families': {'F000002_2': 2}},
-                'ENSG00000277972': {'total': 1, 'families': {'F000002_2': 1}},
+                'ENSG00000275023': {'total': 3, 'families': {'F000002_2': 3}},
```
hanars, line 303: Behavior for gene counts changes for how it counts variants thats are included as part of multiple comp het pairs. While different than before, it is still accurate just with a different definition

bpblanken (approved):

PR #3732, opened 2023-11-15
AIP integration

hanars:


seqr/fixtures/1kg_project.json
```diff
@@ -1693,6 +1693,22 @@
         "order": 16.0
     }
 },
+{
+    "model": "seqr.varianttagtype",
```
jklugherz, line 1697: what's the difference between a variant tag and a variant tag type?

jklugherz: we talked offline about data models :)

seqr/views/apis/summary_data_api.py
```diff
@@ -172,6 +183,140 @@ def bulk_update_family_analysed_by(request):
...
+        saved_variant_map.update(_search_new_saved_variants(new_variants, user))
+
+    aip_tag_type = VariantTagType.objects.get(name=AIP_TAG_TYPE, project=None)
+    existing_tags = {
+        tuple(t.saved_variant_ids): t for t in VariantTag.objects.filter(
+            variant_tag_type=aip_tag_type, saved_variants__in=saved_variant_map.values(),
+        ).annotate(saved_variant_ids=ArrayAgg('saved_variants__id', ordering='id'))
+    }
+
+    today = datetime.now().strftime('%Y-%m-%d')
```
bpblanken, line 221: Do we typically do dates in ET or UTC?

seqr/views/utils/variant_utils.py
```diff
@@ -58,6 +59,24 @@ def update_project_saved_variant_json(project, family_id=None, user=None):
     return updated_saved_variant_guids
 
 
+def parse_saved_variant_json(variant_json, family):
+    if 'xpos' not in variant_json:
+        variant_json['xpos'] = get_xpos(variant_json['chrom'], variant_json['pos'])
+    xpos = variant_json['xpos']
```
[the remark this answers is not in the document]
hanars: yeah, this is copy-paste but no harm in cleaning tup while I'm at it!

seqr/views/apis/summary_data_api.py
hanars: Nope! We probably should, but that would be an out of scope change all over the repo

seqr/views/utils/variant_utils.py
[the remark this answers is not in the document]
hanars: I'm actually worried there are cases where there is `'xpos'` in the json but no `'chrom'` or `'pos'` and then  `setdefault` will error. Since this was jsut a copy-paste I am going to leave it as-is instead of spending time investigating

seqr/views/apis/saved_variant_api.py
```diff
@@ -62,12 +61,12 @@ def create_saved_variant_handler(request):
     saved_variant_guids = []
     for single_variant_json in variants_json:
         try:
-            parsed_variant_json = _get_parsed_variant_args(single_variant_json, family)
+            create_json, update_json = parse_saved_variant_json(single_variant_json, family)
```
jklugherz, line 64: when in the seqr user flow are saved variants created?

seqr/views/apis/saved_variant_api.py
```diff
@@ -237,7 +219,7 @@ def update_variant_functional_data_handler(request, variant_guids):
         get_tag_create_data=lambda tag, **kwargs: {'functional_data_tag': tag.get('name')})
 
 
-def _update_variant_tag_models(request, variant_guids, tag_key, model_cls, get_tag_create_data, response_guid_key=None, delete_variants_if_empty=False):
+def _update_variant_tag_models(request, variant_guids, tag_key, model_cls, get_tag_create_data, response_guid_key=None, delete_variants_if_empty=False, protected_tag_type=None):
```
jklugherz, line 222: why is AIP a protected tag? what does protected mean here?

seqr/views/apis/summary_data_api.py
```diff
@@ -172,6 +183,140 @@ def bulk_update_family_analysed_by(request):
     })
 
 
+def _load_aip_data(data, user):
```
jklugherz, line 186: on what cadence will AIP data be loaded? are analysts the users that do that?

seqr/views/apis/summary_data_api.py
```diff
@@ -172,6 +183,140 @@ def bulk_update_family_analysed_by(request):
...
+    category_map = data['metadata']['categories']
+    results = data['results']
+
+    family_id_map = dict(Individual.objects.filter(
+        family__project__in=get_internal_projects(), individual_id__in=results.keys(),
+    ).values_list('individual_id', 'family_id'))
+    missing_individuals = set(results.keys()) - set(family_id_map.keys())
+    if missing_individuals:
+        raise ErrorsWarningsException([f'Unable to find the following individuals: {", ".join(sorted(missing_individuals))}'])
```
jklugherz, line 195: what's the reasoning for a combined error/warning exception?

seqr/views/apis/saved_variant_api.py
hanars: The first time a user adds a tag or a note to a variant

seqr/views/apis/saved_variant_api.py
hanars: Per the ticket:
> Create a new AIP tag with a hover over for metadata, which include the AIP categories and date. This tag should not be editable by users

The fact that it should not be editable by users, and can only be added by the AIP upload and never removed by a user, is what "protected" means

seqr/views/apis/summary_data_api.py
hanars: No idea! Realistically, its going to be Lynn doing the loading for now, but not sure what that looks like in the future. The ticket specified that this upload should be done using the existing "External Analysis" UI, which is restricted to analysts, so for now thats what we are going with

seqr/views/apis/summary_data_api.py
hanars: This is just a generic error type that we defined which has error handling middleware that properly handles this exception if it is raised and returns a well formatted httpresponse to the user with a 400 status code. Using it is just a way to throw an error that will be a 400 error message displayed to the user without having to reinvent the wheel for writing that logic

seqr/views/apis/saved_variant_api.py
jklugherz: ok just making sure my understanding is correct: 

Since it's a protected tag type, the link between a saved variant and an AIP tag will not be editable by a user. The tag is associated with a saved variant (aka variant tag type model is created) when the AIP data is loaded.

seqr/views/apis/summary_data_api.py
```diff
@@ -172,6 +183,140 @@ def bulk_update_family_analysed_by(request):
...
+        SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL,
+        f'{summary_message}:\n```{", ".join(sorted(family_id_map.keys()))}```',
+    )
+
+    return create_json_response({
+        'info': [summary_message],
+    })
+
+
+def _search_new_saved_variants(family_variant_ids, user):
```
jklugherz, line 247: I'm used to adding type hints for fn parameters and return values in Python for the sake of readability and maintainability. We also had a lot of teams making changes in other team's repo, so it was easier for someone new to understand what's happening in the codebase.

Are type hints something you've considered? 

I've seen type hints with custom types that are very verbose `_search_new_saved_variants(family_variant_ids: List[Tuple[ FamilyGuID, VariantGuID]], user: User): -> Dict[Tuple[ FamilyGuID, VariantGuID], Variant]]`

and ones that use the basic data types
`_search_new_saved_variants(family_variant_ids: list, user) -> dict:`

Historically I tend to write type hints that feel too verbose to me, taking up too much space in the function signature and requiring extra code and tests, so I'm open to hearing your thoughts.

seqr/views/apis/summary_data_api_tests.py
```diff
@@ -413,6 +417,82 @@ def test_bulk_update_family_analysed_by(self, mock_load_uploaded_file):
...
+                'HG00731': {
+                    '12-48367227-TC-T': {'categories': ['3', '4'], 'support_vars': ['2-103343353-GAGA-G']},
+                    '1-248367227-TC-T': {'categories': ['1'], 'support_vars': ['12-48367227-TC-T']},
+                },
+                'SAM_123': {
+                    '12-48367227-TC-T': {'categories': ['4', 'support'], 'support_vars': []},
+                },
+            }
+        }
+        mock_load_uploaded_file.return_value = aip_upload
```
jklugherz, line 441: I noticed that single test functions include multiple test 'cases' (error cases, success cases). Is that a django standard or just how it's done in this repo?

hanars: just this repo

seqr/views/apis/summary_data_api.py
hanars: We have considered them, they are in the category of "it would be nice if we had them but adding them everywhere is hard". But probably a good idea to start adding them to new functions like this one!

hanars: added!

jklugherz (approved):

bpblanken (approved):

PR #3735, opened 2023-11-16
Fix AD inheritance in report

hanars:


bpblanken (approved):
Is this reasonable to cover with a test?

hanars:
> Is this reasonable to cover with a test?

It would be hard because this is a really uncommon family structure and we don't have any fixture data for it, either for families with this structure or fake variants with this inheritance pattern, so it would involve modifying or adding a bunch of fixture data which would probably have a lot of impact on other tests

PR #3739, opened 2023-11-20
V3 airflow dag triggering

hanars:


seqr/views/utils/airflow_utils.py
```diff
@@ -63,16 +102,110 @@ def _check_dag_running_state(dag_id):
...
+    else:
+        dag_variables['project_path'] = f'{dag_path}/{projects[0]}/v{datetime.now().strftime("%Y%m%d")}'
     return dag_variables
 
 
+def _construct_v3_dag_variables(projects: list[str], data_path: str, genome_version: str, is_internal: bool,
+                                sample_type: str, **kwargs):
+    return {
+        'projects_to_run': projects,
+        'callset_path': data_path,
```
bpblanken, line 148: This changed literally two days ago to `callset_paths` taking an array :/

seqr/views/utils/airflow_utils.py
```diff
@@ -1,30 +1,52 @@
...
 
 class DagRunningException(Exception):
     pass
 
 
-def trigger_data_loading(dag_name, projects, data_path, additional_dag_variables, user,
-                         success_message, success_slack_channel, error_message):
+def trigger_data_loading(projects: list[Project], sample_type: str, data_path: str, user: User, success_message: str,
+                         success_slack_channel: str, error_message: str,
+                         dataset_type: str = Sample.DATASET_TYPE_VARIANT_CALLS, genome_version: str = GENOME_VERSION_GRCh38,
```
bpblanken, line 33: This is maybe me being paranoid about lack of type safety, but is there a strong reason for us to default to `SNV_INDEL` here rather than force the caller to supply a dataset_type?

seqr/views/utils/individual_utils.py
```diff
@@ -52,7 +52,7 @@ def add_or_update_individuals_and_families(project, individual_records, user, ge
     families_by_id = {f.family_id: f for f in Family.objects.filter(project=project, family_id__in=family_ids)}
 
     missing_family_ids = family_ids - set(families_by_id.keys())
-    for family_id in missing_family_ids:
+    for family_id in sorted(missing_family_ids):
```
bpblanken, line 55: Is this a flappy test fix?

seqr/views/utils/airflow_utils.py
hanars: no strong reason, I can make this explicit

seqr/views/utils/individual_utils.py
hanars: this PR would have resulted in a flappy test if not for this change. Previously, we sorted the sample ids for export in anvil loading. Now, the way I modified the code to share behavior for exporting the sample ids in elasticsearch or the pedigree file on the hail backend, both exports are sorted the same way which is by family db id and then by sample id. So for a test that adds multiple families, the sort order by just sample id was deterministic but the sort by family and sample id was not if the families were not created in a reliable order

seqr/views/utils/airflow_utils.py
```diff
@@ -63,16 +101,110 @@ def _check_dag_running_state(dag_id):
...
+        dag_variables['project_path'] = f'{dag_path}/{projects[0]}/v{datetime.now().strftime("%Y%m%d")}'
     return dag_variables
 
 
+def _construct_v3_dag_variables(projects: list[str], data_path: str, genome_version: str, is_internal: bool,
+                                sample_type: str, **kwargs):
+    return {
+        'projects_to_run': projects,
+        'callset_paths': [data_path],
+        'sample_source': 'Broad_Internal' if is_internal else 'AnVIL',
```
bpblanken, line 148: I'm supporting `Broad_External` still in airflow.  It looks like there's a couple of exome projects in there that have subset files, but that's it.  Do we expect more?

hanars: We should continue to support `Broad_External` in airflow, but its not currently supported being triggered from seqr, in v2 or v3. Currently we have the UI to support triggering either:
 - AnVIL SNV_INDEL DAGs
 - Broad_Internal SV/GCNV/MITO DAGs

Since we do not even have SV/GCNV/MITO for Broad_External, theres no reason to have it supported from the seqr UI/ backend. When we add support for triggering SNV_INDEL to the "internal" data triggering UI, we can think about if we want to support Broad_External from the UI or no.

Out of curiosity, how does the pieplein run differ between `Broad_Internal` and `Broad_External`?

bpblanken: There's no difference beyond where airflow looks for sample subset, pedigree, and remap files.

bpblanken (approved):

PR #3748, opened 2023-11-27
Fix in silico score information and display

jklugherz:
1. Fixes capitalization/spacing of FATHMM, SIFT, CADD, REVEL, MPC, VEST, PolyPhen, MutTaster, MutPred, PrimateAI
2. Removes citation in the hover over for SpliceAI and MPC
3. Only show CADD, REVEL, VEST, Mutpred2, MPC, SpliceAI scores above the fold in a variant view, other scores are shown below. In order to do this, I split the scores into primary and secondary in silico field groups. 

Screenshot 2023-11-27 at 10 08 48 AM
Screenshot 2023-11-27 at 10 11 03 AM
Screenshot 2023-11-27 at 10 11 20 AM


hanars (changes requested):

ui/shared/utils/constants.js
```diff
@@ -1392,10 +1399,13 @@ export const predictorColorRanges = thresholds => (
         </div>
       )
     })}
-    <small>
-      {/* eslint-disable-next-line react/jsx-one-expression-per-line */}
-      Based on 2022 ClinGen recommendations (PMID:&nbsp;{rangeSourceLink})
-    </small>
+    {/* The thresholds for SpliceAI and MPC do not require a citation */}
+    {!([SPLICE_AI_FIELD, MPC_FIELD].includes(field)) && (
```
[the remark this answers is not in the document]
bpblanken: I'd also suggest putting the configuration with the rest of the `PREDICTOR_FIELDS` options as `requiresCitation` or something like that.

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -17,8 +17,6 @@ const PredictionValue = styled.span`
   text-transform: uppercase;
 `
 
-const NUM_TO_SHOW_ABOVE_THE_FOLD = 6 // how many predictors to show immediately
```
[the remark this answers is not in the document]
jklugherz: I added back in the `NUM_TO_SHOW_ABOVE_THE_FOLD` behavior and implemented the sorting approach instead. I added `displaySortOrder` as a new `PREDICTOR_FIELDS` option and sorted the filtered fields according to that value.

ui/shared/utils/constants.js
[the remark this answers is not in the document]
jklugherz: I agree with this. I added `requiresCitation: false` to `PREDICTOR_FIELDS` MPC and SpliceAI. I defaulted it to `true` in  the `predictorColorRanges` function signature (as opposed to where the fn is called, so the default logic lives in once place)

ui/shared/components/panel/variants/Predictions.jsx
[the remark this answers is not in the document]
hanars: is there a reason to add `displaySortOrder` as a field, instead of just using the order of `PREDICTOR_FIELDS` as the display order?

jklugherz:
I added the sort order as a field option because it is more explicit to a
reader that order of the predictors matters somewhere, and the significance
of ordering isn’t hidden in the implementation of the PREDICTOR_FIELDS as
an array. If another field is added anywhere in the array the same order
remains.

However, this means right now that the displaySortOrder values match the
order of
PREDICTOR_FIELDS so the sorting logic I added is redundant.

I’m flexible and open to removing  displaySortOrder if you think
PREDICTOR_FIELDS won’t need to be changed too much in the future.

ui/shared/components/panel/variants/Predictions.jsx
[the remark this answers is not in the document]
jklugherz: I added the sort order as a field option because it is more explicit to a
reader that order of the predictors matters somewhere, and the significance
of ordering isn’t hidden in the implementation of the PREDICTOR_FIELDS as
an array. If another field is added anywhere in the array the same order
remains.

However, this means right now that the displaySortOrder values match the
order of
PREDICTOR_FIELDS so the sorting logic I added is redundant.

I’m flexible and open to removing  displaySortOrder if you think
PREDICTOR_FIELDS won’t need to be changed too much in the future.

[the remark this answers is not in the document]
hanars: My general assumption is that arrays of things to display are ordered lists, and that for things displayed to users order is generally important. So to me, its implied that the order of `PREDICTOR_FIELDS` is meant to be ordered, but thats clearly not universal. I don't really like the idea of needing to maintain the additional `displaySortOrder` since I don't really see an advantage to allowing the list to have a different order than the display, but maybe renaming the array `ORDERED_PREDICTOR_FIELDS` or `DISPLAY_ORDER_PREDICTOR_FIELDS` or something else that makes it clear that the order is important will make it explicit that the order is being used without adding any additional fields to maintain

ui/shared/components/panel/search/constants.js
```diff
@@ -453,40 +453,42 @@ const REQUIRE_SCORE_FIELD = {
   label: 'Require Filtered Predictor',
   labelHelp: 'Only return variants where at least one filtered predictor is present. By default, variants are returned if a predictor meets the filtered value or is missing entirely',
 }
-export const IN_SILICO_FIELDS = [REQUIRE_SCORE_FIELD, ...PREDICTOR_FIELDS.filter(({ displayOnly }) => !displayOnly).map(
-  ({ field, fieldTitle, thresholds, indicatorMap, group, min, max }) => {
-    const label = fieldTitle || snakecaseToTitlecase(field)
-    const filterField = { name: field, label, group }
+export const IN_SILICO_FIELDS = [
+  REQUIRE_SCORE_FIELD,
```
jklugherz, line 457: Made a change to indentation here due to line length limit

ui/shared/components/panel/variants/Predictions.jsx
```diff
@@ -59,6 +59,26 @@ Prediction.propTypes = {
...
+  const mappedFields = ORDERED_PREDICTOR_FIELDS.map(({
+    fieldTitle,
+    getHref,
+    ...predictorField
+  }) => ({
+    field: predictorField.field,
+    fieldTitle,
+    href: getHref && getHref(variant),
+    requiresCitation: predictorField.requiresCitation,
+    displaySortOrder: predictorField.displaySortOrder,
```
hanars, line 75: I thought we had opted to remove `displaySortOrder`

ui/shared/utils/constants.js
```diff
@@ -1326,16 +1328,17 @@ export const PREDICTOR_FIELDS = [
     getHref: ({ chrom, pos, ref, alt, genomeVersion }) => (
       `https://spliceailookup.broadinstitute.org/#variant=${chrom}-${pos}-${ref}-${alt}&hg=${genomeVersion}&distance=1000&mask=1`
     ),
+    requiresCitation: false,
+    displaySortOrder: 6,
```
hanars, line 1332: We should remove this is we are not going to be using it

ui/shared/utils/constants.js
```diff
@@ -1370,7 +1374,7 @@ export const predictionFieldValue = (
 
   return indicatorMap[value[0]] || indicatorMap[value]
 }
-export const predictorColorRanges = thresholds => (
+export const predictorColorRanges = (thresholds, requiresCitation = true) => (
```
hanars, line 1377: I'm not sure that defaulting this to true is the behavior we want - is there any case where we are calling this function without explicitly passing `requiresCitation`?

ui/shared/components/panel/variants/Predictions.jsx
jklugherz: missed this, good catch

ui/shared/utils/constants.js
jklugherz: We call this function when `requiresCitation` is sometimes `undefined`. I only set the value of `requiresCitation` for those predictors in `ORDERED_PREDICTOR_FIELDS` when the value is`false`, and for the rest of the predictors `requiresCitation` is `undefined`. The fn `predictorColorRanges` is called twice, and in both calls I contemplated defaulting `requiresCitation` to `true` if it is `undefined` before the call, but decided to do it here so defaulting only happens once.

Another option is just to set `requiresCitation: true` in each field in `ORDERED_PREDICTOR_FIELDS` and have no default logic, and make `requiresCitation` a required prop on `Prediction`

hanars: I think I am a little confused why we would want to be showing the citation unless we were sure it was a field where the citation was appropriate. So if the function is called twice, shouldn't we explicitly be setting `requiresCitation` in both of those calls, to make sure the behavior is correct in both places? Also, as discussed on the previous review, we really want the citation to be a whitelist and not a blacklist. So setting it to false in a few places opens us up to the likelihood that as we add new predictions we do not remember to set the citation field to false. Since there is a finite list of scores in the paper that we want to cite, I think it is better practice to explicitly set the citation to true in those cases, and ave the default behavior be no citation

jklugherz: ahh I see how setting `requiresCitation` to `false` only for some predictors is still just blacklisting them. If I did the opposite and just added `requiresCitation: true` to those fields that we want citations for, that would be whitelisting?

hanars (approved):

PR #3758, opened 2023-11-28
simplify control flow for variant inheritance

hanars:
the logic for this function was really convoluted and there were a few uncovered branches. This simplifies the logic which fixes the bug and also hopefully makes it more resilient

### DIFF
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -335,15 +335,18 @@ DOMINANT = 'Autosomal dominant'
 
 def update_variant_inheritance(
         variant_json: dict, family_individual_data: tuple[set[str], set[str], set[str], dict[str: list[str]]],
-        update_potential_comp_het_gene: Callable[str, None]
-):
+        update_potential_comp_het_gene: Callable[str, None]) -> None:
     """Compute the inheritance mode for the given variant and family"""
+
+    affected_individual_guids, unaffected_individual_guids, male_individual_guids, parent_guid_map = family_individual_data
     is_x_linked = 'X' in variant_json.get('chrom', '')
+
     genotype_zygosity = {
         sample_guid: _get_genotype_zygosity(genotype, is_hemi_variant=is_x_linked and sample_guid in male_individual_guids)
         for sample_guid, genotype in variant_json.get('genotypes', {}).items()
     }
-    inheritance_model, possible_comp_het = _get_inheritance_model(genotype_zygosity, family_individual_data, is_x_linked)
+    inheritance_model, possible_comp_het = _get_inheritance_model(
+        genotype_zygosity, affected_individual_guids, unaffected_individual_guids, parent_guid_map, is_x_linked)
 
     if possible_comp_het:
         for gene_id in variant_json.get('transcripts', {}).keys():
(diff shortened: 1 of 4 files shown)


seqr/views/utils/variant_utils.py
```diff
@@ -329,56 +329,36 @@ def get_saved_discovery_variants_by_family(variant_filter, format_variants, get_
...
+    unaffected_zygosities = {genotype_zygosity[g] for g in unaffected_individual_guids if g in genotype_zygosity}
 
-    return inheritance_models, potential_compound_het_gene_ids, genotype_zygosity
+    inheritance_model = None
+    potential_compound_het_gene_ids = set()
+    if any(zygosity in unaffected_zygosities for zygosity in {HOM_ALT, HEMI}):
+        # No valid inheritance modes for hom alt unaffected individuals
+        inheritance_model = None
+    elif any(zygosity in affected_zygosities for zygosity in {HOM_ALT, HEMI}):
+        inheritance_model = 'X-linked' if is_x_linked else 'AR-homozygote'
```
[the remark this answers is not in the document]
hanars: fair point, see this lovely exchange on the PR where the bug was introduced... 
https://github.com/broadinstitute/seqr/pull/3735#issuecomment-1819813663

seqr/views/utils/variant_utils.py
```diff
@@ -329,56 +329,36 @@ def get_saved_discovery_variants_by_family(variant_filter, format_variants, get_
...
+                guid for guid in affected_individual_guids
+                if genotype_zygosity.get(guid) == HET and
+                any(genotype_zygosity.get(parent_guid) == HET for parent_guid in parent_guid_map[guid])
+            )
+            inheritance_model = 'AD' if inherited else 'de novo'
+
+        if (len(unaffected_individual_guids) < 2 or HET in unaffected_zygosities) and 'transcripts' in variant_json:
+            potential_compound_het_gene_ids.update(list(variant_json['transcripts'].keys()))
+
+    return inheritance_model, potential_compound_het_gene_ids, genotype_zygosity
```
jklugherz, line 361: Have you considered splitting this function that returns 3 values into separate functions that return 1, or returning a namedtuple or dictionary instead here? Then we'd never get the vague "ValueError: not enough values to unpack (expected 3, got 2)" again.

seqr/views/utils/variant_utils.py
```diff
@@ -329,56 +329,36 @@ def get_saved_discovery_variants_by_family(variant_filter, format_variants, get_
 
 def get_variant_inheritance_models(variant_json, family_individual_data):
```
jklugherz, line 330: Can you add a comment with a summary of what the function is doing, and maybe type hints for args and return value?

seqr/views/utils/variant_utils.py
hanars: I hear that, and I see how that could make things more resilient in some ways, but I would rather leave this function signature as is. I prefer the explicit return of the different things we expect instead of a dictionary with magic string keys, or the maintenance overhead of a named tuple thats making it seem like theres a real relationship between these parsed values when in fact they are just 3 different things that happen to be computed at the same time. This issue arose not because the calling functions were not expecting the right number of values, but because there was a branch in this fucntion that returned early on an edge case. To me, the solution is to have a single return statement per function, so in the future any changes to the signature only need one place to change instead of remembering to track down multiple possible return branches

hanars: Actually, looking more into the usage, I think a dictionary would work better bcause I can clean up some of the logic that way, as these fields are used to update a dictionary in the end

bpblanken:
Tests look great!  Ping on [inheritance model](https://github.com/broadinstitute/seqr/pull/3758#discussion_r1409445205) enum if you're up for it.

hanars:
> Ping on [inheritance model](https://github.com/broadinstitute/seqr/pull/3758#discussion_r1409445205) enum if you're up for it.

I went with constants instead of an Enum

jklugherz (approved):

PR #3762, opened 2023-11-30
Correct length of cram block is returned by dd.

barslmn:
Alright, I think I got it. It is because this [length here](https://github.com/broadinstitute/seqr/blob/6bd1328dad35c2182afb64b9756bd0a46cd13461/seqr/utils/file_utils.py#L57C28-L57C28) does not match [this length here](https://github.com/broadinstitute/seqr/blob/6bd1328dad35c2182afb64b9756bd0a46cd13461/seqr/views/apis/igv_api.py#L243).

Off by one error.

Discussion here: https://github.com/broadinstitute/seqr/discussions/3755

I also kept the status="none" in dd command.

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -52,9 +52,9 @@ def file_iter(file_path, byte_range=None, raw_content=False, user=None):
         for line in _google_bucket_file_iter(file_path, byte_range=byte_range, raw_content=raw_content, user=user):
             yield line
     elif byte_range:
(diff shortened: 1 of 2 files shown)


barslmn:
IGV can open bam files with this change. I suggest one more independent test would be apt before merging.

hanars:
>I suggest one or more independent tests would be apt before merging.

Yeah, it looks like "test_proxy_local_to_igv" in igv_api_tests.py is failing, you will need to update that in order to accept this PR

hanars:
thanks so much for fixing this!

PR #3767, opened 2023-12-04
validate vcf meta reference version

hanars:
If a user requests data loading through AnVIL, this adds an additional validation that if the provided VCF specifies the genome build in its VCF meta we confirm that the user-provided genome build matches. Not all VCFs have this metadata, and we do not throw an error if it is mssing

### DIFF
--- a/seqr/utils/vcf_utils.py
+++ b/seqr/utils/vcf_utils.py
@@ -26,7 +26,7 @@ def _validate_vcf_header(header):
         raise ErrorsWarningsException([f'Missing required VCF header field(s) {missing_fields}.'], [])
 
 
-def _validate_vcf_meta(meta):
+def _validate_vcf_meta(meta, genome_version):
     errors = []
     for field, sub_field_meta in EXPECTED_META_FIELDS.items():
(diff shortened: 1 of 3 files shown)


seqr/utils/vcf_utils.py
```diff
@@ -72,7 +83,7 @@ def validate_vcf_and_get_samples(vcf_filename):
     _validate_vcf_header(header)
     if not samples:
         raise ErrorsWarningsException(['No samples found in the provided VCF.'], [])
-    _validate_vcf_meta(meta)
+    _validate_vcf_meta(meta, genome_version)
```
jklugherz, line 86: is `meta` ever an empty dict by this point? we could check to see if it's empty and skip calling this function when it is to avoid executing the first 10 lines of `_validate_vcf_meta()`

jklugherz (approved):

seqr/utils/vcf_utils.py
hanars: no this will always be populated, the way VCFs work is if there is a FORMAT field there are also FORMAT meta fields, and we already validate above that the field exists overall

bpblanken (approved):

PR #3769, opened 2023-12-04
Sample metdata field additions

hanars:
Adds some fields to the sample metadata export to support the RNA and lirical pipelines being able to run without access to a full copy of the seqr database. Since these fields are generally useful if you are computationally processing the data but not useful to a human reader looking at the data, I chose to have them included in the export but not shown in the current table view


ui/pages/SummaryData/components/SampleMetadata.test.js
```diff
@@ -0,0 +1,90 @@
...
+    displayName: '12',
+    MME: 'Y',
+    subject_id: 'NA20889',
+    individual_guid: 'I000017_na20889',
+    proband_relationship: 'Self',
+    consanguinity: 'None suspected',
+  },
+]
+
+test('SampleMetadataDataTable render and export', () => {
```
hanars, line 68: we do not generally have many or robust client side unit tests so this is atypical for the repo, but given that this file will be a dependency for the computational biologists who support seqr we want to ensure this file is stable (or at least does not change unintentionally) it makes sense to add testing for it

jklugherz: I was curious about the enzyme adapter and found some fun developer drama wrt enzyme support for react.. here's a blog post from the adapter maintainer: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl. Is this on your radar? Should we avoid writing enzyme tests in the future and try jest or React Testing Library instead?

ui/shared/components/table/DataTable.jsx
```diff
@@ -41,8 +41,15 @@ const StyledDataTable = styled(Table)`
 const ASCENDING = 'ascending'
 const DESCENDING = 'descending'
 
-const getRowColumnContent = (row, isExport) => (col, formatProps) => (
-  (col.format && !(isExport && col.noFormatExport)) ? col.format(row, isExport, formatProps) : row[col.name])
+const getRowColumnContent = (row, col, isExport, formatProps) => {
```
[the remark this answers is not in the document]
jklugherz: do you remember why it was bound initially?

ui/pages/SummaryData/components/SampleMetadata.test.js
hanars: So thats a really good point, but I think realistically we do not have the bandwidth to invest any substantive dev time in  UI testing given our current workload. So our options in the meantime are basically "continue to write new enzyme tests and then when we do migrate have more tests to migrate" or "do not write any new UI tests". Given those options, I think I prefer the former. We write UI tests so infrequently that the time we will save by having a couple tests already moved over is pretty minimal, and it would be better to just switch over all our tests and deprecate enzyme in one big push.

ui/shared/components/table/DataTable.jsx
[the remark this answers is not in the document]
hanars: I think because somebody though `columns.map(getRowColumnContent (row, true))` looked neater than `columns.map(col => getRowColumnContent (row, col, true))`

hanars:
> what are the RNA and lirical pipelines?

These are pipelines that computational biologists run outside of seqr, the results of which are then uploaded back to seqr to support analysis. RNA involves looking into RNA data that was collected for some of our samples, which is just a totally different biological thing than exome/genome. Lirical is like an algorithm to prioritize genes and/or variants within a given family based on patient phenotype

You can see examples of this on this family, on the right side of the page for the individual you can select "RNAseq Results"
and "Show Phenotype Prioritized Genes"
https://seqr.broadinstitute.org/project/R0384_rare_genomes_project_gen/family_page/F027968_rgp_1017

ui/pages/SummaryData/components/SampleMetadata.test.js
jklugherz: I agree with your conclusion. I like how you list out the concrete options for what we can do Right Now in addition to thinking about something for the future.

jklugherz (approved):

bpblanken (approved):

PR #3771, opened 2023-12-06
Strip whitespace from anvil upload

jklugherz:


jklugherz:
> but they result in a fragile test because they only test the behavior on isolated and synthetic data, so it does not necessarily cover the way the function is used well. Also, if the way the helper is called changes the tests need to be updated even if no behavior changed which makes it hard to see if a change has consequences or not. If there are changes that can not be tested through existing tests that test the calling functions, adding new utility-specifc tests is needed, but in this case it would be better to actually test the behavior using the existing tests that call this function downstream.

This is a great point and something I hadn't considered before. Testing utility functions by testing the business logic from where they're called also keeps the test suite smaller and faster. I'll update the existing tests instead.

hanars (approved):

PR #3772, opened 2023-12-06
Fiel upload in progress UI improvement

hanars:
A user got a confusing error message caused by an edge case where he submitted a form while a required file was in the process of re-uploading. This updates behavior to prevent form submission while a fiel is uploading, and also ensures the file upload progress bar never shows a complete upload before teh fiel actually finishes, which currently happens frequently

### DIFF
--- a/ui/shared/components/form/XHRUploaderWithEvents.jsx
+++ b/ui/shared/components/form/XHRUploaderWithEvents.jsx
@@ -11,6 +11,7 @@ import XHRUploader from 'react-xhr-uploader'
 
 const NO_DISPLAY_STYLE = { display: 'none' }
 const POINTER_CURSOR_STYLE = { cursor: 'pointer' }
+const MAX_UNCOMPLETED_PROGRESS = 80
 
 const onClickInput = (event) => {
   // allows the same file to be selected more than once (see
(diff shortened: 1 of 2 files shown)


ui/shared/components/form/XHRUploaderWithEvents.jsx
```diff
@@ -81,9 +81,8 @@ class XHRUploaderWithEvents extends XHRUploader {
         }
       }
       xhr.upload.onprogress = (e) => {
-        if (e.lengthComputable) {
-          progressCallback(e.loaded / e.total * 100) // eslint-disable-line no-mixed-operators
-        }
+        const progress = e.lengthComputable ? (e.loaded / e.total * 100) : 50 // eslint-disable-line no-mixed-operators
+        progressCallback(progress === 100 ? 80 : progress)
```
jklugherz, line 85: why do you send 80 back to `progressCallback` if `progress` is 100? 

how about making 80 an all-caps constant since it's a specific value that has some meaning?

hanars: because it is incorrectly reporting progress as 100% in many cases, and we have an explicit `progressCallback(100)` on the actual upload finish event, so having it hover at 80% until that gets triggered means the user never is shown that its finished when it isn't. I chose 80 because thats what looks like a lot has loaded but is still very obvious that its not done, much higher than that and it looks done at. glance

jklugherz (approved):

bpblanken (approved):

PR #3773, opened 2023-12-07
fix unit test uing real time

hanars:
We were using real world time in one of our unit tests instead of the mocked time, and when we actually hit the mock date of 12/5/2023 IRL it caused a test to break 🤦


bpblanken (approved):

jklugherz (approved):
Thank you!!!

PR #3776, opened 2023-12-08
Anvil tracking update

hanars:
Updates the anvil loading tracker when data is added to a project and when a project is deleted

### DIFF
--- a/seqr/views/utils/airtable_utils.py
+++ b/seqr/views/utils/airtable_utils.py
@@ -47,7 +47,7 @@ class AirtableSession(object):
         except Exception as e:
             logger.error(f'Airtable create "{record_type}" error: {e}', self._user)
 
-    def safe_patch_record(self, record_type, record_or_filters, record_and_filters, update, max_records=PAGE_SIZE-1):
+    def safe_patch_records(self, record_type, record_or_filters, record_and_filters, update, max_records=PAGE_SIZE - 1):
         try:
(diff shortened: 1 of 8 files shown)


jklugherz (approved):
looks good to me. is it possible to test this manually in dev?

hanars:
> looks good to me. is it possible to test this manually in dev?

yeah it would be!

seqr/views/utils/airtable_utils.py
```diff
@@ -44,15 +47,47 @@ def safe_create_record(self, record_type, record):
...
+            record_type, fields=record_or_filters.keys(), or_filters=record_or_filters, and_filters=record_and_filters,
+            page_size=max_records+1,
+        )
+        if not records or len(records) > max_records:
+            raise ValueError('Unable to identify record to update')
+
+        self._session.params = {}
+        errors = []
+        for record_id in records.keys():
+            response = self._session.patch(f'{self._url}/{record_type}/{record_id}', json={'fields': update})
```
bpblanken, line 69: Definitely a nit, but is there a reason not to put the `patch()` call itself inside the `try/except`?  I think the `patch()` on its own can throw connection errors/timeouts etc. we'd want to continue the loop right?

bpblanken (approved):
small question but don't want to hold this up!

bpblanken (approved):

PR #3784, opened 2023-12-13
fix anvil dag construction path

hanars:


seqr/views/utils/airflow_utils.py
```diff
@@ -60,15 +60,15 @@ def trigger_data_loading(projects: list[Project], sample_type: str, dataset_type
 
 def write_data_loading_pedigree(project: Project, user: User):
     match = next((
-        (callset, sample_type) for callset, sample_type in itertools.product(['Internal', 'External', 'AnVIL'], ['WGS', 'WES'])
```
bpblanken, line 63: is this `AnVIL -> False` actually needed if we keep the `is_internal=callset != 'AnVIL'` on line 65 and also add to line 71?

bpblanken: yeah... I'd vote for leaving as is.  I think conceptually having `callset` come from a `list[str]` makes more sense than allowing `callset` to be a boolean.

bpblanken (approved):

PR #3785, opened 2023-12-13
Family metadata

hanars:
Adds the new Family Metadata report and removes the old discovery sheet from the UI. Abstracts the code for the Sample Metadata report so it can be more easily shared by this report, and by the forthcoming variant metadata report
Backend code cleanup for the discovery sheet to follow in a separate PR.


seqr/views/apis/report_api.py
```diff
@@ -984,6 +985,69 @@ def _get_row_id(row):
...
+        if proband:
+            known_ids = {
+                'proband_id': proband['subject_id'],
+                'paternal_id': proband['paternal_id'],
+                'maternal_id': proband['maternal_id'],
+            }
+            f.update(known_ids)
+            individuals_ids -= set(known_ids.values())
+            if proband['paternal_id'] and proband['maternal_id'] and len(individuals_ids) < 2:
+                family_structure = 'quad' if individuals_ids else 'trio'
```
bpblanken, line 1032: can `quad`, `trio`, `duo` be constants?

bpblanken (approved):

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -96,22 +112,23 @@ def parse_anvil_metadata(projects, max_loaded_date, user, add_row, omit_airtable
         })
         family_data_by_id[family_id] = f
 
-    samples_by_family_id = defaultdict(list)
+    individuals_by_family_id = defaultdict(list)
```
jklugherz, line 115: can you remind me of the difference btw samples and individuals, and why you changed this data structure to the latter?

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -138,56 +155,69 @@ def parse_anvil_metadata(projects, max_loaded_date, user, add_row, omit_airtable
...
-            discovery_row = _get_discovery_rows(sample, parsed_variants)
-            add_row(discovery_row, family_id, DISCOVERY_ROW_TYPE)
+            if not no_variant_zygosity:
+                discovery_row = _get_discovery_rows(individual, sample, parsed_variants)
+                add_row(discovery_row, family_id, DISCOVERY_ROW_TYPE)
 
 
-def parse_family_sample_affected_data(family_samples):
-    indiv_id_map = {s.individual.id: s.individual.guid for s in family_samples}
+def parse_family_sample_affected_data(family_individuals):
```
jklugherz, line 211: nitpick, but
```suggestion
def parse_family_individual_affected_data(family_individuals):
```

ui/pages/SummaryData/components/SampleMetadata.jsx
```diff
@@ -124,68 +96,19 @@ const getColumns = (data) => {
...
-    childComponent={SampleMetadata}
-    projectGuid={match.params.projectGuid}
-    user={user}
-  />
-)
-
-LoadedSampleMetadata.propTypes = {
-  match: PropTypes.object,
-  user: PropTypes.object,
+const mapStateToProps = (state, ownProps) => {
```
jklugherz, line 102: this feels like it could be a silly question, but why can't you just export a component that returns `LoadReportTable`?

seqr/views/utils/anvil_metadata_utils.py
hanars: Individuals represent a person. Samples represent a biological sample that was processed for genetic information. There is a one-to-many relationship between Individuals and Samples, and not all Individuals have any Samples. The change is because this helper was previously used for reports that only included Individuals with loaded data, so we were fetching the sample models and then using the associated Individual when needed. But this report needs to include Individuals that do not have  and loaded data (i.e no associated Samples) so that approach results in data being omitted. We actually were already storing `individual_samples` as a mapping between individuals and their sample (for the purposes of this report we only use the latest Sample for each Individual), so changing this structure to keep track of all Individuals instead of Samples means no one is left out, but we can still easily get the sample in cases where it is needed

ui/pages/SummaryData/components/SampleMetadata.jsx
hanars: I'm not sure I understand the question. At the bottom of the file we export a component that binds these props to `LoadReportTable`, resulting in a Report table with all the desired props being exported
```
export default connect(mapStateToProps)(LoadReportTable)
```

seqr/views/utils/anvil_metadata_utils.py
jklugherz: this clears things up, thank you!

ui/pages/SummaryData/components/SampleMetadata.jsx
jklugherz: I guess I meant why use `mapStateToProps` instead of exporting the component like it's done [here](https://github.com/broadinstitute/seqr/pull/3785/files/aa0f3bc81387b0dcd663854d92052e18481eecd0#diff-94ee2491f4eb1a73f4d1de4a532827b7c7b785059d3023f2cb1c6ab9588c935bR33), but I just realized that the global state is accessed to get the user so that is probably why

hanars: yeah, thats exactly it, thsi componenet needs the `user` while that other example does not

jklugherz (approved):

PR #3787, opened 2023-12-13
Discovery sheet cleanup

hanars:
Removes unused discovery sheet code. There are a few utility functions that were used exactly twice, once in the discovery sheet and once elsewhere, so I moved those to be private helper methods for the one remaining case where ttheyre used, instead of leaving them as generic helpers


jklugherz (approved):
I love to delete stuff!!!

bpblanken (approved):

PR #3790, opened 2023-12-15
remove GC from MITO variant search and hover over

jklugherz:
Variant hover over before:
Screenshot 2023-12-15 at 1 52 25 PM

After:
Screenshot 2023-12-15 at 4 25 57 PM

Variant search before this change used the GQ filter and excluded MITO variants. After the change, the GQ filter no longer applies to MITO variants even if it's included in the request. Here's a screenshot from debugging, see that `field` is `None` even though `min_qc` is included in the filters:
Screenshot 2023-12-15 at 4 31 56 PM

### DIFF
--- a/hail_search/test_utils.py
+++ b/hail_search/test_utils.py
@@ -755,7 +755,7 @@ MITO_VARIANT1 = {
     'familyGuids': ['F000002_2'],
     'genotypes': {'I000006_hg00733': {
         'sampleId': 'HG00733', 'sampleType': 'WGS', 'individualGuid': 'I000006_hg00733', 'familyGuid': 'F000002_2',
-        'numAlt': 1, 'dp': 3955, 'gq': 53.0, 'hl': 0.083, 'mitoCn': 230, 'contamination': 0.0,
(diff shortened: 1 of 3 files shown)


hail_search/test_utils.py
```diff
@@ -755,7 +755,7 @@
     'familyGuids': ['F000002_2'],
     'genotypes': {'I000006_hg00733': {
         'sampleId': 'HG00733', 'sampleType': 'WGS', 'individualGuid': 'I000006_hg00733', 'familyGuid': 'F000002_2',
-        'numAlt': 1, 'dp': 3955, 'gq': 53.0, 'hl': 0.083, 'mitoCn': 230, 'contamination': 0.0,
+        'numAlt': 1, 'dp': 3955, 'hl': 0.083, 'mitoCn': 230, 'contamination': 0.0,
```
jklugherz, line 758: This variant `MITO_VARIANT1` should be returned in the MITO test case because it passes the `min_hl` filter.

hanars:
PRs generally need to be against dev and not master, but otherwise looks awesome!

hanars (approved):

PR #3791, opened 2023-12-18
Omim regions

hanars:
Include OMIM conditions in seqr that are associated with genomic regions but no specific gene(s)


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -18,52 +19,76 @@ import { ButtonLink, HelpIcon } from '../../StyledComponents'
...
-  getOverlappedSpliceOutliers,
+  SPLICE_OUTLIER_OVERLAP_ARGS,
 } from './VariantUtils'
 import {
   GENOME_VERSION_37, GENOME_VERSION_38, getVariantMainTranscript, SVTYPE_LOOKUP, SVTYPE_DETAILS, SCREEN_LABELS,
 } from '../../../utils/constants'
 
-const BaseSpliceOutlierLabel = React.memo(({ variant, spliceOutliersByFamily }) => {
-  if (!spliceOutliersByFamily || spliceOutliersByFamily.length < 1) {
+const OverlappedIntervalLabels = React.memo(({ groupedIntervals, variant, getOverlapArgs, getLabels }) => {
```
hanars, line 36: Since we are now introducing the third label shown in this component based on whether a variant overlaps an interval, I created a generic helper component to use for the existing 2 labels as well as the new one, instead of just copy-pasting and modifying to create a new componenet for the new label

ui/shared/components/panel/variants/VariantGene.jsx
```diff
@@ -189,7 +189,7 @@ const BaseLocusListLabels = React.memo(({
       paLocusList,
       geneSymbol,
     }) : {
-      description: label,
+      description: locusListDescription,
```
hanars, line 192: this was actually an existing bug where we were not showing the correct description for gene lists in either the geen or variant labels

seqr/fixtures/1kg_project.json
```diff
@@ -2059,7 +2059,7 @@
             "liftedOverChrom": null,
             "svType": "DEL",
             "variantId": "prefix_19107_DEL",
-            "chrom": "12",
+            "chrom": "1",
```
hanars, line 2062: this saved variant fixture actually had a discrepancy between what the xpos represented for its position and what this chrom/pos in the saved_variant_json represented, so this fixes it so they are the same

bpblanken (approved):
I'

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # _seqr_ Changes
 
 ## dev
+* Support OMIM entries with no associated gene and remove phenotypic_series_number (REQUIRES DB MIGRATION)
```
jklugherz, line 4: ah should I update this for every PR I create against dev?

seqr/views/utils/variant_utils.py
```diff
@@ -131,6 +132,16 @@ def _saved_variant_genes_transcripts(variants):
...
 
 
+def get_omim_intervals_query(variants):
+    chroms = {v['chrom'] for v in variants if v.get('svType')}
+    return Q(phenotype_mim_number__isnull=False, gene__isnull=True, chrom__in=chroms)
+
+
+def _get_omim_intervals(variants):
+    omims = Omim.objects.filter(get_omim_intervals_query(variants))
+    return {o['phenotypeMimNumber']: o for o in get_json_for_queryset(omims)}
```
jklugherz, line 142: I just went to the `Omim` model definition and found
```
    class Meta:
        # ('mim_number', 'phenotype_mim_number') is not unique - for example ('124020', '609535')
        unique_together = ('mim_number', 'phenotype_mim_number', 'phenotype_description')
``` 
Do you think we'll get multiple models with the same `phenotypeMimNumber`, and would that mean not returning all possible omims?

CHANGELOG.md
hanars: no, generally we only update this with breaking changes for local seqr installations

seqr/views/utils/variant_utils.py
hanars: oh yeah, good catch. I will look into this further.

hanars: Okay so this is an issue, but its an issue already in how we handle existing omim conditions associated with genes for reporting and discovery. I would like to fix it but I think its out of scope for this PR, so my proposed solution is that this PR goes forward with the current logic here for collapsing multiple phenotype_mim_number into one for conditions with and without genes, and then I will put up a follow up PR later today that handles this correctly everywhere in seqr

jklugherz: sounds great!

jklugherz (approved):

PR #3792, opened 2023-12-18
Dup comp het hail backend

hanars:


hail_search/queries/base.py
```diff
@@ -739,22 +739,32 @@ def _filter_compound_hets(self):
...
-        ch_ht = self._filter_grouped_compound_hets(ch_ht)
 
-        # Format pairs as lists and de-duplicate
-        ch_ht = ch_ht._key_by_assert_sorted(key_pair=hl.sorted([
-            ch_ht[v][self.KEY_FIELD[0]] if len(self.KEY_FIELD) == 1 else hl.tuple([ch_ht[v][k] for k in self.KEY_FIELD])
-            for v in ['v1', 'v2']
-        ]))
-        ch_ht = ch_ht.distinct().key_by()
+        # Compute all distinct variant pairs
+        key_expr = lambda v: v[self.KEY_FIELD[0]] if len(self.KEY_FIELD) == 1 else hl.tuple([v[k] for k in self.KEY_FIELD])
```
bpblanken, line 744: Is it possible for us to just pull the key off the table?
```
In [7]: tuple(ch_ht.key.keys())
Out[7]: ('locus', 'alleles')
```

jklugherz (approved):

hail_search/queries/base.py
hanars: maybe, but we have this constant defined anyways because we neeed it for some of the variant key filtering:  
https://github.com/broadinstitute/seqr/blob/master/hail_search/queries/base.py#L540

It seems like better practice to consitently use this constant if we have it defined, rather than use it in some places and not in others

bpblanken (approved):

PR #3795, opened 2023-12-20
Fix gnomad v2 links

tommyli:
(cherry picked from commit 38b4bbdb98883fc6909ba82d09db71b264c9084e)

Gnomad website have defaulted to V4 so the Gnomad V2 links are broken unless `dataset=gnomad_r2_1` query param is provided in the URL.


hanars (approved):
Thanks for fixing!

PR #3818, opened 2024-01-11
Project overview performance

hanars:
Addresses a longstanding issue with slow loading for the RGP project page(~9 seconds). The major change is to stop returning all Samples as part of the project overview.

RGP currently has ~3500 active samples and ~50,000 inactive samples. On the project page, we use samples for the following:
- **All samples** to show the "Datasets" summary section, which shows all the samples/dataset types broken down by the number of samples loaded on each date
- Active samples to show the "Families/Individuals With Data" summary section (was using all samples but that was a bug, it should always have been active only)
- **All samples** for data export
- The first samples loaded for each family, to show on the family table an to sort it
- Active samples to display/sort/filter the family table

Returning the full samples data for ~50,000 samples was giving both the server side endpoint and client side rendering a performance hit, and for  no real advantage as for most project page renders the details of inactive samples are not important.  This changes the logic to only return active samples and the first loaded sample per family which covers most of the use cases for samples on the project page. For the 2 use cases that currently require all the samples data, this makes the following updates:
- Returns project-level summary statistics to use for the Datasets section so that the full summary can be shown without access to all the inactive samples
- Fetches all the samples when the user goes to export data. We were already fetching the individual data at that point and since the requests run in parallel this does not change the performance for download, and means that in the >99% of cases where a user is on the project page and not downloading anything, we do not need to have all the data ready

The final performance enhancement is an improvement to how we query the number of variant notes in a project- it turns out the query django was constructing was wildly inefficient and took over a second

### DIFF
--- a/seqr/views/apis/project_api.py
+++ b/seqr/views/apis/project_api.py
@@ -11,7 +11,7 @@ from django.db.models.functions import JSONObject
 from django.utils import timezone
 
 from matchmaker.models import MatchmakerSubmission
-from seqr.models import Project, Family, Individual, Sample, IgvSample, VariantTag, SavedVariant, \
+from seqr.models import Project, Family, Individual, Sample, IgvSample, VariantTag, VariantNote, \
     FamilyNote, CAN_EDIT
 from seqr.views.utils.airtable_utils import AirtableSession, ANVIL_REQUEST_TRACKING_TABLE
 from seqr.views.utils.individual_utils import delete_individuals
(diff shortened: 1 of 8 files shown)


seqr/views/apis/project_api.py
```diff
@@ -314,15 +341,13 @@ def _add_tag_type_counts(project, project_variant_tags):
         'description': '',
         'color': 'grey',
         'order': 100,
-        'numTags': SavedVariant.objects.filter(family__project=project, variantnote__isnull=False).distinct().count(),
+        'numTags': project_notes.aggregate(count=Count('saved_variants__guid', distinct=True))['count'],
```
hanars, line 344: this is getting at the same data (number of distinct saved variant rows associated with at least one variant note), but the way it was constructed before was running a full subquery on the saved variant table and actually getting all the records and then counting those, whereas this just runs a `COUNT` query on the database and does not get any records, which is much faster

bpblanken (approved):

PR #3822, opened 2024-01-16
fix GTEx displays

jklugherz:
Fixes expression chart on the gene page and expression data boxplot for specific genes on variants.

The gtex repos are private, but I was given access by a gtex team member (Jared Nedzel)
https://github.com/broadinstitute/gtex-server/tree/master
https://github.com/broadinstitute/gtex-viz
(here's a link to the slack thread I started if you also want to request access: https://broadinstitute.slack.com/archives/C1C22V6FN/p1705085347353619)

With regards the change to 'httpRequestHelper.ts' to include credentials on the fetch request, I did this because I was getting these CORS errors when I switched to v2:
image

I looked into the gtex source code myself, and it looks like when the gtex server changed from v1 to v2, they might have also switched from flask to fastapi, and their CORS settings changed. This may explain why we were able to send credentials when using the v1 api.

flask cors setting:
image

fastapi cors setting:
image

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -83,7 +83,7 @@
     "final-form": "^4.20.6",
     "final-form-arrays": "^3.0.2",
     "final-form-calculate": "^1.3.2",
-    "gtex-d3": "github:broadinstitute/gtex-viz#8d65862",
+    "gtex-d3": "github:broadinstitute/gtex-viz-beta#13510d6",
     "igv": "^2.10.5",
     "jquery": "^3.6.0",
     "js-cookie": "^2.2.1",
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -23,7 +23,7 @@
         "final-form": "^4.20.6",
         "final-form-arrays": "^3.0.2",
         "final-form-calculate": "^1.3.2",
-        "gtex-d3": "github:broadinstitute/gtex-viz#8d65862",
+        "gtex-d3": "github:broadinstitute/gtex-viz-beta#13510d6",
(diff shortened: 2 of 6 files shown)


ui/package.json
```diff
@@ -83,7 +83,7 @@
     "final-form": "^4.20.6",
     "final-form-arrays": "^3.0.2",
     "final-form-calculate": "^1.3.2",
-    "gtex-d3": "github:broadinstitute/gtex-viz#8d65862",
+    "gtex-d3": "github:broadinstitute/gtex-viz-beta#13510d6",
```
jklugherz, line 86: since `gtex-viz-beta` is private, need to figure out if we can/how to access these in github workflows and dev and prod

jklugherz: unfortunately the components in the public repo do not work with the new api, and the components that do work with v2 are in this private repo for some reason. I asked Jared about using them in production and here's the response I got: 

> Yes, it really isn’t beta. It is what we use in production for the GTEx Portal. So just a bad name for the repo. We only deploy from the master branch so we recommend that you do not deploy from a branch below master.

hanars: > need to figure out if we can/how to access these in github workflows and dev and prod

This is a really substantive issue and I cannot review the code until we confirm that this problem has a solution. I also do not believe that we can just add code from a private repo and add it to a public repo and have that just be okay with the gtex team, this feels like a huge problem. I highly doubt they gave you access because they were cool with us releasing all of their code under the broadest possible GPL liscense

PR #3824, opened 2024-01-16
Restrict no location hail backend search

hanars:
Do not allow unrestricted searches across large SNV/Indel projects

### DIFF
--- a/seqr/utils/search/hail_search_utils.py
+++ b/seqr/utils/search/hail_search_utils.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from django.db.models import F, Min
+from django.db.models import F, Min, Count
 
 import requests
 from reference_data.models import Omim, GeneConstraint, GENOME_VERSION_LOOKUP, GENOME_VERSION_GRCh38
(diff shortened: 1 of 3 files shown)


seqr/utils/search/hail_search_utils.py
```diff
@@ -186,3 +186,18 @@ def _validate_expected_families(results, expected_families):
...
+
+MAX_FAMILY_COUNTS = {Sample.SAMPLE_TYPE_WES: 200, Sample.SAMPLE_TYPE_WGS: 35}
+
+
+def validate_hail_backend_no_location_search(samples):
+    sample_counts = samples.filter(dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS).values('sample_type').annotate(
+        family_count=Count('individual__family_id', distinct=True),
+        project_count=Count('individual__family__project_id', distinct=True),
+    )
+    from seqr.utils.search.utils import InvalidSearchException
```
bpblanken, line 199: Should this import be at the top of the file now that we use the exception twice?

hanars: no I tried and we end up in circular dependency hell, and I thought refactoring where it lived was out of scope of this PR

bpblanken: Ah.  Alright.  We should make a ticket though!

seqr/utils/search/utils.py
```diff
@@ -404,7 +410,8 @@ def _parse_inheritance(search, samples, previous_search_results):
                     f'Unable to search for comp-het pairs with dataset type "{invalid_type}". This may be because inheritance based search is disabled in families with no loaded affected individuals'
                 )
 
-    return samples
+    if not has_location_filter:
```
jklugherz, line 413: is the technical reason for requiring a location filter for multiple project/family searches because the search takes too long without it?

bpblanken (approved):

PR #3835, opened 2024-01-24
Enable build 37 for hail backend

hanars:
Actually enables build 37 search for SNV_INDEL data. Originally I had envisioned each class having a list of supported genome versions, but I realized that since each genome build for any given data type is going to have different reference data available, its probably easier to explicitly make classes whenever we add build support to a data type, instead of adding a lot of rarely-used conditional logic


bpblanken (approved):
this got a brief once over... don't see anything unexpected.

jklugherz (approved):

PR #3836, opened 2024-01-25
explicitly set spark driver memory

hanars:


hail_search/web_app.py
```diff
@@ -8,6 +8,9 @@
 
 logger = logging.getLogger(__name__)
 
+MACHINE_MEM = 128
```
bpblanken, line 11: should we use something like `psutil.virtual_memory().total` so that this is dynamic in case we have to also switch out the instance size?

hanars: I'd love to not have to introduce runtime python requirements, since we currently only use whats packaged with the default docker image. We can make this an env variable so its more easily configurable

bpblanken: alright.  this is ok by me

PR #3837, opened 2024-01-25
Use variant lookup page for all-project single variant search

hanars:
Currently, when a user searches for a single variant across all projects its treated as a normal search. However, we also have an endpoint in the hail backend designed to specifically query a single variant, which currently does not query project tables at all. Using this endpoint is much more efficient for this type of search, however, because in the usual search we do a full outer join of all the project tables and the annotate from the annotations table, while in the variant_lookup endpoint we start by querying the single row from the annotations table. Doing an inner join of the one field we need from the projects tables is much faster than doing an outer join on all the tables, and while we can't to that for most searches because we will lose variants that are only in some projects, in this particular case we can because we only care about the one variant

The longer term plan for the "seqr" variant search link was always to switch to the variant lookup page, I was just initially intending to wait until more of the internal matching functionality had been build out, but theres no harm in doing this sooner


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -313,19 +315,28 @@ const variantSearchLinks = (variant, mainTranscript, genesById, user) => {
...
+      buttonText="seqr"
+      genomeVersion={genomeVersion}
+      svType={svType}
+      variantId={svType ? null : variantId}
+      location={svType && (
+        (endChrom && endChrom !== chrom) ? `${chrom}:${pos - 50}-${pos + 50}` : `${chrom}:${pos}-${end}%20`)}
+    />
+  ) : (
+    <NavLink
+      to={`/summary_data/variant_lookup?variantId=${variantId}&genomeVersion=${genomeVersion}&include_genotypes=true`}
```
jklugherz, line 329: I don't see `summary_data/variant_lookup` in `urls.py,`  should this just be `/variant_lookup?...`?

hanars: So the distinction between api routes for fetching data and top-level routes for rendering the seqr app can be a bit confusing. 

This is not linking to an `/api` endpoint for fetching data, its essentially an anchor tag linking to an rendered seqr app page. urls.py includes `'summary_data/.*'` as a route under the `react_app_pages` which means this page will render the main app html with js and then the exact page routing will be handled using the react apps routing.

If you lookin the SummaryData.jsx component you can see how all the subpaths are routed, and `variant_lookup` routes to the `VariantLookup` component. In that component, you can see that it sets `'/api/variant_lookup'` as the url for data fetching: https://github.com/broadinstitute/seqr/blob/master/ui/pages/SummaryData/components/VariantLookup.jsx#L48

jklugherz: riiiight I remember this, I should've been looking in `app.jsx` to where the component routes are defined.

PR #3839, opened 2024-01-26
add env variable to hail search tests

hanars:


bpblanken (approved):
good whenever the tests pass

PR #3850, opened 2024-02-02
Rna seq optimization

hanars:
This PR improves performance for loading very large files of RNA data. While the control flow changes quite a lot, the actual behavior of loading does not change in general.

The one exception is that previously if a sample was present in a file with multiple tissue types this raised an error. Since we know that FRASER2 will have this and maintaining the old behavior was hard with the new code, this updates the behavior to instead load the data grouped by sample-tissue_type, so if there are multiple tissue types data will be loaded for each. Seqr has always supported having data from multiple tissue types, its just that previously they had to be loaded from separate files

The major change relates to how much of the loading data is held in memory, given that the file we are anticipating is ~12Gb. Previously the flow went roughly as follows:
- Read the whole file into memory
- Iterate through the whole file, parsing each row and storing the parsed row in a dictionary
- Validate the parsed data and return errors if needed
- Write all the parsed data to a single json file
- Once the user confirms loading, trigger `load_rna_seq_sample_data` for each sample. This reads in the whole file, filters to the specified sample, and then creates the data models for that samples

The updated flow is as follows:
- Use an iterator to iterate through the file one line at a time
- Iterate through the whole file, parsing *and validating* each row and storing the valid parsed row in a dictionary
- If sufficient data for a given sample has been parsed, write the sample-specific data to the file and clear the dictionary entry for it
- If file has failed validation, continue to validate the remainder of the file but do not store the parsed rows
- Final validation and return errors if needed
- Write all remaining parsed data to sample-specific files
- Once the user confirms loading, trigger `load_rna_seq_sample_data` for each sample. This now reads in just the sample-specific file and loads that data


seqr/views/utils/file_utils.py
```diff
@@ -54,7 +65,7 @@ def parse_file(filename, stream):
 
         return rows
 
-    elif filename.endswith('.json'):
+    elif filename.endswith('.json') and not iter_file:
         return json.loads(stream.read())
 
     raise ValueError("Unexpected file type: {}".format(filename))
```
jklugherz, line 71: is it worth worth adding the value of `iter_file` in this message? there's a possible case now where the file is a valid format but `iter_file` is True and this error is raised, in this case the error message would be incorrect

seqr/views/utils/dataset_utils.py
```diff
@@ -298,67 +320,103 @@ def _parse_rna_row(row, column_map, required_column_map, missing_required_fields
...
 
     sample_id_to_individual_id_mapping = {}
     if mapping_file:
         sample_id_to_individual_id_mapping = load_mapping_file_content(mapping_file)
 
-    samples_by_id = defaultdict(dict)
+    samples_by_guid = defaultdict(dict)
     f = file_iter(file_path, user=user)
-    header = _parse_tsv_row(next(f))
+    parsed_f = parse_file(file_path.replace('.gz', ''), f, iter_file=True)
```
jklugherz, line 345: ok I'm reviewing generators and iterators and yield in python as we speak so this might be a silly question, but why do we need to `yield` the parsed line (`file_utils.py` line 38) if the un-parsed line is already `yield`ed from `file_iter`?

seqr/views/utils/dataset_utils.py
```diff
@@ -298,67 +320,103 @@ def _parse_rna_row(row, column_map, required_column_map, missing_required_fields
...
+
+            if missing_required_fields or (unmatched_samples and not ignore_extra_samples) or (sample_key in unmatched_samples):
+                # If there are definite errors, do not process/save data, just continue to check for additional errors
+                continue
+
+            if current_sample != sample_guid:
+                if len(samples_by_guid[current_sample]) > 5000:
+                    save_sample_data(current_sample, samples_by_guid[current_sample])
+                    del samples_by_guid[current_sample]
+                current_sample = sample_guid
```
jklugherz, line 383: I don't follow what's happening here - why are you checking `if len(samples_by_guid[current_sample]) > 5000`? should `5000` be a constant?

seqr/views/utils/file_utils.py
hanars: agreed

seqr/views/utils/dataset_utils.py
hanars: I'm not seeing the unparsed line being yielded, just the parsed line

seqr/views/utils/dataset_utils.py
hanars: It should be a constant. The idea here is we do not want to be holding so much in memory, so if a substantive amount of data has been parsed for a sample we should save that data and continue. I'll add a comment to make it clearer

seqr/views/utils/dataset_utils.py
jklugherz: here's the unparsed yield line I'm referring to (in `file_iter()`) https://github.com/broadinstitute/seqr/blob/b57c19e2615fd7669e2f27d046ff5c2e4b6f667a/seqr/utils/file_utils.py#L68

hanars: Ah. So that part is a generator/iterator, but then what happens is that generator is passed into `parse_file` and then the whole thing is read into a list, so the whole thing is then in memory and no longer an iterator:
https://github.com/broadinstitute/seqr/blob/b57c19e2615fd7669e2f27d046ff5c2e4b6f667a/seqr/views/utils/file_utils.py#L46

So what this change is doing is saying that in this case we do not want to actually evaluate the whole iterator here, we want to continue to keep it as a generator/ iterator (I sometimes am fuzzy on the distinction between those)

jklugherz: ahhhhh I see thanks for the clarification

jklugherz (approved):

PR #3854, opened 2024-02-05
Project entries schema update

hanars:
All the code changes needed to use the updated project entries schema. Blocked on reloading data with new schema


hail_search/queries/base.py
```diff
@@ -354,67 +363,64 @@ def _filter_entries_table(self, ht, sample_data, inheritance_mode=None, inherita
...
-        sample_id_family_index_map = {}
-        for sample_id, family_guid in sorted(sample_id_family_map.items()):
-            sample_index = sample_id_index_map[sample_id]
-            family_index = family_index_map[family_guid]
-            if not family_sample_indices[family_index]:
-                family_sample_indices[family_index] = []
-            sample_id_family_index_map[sample_id] = (family_index, len(family_sample_indices[family_index]))
-            family_sample_indices[family_index].append(sample_index)
-        family_sample_indices = hl.array(family_sample_indices)
+        sorted_family_sample_data = [[hl.eval(sample) for _, sample in samples] for _, samples in family_sample_index_data]
```
bpblanken, line 397: This `hl.eval` in a loop looks like it's gonna be slow.  iiuc, `family_sample_index_data` is a mapping of the index of the family_guid to a list of the indexes of the samples?  is it not already `eval`ed?

hanars: so they are a series of structs and somehow they get like un-evaled and you can't use them directly later on without evaling them. they are just a flat array of structs they aren't like a table on disk so the eval is pretty fast, but I can probably get it to work without it

bpblanken (approved):

PR #3855, opened 2024-02-05
RNA-seq manage command cleanup

hanars:
cretes a single manage command for loading rna, instead of needing one command per data type. Makes maintenance easier


jklugherz (approved):
lgtm !

PR #3858, opened 2024-02-06
feat: Clearly show when variant transcript is not Canonical nor MANE

tommyli:
Our analysts/curators were a little confused when the Seqr variant search results did not return the Canonical nor MANE transcripts.  We subsequently added logic to highlight this in the UI.  We've had a fair bit of discussion around this and @drtconway in our group could can provide better context and why it's useful.  I initially thought of raising a discussion first but I think a PR is easier, just reject it if your group doesn't find it useful.

ScreenShot 2024-02-06 at 4 18 51 pm

ScreenShot 2024-02-06 at 4 19 16 pm

### DIFF
--- a/ui/shared/components/panel/variants/Annotations.jsx
+++ b/ui/shared/components/panel/variants/Annotations.jsx
@@ -7,6 +7,7 @@ import { Popup, Label, Icon } from 'semantic-ui-react'
 
 import {
   getGenesById,
+  getTranscriptsById,
   getLocusListIntervalsByChromProject,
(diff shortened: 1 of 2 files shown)


hanars:
@lynnpais I think this functionality would be useful for many seqr users, but let me know if you disagree or if you would like us to adjust the actual UI in any way

lynnpais:
The analyst team is in agreement that this is a useful feature. Thanks Tommy Li!

hanars (approved):

PR #3861, opened 2024-02-06
rna seq parsing clean up

hanars:
fixes a bug/inefficiency introduced in my last PR, and also reformats a confusing helper function

### DIFF
--- a/seqr/views/utils/dataset_utils.py
+++ b/seqr/views/utils/dataset_utils.py
@@ -329,26 +329,10 @@ def _validate_rna_header(header, column_map):
     return required_column_map
 
 
-def _parse_rna_row(row, column_map, required_column_map, missing_required_fields, allow_missing_gene, should_skip=None, format_fields=None):
-    if not (should_skip and should_skip(row)):
-        row_dict = {mapped_key: row[col] for mapped_key, col in column_map.items()}
-        for mapped_key, format_func in (format_fields or {}).items():
-            row_dict[mapped_key] = format_func(row_dict[mapped_key])
-
-        missing_cols = {col_id for col, col_id in required_column_map.items() if not row.get(col)}
-        if allow_missing_gene:
-            missing_cols.discard(GENE_ID_COL)
-        sample_id = row_dict.pop(SAMPLE_ID_COL) if SAMPLE_ID_COL in row_dict else row[SAMPLE_ID_COL]
-        if missing_cols:
-            for col in missing_cols:
-                missing_required_fields[col].append(sample_id)
-        if not missing_cols:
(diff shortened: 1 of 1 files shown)


seqr/views/utils/dataset_utils.py
```diff
@@ -366,44 +350,59 @@ def _load_rna_seq_file(
     missing_required_fields = defaultdict(list)
     gene_ids = set()
     current_sample = None
-    for line in tqdm(f, unit=' rows'):
-        row = dict(zip(header, _parse_tsv_row(line)))
-        for sample_id, row_dict in _parse_rna_row(
```
hanars, line 371: `_parse_rna_row` is a generator so this was executing a `for` loop over it, but thats confusing and implies that there can be multiple sample/row_dicts per row. In fact, there are either 0 or 1 valid row_dicts per row, so it makes more sense to remove this helper and the unnecessary level of nesting it introduces

seqr/views/utils/dataset_utils.py
```diff
@@ -366,44 +350,59 @@ def _load_rna_seq_file(
...
-                row, column_map, required_column_map, missing_required_fields, allow_missing_gene, **kwargs):
-            tissue_type = TISSUE_TYPE_MAP[row[TISSUE_COL]]
-            project = row[PROJECT_COL]
-            sample_key = (sample_id, project, tissue_type)
-
-            if sample_key in potential_loaded_samples:
-                loaded_samples.add(sample_key)
-                continue
+    for line in tqdm(parsed_f, unit=' rows'):
+        row = dict(zip(header, line))
```
hanars, line 354: I created `parsed_f` in my last PR but forgot to actually use it in the iteration. ops

bpblanken (approved):

seqr/views/utils/dataset_utils.py
```diff
@@ -329,26 +329,10 @@ def _validate_rna_header(header, column_map):
...
-                missing_required_fields[col].append(sample_id)
-        if not missing_cols:
-            yield sample_id, row_dict
-
-
 def _load_rna_seq_file(
         file_path, user, potential_loaded_samples, update_sample_models, save_sample_data, get_matched_sample, mismatches,
-        column_map, mapping_file=None, get_unique_key=None, allow_missing_gene=False, ignore_extra_samples=False, **kwargs,
+        column_map, mapping_file=None, get_unique_key=None, allow_missing_gene=False, ignore_extra_samples=False,
+        should_skip=None, format_fields=None,
```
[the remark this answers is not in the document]
hanars: thats fair, but I think out of scope for this change

PR #3870, opened 2024-02-09
use variant lookup for single variant id search

hanars:


hanars:
thinking about this more, I think we do not want to make this change for now as it actually makes futureupdates to internal variant matching harder

PR #3873, opened 2024-02-12
Recessive search split tables

hanars:
This changes the control flow for recessive searches. In the search included in the linked ticket, this reduced the search time by over 60seconds on my local machine, from > 300 seconds that could not even complete on production to 245 seconds which can

Previously, we read in the entries table(s), filtered them to variants that were valid for either hom recessive or comp het, annotated and filtered them, and then at the end ran a filter for hom recessive only and a filter for comp het only to separate it out. What this instead does is that the entry table is initally filtered into 2 different tables, one for hom recessive and one for comp het, and then each table is independently annotated and filtered for the duration of the search.

 Originally I did not go this route because it involved calling `query_table` on the annotations table twice which intuitively seems slower. However, since `query_table` runs roughly linearly to the number of variants to query regardless of how many distinct calls its broken into, and because within a single family possible SNV_INDELs can only match one of the 2 modes of inheritance but not both, that means that the subsetted tables are non-overlapping so the `query_table` time stays essentially the same and then all the additional annotating to maintain the 2 possible inheritance modes and all the filtering at the end to seperate the tables back out is fully removed, saving a bunch of time.

Note that this does slightly slow down SV only comp het search, as the inheritance for those means a variant can both be hom recessive and comp het so we do double query some variants, but SV only search was always much faster to begin with (when testing this change the SV-only part of the search went from 15 to 20 seconds, while the SNV__INDEL portion went from 170 to 105. The SV slow down seems very reasonable in that context). Also, its possible that in a multi-family comp het search it could result in double querying some variants that are recessive in some families and comp het in others. However, multi family comp het search is required to be gene-restircted so its already much faster, and these single family searches are substantially more common than thse, so the treade off feels worth it

### DIFF
--- a/hail_search/queries/base.py
+++ b/hail_search/queries/base.py
@@ -233,20 +233,6 @@ class BaseHailTableQuery(object):
         self.import_filtered_table(
             sample_data, parsed_intervals=parsed_intervals, **kwargs)
 
-        if self._has_comp_het_search:
-            self._comp_het_ht = self._filter_compound_hets()
-            if self._is_recessive_search:
-                self._ht = self._ht.filter(self._ht.family_entries.any(hl.is_defined))
-                if self._has_secondary_annotations:
-                    annotation_filters = self._get_annotation_filters(self._ht)
-                    if annotation_filters:
-                        self._ht = self._ht.filter(hl.any(annotation_filters))
-                    else:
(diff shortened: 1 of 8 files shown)


bpblanken (approved):
logic seems ok to me!

PR #3877, opened 2024-02-13
No phasing comp het

hanars:
This adds a few minor performance/control flow optimizations to avoid extra computations in proband only and duo families. However, the bulk of the performance savings come from setting the `use_new_shuffle` flag, which reduced the runtime of a problematic recessive search from 385 seconds to 230 seconds


hail_search/web_app.py
```diff
@@ -66,6 +66,7 @@ async def init_web_app():
     if JAVA_OPTS_XSS:
         spark_conf.update({f'spark.{field}.extraJavaOptions': f'-Xss{JAVA_OPTS_XSS}' for field in ['driver', 'executor']})
     hl.init(idempotent=True, spark_conf=spark_conf or None)
+    hl._set_flags(use_new_shuffle='1')
```
bpblanken, line 69: hello darkness my old friend

bpblanken (approved):

PR #3883, opened 2024-02-14
Sv lookup interval

hanars:
We were previously looking up matching SVs with a different sample type by just checking if there is any overlap between them. In fact, we should be checking that the breakpoints are within a specified tolerance of each other
Comparable ES functionality: https://github.com/broadinstitute/seqr/blob/master/seqr/utils/search/elasticsearch/es_search.py#L1427

### DIFF
--- a/hail_search/queries/sv.py
+++ b/hail_search/queries/sv.py
@@ -56,7 +56,7 @@ class SvHailTableQuery(BaseHailTableQuery):
     def _get_sample_type(cls, *args):
         return cls.DATA_TYPE.split('_')[-1]
 
-    def _filter_annotated_table(self, ht, *args, parsed_intervals=None, exclude_intervals=False, **kwargs):
+    def _filter_annotated_table(self, ht, *args, parsed_intervals=None, exclude_intervals=False, padded_interval=None, **kwargs):
         if parsed_intervals:
             interval_filter = hl.array(parsed_intervals).any(lambda interval: hl.if_else(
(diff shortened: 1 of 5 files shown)


hail_search/queries/sv.py
```diff
@@ -76,9 +77,20 @@ def _filter_annotated_table(self, ht, *args, parsed_intervals=None, exclude_inte
...
+            ht = ht.filter(hl.all([
+                ht.start_locus.contig == f"chr{padded_interval['chrom']}",
+                self._locus_in_range(ht.start_locus, padded_interval['start'], padding),
+                self._locus_in_range(ht.end_locus, padded_interval['end'], padding)
+            ]))
 
         return super()._filter_annotated_table(ht, *args, **kwargs)
 
+    @staticmethod
+    def _locus_in_range(locus, position, padding):
```
bpblanken, line 91: Is [this](https://hail.is/docs/0.2/hail.expr.LocusExpression.html#hail.expr.LocusExpression.window) something that might be of use here?

Also, where does `3e8` come from?  Is it just "big number that's larger than the max pos on any config"?

bpblanken (approved):
don't wanna hold this up on style questions

hail_search/queries/sv.py
hanars: its copied from xpos_utils in seqr, its what we use in ES to validate. `window` seems better but I will add it is a subsequent PR to unblock this

PR #3888, opened 2024-02-15
improve performance for search with many intervals

hanars:
Changed a Mendiliome search that failed locally after 500 seconds to finishing in 90 seconds 🤷‍♀️

### DIFF
--- a/hail_search/queries/mito.py
+++ b/hail_search/queries/mito.py
@@ -5,6 +5,7 @@ from hail_search.constants import ABSENT_PATH_SORT_OFFSET, CLINVAR_KEY, CLINVAR_
     PATHOGENICTY_HGMD_SORT_KEY
 from hail_search.queries.base import BaseHailTableQuery, PredictionPath, QualityFilterFormat
 
+MAX_LOAD_INTERVALS = 1000
 
 def _clinvar_sort(clinvar_field, r):
     return hl.or_else(r[clinvar_field].pathogenicity_id, ABSENT_PATH_SORT_OFFSET)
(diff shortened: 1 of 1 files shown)


hail_search/queries/mito.py
```diff
@@ -197,6 +198,8 @@ def _parse_variant_keys(self, variant_ids=None, **kwargs):
     def _prefilter_entries_table(self, ht, parsed_intervals=None, exclude_intervals=False, **kwargs):
         if exclude_intervals and parsed_intervals:
             ht = hl.filter_intervals(ht, parsed_intervals, keep=False)
+        elif len(parsed_intervals or []) >= MAX_LOAD_INTERVALS:
```
bpblanken, line 201: do we know why this is faster?

[the docs](https://hail.is/docs/0.2/guides/genetics.html#using-hl-filter-intervals) say this is `suitable for a small list of intervals`, and we're doing the opposite.

hanars: No idea, its very counterintuitive. I asked on zulip - https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/search/_intervals.20filter_intervals

hail_search/queries/mito.py
```diff
@@ -135,7 +136,7 @@ def __init__(self, *args, **kwargs):
 
     def _parse_intervals(self, intervals, exclude_intervals=False, **kwargs):
         parsed_intervals = super()._parse_intervals(intervals,**kwargs)
-        if parsed_intervals and not exclude_intervals:
+        if parsed_intervals and not exclude_intervals and len(parsed_intervals) < MAX_LOAD_INTERVALS:
```
bpblanken, line 139: is `load_table_kwargs` only passed to the annotations read table and not the entries read table?

hanars: no its *only* passed to the entries read table. the annotations table is not read in, its just accessed via query_table

bpblanken (approved):

PR #3894, opened 2024-02-15
Numeric type error

hanars:


hail_search/queries/base.py
```diff
@@ -976,7 +976,7 @@ def _get_sort_expressions(self, ht, sort):
 
         if sort in self.PREDICTION_FIELDS_CONFIG:
             prediction_path = self.PREDICTION_FIELDS_CONFIG[sort]
-            return [-hl.float64(ht[prediction_path.source][prediction_path.field])]
+            return [hl.or_else(-hl.float64(ht[prediction_path.source][prediction_path.field]), 0)]
```
bpblanken, line 979: Is there a reason this is float64 instead of 32?

hanars: So the sort list always ends with `xpos` which is an `int64` field that is actually too big to be an `int32`, so what I found that if this is a `float64` and it gets added to a list with the `int64` they both end up as `float64` but if this is `float34` they both end up as `float34` and `xpos` gets truncated

bpblanken (approved):

PR #3898, opened 2024-02-16
do not use interval prefilter for large gene lists

hanars:
Substantially sped up Mendeliome search


hail_search/queries/base.py
```diff
@@ -582,6 +582,9 @@ def _parse_intervals(self, intervals, **kwargs):
             reference_genome = hl.get_reference(self.GENOME_VERSION)
             intervals = (intervals or []) + [reference_genome.x_contigs[0]]
 
+        if len(intervals) > 100 and len(intervals) == len(gene_ids or []):
```
bpblanken, line 585: Can we make this a constant and add a comment about how it was determined?

bpblanken (approved):

PR #3899, opened 2024-02-16
fix performace for many project search

hanars:


hail_search/queries/base.py
```diff
@@ -245,13 +245,19 @@ def _read_table(self, path, drop_globals=None, use_ssd_dir=False, skip_missing_f
...
+            if self._should_skip_ht(ht, skip_missing_field):
                 return None
             ht_globals = hl.read_table(table_path).globals
             if drop_globals:
                 ht_globals = ht_globals.drop(*drop_globals)
             return ht.annotate_globals(**hl.eval(ht_globals))
-        return hl.read_table(table_path, **self._load_table_kwargs)
+
+        ht = hl.read_table(table_path, **self._load_table_kwargs)
+        return None if self._should_skip_ht(ht, skip_missing_field) else ht
```
bpblanken, line 256: Is it possible to just lift

```
if self._should_skip_ht(ht, skip_missing_field):
  return None
```

above `if 'variant_ht' in self._load_table_kwargs:` so that we only check that logic a single time?

bpblanken (approved):

PR #3905, opened 2024-02-20
Run hail in separate thread (to keep status checks alive).

bpblanken:


hanars (commented):
I'm a little confused by this approach. If we have `max_workers=1` doesn't that mean that there is only one thread which everything is running on regardless of whether we explicitly set the executor or use the default? Also, wouldn't it be more straightforward to let everything *except* that status checks run on the "main" thread and have a reserved secondary executor just for the status checks?

```

```

bpblanken:
yeah... this one is super tricky.  So the `max_workers=1` bit creates a second worker thread off the "main" thread.  So there's definitely two threads here, one for the server process and a second for executing the blocking hail queries.

for

> Also, wouldn't it be more straightforward to let everything except that status checks run on the "main" thread and have a reserved secondary executor just for the status checks?

Yeah, it'd definitely be conceptually easier to reason about.  The issue though, is that a hail query, while running, blocks the main server thread from doing server things like responding to requests.  The `/status` check does not, as it's a native `aiohttp` object, which under the hood is using actual non-blocking io utils. 

The best way to see this in action is adding something like: 

```
import time
async def sleep_sync(request: web.Request) -> web.Response:
    time.sleep(10)
    return web.json_response({'success': True})

async def sleep_async(request: web.Request) -> web.Response:
    await asyncio.sleep(10)
    return web.json_response({'success': True})

async def sleep_pseudoasync(request: web.Request) -> web.Response:
    loop = asyncio.get_running_loop()
    await loop.run_in_executor(request.app.pool, functools.partial(time.sleep, 10))
    return web.json_response({'success': True})
...    

web.get('/sleep_sync', sleep_sync),
web.get('/sleep_async', sleep_async),
web.get('/sleep_pseudoasync', sleep_pseudoasync),
```

to the web_app and trying to `curl` `/status` while `/sleep_sync` or `/sleep_async` `/sleep_pseudoasync` is `curl`-ed and running.

hanars:
okay thanks, that makes sense. In that case, could we add a helper function thats used by all the requests on the secondary worker, so that theres less boiler plate and it'll be easier to maintain with new endpoints going forward?
```
async def run_request_off_main_thread(request: web.Request, func: Callable, **kwargs) -> web.Response:
    loop = asyncio.get_running_loop()
    response = await loop.run_in_executor(request.app.pool, functools.partial(func, await request.json(), **kwargs))
    return web.json_response(response, dumps=hl_json_dumps)
```

hanars (approved):

PR #3908, opened 2024-02-21
All project gene search optimization

hanars:


hail_search/queries/base.py
```diff
@@ -249,19 +249,13 @@ def _read_table(self, path, drop_globals=None, use_ssd_dir=False, skip_missing_f
...
         if 'variant_ht' in self._load_table_kwargs:
             ht = self._query_table_annotations(self._load_table_kwargs['variant_ht'], table_path)
-            if self._should_skip_ht(ht, skip_missing_field):
+            if skip_missing_field and not ht.any(hl.is_defined(ht[skip_missing_field])):
                 return None
             ht_globals = hl.read_table(table_path).globals
             if drop_globals:
                 ht_globals = ht_globals.drop(*drop_globals)
             return ht.annotate_globals(**hl.eval(ht_globals))
-
```
hanars, line 258: reverts https://github.com/broadinstitute/seqr/pull/3899

hail_search/queries/base.py
```diff
@@ -343,6 +344,9 @@ def _add_project_ht(self, families_ht, project_ht, default, default_1=None):
         if default_1 is None:
             default_1 = default
 
+        if not project_ht.any(hl.is_defined(project_ht.family_entries)):
+            return families_ht, False
```
jklugherz, line 348: when would a family be included in a search but not in project?

jklugherz (approved):

hail_search/queries/base.py
hanars: so this case is reached when there is data loaded for a family but there are no rows in the table that match the search after its been filtered for inheritance, quality, and location. In cases where you are doing a restrictive search in a single gene across hundreds of projects, many of them will end up not having any hits so skipping the join improves performance

bpblanken (approved):

PR #3911, opened 2024-02-22
Better data type restrictions for searches

hanars:
Improves search performance by adding additional prefiltering for which data types can be included in search. In particular, removes mito search from several places where no mito variants are possible

### DIFF
--- a/seqr/utils/search/utils.py
+++ b/seqr/utils/search/utils.py
@@ -6,7 +6,7 @@ from reference_data.models import GENOME_VERSION_LOOKUP, GENOME_VERSION_GRCh38
 from seqr.models import Sample, Individual, Project
 from seqr.utils.redis_utils import safe_redis_get_json, safe_redis_set_json
 from seqr.utils.search.constants import XPOS_SORT_KEY, PRIORITIZED_GENE_SORT, RECESSIVE, COMPOUND_HET, \
-    MAX_NO_LOCATION_COMP_HET_FAMILIES, SV_ANNOTATION_TYPES, ALL_DATA_TYPES, MAX_EXPORT_VARIANTS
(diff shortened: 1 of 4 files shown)


seqr/utils/search/utils.py
```diff
@@ -255,8 +251,11 @@ def _query_variants(search_model, user, previous_search_results, sort=None, num_
     dataset_type, secondary_dataset_type, lookup_dataset_type = _search_dataset_type(parsed_search)
     parsed_search.update({'dataset_type': dataset_type, 'secondary_dataset_type': secondary_dataset_type})
     search_dataset_type = None
-    if dataset_type and dataset_type != ALL_DATA_TYPES and (secondary_dataset_type is None or secondary_dataset_type == dataset_type):
-        search_dataset_type = lookup_dataset_type or dataset_type
+    if dataset_type and dataset_type != ALL_DATA_TYPES:
+        if secondary_dataset_type is None or secondary_dataset_type == dataset_type:
+            search_dataset_type = lookup_dataset_type or dataset_type
+        elif dataset_type == Sample.DATASET_TYPE_SV_CALLS:
+            search_dataset_type = DATASET_TYPE_NO_MITO
```
jklugherz, line 258: what's the purpose of this `DATASET_TYPE_NO_MITO` `search_dataset_type`?

jklugherz (approved):

seqr/utils/search/utils.py
hanars: also, there are no mitochondrial SVs so we do not look for compound hets that are an SV/MITO pair. So if its a recessive search where the primary data type is SV, then no mito variants can be the primary variant, and no mito variants can be the secondary variant because they will not share a gene by definition, so this is handling that case and saying we should not include the MITO data here either

PR #3913, opened 2024-02-23
update mito annotations

jklugherz:
- updates mito clinvar pathogenicity filter range for 'vus_or_conflicting'
- updates annotations source for sift` score


hanars:
Yeah, lets think of it as theres gonna be a single release in which the following things will all be released simultaneously:
- An updated MITO annotations table
- An updated SNV_INDEL annotations table
- This PR, which includes all the changes neccessary for both those updated annotations tables

hanars (commented):
generally looks good, although we should hold off on merging until we have a release plan. One point of clarification, are we planning to update clinvar for SNV_INDEL for both 37 and 38 at the same time?

jklugherz:
Since the clinvar enum code is shared between SNV_INDEL 37 and 38 and ONT_SNV_INDEL, I think we either need also need to update those annotations, or handle the old enums in this PR. 
@bpblanken is it feasible to run the pipeline for all 3 data types tomorrow?

hanars (changes requested):
actualy should hold off to prevent breaking dev

hanars (approved):
jk its ready!

PR #3914, opened 2024-02-23
Add Notifications

hanars:


seqr/utils/search/add_data_utils.py
```diff
@@ -49,48 +49,43 @@ def add_new_es_search_samples(request_json, project, user, notify=False, expecte
...
-        date=project.created_date.date().strftime('%B %d, %Y'),
-        anvil_url=ANVIL_UI_URL,
-        namespace=project.workspace_namespace,
-        name=project.workspace_name,
-        base_url=BASE_URL,
-        guid=project.guid,
-        project_name=project.name,
-        num_sample=num_samples,
-    ),
+        email = f'This is to notify you that {sample_summary} have been loaded in seqr project {project_link}'
```
jklugherz, line 70: why don't anvil project users get notification emails?

seqr/views/apis/project_api.py
```diff
@@ -329,6 +330,55 @@ def project_mme_submisssions(request, project_guid):
     })
 
 
+@login_and_policies_required
+def project_notifications(request, project_guid, read_status):
+    project = get_project_and_check_permissions(project_guid, request.user)
+    is_subscriber = project.subscribers.user_set.filter(id=request.user.id).exists()
+    if not is_subscriber:
+        max_loaded = _project_notifications(
+            project, request.user.notifications).aggregate(max_loaded=Max('timestamp'))['max_loaded']
```
jklugherz, line 339: can't you just do 
`max_loaded = _project_notifications(project, request.user.notifications).aggregate(Max('timestamp'))`

seqr/fixtures/users.json
```diff
@@ -31,6 +31,22 @@
...
+    "fields": {
+        "name": "subscribers",
+        "permissions": []
+    }
+},
+{
+    "model": "auth.group",
+    "pk": 7,
+    "fields": {
+        "name": "subscribers_external",
```
jklugherz, line 46: what differentiates a subscriber from an external subscriber?

jklugherz (approved):

seqr/fixtures/users.json
hanars: oh it was just a naming choice I made because this group is assigned to a project thats referred to in the tests as "external". In real life every project has its own distinct subscriber group and the memebers can be different, so I wanted to be able to test that the correct project-specific notifications were going to the correct subscribers

PR #3921, opened 2024-02-26
Individual ID should be unique to project

hanars:
There was an edge case that was allowing individuals with duplicate IDs to be created in different families in the same project. It turns there is no way in postgres to add a constraint at the database level that references multiple tables, but I added a fail-safe behavior to the `save` method of the Individual model so database saves will error if duplicates are created. I also added better error handling so users will be informed up front if they are including individuals that already exist. I also added some vlaidation to prevent renaming or moving individuals if they already have data loaded, as that could break search


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -601,6 +602,15 @@ def test_add_workspace_data(self, mock_compute_indiv_guid):
         url = reverse(add_workspace_data, args=[PROJECT1_GUID])
         self._test_errors(url, ['uploadedFileId', 'fullDataPath', 'vcfSamples'], TEST_WORKSPACE_NAME)
 
+        # Test Individual ID exists in an omitted family
+        self.mock_load_file.return_value = LOAD_SAMPLE_DATA + INVALID_ADDED_SAMPLE_DATA
```
hanars, line 606: this is covering the edge case that was allowing duplicates to be created

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -77,9 +77,9 @@ def parse_pedigree_table(parsed_file, filename, user, project):
     return json_records, warnings
 
 
-def parse_basic_pedigree_table(parsed_file, filename, required_columns=None):
+def parse_basic_pedigree_table(project, parsed_file, filename, required_columns=None):
     rows, header = _parse_pedigree_table_rows(parsed_file, filename)
-    return _parse_pedigree_table_json(rows, header=header, fail_on_warnings=True, required_columns=required_columns)
+    return _parse_pedigree_table_json(project, rows, header=header, fail_on_warnings=True, required_columns=required_columns, allow_id_update=False)
```
hanars, line 82: this is only used when parsing anvil uploaded pedigree, and there is no reason why we would ever want to allow anvil users to change the IDs of existing individuals within their projects

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -260,6 +267,11 @@ def validate_fam_file_records(records, fail_on_warnings=False, errors=None):
         individual_id_counts[individual_id] += 1
         family_id = r.get(JsonConstants.FAMILY_ID_COLUMN) or r['family']['familyId']
 
+        if loaded_individual_families.get(r.get(JsonConstants.PREVIOUS_INDIVIDUAL_ID_COLUMN)):
```
jklugherz, line 270: what is the `previousIndividualId` and how is it different from `individualId`?

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -211,6 +212,9 @@ def _parse_header_columns(header):
             column = JsonConstants.FAMILY_NOTES_COLUMN
         elif key.startswith("notes"):
             column = JsonConstants.NOTES_COLUMN
+        elif 'indiv' in key and 'previous' in key:
+            if allow_id_update:
```
jklugherz, line 216: what's the use case for updating an individual id?

jklugherz (approved):

seqr/views/utils/pedigree_info_utils.py
hanars: Usually, its that someone on the PM team had a typo in their sample manifest so the ID in seqr is wrong. It doesn't happen in the AnVIL case, because we validate their pedigree against the VCF they provide so the IDs have to match. But for internal projects they get created first and loaded later, so you can find out at load time that the ID in seqr is wrong

Also, for RGP we create IDs like RGP_#_1, RGP_#_2 but if we later learn we are not going to get a samples for a parent they will change the name to RGP_#_mom or RGP_#_dad

PR #3925, opened 2024-02-27
Variant metadata tags

hanars:


jklugherz (commented):
looks good. I have some context on this since Stephanie showed me this issue while I was shadowing her using seqr.

jklugherz (approved):

PR #3929, opened 2024-02-28
Reload saved variants after new samples loaded

hanars:
After new data is loaded, seqr allele frequencies (plus possibly other annotations) will be updated for all variants of the data type. This will cause the saved variants to be updated to stay in sync with the latest annotations

### DIFF
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -88,7 +88,7 @@ def update_project_saved_variant_json(project_id, family_ids=None, dataset_type=
     variant_ids = sorted(variant_ids)
     families = sorted(families, key=lambda f: f.guid)
     variants_json = []
-    for sub_var_ids in [variant_ids[i:i+MAX_VARIANTS_FETCH] for i in range(0, len(variant_ids), MAX_VARIANTS_FETCH)]:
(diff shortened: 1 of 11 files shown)


hanars:
Realized I missed an edge case on this, will resubmit once functionality is accurate

seqr/views/utils/variant_utils.py
```diff
@@ -27,13 +29,53 @@
...
-    if family_id:
-        saved_variants = saved_variants.filter(family__family_id=family_id)
+def update_projects_saved_variant_json(projects, user_email, **kwargs):
+    success = {}
+    skipped = {}
+    error = {}
+    logger.info(f'Reloading saved variants in {len(projects)} projects')
+    for project_id, project_name, family_ids in tqdm(projects, unit=' project'):
+        try:
+            updated_saved_variant_guids = update_project_saved_variant_json(
```
bpblanken, line 39: Just to make sure I've followed this correctly, this makes a single search call to the hail search service per project, with all of the saved variants embedded in the POST body?

hanars: yes, although if theres more than 1000 variants it does them in batches

hanars: There is a better way to do it but it involves some changes to the hail backend, so I was going ot do that as a separate PR: https://github.com/broadinstitute/seqr-private/issues/1362#issuecomment-1971344093

bpblanken (approved):

PR #3933, opened 2024-02-29
fix lookup page js

hanars:


bpblanken (approved):
Maybe not the right place to ask this, but is there a reason we're using the minimized build on dev?  I went to investigate [here](https://seqr-dev.broadinstitute.org/summary_data/variant_lookup) and didn't make it too far.

hanars:
Um I never bothered to set up a different configuration in dev because I always need to debug JS errors locally? Not a super compelling reason

PR #3937, opened 2024-03-01
Ehigham comp het optimization

hanars:


hail_search/queries/base.py
```diff
@@ -824,108 +827,103 @@ def _filter_compound_hets(self):
...
-        """ After grouping by v1:
-         v1 | v2     | comp_het_gene_ids
-         1  | [2, 3] | {A}
-         2  | [3]    | {A, B}             
-        """
+        variants = variants.group_by(lambda v: hl.tuple([v[1].key_, v[2].key_]))
+        variants = variants.values().map(lambda v: hl.rbind(
+            hl.set(v.map(lambda v: v[0])),
+            lambda comp_het_gene_ids: hl.tuple([v[0][1], v[0][2], comp_het_gene_ids])
+        ))
```
ehigham, line 884: the `rbind` here is unnecessary through maybe makes things a bit clearer as to what this represents.
I'd stick with a `struct` so you can maintain names for what the various fields represent. `comp_het_gene_ids` is clearer and easier to remember in the future than `v[2]`

hail_search/queries/base.py
```diff
@@ -824,108 +827,103 @@ def _filter_compound_hets(self):
...
+        return hl.Table.parallelize(
+            variants.map(lambda v: hl.struct(**{GROUPED_VARIANTS_FIELD: v}))
+        )
 
-        return ch_ht
+    def _filter_comp_het_families(self, variants, set_secondary_annotations=True):
+        return variants.map(lambda v: hl.rbind(
+            hl.enumerate(v[0].family_entries).filter(
+                lambda x: self._is_valid_comp_het_family(v[0], v[1], x[0])
+            ).map(lambda x: x[0]),
```
ehigham, line 902: swap the `map` and `filter` - we don't do a great job at eliminating common subexpressions at the moment (`x[0]`) (hopefully coming soon!)

hail_search/queries/base.py
```diff
@@ -824,108 +827,103 @@ def _filter_compound_hets(self):
...
+            lambda valid_family_indices: hl.tuple([*v, valid_family_indices])
+        )).filter(
+            lambda v: v[3].any(hl.is_defined)
```
ehigham, line 905: would any of the indices not be defined?

hail_search/queries/base.py
ehigham: We also have a few more optimisations that target structs which can help reduce the amount of unpacking/packing of data. You may not see much benefit here though

ehigham (commented):
This looks great! Glad we were able to help. 
A few minor comments, mostly about readability.
If you can, prefer structs and use annotate/select as this can reduce the amount of data that we copy around from unpacking/repacking tuples.

hail_search/queries/base.py
hanars: No, I guess this was just a weird way of checking if the list is empty or not. Is there an optimal way to check for that? Wold `v[3].size() > 0` be better?

ehigham: That might be clearer

hail_search/queries/base.py
```diff
@@ -824,108 +827,105 @@ def _filter_compound_hets(self):
...
 
-    def _filter_grouped_compound_hets(self, ch_ht):
-        # Filter variant pairs for family and genotype
-        ch_ht = ch_ht.annotate(valid_families=hl.enumerate(ch_ht.v1.family_entries).map(
-            lambda x: self._is_valid_comp_het_family(ch_ht, x[1], ch_ht.v2.family_entries[x[0]])
+    def _filter_comp_het_families(self, variants, set_secondary_annotations=True):
+        return variants.map(lambda v: v.annotate(
+            valid_family_indices=hl.enumerate(v.v1.family_entries).map(lambda x: x[0]).filter(
+                lambda i: self._is_valid_comp_het_family(v.v1, v.v2, i)
+            )
```
ehigham, line 906: Is `len(v1.family_entries) <= len(v2.family_entries)`

hanars: no they are always the same length

hail_search/queries/base.py
```diff
@@ -824,108 +827,105 @@ def _filter_compound_hets(self):
...
+        2    | B        | true       | true
+        3    | A        | false      | true
+        3    | B        | false      | true
         """
-        key_expr = lambda v: v[self.KEY_FIELD[0]] if len(self.KEY_FIELD) == 1 else hl.tuple([v[k] for k in self.KEY_FIELD])
-        ch_ht = ch_ht.annotate(
-            v1_set=hl.set(ch_ht.v1.map(key_expr)),
-            v2=ch_ht.v2.group_by(key_expr).map_values(lambda x: x[0]),
+
+        variants = ch_ht.collect(_localize=False)
```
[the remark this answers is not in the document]
hanars: @ehigham ?

jklugherz (approved):

bpblanken (approved):

PR #3945, opened 2024-03-04
Improve AF prefilter

hanars:
reduces the number of variants in the filtered entry ht for comp hets from ~270,000 to ~85,000


hail_search/queries/snv_indel.py
```diff
@@ -90,10 +98,25 @@ def _get_gnomad_af_prefilter(self, frequencies=None, pathogenicity=None, **kwarg
...
+        af_filter = True if af_cutoff_field is True else lambda ht: ht[af_cutoff_field]
+
+        if af_cutoff < PATH_FREQ_OVERRIDE_CUTOFF:
+            clinvar_path_ht = self._get_loaded_clinvar_prefilter_ht(pathogenicity)
+            if clinvar_path_ht is not False:
+                path_cutoff_field = self._get_af_prefilter_field(PATH_FREQ_OVERRIDE_CUTOFF)
+                non_clinvar_filter = lambda ht: hl.is_missing(clinvar_path_ht[ht.key])
+                if af_filter is not True:
+                    non_clinvar_filter = lambda ht: non_clinvar_filter(ht) & af_filter(ht)
+                af_filter = lambda ht: ht[path_cutoff_field] | non_clinvar_filter(ht)
```
hanars, line 114: For clinvar path and likely path variants, seqr will return variants with up to 5% frequency even if the filter is set lower than that. To address this in the prefilter, we set it so if the clinvar path filter existed we just use the higher 5% cutoff for the prefilter for all variants. However, since we actually also have a clinvar prefilter table that we are already reading in to help with the quality filtering, we can add logic here so variants in clinvar are filtered out if they have greater than 5% frequency, and variants not in clinvar are prefiltered to the actual threshhold

hail_search/queries/snv_indel.py
```diff
@@ -73,6 +74,13 @@ class SnvIndelHailTableQuery(MitoHailTableQuery):
         PATHOGENICTY_HGMD_SORT_KEY: lambda r: [MitoHailTableQuery.CLINVAR_SORT(CLINVAR_KEY, r), r.hgmd.class_id],
     }
 
+    FREQUENCY_PREFILTER_FIELDS = OrderedDict([
+        (True, PREFILTER_FREQ_CUTOFF),
```
jklugherz, line 78: why is af of 1% handled differently than 3, 5, and 10%? why don't we store it in our reference data?

hanars: this isn't reading the reference data, its reading the `high_af_variants` prefilter table. This is a way of quickly filtering out high frequency variants without reading in the whole table by instead anti-joining with a table that has all the varianst that are too frequent to return in search. Everything in the prefilter table has a frequency of at least 0.1%, so that threshhol is treated differently because you can just filter on the existance of the row in the table and not a specific annotation on the row

jklugherz (approved):

bpblanken (approved):

PR #3955, opened 2024-03-07
Probably solved

hanars:


bpblanken (approved):

seqr/migrations/0060_alter_family_analysis_status.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('seqr', '0059_project_subscribers'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='family',
+            name='analysis_status',
+            field=models.CharField(choices=[('S', 'S'), ('S_kgfp', 'S'), ('S_kgdp', 'S'), ('S_ng', 'S'), ('ES', 'E'), ('Sc_kgfp', 'S'), ('Sc_kgdp', 'S'), ('Sc_ng', 'S'), ('Rcpc', 'R'), ('Rncc', 'R'), ('C', 'C'), ('PB', 'P'), ('P', 'P'), ('I', 'A'), ('Q', 'W'), ('N', 'N')], default='Q', max_length=10),
```
jklugherz, line 16: what does the second value in these tuples represent?

hanars: the django display value, so if you use the `family.get_analysis_status_display()` function it will return it. But since we tend not to use those they're pretty irrelevent
https://docs.djangoproject.com/en/3.2/ref/models/fields/#choices

jklugherz (approved):

PR #3961, opened 2024-03-11
specify which notification failed when reporting notification error

hanars:
If a MME match comes in and matches many submissions, we send one email per submission to let the users know. However, if any of those emails fail to send we combine those failures into one error which lists all the results that matched but does not make clear which submissions were correctly notified and which were not

Example of unclear notification: https://console.cloud.google.com/errors/detail/CNLmz4aq-YiVjwE;time=P30D?project=seqr-project&utm_source=error-reporting-notification&utm_medium=slack&utm_content=resolved-error

### DIFF
--- a/matchmaker/views/external_api.py
+++ b/matchmaker/views/external_api.py
@@ -79,14 +79,7 @@ def mme_match_proxy(request, originating_node_name):
         patient_data=query_patient_data, origin_request_host=originating_node_name,
     )
 
-    try:
-        _generate_notification_for_incoming_match(results, incoming_query, originating_node_name, query_patient_data)
-    except Exception as e:
-        logger.error('Unable to create notification for incoming MME match request for {} matches{}: {}'.format(
-            len(results),
-            ' ({})'.format(', '.join(sorted([result.get('patient', {}).get('id') for result in results]))) if results else '',
-            str(e)),
-        )
+    _safe_generate_notification_for_incoming_match(results, incoming_query, originating_node_name, query_patient_data)
 
     return create_json_response({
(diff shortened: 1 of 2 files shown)


matchmaker/views/external_api.py
```diff
@@ -198,9 +197,7 @@ def _generate_notification_for_incoming_match(results, incoming_query, incoming_
             )
             email_message.send()
         except Exception as e:
-            email_errors.append(str(e))
-    if email_errors:
-        raise ValueError('\n'.join(email_errors))
+            logger.error(f'Unable to send notification email for incoming MME match with {submission_id}: {e}')
```
jklugherz, line 200: since we're no longer re-raising a `ValueError` and logging instead we may never reach the original unclear error [here](https://github.com/broadinstitute/seqr/pull/3961/files#diff-45104ab8be01f23b15bb10bfdd314898603426372f37d7c412a50c93f713017fR85), I'd get rid of the original error unless logging the patient ID in addition to submission ID is also helpful

hanars: patient ID is actually derived form submission ID, so agreed that thats not necessary anymore

hanars: however, we do want to keep they try/except because we want to be 100% sure that the end user gets a 200 response in this situation. But I can update that error to be clearer

hanars: okay I could not figure out a way to get any of the rest of the code to raise an exception, so in the end its fine to remove it

jklugherz (approved):

PR #3981, opened 2024-03-19
feat: Show more details on hover over data point of RNASeq Outliers plot

tommyli:
Hi,

We made some small improvements to the RNASeq Outliers plot so that each data point shows extra details on hover.  See screenshot below.

In order to show the gene symbols, it meant the `isSignificant` genes filter had to be removed from the `get_individual_rna_seq_data()` API call.  This increased the API call time by around 20% on our instance.  However, since this plot already has a long delay to get the rest of the data, we didn't think it would make the experience much worse.

ScreenShot 2024-03-19 at 1 12 17 PM

A question I had was any future plans around RNASeq in Seqr.  Will there be much more significant changes to how Seqr supports RNASeq?  We're asking because we have other improvements to this plot that have more involved changes and if this plot is not going to stay around then we may not bother.  The improvement is to add a new drop down for gene lists selection such that on drop down change, the data points in the selected gene list gets a darker shade.  Would this feature be useful for your team?

### DIFF
--- a/ui/pages/Project/components/RnaSeqOutliers.jsx
+++ b/ui/pages/Project/components/RnaSeqOutliers.jsx
@@ -52,12 +52,16 @@ class RnaSeqOutliersGraph extends React.PureComponent {
     const dataPoints = svg.append('g').selectAll('dot').data(dataArray).enter()
       .append('g')
 
+    console.debug('genesById', genesById)
+
     dataPoints.append('circle')
-      .attr('cx', d => x(d[xField]))
-      .attr('cy', d => y(d[yField]))
-      .attr('r', 3)
-      .style('fill', 'None')
+      .attr('cx', d => x(d.zScore))
+      .attr('cy', d => y(d.pValue))
+      .attr('r', d => (d.deltaPsi === undefined ? 3 : r(Math.abs(d.deltaPsi))))
+      .style('fill-opacity', '0%')
       .style('stroke', d => (d.isSignificant ? 'red' : 'lightgrey'))
+      .append('svg:title')
+      .text(d => `${genesById[d.geneId] ? genesById[d.geneId].geneSymbol : d.geneId}\nZ = ${d.zScore.toPrecision(2)}\n${'P-value = 10^-'}${(-Math.log10(d.pValue)).toPrecision(2)}`)
 
     dataPoints.append('text')
       .text(d => (d.isSignificant ? (genesById[d.geneId] || {}).geneSymbol : null))
--- a/ui/pages/Project/components/RnaSeqOutliers.jsx
+++ b/ui/pages/Project/components/RnaSeqOutliers.jsx
@@ -52,8 +52,6 @@ class RnaSeqOutliersGraph extends React.PureComponent {
     const dataPoints = svg.append('g').selectAll('dot').data(dataArray).enter()
       .append('g')
 
-    console.debug('genesById', genesById)
-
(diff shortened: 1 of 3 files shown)


hanars:
I will discuss with our team and get back to you!

ui/pages/Project/components/RnaSeqOutliers.jsx
```diff
@@ -50,12 +50,16 @@ class RnaSeqOutliersGraph extends React.PureComponent {
     const dataPoints = svg.append('g').selectAll('dot').data(dataArray).enter()
       .append('g')
 
+    console.debug('genesById', genesById)
```
hanars, line 53: This needs to be removed before merging to the main seqr branch

hanars (changes requested):
Discussed with our team and we agree the performance trade off is worth it for the enhanced functionality. we will be happy to upstream this once debug code has been removed!

> Will there be much more significant changes to how Seqr supports RNASeq?

We are currently changing the splice outlier code to work with FRASER2 instead of FRASER 1, but outside of that we have no planned RNA work, and no planned UI enhancements. This plot will be staying around, we have no plans to remove it, but we also have no plans to improve it either. For the example new feature you provided, our team feels like this is not high priority for us but if your team made it available we would probably want to adopt it into the main seqr branch

hanars:
We would be happy to integrate this branch once merge conflicts are resolved and debug code is removed!

tommyli:
I've rebased from latest dev and have updated the feat branch.  The debug statement is gone and there should be no merge conflicts going into the dev branch.

ui/pages/Project/components/RnaSeqOutliers.jsx
```diff
@@ -53,11 +53,13 @@ class RnaSeqOutliersGraph extends React.PureComponent {
       .append('g')
 
     dataPoints.append('circle')
-      .attr('cx', d => x(d[xField]))
-      .attr('cy', d => y(d[yField]))
-      .attr('r', 3)
-      .style('fill', 'None')
+      .attr('cx', d => x(d.zScore))
+      .attr('cy', d => y(d.pValue))
+      .attr('r', d => (d.deltaPsi === undefined ? 3 : r(Math.abs(d.deltaPsi))))
```
hanars, line 58: Unfortunately this is actually a merge conflict and it is changing the behavior (using hardcoded `zScore` and `pValue` instead of `xField` and `yField`, checking `deltaPsi` which is no longer a supported field)

ui/pages/Project/components/RnaSeqOutliers.jsx
```diff
@@ -53,11 +53,13 @@ class RnaSeqOutliersGraph extends React.PureComponent {
...
+      .style('fill-opacity', '0%')
       .style('stroke', d => (d.isSignificant ? 'red' : 'lightgrey'))
+      .append('svg:title')
+      .text(d => `${genesById[d.geneId] ? genesById[d.geneId].geneSymbol : d.geneId}\nZ = ${d.zScore.toPrecision(2)}\n${'P-value = 10^-'}${(-Math.log10(d.pValue)).toPrecision(2)}`)
```
hanars, line 62: this will need to use the `xField` and `yField` logic instead of hardcoding fields

hanars (changes requested):

tommyli:
I see.  I'm gonna close this PR and raise a new one in the future after we do another round of merging.

PR #3982, opened 2024-03-19
gene_known_for_phenotype from tags

hanars:


jklugherz (approved):

jklugherz (approved):
what was the improvement here? it looks like the logic for assigning the `gene_known_for_phenotype` tag is almost same, just in a different spot, but I know there's something else in here

hanars:
The logic has changed, its now determined based on variant tags instead of by the family omim_number/  mondo_id. It was being set as 
```
Case(When(
            Q(family__post_discovery_omim_numbers__len=0, family__mondo_id__isnull=True),
            then=Value('Candidate')), default=Value('Known')
        )
```
and is now being set as ` 'Known' if 'Known gene for phenotype' in variant.tags else 'Candidate',`

PR #3990, opened 2024-03-21
Populate from gregor metadata

hanars:


seqr/views/utils/project_context_utils.py
```diff
@@ -162,3 +163,42 @@ def add_project_tag_types(projects_by_guid):
             ),
             'variantFunctionalTagTypes': VariantFunctionalData.FUNCTIONAL_DATA_TAG_TYPES,
         })
+        if add_counts:
+            family_counts.update(_add_tag_type_counts(project_guid, project_json['variantTagTypes']))
+
+    return family_counts
+
+
+def _add_tag_type_counts(project_guid, project_variant_tags):
```
hanars, line 172: copy-pasted from proect_api, does not need review

seqr/views/utils/variant_utils.py
```diff
@@ -124,11 +126,95 @@ def parse_saved_variant_json(variant_json, family):
...
+        if updated_tag:
+            update_tags.append(updated_tag)
+        else:
+            num_new += 1
+
+    VariantTag.bulk_update_models(user, update_tags, ['metadata'])
+    return num_new, len(update_tags)
+
+
+def _set_updated_tags(key: tuple[int, str], metadata: dict[str, dict], support_var_ids: list[str],
```
hanars, line 181: copy-paste from saummary_data_api, does not need review

seqr/views/utils/variant_utils.py
```diff
@@ -124,11 +126,95 @@ def parse_saved_variant_json(variant_json, family):
...
         'ref': ref,
         'alt': alt,
-        'family': family,
-        'variant_id': variant_json['variantId']
+        'family_id': family_id,
+        'variant_id': variant_json.get('variantId', variant_id),
     }, update_json
 
 
+def bulk_create_tagged_variants(family_variant_data, tag_name, get_metadata, user, project=None, load_new_variant_data=None):
```
hanars, line 134: adapted from summary_data_api, needs less thorough review

seqr/utils/file_utils.py
```diff
@@ -12,12 +12,12 @@ def run_command(command, user=None, pipe_errors=False):
...
 
 
-def _run_gsutil_command(command, gs_path, gunzip=False, user=None, pipe_errors=False):
+def _run_gsutil_command(command, gs_path, gunzip=False, user=None, pipe_errors=False, no_project=False):
     if not is_google_bucket_file_path(gs_path):
         raise Exception('A Google Storage path is expected.')
 
     #  Anvil buckets are requester-pays and we bill them to the anvil project
-    google_project = get_google_project(gs_path)
+    google_project = get_google_project(gs_path) if not no_project else None
```
jklugherz, line 20: why aren't we including a billing project when importing the gregor metadata file?

ui/pages/Project/reducers.js
```diff
@@ -167,6 +167,8 @@ export const updateIndividuals = values => (dispatch, getState) => {
     action = `save_individuals_table/${values.uploadedFileId}`
   } else if (values.delete) {
     action = 'delete_individuals'
+  } else if (values.workspaceName) {
```
jklugherz, line 170: Can you make `uploadedFileId`, `delete`, and `workspaceName` constants and use them wherever they're being defined originally? I know that `workspaceName` comes from a field on the uploaded Gregor metadata form but I haven't figured out yet where `uploadedFileId` comes from.

seqr/views/apis/individual_api.py
```diff
@@ -806,6 +815,170 @@ def save_individuals_metadata_table_handler(request, project_guid, upload_file_i
...
+
+    response_json['projectsByGuid'] = {project_guid: {}}
+    response_json['familyTagTypeCounts'] = add_project_tag_types(response_json['projectsByGuid'], add_counts=True)
+
+    response_json['importStats'] = {'gregorMetadata': {'info': info, 'warnings': warnings}}
+    return create_json_response(response_json)
+
+
+def _iter_metadata_table(file_path, table_name, user, filter_row):
+    file_name = f'{file_path}/{table_name}.tsv'
```
jklugherz, line 947: Is file the same as the one generated by the gregor report page in seqr?

seqr/views/apis/individual_api_tests.py
```diff
@@ -1087,6 +1342,10 @@ class LocalIndividualAPITest(AuthenticationTestCase, IndividualAPITest):
     fixtures = ['users', '1kg_project', 'reference_data']
     HAS_EXTERNAL_PROJECT_ACCESS = False
 
+    def test_import_gregor_metadata(self, *args):
+        # Importing gregor metadata does not work in local environment
```
jklugherz, line 1346: Is it too much to mock the return value of `file_iter` to yield rows of 'mock' tsv files? I'm curious about what the files look like 😭

seqr/utils/file_utils.py
hanars: because that workspace is a special snowflake somehow and including the billing project causes it to fail *shrug*

ui/pages/Project/reducers.js
hanars: I think fully refactoring a helper function used in several different place plus updating a whole lot of forms unrelated to this change to use those constants is outside the scope of this PR,  but I am happy to add comments to make it clearer

seqr/views/apis/individual_api.py
hanars: Kind of. All the centers submit the same reports and then the DCC makes them into one big report. But in general, we want the schema for the report we generate to match the schema for the metadata we import here

seqr/views/apis/individual_api_tests.py
hanars: So the way this test file is structured it has a base `IndividualAPITest` object thats not actually a runnable test case, and then it has `LocalIndividualAPITest` and `AnvilIndividualAPITest` that both inherit from that object plus from a different auth base class so that every test in the `IndividualAPITest` is by default run under both local install and anvil auth conditions. However, this test fails under local conditions because you can't import metadata from anvil buckets without anvil, so for the `LocalIndividualAPITest` we override this function so it doesn't run

ui/pages/Project/reducers.js
jklugherz: I agree that's out of scope, comments about where those fields come from would help too

seqr/views/apis/individual_api_tests.py
jklugherz: oh I missed that `test_import_gregor_metadata` is in a different `*APITest` class. thanks for explaining that

jklugherz (approved):

PR #3994, opened 2024-03-27
Fraser2

hanars:


ui/pages/Project/components/RnaSeqOutliers.jsx
```diff
@@ -33,16 +36,15 @@ class RnaSeqOutliersGraph extends React.PureComponent {
   }
 
   initPlot = () => {
-    const { data: dataArray, genesById } = this.props
+    const { data: dataArray, genesById, xField, yField = 'pValue' } = this.props
```
jklugherz, line 39: nit, but can you make 'pValue' a constant?
also, why doesn't `xField` get a default?

ui/pages/Project/components/RnaSeqOutliers.jsx
```diff
@@ -51,20 +53,20 @@ class RnaSeqOutliersGraph extends React.PureComponent {
       .append('g')
 
     dataPoints.append('circle')
-      .attr('cx', d => x(d.zScore))
-      .attr('cy', d => y(d.pValue))
-      .attr('r', d => (d.deltaPsi === undefined ? 3 : r(Math.abs(d.deltaPsi))))
+      .attr('cx', d => x(d[xField]))
+      .attr('cy', d => y(d[yField]))
+      .attr('r', 3)
```
jklugherz, line 58: what does 3 mean here?

hanars: The size for the dot. The size was previously `d.deltaPsi === undefined ? 3 : r(Math.abs(d.deltaPsi)))` so for the cases where we had `deltaPsi` the size was variable based on the data and for the places we didn't it was always `3`. Now that there is no data where we want the dot size to be dynamic, its just always the default, or `3`

ui/pages/Project/components/RnaSeqOutliers.jsx
jklugherz: I edited my comment to remove the part of the question about `xField` because a second later I saw it defined in `RnaSeqResultPage`, I think you made the right call to make `yField` configurable with a default

ui/shared/components/table/RnaSeqJunctionOutliersTable.jsx
```diff
@@ -72,7 +72,26 @@ const OTHER_SPLICE_COLUMNS = [
     }
   )),
   {
-    name: 'rareDiseaseSamplesWithJunction',
+    name: 'counts',
```
jklugherz, line 75: not a blocker for this PR, but would you be open to including screenshots/recordings of screen interactions in the PR description when there are UI changes? I'm curious to see what the changes look like but I'll likely only interact with these changes in seqr if I remember to after this gets merged

jklugherz (approved):

ui/shared/components/table/RnaSeqJunctionOutliersTable.jsx
hanars: yeah that seems very reasonable!

PR #3995, opened 2024-03-28
Update airflow API auth

bpblanken:
The auth pattern has changed with the upgrade, docs [here](https://cloud.google.com/composer/docs/composer-2/access-airflow-api)


bpblanken:
(working on fixing the tests 🤦 )

seqr/views/utils/airflow_utils.py
```diff
@@ -172,13 +171,12 @@ def _trigger_dag(dag_id):
...
     # account.
-    google_open_id_connect_token = id_token.fetch_id_token(Request(), AIRFLOW_API_AUDIENCE)
-
     webserver_url = f'{AIRFLOW_WEBSERVER_URL}/api/v1/{endpoint}'
-    resp = requests.request(
-        method, webserver_url,
-        headers={'Authorization': 'Bearer {}'.format(
-            google_open_id_connect_token)}, **kwargs)
-
+    authed_session = AuthorizedSession(SERVICE_ACCOUNT_CREDENTIALS)
```
hanars, line 175: I don't think wehould assume that the service account we use for core seqr app functionality is the same account that is authorized to access composer. It in fact seems like a reversion that we are switching from having a designated account for airflow to not having one, since we are trying to move toward having more service accoutns that are all well scoped for their functionality

seqr/views/utils/test_utils.py
```diff
@@ -671,28 +668,23 @@ def _assert_airflow_calls(self, dag_name, dag_variables, call_count, secondary_d
...
         if call_count > 5:
             self.assertEqual(responses.calls[offset+4].request.url, f'{dag_url}/tasks')
             self.assertEqual(responses.calls[offset+4].request.method, 'GET')
-            self.assertEqual(responses.calls[offset+4].request.headers['Authorization'], self.auth_header)
 
         # trigger dag
         self.assertEqual(responses.calls[offset+call_cnt].request.url, f'{dag_url}/dagRuns')
         self.assertEqual(responses.calls[offset+call_cnt].request.method, 'POST')
         self.assertDictEqual(json.loads(responses.calls[offset+call_cnt].request.body), {})
-        self.assertEqual(responses.calls[offset+call_cnt].request.headers['Authorization'], self.auth_header)
```
hanars, line 695: These requests do not have this header anymore, but they are being run from an authenticated session which means *something* is different about them than if they weren't, and we should be testing that if thats at all possible

seqr/views/utils/airflow_utils.py
bpblanken: 👍 fair point.  My understanding of what's going on is a little shaky, but I don't know if it's a regression... just different.  

The audience token we've been using is a "client ID" used by the "Identity Aware Proxy", which I was eventually able to track down by running [this script](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/composer/rest/get_client_id.py).  `fetch_id_token` then uses that token in combination with the service account of the gce environment to auth itself.  At some point we manually added the default seqr service account to the set of authorized users in airflow itself (which I have also done in in the new environment.. docs pending).  

Composer 2 appears to not need the "client ID" because "the Airflow Webserver is not in the tenant project", so we're auth-ing just with the service account credentials.  

it does seem reasonable to use a better-scoped service account for this, but I think that's going to be a challenge.

seqr/views/utils/test_utils.py
bpblanken: I think it's baked into whatever `AuthorizedSession` is doing under the hood... I do think worth investigating further though.

seqr/views/utils/airflow_utils.py
hanars: Okay, if composer 1 was using our default service account than there is no harm continuing that behavior with composer 2 for now and switching to finer grained auth is out of scope for this change. I assumed since we had a different auth mechanism we were using a different auth account, but I suppose that was too much to hope for *sigh*

hanars (approved):

PR #3999, opened 2024-04-01
airflow api auth -> master

bpblanken:


hanars:
We do not do cherry picking directly to master as part of our standard deployment process. Is there a reason to do this and not just cut a normal seqr release with everything on dev and go through our standard release process?

bpblanken:
ah.... I thought this was easier to reason about (for me at least) because I don't have the ability to test or validate the other work in progress.  No prob to wait though.

PR #4000, opened 2024-04-01
Multi project multi way zip join

hanars:
For searching across many projects, construct a single merged entries table that is then filtered, instead of sequentially filtering each project table and then merging them together. This allows us to take advantage of hails parallelization and allows this query to complete in the allotted time


hail_search/queries/base.py
```diff
@@ -278,39 +278,48 @@ def _parse_sample_data(self, sample_data):
...
+    def _load_filtered_project_hts(self, project_samples, skip_all_missing=False, n_partitions=MAX_PARTITIONS, **kwargs):
+        if len(project_samples) == 1:
+            project_guid = list(project_samples.keys())[0]
+            project_ht = self._read_table(f'projects/{project_guid}.ht', use_ssd_dir=True)
+            return self._filter_entries_table(project_ht, project_samples[project_guid], **kwargs)
+
+        # Need to chunk tables or else evaluating table globals throws LineTooLong exception
+        # However, minimizing number of chunks minimizes number of aggregations/ evals and improves performance
+        # Adapted from https://discuss.hail.is/t/importing-many-sample-specific-vcfs/2002/8
+        chunk_size = 64
```
jklugherz, line 290: Why did you choose a chunk size of 64 and 12 max partitions?

hail_search/queries/base.py
```diff
@@ -321,60 +330,50 @@ def import_filtered_table(self, project_samples, num_families, intervals=None, *
...
-            return families_ht, False
-
-        families_ht = families_ht.join(project_ht, how='outer')
-        families_ht = families_ht.select_globals(
-            family_guids=families_ht.family_guids.extend(families_ht.family_guids_1)
+    @staticmethod
+    def _merge_project_hts(project_hts, n_partitions, include_all_globals=False):
+        if not project_hts:
+            return None
+        ht = hl.Table.multi_way_zip_join(project_hts, 'project_entries', 'project_globals')
```
jklugherz, line 358: I'm not positive on this, but you might want to ensure that all tables in `project_hts` have the same # of partitions for `multi_way_zip_join` to work the best it can - you could pass some constant value as `n_partitions` to `_read_table`

hail_search/queries/base.py
hanars: chunk size of 64 was recommended in the linked hail discussion, and I did some testing with larger sizes and they did not work. 12 was determined experimentally and the ticket where the testing was done is linked in the comments: https://github.com/broadinstitute/seqr/blob/master/hail_search/queries/base.py#L22

hail_search/queries/base.py
hanars: All the tables are read in with a specified interval, and there is no way to specify the interval and the number of partitions, hail throws an error. But they generally are all 1 partition because the interval is pretty small, even for the largest of tables

hail_search/queries/base.py
jklugherz: nice, thank you

jklugherz (approved):

PR #4006, opened 2024-04-04
Lookup all projects mvp

hanars:


seqr/views/apis/variant_search_api.py
```diff
@@ -534,19 +534,78 @@ def _flatten_variants(variants):
...
+
     return create_json_response(response)
+
+
+def _update_lookup_variant(variant, response):
+    individual_guid_map = {
+        (i['familyGuid'], i['individualId']): i['individualGuid'] for i in response['individualsByGuid'].values()
+    }
+
+    no_access_families = set(variant['familyGenotypes']) - set(variant['familyGuids'])
```
jklugherz, line 575: what's the difference between what `variant['familyGenotypes']` and `variant['familyGuids']` are representing?

hail_search/queries/mito.py
```diff
@@ -297,3 +302,37 @@ def _omim_sort(cls, r, omim_gene_set):
     @classmethod
     def _gene_rank_sort(cls, r, gene_ranks):
         return [gene_ranks.get(r.selected_transcript.gene_id)] + super()._gene_rank_sort(r, gene_ranks)
+
+    def _add_project_lookup_data(self, ht, annotation_fields, *args, **kwargs):
+        # Get all the project-families for the looked up variant formatted as a dict of dicts:
+        # {<project_guid>: {<family_guid>: True, <family_guid_2>: True}, <project_guid_2>: ...}
+        lookup_ht = self._read_table('lookup.ht', use_ssd_dir=True, skip_missing_field='project_stats')
+        if lookup_ht is None:
+            raise HTTPNotFound()
```
jklugherz, line 311: This might make more sense as a 50x error because if there's no lookup table something's wrong on our end, not the client's request.

hanars: no, so this is the expected behavior if someone provides a valid variant that happens to be absent from our annotations/ lookup table. That case occurs for any variant that was never seen in any seqr sample, and in fact our reference data table is like 10x bigger than our annotations table because there are so many variants that fall into that category. Therefore, a 404 is the appropriate response here

jklugherz (approved):

PR #4009, opened 2024-04-05
Report sex/relatedness failures

hanars:


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -107,6 +109,28 @@ def handle(self, *args, **options):
             updated_families.update(project_families)
             updated_project_families.append((project.id, project.name, project_families))
 
+        # Send failure notifications
+        failure_checks = ['relatedness_check', 'sex_check']
```
bpblanken, line 113: Is it the worst idea to also include `missing_samples` here?  Too noisy?

hanars: not the worst idea, I just assumed that the pipeline was already reporting those?

bpblanken: not to slack... just the dataproc logs or this file.

bpblanken (approved):

PR #4013, opened 2024-04-08
return NEAREST_TSS SVs in gene search

hanars:


jklugherz (commented):
what does `NEAREST_TSS` mean and why is it important for SV search?

hanars:
> what does `NEAREST_TSS` mean and why is it important for SV search?

Thats actually a question for @lynnpais, I just have a note on the ticket saying we decided at an analyst meeting that we did want these coming back in search so I implemented this fix but I don't remember why thats what we decided

bpblanken (approved):

lynnpais:
@jklugherz it tells us if the variant is near the transcription start site of a gene and hence if it might be affecting the function of that gene.

PR #4014, opened 2024-04-08
restrict anvil loading when loading pending

hanars:


seqr/views/apis/anvil_workspace_api.py
```diff
@@ -128,6 +128,15 @@ def validate_anvil_vcf(request, namespace, name, workspace_meta):
         error = 'Field(s) "{}" are required'.format(', '.join(missing_fields))
         return create_json_response({'error': error}, status=400, reason=error)
 
+    # Validate no pending loading projects
+    pending_project = Project.objects.filter(
```
jklugherz, line 132: For how long usually is a project in this pending state, awaiting loading?

hanars: so our communicated turnaround time is up to a week, but now that we've fully automated it its generally a day or two

jklugherz (approved):
I'm cool with this as long as @lynnpais and @bpblanken are too

bpblanken (approved):

PR #4029, opened 2024-04-16
Environment lock

nevoodoo:
## Description
In a bid to introduce better security permissions, we have replaced the use of SA keys to workload identity providers. This change will also implement the use of environments to ensure deployments are going through the right channels.

## Changes
- [x]  Updated the workflow to use WIP over SA keys
- [x]  Added `environment` attribute


nevoodoo:
Sorry wrong repo :(

PR #4030, opened 2024-04-16
clinvar submitter

jklugherz:
Screenshot 2024-04-17 at 10 50 14 AM


hanars:
@lynnpais I know this is the UI we put in the ticket, but are we sure we want it to be a separate tag, and not part of the existing clinvar tag? I feel like its not clear form this UI that this specific clinvar submission is the one the Broad submitted, but maybe thats by design

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -96,6 +110,7 @@ const Pathogenicity = React.memo(({ variant, showHgmd }) => {
       href: clinvarUrl(clinvar),
       goldStars: clinvar.goldStars,
       popup: clinvar.version && `Last Updated: ${new Date(clinvar.version).toLocaleDateString()}`,
```
hanars, line 112: conditions should be shown on hover for the main label, not the Broad submission label. Sorry for not making that clear in the ticket

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -65,6 +65,16 @@ PathogenicityLink.propTypes = {
   popup: PropTypes.string,
 }
 
+const ClinvarSubmissions = React.memo(({ submissions }) => submissions != null &&
```
hanars, line 68: rather than iterating the list and returning null for all but one value, it might be neater to instead just get the broad submission and show the label based on that.
So in the main component something like 
```
const broadSubmission = submissions?.find(({ submitter }) => submitter === 'Broad...')
...
{broadSubmission && <BroadSubmissionLabel />}
```

ui/shared/components/panel/variants/Pathogenicity.jsx
```diff
@@ -117,6 +132,14 @@ const Pathogenicity = React.memo(({ variant, showHgmd }) => {
       <b>{`${title}:`}</b>
       <HorizontalSpacer width={5} />
       <PathogenicityLink {...linkProps} />
+      {
+        title === 'ClinVar' && (
```
hanars, line 136: I think making this conditional on a magic string is not great. I think it would be best to do it based on the presence of `linkProps.submissions` but if we really want it to be based on title we should at least make that a constant

hanars (changes requested):
So one thing about this change is that it needed no hail backend changes because we are just passing this through. Thats great, but I do worry that down the line not testing this at all could lead to issues of us messing up either seqr or the hail backend to not continue to return these. I think we should update the annotations fixture table (38 SNV_INDEL is probably sufficient) to include submitters and conditions and update our unit tests accordingly

lynnpais:
It's fine to have it part of the existing ClinVar tag.

jklugherz:
@lynnpais @hanars can one of you provide a quick sketch of what you mean by part of the existing clinvar tag? 

I'm not sure if that means updating the label or just adding to the hover.

ui/shared/components/panel/variants/Pathogenicity.jsx
jklugherz: there's no reason not to use `linkProps.submissions`, I'll do that instead

hanars:
Maybe something like this
Screenshot 2024-04-16 at 4 25 07 PM

jklugherz:
image
@hanars how does this look for the condition hover?

(theoretically there can be multiple submissions for different conditions by one submitter, however I'm not sure if that happens irl)

hanars:
Since we are not listing all the submitters and are not actually associating conditions with submitters in the backend, I would not worry about that. We should always show conditions in the hover, regardless of whether it is Broad submitted or not. The UI is fine, although a bulleted list might be more readable

jklugherz:
@hanars Sometimes a condition is "not specified" or "not provided", and I didn't filter those out in the pipeline. Should I explicitly ignore those in the UI?

hail_search/test_utils.py
```diff
@@ -256,7 +270,16 @@
...
-    'clinvar': None,
+    'clinvar': {
+        'alleleId': None,
+        'conflictingPathogenicities': None,
+        'goldStars': None,
+        'submitters': None,
+        'conditions': None,
+        'pathogenicity': None,
+        'assertions': None,
+        'version': '2024-02-21',
```
jklugherz, line 281: I'm not sure why the entire clinvar dict was not returned in search before I changed the annotations table schema to add submitters/conditions

hanars: Did you possibly take some of the values that were fully missing and change them to be defined but with all the subfields missing?

jklugherz: this was my script to update one row: 
```
ht = ht.annotate(
    clinvar=hl.Struct(
        **dict(ht.clinvar), 
        submitters=hl.if_else(
            (ht.locus == hl.locus('chr1', 38724419, reference_genome='GRCh38')),
            ['Broad Center for Mendelian Genomics, Broad Institute of MIT and Harvard', 'Illumina Laboratory Services, Illumina', 'Blueprint Genetics','GenomeConnect, ClinGen'],
            hl.missing(hl.tarray(hl.tstr))
        ), 
        conditions=hl.if_else(
            (ht.locus == hl.locus('chr1', 38724419, reference_genome='GRCh38')),
            ['ABCA4-Related Disorders','Severe early-childhood-onset retinal dystrophy','not specified','not provided'],
            hl.missing(hl.tarray(hl.tstr))
        )
    )
```
it might be the `**dict(ht.clinvar)` that's wrong or the `hl.missing(hl.tarray(hl.tstr))`

hanars: yeah so assigning it as a new struct is probably the issue. I would try just annotating the struct, instead of replacing it. If you annotate a missing struct it should continue to be missing
```
ht.annotate(clinvar=ht.clivar.annotate(
  submitters=hl.if_else(
            (ht.locus == hl.locus('chr1', 38724419, reference_genome='GRCh38')),
            ['Broad Center for Mendelian Genomics, Broad Institute of MIT and Harvard', 'Illumina Laboratory Services, Illumina', 'Blueprint Genetics','GenomeConnect, ClinGen'],
            hl.missing(hl.tarray(hl.tstr))
        ),
...
))
```

jklugherz: thank you, this worked!

PR #4037, opened 2024-04-17
UI code cleanup for dyanmic analysis groups

hanars:
We need to repurpose some of the family table filtering logic in the seqr UI in order to support the new dynamic analysis group functionality. This moves the functionality that will need to be shared into a shared selector and does soem overall cleanup to how the filtering is performed to make it possible to split it out


jklugherz (approved):

bpblanken (approved):
not even a nit!

PR #4042, opened 2024-04-19
Dynamic analysis group

hanars:


ui/pages/Search/VariantSearch.jsx
```diff
@@ -11,8 +11,8 @@ const RESULTS_PATH = 'results/:searchHash'
 const SINGLE_VARIANT_RESULTS_PATH = 'variant/:variantId/family/:familyGuid'
 
 const SEARCH_FORM_PAGES = [
+  'project/:projectGuid/analysis_group/:analysisGroupGuid',
```
hanars, line 14: dynamic analysis groups are not necessarily associated with a specific project which makes the URL for this search page ambiguous as it could be for that group in any project. Therefore, rework that URL to require the relevant project

seqr/migrations/0063_dynamicanalysisgroup.py
```diff
@@ -0,0 +1,33 @@
...
+                ('guid', models.CharField(db_index=True, max_length=30, unique=True)),
+                ('created_date', models.DateTimeField(db_index=True, default=django.utils.timezone.now)),
+                ('last_modified_date', models.DateTimeField(blank=True, db_index=True, null=True)),
+                ('name', models.TextField()),
+                ('criteria', models.JSONField()),
+                ('created_by', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to=settings.AUTH_USER_MODEL)),
+                ('project', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='seqr.project')),
+            ],
+            options={
+                'unique_together': {('project', 'name')},
```
jklugherz, line 30: how does this constraint behave if project is null?

ui/pages/Search/constants.js
```diff
@@ -5,7 +5,7 @@ import {
 
 export const getSelectedAnalysisGroups = (analysisGroupsByGuid, familyGuids) => Object.values(
   analysisGroupsByGuid,
-).filter(group => group.familyGuids.every(familyGuid => familyGuids.includes(familyGuid)))
+).filter(group => group.familyGuids?.every(familyGuid => familyGuids.includes(familyGuid)))
```
jklugherz, line 8: nitpick, but this function `getSelectedAnalysisGroups` is used only once now in `PageHeader.jsx`, can you move it there?

seqr/models.py
```diff
@@ -1040,6 +1040,23 @@ class Meta:
         json_fields = ['guid', 'name', 'description']
 
 
+class DynamicAnalysisGroup(ModelWithGUID):
+    project = models.ForeignKey('Project', on_delete=models.CASCADE, null=True, blank=True)
+    name = models.TextField()
+    criteria = JSONField()
+
+    def __unicode__(self):
```
bpblanken, line 1048: Oooh, is this still needed in whatever Django version we're on?  I think this can be `__str__` in python3 if it's not a Django thing.

seqr/migrations/0063_dynamicanalysisgroup.py
hanars: It means you can only have a name associated with one list with no project. So fi there is already a list of that name associated with a different project its fine, but if it is associated with no project it would not let you make a duplicate

seqr/models.py
hanars: its a Django thing 😿

hanars: At some point we should probably do a django upgrade and clean up this sort of thing all over the repo

jklugherz (approved):

PR #4044, opened 2024-04-23
Allow optional "dataset_version" in GitHub action.

bpblanken:
PR into `master` on purpose.

EDIT: PR no longer into `master`.


hanars:
If we are going to PR direct to master we need to include the [release checklist ](https://github.com/broadinstitute/seqr-private/blob/master/release_pr_checklist.md) and it also needs a PR to dev to make sure it doesn't get lost. But as always, wouldn't it just be easier to take this to dev and then do a release to prod once its in?

bpblanken:
yeah, this one was special because I thought the action needed to be on `master` for it to be runnable and testable.  It turns out that I was wrong!  The action just needs to have been run a single time with its current name on `master`, but then we can run it on the dev branch with the changes.  I will fix and also fix our hardcoded reference [here](https://github.com/broadinstitute/seqr-pipeline-airflow/blob/main/dags/dependencies/operators.py#L50)

hanars (approved):

PR #4047, opened 2024-04-24
panelapp use genes api endpoint

jklugherz:
let's try calling the /genes endpoint instead of /panels/panel_id/genes for every panel_id


panelapp/panelapp_utils.py
```diff
@@ -105,16 +109,22 @@ def _get_all_panels(panels_url, all_results):
...
-    resp = requests.get(panel_genes_url)
+def _get_all_genes(genes_url: str, results_by_panel_id: dict):
+    resp = requests.get(genes_url, timeout=REQUEST_TIMEOUT_S)
     resp_json = resp.json()
-    all_results += resp_json.get('results', [])
+
+    for result in resp_json.get('results', []):
+        if result.get('panel'):
+            panel_id = result['panel']['id']
+            if panel_id not in results_by_panel_id:
```
hanars, line 119: you do not have to change this, but we use `defaultdict` in a lot of places in the seqr codebase to avoid having to handle this edge case
```
results_by_panel_id = defaultdict(list)
...
results_by_panel_id[panel_id].append(result). // no need to check if key is present
```

hanars (commented):
Awesome that this works! Did you do any sort testing on the real data to see if the results changed, or were you pretty confident that  it will work as is?

panelapp/panelapp_utils.py
jklugherz: I updated the code to use `defaultdict`!

hanars (approved):

jklugherz:
@hanars I spot checked the list of genes between the old endpoint and the new endpoint for a few genes and didn't see any differences, so I'm confident that that the data is the same.

PR #4051, opened 2024-04-26
UI snv indel loading

hanars:
Ads the option to load SNV/INDEL data from the UI, and adds a check with airtable to only load projects that are ready for loading in airtable, and only load the subset of families that are actually updated. When this goes live it should work redundantly with the current airtable checks in airflow (so no issues should arise) and then after this is live those can be safely removed

Project dropdown UI:
Screenshot 2024-04-26 at 5 05 18 PM

### DIFF
--- a/seqr/views/apis/data_manager_api.py
+++ b/seqr/views/apis/data_manager_api.py
@@ -446,14 +446,14 @@ def get_loaded_projects(request, sample_type, dataset_type):
 
     if project_samples:
         for project in projects:
-            project['sampleIds'] = project_samples[project['projectGuid']]
+            project['sampleIds'] = sorted(project_samples[project['projectGuid']])
 
     return create_json_response({'projects': list(projects)})
 
 
 def _fetch_airtable_loadable_project_samples(user):
     pdos = AirtableSession(user).fetch_records(
-        'PDO', fields=['PassingCollaboratorSampleIDs', 'SeqrIDs', 'PDOName', 'PDOStatus', 'SeqrProjectURL'],
+        'PDO', fields=['PassingCollaboratorSampleIDs', 'SeqrIDs', 'SeqrProjectURL'],
         or_filters={'PDOStatus': LOADABLE_PDO_STATUSES}
     )
     project_samples = defaultdict(set)
--- a/seqr/views/apis/data_manager_api.py
+++ b/seqr/views/apis/data_manager_api.py
@@ -436,6 +436,7 @@ def get_loaded_projects(request, sample_type, dataset_type):
         project_samples = _fetch_airtable_loadable_project_samples(request.user)
         projects = projects.filter(guid__in=project_samples.keys())
         exclude_sample_type = Sample.SAMPLE_TYPE_WES if sample_type == Sample.SAMPLE_TYPE_WGS else Sample.SAMPLE_TYPE_WGS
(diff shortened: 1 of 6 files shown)


seqr/views/apis/data_manager_api.py
```diff
@@ -424,31 +430,61 @@ def validate_callset(request):
...
-    ).distinct().order_by('name').values('name', projectGuid=F('guid'), dataTypeLastLoaded=Max(
+    projects = get_internal_projects().filter(is_demo=False)
+    project_samples = None
+    if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS:
+        project_samples = _fetch_airtable_loadable_project_samples(request.user)
+        projects = projects.filter(guid__in=project_samples.keys())
+        exclude_sample_type = Sample.SAMPLE_TYPE_WES if sample_type == Sample.SAMPLE_TYPE_WGS else Sample.SAMPLE_TYPE_WGS
+        projects = projects.exclude(family__individual__sample__sample_type=exclude_sample_type)
+    else:
+        projects = projects.filter(family__individual__sample__sample_type=sample_type)
```
jklugherz, line 441:
```suggestion
    if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS:
        project_samples = _fetch_airtable_loadable_project_samples(request.user)
        projects = projects.filter(guid__in=project_samples.keys())
       
   projects = projects.filter(family__individual__sample__sample_type=sample_type)
```
This jumped out to me as a possible simplification but I bet using `exclude` here was intentional - why? SNV_INDEL projects should also return RNA samples?

hanars: so the difference is that explicitly filtering for the sample type only includes projects that already have data loaded for that sample type. Excluding the opposite sample type includes projects that are loaded for the sample type AND projects that have no loaded data at all. For SV and MITO we only want to be able to add to projects that have existing data so they use the `filter` logic, but for SNV_INDEL data at some point we do need to load it for the first time, and so we need to instead use this `exclude` logic

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1309,27 +1351,43 @@ def test_validate_callset(self, mock_subprocess):
...
+        snv_indel_url = url.replace('SV', 'SNV_INDEL')
+        response = self.client.get(snv_indel_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertDictEqual(response.json(), {'projects': [EMPTY_PROJECT_OPTION, PROJECT_SAMPLES_OPTION]})
+        self.assert_expected_airtable_call(
+            call_index=0, filter_formula="OR(PDOStatus='Methods (Loading)',PDOStatus='On hold for phenotips, but ready to load')",
+            fields=['PassingCollaboratorSampleIDs', 'SeqrIDs', 'SeqrProjectURL'],
+        )
+
+        # test projects with no data loaded are returned for any sample type
```
hanars, line 1386: @jklugherz this tests the `exclude` behavior you were asking about - changing the logic the way you suggested would cause the empty project to not be returned here

seqr/views/apis/data_manager_api.py
jklugherz: thanks for the explanation. can you add this as a comment?

jklugherz (approved):

PR #4052, opened 2024-04-29
add gene field to manual sv form

jklugherz:
Screenshot 2024-04-29 at 10 37 35 AM

### DIFF
--- a/ui/pages/Project/components/CreateVariantButton.jsx
+++ b/ui/pages/Project/components/CreateVariantButton.jsx
@@ -115,6 +115,15 @@ const POS_FIELD = {
 const START_FIELD = { name: 'pos', label: 'Start Position', ...POS_FIELD }
 const END_FIELD = { name: 'end', label: 'Stop Position', ...POS_FIELD }
 
+const GENE_FIELD = {
+  name: GENE_ID_FIELD_NAME,
+  label: 'Gene',
+  control: AwesomeBarFormInput,
+  categories: ['genes'],
+  fluid: true,
+  placeholder: 'Search for gene',
+}
+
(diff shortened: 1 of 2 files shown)


ui/pages/Project/components/CreateVariantButton.jsx
```diff
@@ -205,7 +206,7 @@ const SV_FIELDS = [
       max: 12,
     },
   },
-].map(formatField)
+].map(formatField).map(field => (field.validate ? { ...field, label: `${field.label}*` } : field))
```
hanars, line 209: this looks like its copy/pasted from above so you could move it into the shared `formatField` helper

ui/pages/Project/components/CreateVariantButton.jsx
```diff
@@ -139,24 +148,21 @@ const validateHasTranscriptId = (value, allValues, props, name) => {
...
 }
 
-const formatField = field => ({ inline: true, width: 16, ...field })
+const formatField = (field) => {
+  let formattedField = { inline: true, width: 16, ...field }
+  if (field.validate && field.validate !== validateHasTranscriptId) {
+    formattedField = { ...formattedField, label: `${field.label}*` }
+  }
+  return formattedField
+}
```
hanars, line 157: you can use a `const` here because javascript objects are mutable. A cleaner way to do it might be
```suggestion
const formatField = (field) => {
  const formattedField = { inline: true, width: 16, ...field }
  if (field.validate && field.validate !== validateHasTranscriptId) {
    formattedField.label =`${field.label}*`
  }
  return formattedField
}
```

hanars (approved):

PR #4056, opened 2024-05-01
update phenotype prioritization notifs, filtering, display

jklugherz:
new filter option:
Screenshot 2024-05-01 at 4 18 35 PM

family table filtering actually happens:
Screenshot 2024-05-01 at 4 18 40 PM

phenotype prioritization displayed similar to loaded sample:
image

notification on project page:
Screenshot 2024-05-06 at 11 17 22 AM


seqr/views/apis/project_api.py
```diff
@@ -197,6 +197,10 @@ def project_families(request, project_guid):
             JSONObject(paternalGuid='individual__father__guid', maternalGuid='individual__mother__guid'),
             filter=Q(individual__mother__isnull=False) | Q(individual__father__isnull=False), distinct=True,
         ),
+        phenotypePrioritizationTools=ArrayAgg(
```
jklugherz, line 200: I added phenotypePrioritizationTools to two APIs -  project_families and details ([seqr/views/utils/orm_to_json_utils.py](https://github.com/broadinstitute/seqr/pull/4056/files#diff-91b87a6b0addc52510c5b65dadb7b810a7d4863b997db1ca751554de1f62f4cf) , but I also considered adding it as an attribute of a Sample instead so I could follow the `hasMatchingSampleFilter` pattern [here](https://github.com/broadinstitute/seqr/pull/4056/files#diff-8fbcc475f318e2396a365a6943ebb0a75e524d8099d133ea1c51b8447d450459R382) but it would make handling phenotypePrioritizationTools  on `loadedSamples` [here](https://github.com/broadinstitute/seqr/pull/4056/files#diff-4dc8b572f8d236793259125c7319e39372f2e269d33bbf87ac9ae403dc2aa1b3R151) weird. And conceptually, they're not really attributes of samples, they're attributes of individuals.

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -9,6 +9,7 @@ import { BaseVariantGene } from 'shared/components/panel/variants/VariantGene'
 import { camelcaseToTitlecase } from 'shared/utils/stringUtils'
 import { loadPhenotypeGeneScores } from '../reducers'
 import { getPhenotypeDataLoading, getIndividualPhenotypeGeneScores } from '../selectors'
+import {PHENOTYPE_PRIORITIZATION_FIELDS} from "../constants";
```
hanars, line 12: this should be failing linting - make sure to check that there are no linter errors before putting up the final PR?

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
```diff
@@ -27,7 +28,12 @@ const PHENOTYPE_GENE_INFO_COLUMNS = [
       />
     ),
   },
-  { name: 'tool', width: 1, content: 'Tool' },
+  {
+    name: 'tool',
+    width: 1,
+    content: 'Tool',
+    format: ({ tool }) => PHENOTYPE_PRIORITIZATION_FIELDS[tool],
```
hanars, line 35: what is this changing?

ui/pages/Project/constants.js
```diff
@@ -504,3 +504,8 @@ export const TAG_FORM_FIELD = {
   parse: value => (value || []).map(name => ({ name })),
   validate: value => (value && value.length ? undefined : 'Required'),
 }
+
+export const PHENOTYPE_PRIORITIZATION_FIELDS = {
+  exomiser: 'Exomiser',
+  lirical: 'LIRICAL',
```
hanars, line 510: I don't think it look good to use different casing strategies for the different tools, so I think we should pick one and be consistent and then you can just apply whatever function directly to the tool instead of needing to define a mapping

seqr/views/apis/project_api.py
```diff
@@ -197,6 +197,10 @@ def project_families(request, project_guid):
...
+            'individual__phenotypeprioritization__tool', distinct=True,
+            filter=Q(individual__phenotypeprioritization__tool__isnull=False),
+        ),
```
hanars, line 203: It looks like the only place this is used in the UI is to check `phenotypePrioritizationTools.length > 0`. If so,l it would be cleaner to define a boolean here, similar to `hasRequiredMetadata`
```
family_models = Family.objects.filter(project=project).annotate(
    pp_individual_count=Count('individual', filter=Q(individual__phenotypeprioritization__tool__isnull=False))
...
family_annotations = dict(
    hasPhenotypePrioritization=Case(When(pp_individual_count__gt=0, then=Value(True)), default=Value(False)),
...
```

hanars (commented):
Overall looks really good! Added a bunch of comments but in general this looks like its moving in the right direction!

Would also love to see a screenshot for how the notification looks on the project page as well

seqr/models.py
```diff
@@ -1205,8 +1205,11 @@ class PhenotypePrioritization(BulkOperationBase):
     disease_name = models.TextField()
     scores = models.JSONField()
 
+    created_date = models.DateTimeField(default=timezone.now, db_index=True)
```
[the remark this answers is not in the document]
jklugherz: Do you have a suggestion for how to handle the value of `guid` for existing rows? `makemigrations` returns `You are trying to add a non-nullable field 'guid' to phenotypeprioritization without a default; we can't do that (the database needs something to populate existing rows).` I could default it to an empty string, then create a data migration that turns those values into the result of `_compute_guid`.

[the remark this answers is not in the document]
hanars: yeah that should work. You can use one of the existing migrations as a model for how to execute python as part of the migration (although you won't need a `reverse_code` function): https://github.com/broadinstitute/seqr/blob/master/seqr/migrations/0056_alter_family_post_discovery_omim_number.py

ui/pages/Project/constants.js
jklugherz: I added this as an extra change so that the our display for the tool names matches how they're actually capitalized - LIRICAL is all [uppercase](https://lirical.readthedocs.io/en/latest/) and exomiser [isn't](https://www.sanger.ac.uk/tool/exomiser/). I can take this out though as it's not an explicit part of the ticket.

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
jklugherz: this is related to my comment here: https://github.com/broadinstitute/seqr/pull/4056/files#r1588101844

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -296,6 +296,14 @@ def _get_json_for_individuals(individuals, user=None, project_guid=None, family_
         'maternalId': F('mother__individual_id'),
         'paternalId': F('father__individual_id'),
         'displayName': INDIVIDUAL_DISPLAY_NAME_EXPR,
+        'phenotypePrioritizationTools':  ArrayAgg(
```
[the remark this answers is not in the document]
jklugherz: yup, was thinking this too

seqr/views/apis/project_api.py
jklugherz: agree!

seqr/views/utils/orm_to_json_utils.py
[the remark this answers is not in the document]
jklugherz: actually, I was thinking of removing the logic for hasPhenotypeGeneScores on the `family_page_data` api, and keeping the code I have here (which annotates individuals, not families). I agree that there's no need for both.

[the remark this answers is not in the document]
hanars: so those are 2 different things. In one case you are on the family page and have fetched all the individuals with all their details. In the other case you are on the project page and have fetched only the family metadata, which you need to use to filter the family table

seqr/models.py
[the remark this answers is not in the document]
jklugherz: alright, the best way I could come up to accomplish this was to 
- add the field with a one time default as an empty string and no index/uniqueness constraint (not sure if excluding the index is necessary here)
- run python to set guids to what they should be
- alter the field to have index and be unique

this feels okay, but the migration does take a couple minutes and can probably be optimized

seqr/utils/search/add_data_utils.py
```diff
@@ -9,6 +9,9 @@
     SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL
 
 
+BASE_EMAIL_TEMPLATE = 'Dear seqr user,\n\n{}\n\nAll the best,\nThe seqr team'
```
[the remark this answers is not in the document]
jklugherz: moved the constant into the `send_project_notification` file!

seqr/utils/search/add_data_utils.py
```diff
@@ -87,6 +90,22 @@ def notify_search_data_loaded(project, dataset_type, sample_type, inactivated_sa
     send_project_notification(
         project,
         notification=f'Loaded {sample_summary}',
-        email_body=f'Dear seqr user,\n\n{email}\n\nAll the best,\nThe seqr team',
+        email_body=BASE_EMAIL_TEMPLATE.format(email),
         subject='New data available in seqr',
     )
+
+
+def notify_phenotype_prioritization_loaded(project, tool, num_samples, file_path, user):
```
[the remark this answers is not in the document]
jklugherz: I moved this back into data_manager_api and prefixed it by an underscore

seqr/utils/search/add_data_utils.py
```diff
@@ -87,6 +90,22 @@ def notify_search_data_loaded(project, dataset_type, sample_type, inactivated_sa
...
+    url = f'{BASE_URL}project/{project.guid}/project_page'
+    project_link = f'<a href={url}>{project.name}</a>'
+    email = (
+        f'This is to notify you that {tool.title()} data for {num_samples} samples '
+        f'has been loaded in seqr project {project_link} by {user.get_full_name()}'
+    )
+
+    send_project_notification(
+        project,
+        notification=f'Loaded {tool.title()} data from {file_path} for {num_samples} samples',
```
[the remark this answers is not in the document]
jklugherz: this makes sense, I updated the notification

seqr/views/utils/orm_to_json_utils.py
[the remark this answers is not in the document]
jklugherz: >  In one case you are on the family page and have fetched all the individuals with all their details

For this case, I replaced `hasPhenotypeGeneScores` with a list `phenotypePrioritizationTools` on the family page api. 
 
>  In the other case you are on the project page and have fetched only the family metadata

For this case, I added a boolean field `hasPhenotypePrioritization` to use to filter the family table.

seqr/views/apis/data_manager_api.py
```diff
@@ -380,15 +397,26 @@ def load_phenotype_prioritization_data(request):
...
 
     if to_delete:
         PhenotypePrioritization.bulk_delete(request.user, to_delete)
 
-    PhenotypePrioritization.bulk_create(request.user, [PhenotypePrioritization(**data) for data in all_records])
+    models_to_create = []
+    for indiv_records in all_records_by_project_name.values():
+        for record in indiv_records:
+            model = PhenotypePrioritization(**record)
+            model.guid = f'PP{random.randint(10 ** 8, 10 ** 9)}_{model.individual.individual_id}_{model.gene_id}_{model.disease_id}'[:PhenotypePrioritization.MAX_GUID_SIZE]  # nosec
```
jklugherz, line 412: `bulk_update` doesn't call model.save() so model._compute_guid() is never called. I found one other example in the repo where random.randint is used to help create the guid during a bulk_create operation.

ui/pages/Project/constants.js
jklugherz: removed this change from this PR 😆

ui/pages/Project/components/PhenotypePrioritizedGenes.jsx
jklugherz: removed this change from this PR 😆

seqr/views/apis/data_manager_api.py
hanars: If you wanted to get fancy you could move the guid population to `ModelWithGUID.bulk_create` so theres no need to copy paste this again, but also thats kind of scope creep so this is also okay to leave as is

jklugherz:
@hanars I updated the PR description with a screenshot of a notification on the project page!

seqr/migrations/0064_alter_phenotypeprioritization.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+
+    def update_guids(apps, schema_editor):
+        PhenotypePrioritization = apps.get_model('seqr', 'PhenotypePrioritization')
+        db_alias = schema_editor.connection.alias
+
+        pps = PhenotypePrioritization.objects.using(db_alias).all()
+        for pp in pps:
+            ids_as_str = "%s:%s:%s" % (pp.individual.individual_id, pp.gene_id, pp.disease_id)
+            pp.guid = 'PP%07d_%s' % (pp.id, _slugify(str(ids_as_str)))[:MAX_GUID_SIZE]
```
hanars, line 25: so the reason why we can't use the regular `_compute_guid` functionality for bulk create most of the time is because the `id` does not exist until after the model is created. However here the model and the id already exists so you should be able to just do `pp.guid = pp._compute_guid()` 

Also, when changing around migrations if you want to test them even though you alrwady applied them, its pretty straightforward to unapply and then reapply them in django. What you do is run 
```
./manage.py migrate seqr 0063
```
That will tell django to migrate to whatever number migration you specify, which in this case is the previous migration so it will unapply this migration. Then you can change the migration and then run the migrate command as normal so you can confirm it works

seqr/models.py
```diff
@@ -1193,7 +1193,7 @@ class Meta:
                        'delta_intron_jaccard_index', 'mean_counts', 'total_counts', 'mean_total_counts']
 
 
-class PhenotypePrioritization(BulkOperationBase):
```
hanars, line 1196: Is `BulkOperationBase` no longer needed?

seqr/views/apis/data_manager_api.py
```diff
@@ -331,6 +333,21 @@ def load_rna_seq_sample_data(request, sample_guid):
     return create_json_response({'success': True})
 
 
+def _notify_phenotype_prioritization_loaded(project, tool, num_samples, user):
+    url = f'{BASE_URL}project/{project.guid}/project_page'
+    project_link = f'<a href={url}>{project.name}</a>'
+    email = (
+        f'This is to notify you that {tool.title()} data for {num_samples} sample(s) '
+        f'has been loaded in seqr project {project_link} by {user.get_full_name()}'
```
hanars, line 341: I do not think we want to tell external users who is doing the data loading. Its always an internal data manger that they don't know, so I think its just a confusing extra piece of information. I would take out `by {user.get_full_name()}` from the email body and leave it as is

seqr/views/apis/family_api.py
```diff
@@ -80,15 +80,26 @@ def family_page_data(request, family_guid):
...
         response['individualsByGuid'][individual_guid]['hasRnaOutlierData'] = True
 
-    has_phentoype_score_indivs = PhenotypePrioritization.objects.filter(individual__family=family).values_list(
-        'individual__guid', flat=True)
-    for individual_guid in has_phentoype_score_indivs:
-        response['individualsByGuid'][individual_guid]['hasPhenotypeGeneScores'] = True
+    phentoype_prioritization_tools = (
+        PhenotypePrioritization.objects.filter(individual__family=family).values(
+            'individual__guid',
+            'individual__phenotypeprioritization__tool',
```
hanars, line 86: So here you are getting the values from the `PhenotypePrioritization` model but then starting every field with `individual`, which is an antipattern. It seems like a better approach would be to use [djangos aggereation functionality ](https://docs.djangoproject.com/en/3.2/topics/db/aggregation/#values)to group these by individual and then get the values. In this case it would look something like
```
tools_agg = PhenotypePrioritization.objects.filter(individual__family=family).values('individual__guid').annotate(
  phenotypePrioritizationTools=ArrayAgg(JSONObject(tool='tool', createdDate='created_date''))
)
```

ui/shared/utils/constants.js
```diff
@@ -376,6 +376,12 @@ export const CATEGORY_FAMILY_FILTERS = {
         ({ sampleType, datasetType }) => sampleType !== SAMPLE_TYPE_RNA && datasetType === dataType,
       ),
     })),
+    {
+      value: `${SHOW_DATA_LOADED}_PHENO`,
+      name: 'Data Loaded - Phenotype Prioritization',
+      // eslint-disable-next-line no-unused-vars
+      createFilter: (family, _user, _samplesByFamily) => family.hasPhenotypePrioritization,
```
hanars, line 383: javascript does not care if you pass extra positional arguments, so you can just do 
```
createFilter: family => family.hasPhenotypePrioritization,
```

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -544,6 +548,10 @@ class IndividualRow extends React.PureComponent {
     // only show active or first/ last inactive samples
     loadedSamples = loadedSamples.filter((sample, i) => sample.isActive || i === 0 || i === loadedSamples.length - 1)
 
+    const phenotypePrioritizationTools = individual.phenotypePrioritizationTools.map(
```
hanars, line 551: So constructing arrays for props can cause unwanted updates in react. It would be better to pass `individual.phenotypePrioritizationTools` to the `DataDetails` component and do any transformation/m mapping in the render method

hanars (changes requested):

seqr/migrations/0064_alter_phenotypeprioritization.py
jklugherz: 🙏 thank you for the tip on unapplying migrations, I would've probably just re-loaded from my most recent db backup

jklugherz: It's probably worth adding a reverse_code function then!

seqr/views/apis/data_manager_api.py
jklugherz: ah this makes sense, taking out {user.get_full_name()}

seqr/migrations/0064_alter_phenotypeprioritization.py
jklugherz: I'm getting `AttributeError: 'PhenotypePrioritization' object has no attribute '_compute_guid'` when applying, `_compute_guid` must not exist yet on the PhenotypePrioritization model that's in the django app registry. 

I might try doing the data migration in a separate migration that runs after 0064

hanars: For the reverse migration, you can just do `reverse_code=migrations.RunPython.noop` as theres no need to actually change the guid in reverse because it will be dropped. 

And oh well, I guess we have to leave this as is.  The code you have here for computing the guid will work fine, I don;t think its worth the engineering time to figure out how to get `_compute_guid` to work here

jklugherz: agree with both of these

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
jklugherz: I was doing this originally inside of `DataDetails`:
```
    {individual.phenotypePrioritizationTools.map(
      pp => (
        <div key={pp.tool}>
          <Sample loadedSample={{ sampleType: pp.tool.charAt(0).toUpperCase() + pp.tool.slice(1), loadedDate: pp.createdDate }}/>
        </div>
      ),
    )}
```
but was getting lint error `JSX attribute values should not contain objects created in the same scope` so I moved it out into `IndividualRow` render. 
 
which approach is.. less bad?

hanars: So honestly, I think the best approach is just to return this from the server formatted the way it needs to be formatted for the display, instead of transforming it. IYou would just change the field names you use in the `JSONObject` aggregation

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1214,22 +1217,28 @@ def test_load_phenotype_prioritization_data(self, mock_subprocess):
...
+            ('delete 1 PhenotypePrioritizations', {'dbUpdate': {
+                'dbEntity': 'PhenotypePrioritization', 'updateType': 'bulk_delete',
+                'entityIds': ['PP000003_NA19678_ENSG000002689'],
             }}),
-            ('create PhenotypePrioritizations', {'dbUpdate': {
-                'dbEntity': 'PhenotypePrioritization', 'numEntities': 2, 'updateType': 'bulk_create',
-                'parentEntityIds': ['I000002_na19678', 'I000015_na20885'],
+            ('create 2 PhenotypePrioritizations', {'dbUpdate': {
+                'dbEntity': 'PhenotypePrioritization', 'updateType': 'bulk_create',
+                "entityIds": [ANY, ANY],
```
hanars, line 1226: In most of our tests that use this kind of behavior we mock the call to random so that we can deterministically know what the guids will be, so that we can be sure they are being created the way we want them to

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
jklugherz: yeah that's the right call

hanars:
So because of the release the other day, I think this is going to need some merge conflict resolution. In particular, I think the changelog will be messed up, and I think the migrations will be messed up because the dynamic analysis group migration was added. For the migration, you will want to change the dependencies so that your new migration will be run cleanly after the dynamic analysis group migration

jklugherz:
@hanars I didn't have to change the migration dependency but I did update the changelog

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## dev
 
+## 5/14/24
+* Adds guid and created fields to PhenotypePrioritization model (REQUIRES DB MIGRATION)
```
hanars, line 6: PRs to dev should have new changelog entries go in the dev section, not have a date. When seqr is released from dev to master we add a date to the changelog to indicate the dte all the rleevent features were actually released

seqr/migrations/0064_alter_phenotypeprioritization.py
```diff
@@ -0,0 +1,56 @@
...
+from seqr.models import _slugify
+
+MAX_GUID_SIZE = 30
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('seqr', '0063_dynamicanalysisgroup'),
```
hanars, line 15: This is what I was referring to when I said you might have to update the migration, but since it looks like it was already depending on the last migration I added it makes sense that there was no further change needed!

CHANGELOG.md
jklugherz: ah ok I just fixed it

hanars (approved):

PR #4076, opened 2024-05-09
Multi pod rna loading bug

hanars:
For RNA loading we parse the input file and split it into one file per sample, and then we dispatch one request per sample to load from each subset file in order to achieve parallelism and prevent server timeout (actually parsing and then writing all the data from the file in one request takes way longer than the browser can handle). 

However we have 2 running seqr pods, so all the sample subsets get written into the filesystem of one pod, but then roughly half the follow up requests get routed to the other pod and that pod's filesystem does not have the expected file. This PR therefore moves all the files off of the pod and into GCS so both pods will have access to them

### DIFF
--- a/seqr/views/apis/data_manager_api.py
+++ b/seqr/views/apis/data_manager_api.py
@@ -15,7 +15,7 @@ from django.views.decorators.csrf import csrf_exempt
 from requests.exceptions import ConnectionError as RequestConnectionError
 
 from seqr.utils.search.utils import get_search_backend_status, delete_search_backend_data
-from seqr.utils.file_utils import file_iter, does_file_exist
+from seqr.utils.file_utils import file_iter, does_file_exist, mv_file_to_gs
 from seqr.utils.logging_utils import SeqrLogger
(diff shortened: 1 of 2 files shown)


seqr/views/apis/data_manager_api.py
```diff
@@ -289,6 +292,8 @@ def _save_sample_data(sample_guid, sample_data):
     except ValueError as e:
         return create_json_response({'error': str(e)}, status=400)
 
+    mv_file_to_gs(f'{file_dir}/*', f'{TEMP_GS_BUCKET}/{file_name_prefix}', request.user)
```
jklugherz, line 295: could you write the files directly to gcs without saving them to the temp directory first?

hanars: Unfortunately no, that would be way to slow. This endpoint is already pushing too slow on a local filesystem, having to create and independent transaction to cloud storage for every row would be prohibitively slow

jklugherz (approved):

bpblanken (approved):

PR #4080, opened 2024-05-10
load all phenotpye priorities after viewing variant page

hanars:
If a user goes to the saved variants/ variant search page and then goes to the family page and views phenotype prioritized data, the data has been loaded just for the genes in the search page which was causing it to not load all the data and instead only show a subset. This updates the code to check and ensure all the data is loaded and fetch if not. 

Once this issue was fixed, the genes for the laoded data had all the locus list info loaded but the other genes did not which was making the table look weird, so I also updated the display to explicitly not show locus lists in the prioritized phenotype table

### DIFF
--- a/ui/pages/Project/reducers.js
+++ b/ui/pages/Project/reducers.js
@@ -330,7 +330,16 @@ export const loadRnaSeqData = individualGuid => (dispatch, getState) => {
 export const loadPhenotypeGeneScores = individualGuid => (dispatch, getState) => {
   const state = getState()
   const { familyGuid } = state.individualsByGuid[individualGuid]
-  if (!state.phenotypeGeneScoresByIndividual[individualGuid]) {
+  const loadedToolCounts = Object.values(state.phenotypeGeneScoresByIndividual[individualGuid] || {}).reduce(
+    (acc, dataByTool) => (
+      Object.entries(dataByTool).reduce((acc2, [tool, data]) => ({
+        ...acc2,
+        [tool]: (acc2[tool] || 0) + data.length,
+      }), acc)
+    ), {},
+  )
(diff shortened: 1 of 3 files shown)


ui/pages/Project/reducers.js
```diff
@@ -330,7 +330,16 @@ export const loadRnaSeqData = individualGuid => (dispatch, getState) => {
...
+  const loadedToolCounts = Object.values(state.phenotypeGeneScoresByIndividual[individualGuid] || {}).reduce(
+    (acc, dataByTool) => (
+      Object.entries(dataByTool).reduce((acc2, [tool, data]) => ({
+        ...acc2,
+        [tool]: (acc2[tool] || 0) + data.length,
+      }), acc)
+    ), {},
+  )
+  // Data can be loaded for only a subset of genes if previously loaded variant information
+  if (!Object.values(loadedToolCounts).some(val => val >= 10)) {
```
bpblanken, line 342: would you mind clarifying what the `10` threshold means?  Is it possible every value in `loadedToolCounts` is less than `10` and if not how come?

hanars: Yes, it comes from the code in the backend where dat is fetched for this reducer. The endpoint filters the phenotype prioritization models to the top 10, but as described n this Pr description there are also some pages which only load a subset of genes and therefore you do not get all 10, but if you have all 10 that means it is all loaded as we never load more than that
https://github.com/broadinstitute/seqr/blob/master/seqr/views/utils/variant_utils.py#L318

bpblanken: 👍 .  A comment or a constant with a descriptive name would be nice!

bpblanken (approved):

PR #4090, opened 2024-05-15
Compute guid cleanup

hanars:
Currently, the `bulk_create` functionality for the base `ModelWithGUID` class does not work unless you do some finicky manipulation of the models to assign them guids. This PR cleans up how guids are computed for these models (reducing a lot of copy-paste in theodels file) and makes their computation an automatic part of the  `bulk_create` helper.


seqr/views/utils/dataset_utils.py
```diff
@@ -444,19 +448,22 @@ def _load_rna_seq(model_cls, file_path, save_data, *args, user=None, **kwargs):
...
+            new_sample_models = _create_samples(
                 samples_to_create.values(),
                 user=user,
                 data_source=data_source,
                 sample_type=Sample.SAMPLE_TYPE_RNA,
                 dataset_type=Sample.DATASET_TYPE_VARIANT_CALLS,
             )
+            new_sample_ids = [s.id for s in new_sample_models]
+            sample_key_map = _get_sample_models_by_key(Sample.objects.filter(id__in=new_sample_ids), key_fields=key_fields)
+            sample_guid_keys_to_load.update({s['guid']: sample_key for sample_key, s in sample_key_map.items()})
```
hanars, line 466: because guids are now generated after samples are actually bulk created in the database and not when the placeholder sample models are cretaed, we need to refactor how this code relies on sample guids

bpblanken (approved):
looks cleaner to me!

seqr/models.py
```diff
@@ -127,6 +127,7 @@ def bulk_create(cls, user, new_models):
         """Helper bulk create method that logs the creation"""
         for model in new_models:
             model.created_by = user
+            model.guid = model._format_guid(random.randint(10**(cls.GUID_PRECISION-1), 10**cls.GUID_PRECISION))  # nosec
```
jklugherz, line 130: 🥇

seqr/views/utils/dataset_utils.py
```diff
@@ -472,19 +479,19 @@ def update_sample_models():
...
-    def save_sample_data(sample_guid, sample_data):
+    def save_sample_data(sample_key, sample_data):
         if not sample_data:
             return
 
-        sample_guids_to_load.add(sample_guid)
-        save_data(sample_guid, sample_data)
+        save_data(sample_key, sample_data)
 
     def get_matched_sample(sample_key, unmatched_samples, sample_id_to_individual_id_mapping):
```
jklugherz, line 488: Since this fn no longer returns anything, I don't think it should be prefixed by 'get'. Can you rename it?

jklugherz (approved):

PR #4091, opened 2024-05-15
clinvar override pass filter

hanars:


hail_search/queries/mito.py
```diff
@@ -147,8 +147,8 @@ def _parse_intervals(self, intervals, exclude_intervals=False, **kwargs):
             self._load_table_kwargs = {'_intervals': parsed_intervals, '_filter_intervals': True}
         return parsed_intervals
 
-    def _get_family_passes_quality_filter(self, quality_filter, ht=None, pathogenicity=None, **kwargs):
-        passes_quality = super()._get_family_passes_quality_filter(quality_filter)
+    def _get_family_passes_quality_filter(self, quality_filter, ht, pathogenicity=None, **kwargs):
+        passes_quality = super()._get_family_passes_quality_filter(quality_filter, ht)
         clinvar_path_ht = False if passes_quality is None else self._get_loaded_clinvar_prefilter_ht(pathogenicity)
         if not clinvar_path_ht:
```
hanars, line 153: On the line below is the logic for having clinvar override the quality filter. By merging the pass quality filter in with the other quality filters, it allows the existing clinvar override to work as desired

jklugherz (approved):

PR #4092, opened 2024-05-16
Track External Data

hanars:


ui/shared/components/panel/family/Family.jsx
```diff
@@ -89,6 +89,13 @@ const FAMILY_FIELD_RENDER_LOOKUP = {
       <AnalysedBy analysedByList={analysedByList} compact={compact} familyGuid={familyGuid} />
     ),
   },
+  [FAMILY_FIELD_EXTERNAL_DATA]: {
+    internal: true,
+    component: TagFieldView,
+    tagOptions: FAMILY_EXTERNAL_DATA_OPTIONS,
+    simplifiedValue: true,
```
jklugherz, line 96: I see what `simplifiedValue` controls in `TagFieldView` but is there a conceptual rule for which fields get simplifiedValue set to true?

ui/shared/components/panel/view-fields/TagFieldView.jsx
```diff
@@ -100,13 +101,14 @@ export const TagFieldDisplay = React.memo((
...
 TagFieldDisplay.propTypes = {
-  displayFieldValues: PropTypes.arrayOf(PropTypes.object).isRequired,
+  displayFieldValues: PropTypes.arrayOf(PropTypes.oneOfType(PropTypes.object, PropTypes.string)).isRequired,
   popup: PropTypes.func,
   tagAnnotation: PropTypes.func,
   displayAnnotationFirst: PropTypes.bool,
   displayMetadata: PropTypes.bool,
   linkTagType: PropTypes.string,
   tagLinkUrl: PropTypes.string,
+  tagLookup: PropTypes.object,
```
jklugherz, line 111: did you mean to add this prop [here](https://github.com/broadinstitute/seqr/pull/4092/files#diff-274caeecd16940e867f395d6b80a130298f108e7ec0a5f2da1b0440f522326c6L187)?

seqr/migrations/0065_family_external_data.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('seqr', '0064_alter_phenotypeprioritization'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='family',
+            name='external_data',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(blank=True, choices=[('M', 'Methylation'), ('P', 'PacBio lrGS'), ('R', 'PacBio RNA'), ('L', 'ONT lrGS'), ('O', 'ONT RNA'), ('B', 'BioNano')], max_length=1, null=True), default=list, size=None),
```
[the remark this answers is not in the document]
hanars: I guess? We track them on the model, the same as we track all other enumerated fields in seqr

ui/shared/components/panel/family/Family.jsx
hanars: So a rough rule is if your field has a value which is just a list of strings it probably needs the simplified view, whereas if your field has a values thats a list of objects (like variant tags) it need to more complex handling

ui/shared/components/panel/view-fields/TagFieldView.jsx
hanars: `tagLookup` is added as a prop to this `TagFieldDisplay` class on line 78, so it needs a proptype declaration here

jklugherz (approved):

PR #4093, opened 2024-05-16
Minimal project families response

hanars:


seqr/views/apis/project_api.py
```diff
@@ -192,7 +193,15 @@ def project_families(request, project_guid):
...
-    family_annotations = dict(
+    families = family_models.values(
+        'description',
+        **{_to_camel_case(field): F(field) for field in [
+            'family_id', 'analysis_status', 'created_date', 'coded_phenotype', 'mondo_id',
+        ]},
+        familyGuid=F('guid'),
+        projectGuid=Value(project_guid),
+        **FAMILY_ADDITIONAL_VALUES,
+        **INDIVIDUAL_GUIDS_VALUES,
```
jklugherz, line 204: I understand you did this to just return fields that the overview page uses but is there any downside to not calling `_get_json_for_families` and subsequently `get_json_for_queryset` ?

hanars: Yeah so those functions ultimately pull the fields list, at minimum, from `fields = set(model_class._meta.json_fields)` so theres no way to exclude some of the core model fields. The purpose of this PR is to pair down the fields being returned in this summary endpoint as several of them are unused and for large projects the content download time for this request is itself rather long as there is so much data being returned. Removing all the fields that are not needed reduces the response size, and also means that as we add model fields we do not continue to bloat the response

jklugherz (approved):

PR #4096, opened 2024-05-17
Optimize SQL queries for project page performance

hanars:
This PR optimizes the generated SQL queries for a couple of the slower project detail endpoints. Overall, changes work to reduce table joins where possible and reduce duplicating queries to the same table (i.e. get all data in one query instead of multiple). 

A. note on how this sort of optimization is done: django has pretty good debug tools for this. I change the `ENABLE_DJANGO_DEBUG_TOOLBAR` setting in settings.py to True which enables it. The "gotcha" for this tool is it will only profile requests thqt return HTML, so for whatever API endpoint I want to profile I replace the final return statement with the following, and then access the API endpoint directly in my browser 
```
from seqr.views.react_app import render_app_html
return render_app_html(request, include_user=False)
```

The debug toolbar is then shown, I usually focus on the "SQL" tab (you can also look at total time and if the total time is substantially longer than the SQL time then there may be non-SQL related optimization that is needed). From there I usually focus in on the queries that are duplicated and/ or the queries that are the slowest and dig into them to see whats going on and try out other solutions. 

Screenshot 2024-05-16 at 5 06 21 PM

### DIFF
--- a/seqr/views/apis/project_api.py
+++ b/seqr/views/apis/project_api.py
@@ -183,16 +183,16 @@ def project_page_data(request, project_guid):
 @login_and_policies_required
 def project_families(request, project_guid):
     project = get_project_and_check_permissions(project_guid, request.user)
+
     family_models = Family.objects.filter(project=project)
-    family_annotations = dict(
-        _id=F('id'),
-    )
     families = _get_json_for_families(
         family_models, request.user, has_case_review_perm=has_case_review_permissions(project, request.user),
-        project_guid=project_guid, add_individual_guids_field=False, additional_values=family_annotations,
+        project_guid=project_guid, add_individual_guids_field=False, additional_values={'_id': F('id')},
     )
     families_by_id = {f.pop('_id'): f for f in families}
-    # TODO multiple joins against individual table to get parent annotations
+
+    phenotype_priority_families = set(PhenotypePrioritization.objects.filter(
+        individual__family_id__in=families_by_id).values_list('individual__family_id', flat=True).distinct())
     family_individuals = Individual.objects.filter(family_id__in=families_by_id).values('family_id').annotate(
         caseReviewStatuses=ArrayAgg('case_review_status', distinct=True, filter=~Q(case_review_status='')),
         caseReviewStatusLastModified=Max('case_review_status_last_modified_date'),
@@ -203,11 +203,12 @@ def project_families(request, project_guid):
         )),
     )
     for individual_agg in family_individuals:
-        family = families_by_id[individual_agg.pop('family_id')]
+        family_id = individual_agg.pop('family_id')
         parental_ids = individual_agg.pop('parental_ids')
         id_guid_map = {i['id']: i['guid'] for i in parental_ids}
-        family.update({
+        families_by_id[family_id].update({
             'individualGuids': sorted(id_guid_map.values()),
+            'hasPhenotypePrioritization': family_id in phenotype_priority_families,
             'hasRequiredMetadata': individual_agg.pop('metadata_count') > 0,
             'parents': [
                 {'paternalGuid': id_guid_map.get(p['father_id']), 'maternalGuid': id_guid_map.get(p['mother_id'])}
@@ -216,10 +217,6 @@ def project_families(request, project_guid):
             **individual_agg,
         })
 
-    phenotype_priority_family_ids = set(PhenotypePrioritization.objects.filter(
-        individual__family_id__in=families_by_id).values_list('individual__family', flat=True).distinct())
-    for family_id, family in families_by_id.items():
-        family['hasPhenotypePrioritization'] = family_id in phenotype_priority_family_ids
     response = families_discovery_tags(families, project=project)
     return create_json_response(response)
 
--- a/seqr/views/apis/project_api.py
+++ b/seqr/views/apis/project_api.py
@@ -6,13 +6,13 @@ import json
 from collections import defaultdict
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.core.exceptions import PermissionDenied
-from django.db.models import Count, Max, Q, Case, When, Value
+from django.db.models import Count, Max, Q, F, Case, When, Value
 from django.db.models.functions import JSONObject, TruncDate
 from django.utils import timezone
 from notifications.models import Notification
 
 from matchmaker.models import MatchmakerSubmission
-from seqr.models import Project, Family, Individual, Sample, FamilyNote, CAN_EDIT
+from seqr.models import Project, Family, Individual, Sample, FamilyNote, PhenotypePrioritization, CAN_EDIT
 from seqr.views.utils.airtable_utils import AirtableSession, ANVIL_REQUEST_TRACKING_TABLE
 from seqr.views.utils.individual_utils import delete_individuals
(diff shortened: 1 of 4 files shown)


seqr/views/apis/project_api.py
```diff
@@ -184,34 +184,46 @@ def project_page_data(request, project_guid):
...
     families = family_models.values(
-        'description',
+        'id', 'description',
         **{_to_camel_case(field): F(field) for field in [
             'family_id', 'analysis_status', 'created_date', 'coded_phenotype', 'mondo_id',
         ]},
         familyGuid=F('guid'),
         projectGuid=Value(project_guid),
         **FAMILY_ADDITIONAL_VALUES,
-        **INDIVIDUAL_GUIDS_VALUES,
```
hanars, line 204: the join that was being constructed for all of these annotations was far less efficient than running 3 queries directly - on on families, one on individuals, and one on phenotype prioirty. The `parents` annoattons was causing us to do 3 separate joins on the individuals table, and the phenotype priority data is relatively sparse so it turned out the full outer join on the individual tabel is slower than the very targeted aggregation directly on that table

seqr/views/apis/project_api.py
```diff
@@ -221,11 +233,9 @@ def project_overview(request, project_guid):
 
     sample_models = Sample.objects.filter(individual__family__project=project)
 
-    active_samples = sample_models.filter(is_active=True)
-    first_loaded_samples = sample_models.order_by('individual__family', 'loaded_date').distinct('individual__family')
-    samples_by_guid = {}
-    for samples in [active_samples, first_loaded_samples]:
-        samples_by_guid.update({s['sampleGuid']: s for s in get_json_for_samples(samples, project_guid=project_guid)})
+    first_loaded_samples = sample_models.order_by('individual__family', 'loaded_date').distinct('individual__family').values_list('id', flat=True)
```
hanars, line 236: this still assesses to 2 DB queries, but the query that computed `first_loaded_samples` and pulled all the json fields out for it was actually relatively inefficient due to the underlying aggregation and it was cheaper to do the aggegration as one query and then pull out all the json fields for all the relevant samples as a separate query

seqr/views/utils/project_context_utils.py
```diff
@@ -110,11 +110,12 @@ def add_child_ids(response):
         family['individualGuids'] = individual_guids_by_family[family['familyGuid']]
 
 
-def families_discovery_tags(families):
+def families_discovery_tags(families, project=None):
     families_by_guid = {f['familyGuid']: dict(discoveryTags=[], **f) for f in families}
 
+    family_filter = {'family__project': project} if project else {'family__guid__in': families_by_guid.keys()}
```
hanars, line 116: removes a join with the project table when unneded (querying directly on the project uses `family.project_id` so the project table is never joined)

seqr/views/utils/project_context_utils.py
```diff
@@ -132,20 +133,20 @@ def families_discovery_tags(families):
 MME_TAG_NAME = 'MME Submission'
 
 
-def add_project_tag_types(projects_by_guid, add_counts=False):
-    variant_tag_types_models = VariantTagType.objects.filter(Q(project__guid__in=projects_by_guid.keys()) | Q(project__isnull=True))
-    variant_tag_types = _get_json_for_models(variant_tag_types_models)
+def add_project_tag_types(projects_by_guid, project=None):
+    is_single_project = len(projects_by_guid) == 1
+    project_q = dict(project=project) if project else dict(project__guid__in=projects_by_guid.keys())
```
hanars, line 138: removes a join with the project table when unneeded

seqr/views/utils/project_context_utils.py
```diff
@@ -132,20 +133,20 @@ def families_discovery_tags(families):
...
+    variant_tag_types_models = VariantTagType.objects.filter(Q(**project_q) | Q(project__isnull=True))
+    variant_tag_types = get_json_for_queryset(
```
hanars, line 140: `get_json_for_queryset` is more efficient than `_get_json_for_models`, I just haven't gotten around to fully replacing it yet

seqr/views/utils/project_context_utils.py
```diff
@@ -165,42 +165,46 @@ def add_project_tag_types(projects_by_guid, add_counts=False):
             ),
             'variantFunctionalTagTypes': VariantFunctionalData.FUNCTIONAL_DATA_TAG_TYPES,
         })
-        if add_counts:
```
hanars, line 168: this function can get tag types for multiple projects, but it is only ever called with `add_counts` for a single project, and there are a lot of db optimizations that can be made if we can assume a single project. Therefore, I added a separate utility function that adds the counts that only works on a single project

seqr/views/utils/project_context_utils.py
```diff
@@ -165,42 +165,46 @@ def add_project_tag_types(projects_by_guid, add_counts=False):
...
+def add_project_tag_type_counts(project, response_json, project_json=None):
+    response_json['projectsByGuid'] = {project.guid: project_json or {}}
+    add_project_tag_types(response_json['projectsByGuid'], project=project)
 
-def _add_tag_type_counts(project_guid, project_variant_tags):
-    project_tags = VariantTag.objects.filter(saved_variants__family__project__guid=project_guid)
-    project_notes = VariantNote.objects.filter(saved_variants__family__project__guid=project_guid)
+    saved_variants = SavedVariant.objects.filter(family__project=project)
+    project_tags = VariantTag.objects.filter(saved_variants__in=saved_variants)
+    project_notes = VariantNote.saved_variants.through.objects.filter(savedvariant_id__in=saved_variants)
```
hanars, line 176: using the `through` class means we never query the note table itself, just the mapping table between note ids and variant ids. Since we do not actually use any of the note fields, it removes an extra join

seqr/views/utils/project_context_utils.py
```diff
@@ -165,42 +165,46 @@ def add_project_tag_types(projects_by_guid, add_counts=False):
...
+
     note_tag_type = {
         'variantTagTypeGuid': 'notes',
         'name': 'Has Notes',
         'category': 'Notes',
         'description': '',
         'color': 'grey',
         'order': 100,
-        'numTags': project_notes.aggregate(count=Count('saved_variants__guid', distinct=True))['count'],
+        'numTags': project_notes.values_list('savedvariant_id').distinct().count(),
```
hanars, line 187: this appears to be a more efficient query to get a single count

seqr/views/utils/project_context_utils.py
```diff
@@ -165,42 +165,46 @@ def add_project_tag_types(projects_by_guid, add_counts=False):
...
     }
 
-    mme_counts_by_family = project_tags.filter(saved_variants__matchmakersubmissiongenes__isnull=False) \
-        .values('saved_variants__family__guid').annotate(count=Count('saved_variants__guid', distinct=True))
+    mme_counts_by_family = saved_variants.filter(matchmakersubmissiongenes__isnull=False) \
```
hanars, line 190: since this was only querying and counting off the saved variant models and not using the tag models at all, much more efficient to use the saved variant models directly

jklugherz (approved):
🙌 thanks for writing out the steps you took to find the queries that needed optimizations

PR #4097, opened 2024-05-17
do not allow mito 37 variant lookup

jklugherz:


seqr/utils/search/hail_search_utils.py
```diff
@@ -87,6 +87,9 @@ def _execute_lookup(variant_id, data_type,  user, **kwargs):
 
 
 def hail_variant_lookup(user, variant_id, **kwargs):
+    if kwargs.get('genome_version') == 'GRCh37' and variant_id[0] == 'M':
+        from seqr.utils.search.utils import InvalidSearchException
+        raise InvalidSearchException('Mitochondrial variants are not supported in GRCh37')
     variant, _ = _execute_lookup(variant_id, Sample.DATASET_TYPE_VARIANT_CALLS, user, **kwargs)
```
hanars, line 93: So it looks like you are checking here if the variant starts with"M", but we are still sending the lookup query with the wrong data type. I think the solution here actually needs to involve correctly setting the data type for this search, and then validating it. The `_variant_ids_dataset_type` helper function in the main search utils.py is probably a good place to share logic. You can also keep the error handling in this file or moe it into the other file, whichever is easier

hanars: Also, we should probably add this validation to SV lookup as well

hanars (changes requested):

seqr/utils/search/hail_search_utils.py
jklugherz: right, the default of Sample.DATASET_TYPE_VARIANT_CALLS isn't correct for MITO. moving the validation elsewhere makes sense

seqr/utils/search/hail_search_utils.py
```diff
@@ -86,19 +86,19 @@ def _execute_lookup(variant_id, data_type,  user, **kwargs):
...
 
-def hail_variant_lookup(user, variant_id, **kwargs):
-    variant, _ = _execute_lookup(variant_id, Sample.DATASET_TYPE_VARIANT_CALLS, user, **kwargs)
+def hail_variant_lookup(user, variant_id, dataset_type, **kwargs):
+    variant, _ = _execute_lookup(user, variant_id, data_type=dataset_type, **kwargs)
     return variant
 
 
-def hail_sv_variant_lookup(user, variant_id, samples, sample_type=None, **kwargs):
+def hail_sv_variant_lookup(user, variant_id, dataset_type, samples, sample_type=None, **kwargs):
```
jklugherz, line 94: `dataset_type` as an arg is redundant here but I have it in here because the caller https://github.com/broadinstitute/seqr/pull/4097/files#diff-20e96f31f5b1390b1b519bb5fc6cefb42e859489cbc0d4d2f084b80a2ee63342R161 is shared

seqr/utils/search/hail_search_utils.py
jklugherz: @hanars made some changes, lmk if this is more in line with what you're thinking

seqr/utils/search/utils.py
```diff
@@ -149,27 +149,36 @@ def _get_variants_for_variant_ids(families, variant_ids, user, user_email=None,
...
+    variant = lookup_func(user, variant_id, dataset_type, genome_version=GENOME_VERSION_LOOKUP[genome_version], **kwargs)
     safe_redis_set_json(cache_key, variant, expire=timedelta(weeks=2))
     return variant
 
 
-def variant_lookup(*args, **kwargs):
-    return _variant_lookup(hail_variant_lookup, *args, **kwargs)
+def _validate_dataset_type_genome_version(dataset_type, genome_version):
+    if genome_version == GENOME_VERSION_GRCh37 and dataset_type != Sample.DATASET_TYPE_VARIANT_CALLS:
+        raise InvalidSearchException('Only SNV_INDEL variants are available for GRCh37')
```
hanars, line 168: Since this error is surfaced to users I think it would be better to have a slightly more user-friendly message, perhaps 
```
f'{dataset_type} variants are not available for GRCh37'
```

seqr/utils/search/utils.py
```diff
@@ -149,27 +149,36 @@ def _get_variants_for_variant_ids(families, variant_ids, user, user_email=None,
...
+
+
+def variant_lookup(user, parsed_variant_id, **kwargs):
+    dt = _variant_ids_dataset_type([parsed_variant_id])
+    dataset_type = DATASET_TYPES_LOOKUP[dt][0]
```
hanars, line 173: nitpick, but `dt` and `dataset_type` are a bit confusing. as avriable names here, its not lear why exactly you need both. Maybe rename them like `data_type_key` and `single_dataset_type` or something else a bit more descriptive?

jklugherz: think I'll just combine them into one variable

hanars (approved):

PR #4102, opened 2024-05-21
improve mgration performance

hanars:
This migration is actually taking longer than the timeout for the seqr pod in dev, causing it to terminate before the migration can complete. This improves performance by doing one query to map PP models to individuals, instead of one query per model


jklugherz (approved):

bpblanken (commented):
cool this is better!

PR #4105, opened 2024-05-22
Post discovery mondo ID

hanars:


seqr/views/apis/report_api_tests.py
```diff
@@ -1293,7 +1296,9 @@ def test_variant_metadata(self):
             **BASE_VARIANT_METADATA_ROW,
             'alt': None,
             'chrom': '1',
-            'condition_id': 'MONDO:0008788',
+            'condition_id': 'OMIM:616126',
```
hanars, line 1299: test the ability to report OMIM for one variant in a family and MONDO for another - the above test for `['rows'][5]` still reports the Modno condition but this row now reports in OMIM

seqr/migrations/0066_family_post_discovery_mondo_id.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('seqr', '0065_family_external_data'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='family',
+            name='post_discovery_mondo_id',
```
jklugherz, line 15: do we have to support multiple `post_discovery_mondo_ids` for one family?

hanars: It was not in the ticket, so I did not build it. I guess probably in the future we may need to, but also at some point in the future we are going to like ingest MONDO and actually have users pick options from a dropdown instead of storing free text, so it seems like the simplicity of maintaining a single field for now is easier, and we can update it to a multiple field at some point in  the future when we do a full overhaul of the field

ui/shared/components/panel/family/Family.jsx
```diff
@@ -111,13 +121,13 @@ const FAMILY_FIELD_RENDER_LOOKUP = {
...
-    fieldDisplay: value => (
-      <a target="_blank" rel="noreferrer" href={`http://purl.obolibrary.org/obo/MONDO_${value.replace('MONDO:', '')}`}>
-        {value}
-      </a>
-    ),
+  },
+  [FAMILY_FIELD_DISCOVERY_MONDO_ID]: {
+    ...MONDO_FIELD,
+    internal: true,
+    canEditFamily: ({ discoveryTags }) => discoveryTags?.length > 0,
```
jklugherz, line 130: What does it mean in terms of the status of the case if it has discovery tags? is it solved if there are discoveryTags?

jklugherz (approved):

seqr/migrations/0066_family_post_discovery_mondo_id.py
jklugherz: ah ok I was confused about task # 3 (Accommodate reporting for cases with multiple findings across OMIM and MONDO (similar to current logic of multiple findings with multiple OMIMs)) but that seems to be related to the changes in [anvil_metadata_utils.py](https://github.com/broadinstitute/seqr/pull/4105/files#diff-fcba4e6ac6983d5e2f987e729d8b190d3add7e8dd3b75a59b9c274bbdadd7ce6)

ui/shared/components/panel/family/Family.jsx
hanars: yes, discovery tags mean that someone added at tag to a variant indicating that it is (at least part of) the solve. Also, the post discovery OMIM and MONDO numbers are used for reporting purposes, and they are included per discovery variant, so there is no purpose of having them at all if theres no discovery variant

PR #4116, opened 2024-05-28
limit concurrent requests during RNA loading

hanars:


ui/pages/DataManagement/reducers.js
```diff
@@ -54,7 +54,17 @@ export const uploadQcPipelineOutput = values => submitRequest(
...
+const executeMultipleRequests = (requests, onSuccess, warnings) => Promise.all(requests.map(
+  ([entityUrl, entityId, body]) => new HttpRequestHelper(
+    entityUrl,
+    onSuccess,
+    e => warnings.push(`Error loading ${entityId}: ${e.body && e.body.error ? e.body.error : e.message}`),
+  ).post(body),
+))
+
+const loadMultipleData = (
+  path, getUpdateData, dispatchType, formatSuccessMessage, maxConcurrentRequests = 50,
```
bpblanken, line 66: could we add an assertion that `maxConcurrentRequests` is `> 1`?

ui/pages/DataManagement/reducers.js
```diff
@@ -64,15 +74,19 @@ const loadMultipleData = (path, getUpdateData, dispatchType, formatSuccessMessag
...
-      ([entityUrl, entityId, body]) => new HttpRequestHelper(
-        entityUrl,
-        () => {
-          numLoaded += 1
-        },
-        e => warnings.push(`Error loading ${entityId}: ${e.body && e.body.error ? e.body.error : e.message}`),
-      ).post(body),
-    )).then(() => {
+    const updateData = getUpdateData(successResponseJson, values)
+    return updateData.reduce((prevPromise, item, index) => {
```
bpblanken, line 78: it might be cleaner to reason about this by splitting the `batch`-ing from the `Promise` stuff with a `batch(arr, size)` helper.  that's usually how I've seen this... 🤷 but this definitely looks right as is!

ui/pages/DataManagement/reducers.js
hanars: we currently have no js function argument validation anywhere in our code, so it seems pretty out of scope to add it here

ui/pages/DataManagement/reducers.js
hanars: I actually had it like that originally but it was basically a `reduce` followed by another nearly identical `reduce` so it looked messier to me than having the single function call

bpblanken: yep that's fair!  I know it doesn't fit the style of the reducers but something like this w/ a generator might also work:

```
function* batch(arr, size){
  for(let i = 0; i < arr; i += n) yield arr(i, i + size);
}
```

bpblanken (approved):

PR #4117, opened 2024-05-28
add tenacity retry to panelapp function

jklugherz:
I chose to add tenacity as opposed to using urllib3.util.Retry because it's simper to use and is also well supported, lots of stars on github

### DIFF
--- a/panelapp/panelapp_utils.py
+++ b/panelapp/panelapp_utils.py
@@ -3,6 +3,8 @@ from collections import defaultdict
 import requests
 from django.db import transaction
 from django.utils import timezone
+from tenacity import retry, wait_exponential, stop_after_attempt, retry_if_exception_type
+from urllib3.exceptions import MaxRetryError
 
 from panelapp.models import PaLocusList, PaLocusListGene
 from seqr.models import LocusList as SeqrLocusList, LocusListGene as SeqrLocusListGene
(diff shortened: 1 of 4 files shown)


panelapp/panelapp_utils.py
```diff
@@ -113,6 +115,11 @@ def _get_all_panels(panels_url, all_results):
         return _get_all_panels(next_page, all_results)
 
 
+@retry(
+    retry=retry_if_exception_type(MaxRetryError),
+    wait=wait_exponential(multiplier=1, min=4, max=10),
+    stop=stop_after_attempt(5),
```
hanars, line 121: what does this all mean?

jklugherz: > retry_if_exception_type(MaxRetryError)

Retry the function if it throws a `urllib3.exceptions.MaxRetryError` error - this is the error we get in the most recent logs. I can add more than one exception type here.

> wait=wait_exponential(multiplier=1, min=4, max=10)

Wait with exponential backoff between retries. From the [docs](https://tenacity.readthedocs.io/en/latest/): "Wait 2^x * 1 second between each retry starting with 4 seconds, then up to 10 seconds, then 10 seconds afterwards"

> stop=stop_after_attempt(5)

Retry up to 5 times then stop, if it doesn't succeed on the 5th retry then original error will be thrown.

hanars (approved):

PR #4118, opened 2024-05-29
Temp file refactor

hanars:
Expands work done in https://github.com/broadinstitute/seqr/pull/4099/files to apply to all cases where temp files are interacted with. Also, updates that logic to be properly conditional for local vs non-local deployments


seqr/views/utils/file_utils.py
```diff
@@ -110,16 +117,27 @@ def save_uploaded_file(request, process_records=None, allow_json=False):
...
+
     return uploaded_file_id, filename, json_records
 
 
-def load_uploaded_file(upload_file_id):
-    serialized_file_path = _compute_serialized_file_path(upload_file_id)
-    with gzip.open(serialized_file_path, "rt") as f:
-        json_records = json.load(f)
+def persist_temp_file(file_name, user, is_directory=False):
+    if not anvil_enabled():
```
jklugherz, line 131: just to clarify in my mind - are there local users who have anvil enabled? In that case, we are allowing them to store uploads in our seqr-scratch-temp?

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1054,15 +1053,18 @@ def _test_basic_data_loading(data, num_parsed_samples, num_loaded_samples, new_s
...
-            f'gsutil mv tmp/{file_path}/* gs://seqr-scratch-temp/{file_path}',
-        ]])
-        mock_mkdir.assert_called_with(f'tmp/{file_path}')
-        filename = f'tmp/{file_path}/{new_sample_guid}.json.gz'
+        ] + self._additional_expected_loading_subprocess_calls(file_path)
+        self.assertEqual(mock_subprocess.call_count, len(expected_subprocess_calls))
+        mock_subprocess.assert_has_calls([
+            mock.call(command, stdout=-1, stderr=-2, shell=True) for command in expected_subprocess_calls  # nosec
+        ])
+        mock_mkdir.assert_any_call(f'tmp/temp_uploads/{file_path}')
```
jklugherz, line 1064: is gettempdir() called multiple times in the code path under test? if applicable here, it could be better to use `assert_has_calls`

jklugherz (approved):

seqr/views/apis/data_manager_api_tests.py
hanars: so that is a good point, but under the hood there are some additional calls being made from inside `get_temp_file_path` to create the initial temp directory that are tested elsewhere and would be really confusing to test here, as they are called way more often than you would expect due to the directory not actually being created because of the mocking. Since that behavor is tested elsewhere and the only thing we really care about for this file is that the one call to `mkdir` in this file is called correctly, I thought it was neater just to test the relevant call

PR #4122, opened 2024-05-30
Job to update savedvariant annotations

hanars:
Currently all the annotations for saved variants are automatically updated whenever new data is available in seqr. However, there will be times when we may update the annotations table without loading new samples (i.e. for new VEP table) so having the ability to sync annotations with the backend is useful (and also can be useful during local development)

### DIFF
--- a/seqr/management/commands/check_for_new_samples_from_pipeline.py
+++ b/seqr/management/commands/check_for_new_samples_from_pipeline.py
@@ -132,23 +132,27 @@ class Command(BaseCommand):
         # Reload saved variant JSON
         updated_variants_by_id = update_projects_saved_variant_json(
             updated_project_families, user_email=USER_EMAIL, dataset_type=dataset_type)
+
+        data_type = f'{dataset_type}_{sample_type}' if dataset_type == Sample.DATASET_TYPE_SV_CALLS else dataset_type
         self._reload_shared_variant_annotations(
-            updated_variants_by_id, updated_families, dataset_type, sample_type, genome_version)
+            data_type, genome_version, updated_variants_by_id, exclude_families=updated_families)
 
         logger.info('DONE')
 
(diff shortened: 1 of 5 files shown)


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -163,11 +166,11 @@ def _reload_shared_variant_annotations(updated_variants_by_id, updated_families,
         for v in variant_models:
             variants_by_id[v.variant_id].append(v)
 
-        logger.info(f'Reloading shared annotations for {len(variant_models)} saved variants ({len(variants_by_id)} unique)')
+        logger.info(f'Reloading shared annotations for {len(variant_models)} {data_type} {genome_version} saved variants ({len(variants_by_id)} unique)')
 
         updated_variants_by_id = {
             variant_id: {k: v for k, v in variant.items() if k not in {'familyGuids', 'genotypes', 'genotypeFilters'}}
-            for variant_id, variant in updated_variants_by_id.items()
+            for variant_id, variant in (updated_variants_by_id or {}).items()
```
jklugherz, line 173: I have a question about how this works in general (not necessarily related to your pr):

If `updated_variants_by_id` is empty, we won't update any shared annotations right? 
`bulk_update` is called on `updated_variant_models` which comes from `updated_variants_by_id`, so why don't call `bulk_update` on the variants in `variants_by_id` instead?

jklugherz: looking at this more, it seems like `updated_variants_by_id` should be updated with the other variants in the `if fetch_variant_ids:` block

jklugherz (approved):
thank you, this already makes my development easier!
(to get around this I kept re-setting the postgres db between check samples runs!)

PR #4130, opened 2024-05-31
add clingen allele ID to variant

jklugherz:
Screenshot 2024-05-31 at 2 36 34 PM


ui/shared/components/panel/variants/ClinGenVciLink.jsx
```diff
@@ -34,11 +41,17 @@ class ClinGenVciLink extends React.PureComponent {
...
-      <DataLoader contentId={hgvsc} content={alleleId || error} loading={loading} load={this.load}>
+    const clingenInfo = (
+      <div>
         <a href={CLINGEN_VCI_URL} target="_blank" rel="noreferrer">In ClinGen VCI</a>
         <br />
         {error || (alleleId && <CopyToClipboardButton text={alleleId} />)}
+      </div>
+    )
+
+    return alleleId ? clingenInfo : (
```
hanars, line 52: Thjis works, but I think this would be clearer if the conditional existed at a different level. Theres 2 different control flows that happen here - the ID is a prop and we render the info directly, or the ID is in the state and we need to render the data loader. Rather than have this one component that has that conditional check twice to acheive that behavior, it seem like it would be neater to have a `ClingenInfo` component that just handles rendering the loaded content, a `LoadedClinGenVciLink` that handle loading the ID if necessary, and then the actual exported component that would handle the conditional:
```
const ClinGenVciLink = (
  { CAID, ...props }) => (CAID ? <ClingenInfo alleleId={CAID} /> : <LoadedClinGenVciLink {...props } />
)
```

ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -438,7 +438,7 @@ const svSizeDisplay = (size) => {
 const Annotations = React.memo(({ variant, mainGeneId, showMainGene, transcriptsById }) => {
   const {
     rsid, svType, numExon, pos, end, svTypeDetail, svSourceDetail, cpxIntervals, algorithms, bothsidesSupport,
-    endChrom,
+    endChrom, CAID,
```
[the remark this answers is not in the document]
jklugherz: no lint error would you prefer it to be something else?

hail_search/test_utils.py
```diff
@@ -142,6 +142,7 @@
     'mainTranscriptId': None,
     'selectedMainTranscriptId': None,
     '_sort': [1000010439],
+    'CAID': 'CA16717152',
```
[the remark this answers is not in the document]
jklugherz: yes, one of the 38 variants has CAID == None!

Screenshot 2024-06-10 at 4 29 51 PM

ui/shared/components/panel/variants/Annotations.jsx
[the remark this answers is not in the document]
hanars: nope, as lon as there is no linting error I am happy with it!

ui/shared/components/panel/variants/ClinGenVciLink.jsx
jklugherz: that's reasonable and clearer, I split up the components like you suggested

ui/shared/components/panel/variants/ClinGenVciLink.jsx
```diff
@@ -36,13 +56,20 @@ class ClinGenVciLink extends React.PureComponent {
...
-        {error || (alleleId && <CopyToClipboardButton text={alleleId} />)}
+        <ClingenInfo alleleId={alleleId} error={error} />
       </DataLoader>
     )
   }
 
 }
 
+const ClinGenVciLink = ({ CAID, hgvsc }) => (
+  CAID ? <ClingenInfo alleleId={CAID} /> : <LoadedClingenVciLink alleleId={CAID} hgvsc={hgvsc} />
```
hanars, line 67: If there is a `CAID` this conditional will render the `ClingenInfo` so the `alleleId={CAID}` for the `LoadedClingenVciLink` is always going to be undefined, so theres no need have that component accept that prop in the first place

jklugherz: you're right, good catch

hanars (approved):

PR #4133, opened 2024-05-31
Do not hardcode expected columns for gregor reports

hanars:
For the Gregor reports, we fetch the data model and use that to determine which columns we return. However, we want to add warnings if a field we are computing is no longer in the data model so we know we can stop maintaining the logic. We did this by hardcoding a list of all the columns we compute for this table, but this is really unwieldy and hard to maintain, and also unlikely to remain accurate. Therefore, this updates the code to instead actually get the list of all the fields we have computed and validate against that.


seqr/views/apis/report_api.py
```diff
@@ -188,17 +196,6 @@ def _add_row(row, family_id, row_type):
 SMID_FIELD = 'SMID'
 PARTICIPANT_ID_FIELD = 'CollaboratorParticipantID'
 COLLABORATOR_SAMPLE_ID_FIELD = 'CollaboratorSampleID'
-PARTICIPANT_TABLE_COLUMNS = {
```
hanars, line 191: these are the hardcoded lists we no longer need

seqr/views/apis/report_api_tests.py
```diff
@@ -509,10 +509,7 @@
     'notes': None,
     'phenotype_contribution': 'Full',
     'partial_contribution_explained': '',
-    'phenotype_description': None,
-    'pmid_id': None,
     'seqr_chosen_consequence': None,
-    'solve_status': 'Unsolved',
```
hanars, line 515: these fields are not used on the variant metadata page even though they are included in the response, at some point I will want to go back and remove everything we are not using from the response to prevent confusion in cases like this one but thats out of scope here

seqr/views/apis/report_api_tests.py
```diff
@@ -1120,11 +1117,10 @@ def test_family_metadata(self):
         test_row = next(r for r in response_json['rows'] if r['familyGuid'] == 'F000012_12')
         self.assertDictEqual(test_row, {
             'projectGuid': 'R0003_test',
-            'internal_project_id': 'Test Reprocessed Project',
             'familyGuid': 'F000012_12',
             'family_id': '12',
             'displayName': '12',
-            'solve_status': 'Unsolved',
+            'solve_status': 'Partially solved',
```
hanars, line 1123: this is actually a more accurate way to show solve status, and `internal_project_id` was never used on the family metadata page

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -136,7 +138,8 @@ def _get_family_metadata(family_filter, family_fields, include_metadata, include
     family_data_by_id = {}
     for f in family_data:
         family_id = f.pop('id')
-        solve_status = ANALYSIS_SOLVE_STATUS_LOOKUP.get(f['analysisStatus'], Individual.UNSOLVED)
+        analysis_status = f['analysisStatus'] if include_metadata else f.pop('analysisStatus')
```
jklugherz, line 141: Can you tell me why we have the `include_metadata` flag  in `parse_anvil_metadata() ` -  what the 'metadata' is and why we only want to return it sometimes? A lot of fields depend on it and I want to understand it!

seqr/views/apis/report_api.py
```diff
@@ -173,6 +178,9 @@ def _add_row(row, family_id, row_type):
         }},
     )
 
+    for row in parsed_rows[SUBJECT_ROW_TYPE]:
+        row.update(family_diseases[row['family_id']])
```
jklugherz, line 182: `family_diseases` gets populated if `row_type == FAMILY_ROW_TYPE`, so is there ever possibility that `family_diseases` is empty and cause an attribute error here?

seqr/views/utils/anvil_metadata_utils.py
hanars: So this helper function is used 5 times - in 2 places to generate multi-file reports with a strict data schema (the Gregor and Anvil exports) and in 3 places to generate "metadata" reports (family, variant, and individual) that we define and use internally to help us look into our own data. These metadata reports are often close to the gregor report because one of the things we use them for ins QCing those reports, but they are also useful for other purposes and include some additional information not required by the main reports, and also they split apart/aggregate certain fields based on the primary entity for said report (i.e. some fields in the family metadata are the aggregate of that same field in the individual metadata). 

The `include_metadata` flag is True for the 3 metadata reports and False for the other 2 usages, so "metadata" here refers to the name of the reports where this additional information is valuable. 

TBH, there is probably a bunch of fields that are included when `include_metadata` is enabled thats not needed for all 3 metadata reports, and at some point I would love to split out as much of the `include_metadata` logic as possible from this one boolean flag into more targeted per-function logic, but the metadata reports are incredibly tolerant of having extra fields returned in the server response so it does not break anything for now, and I thought that was out of scope for what I wanted to accomplish here. But its definitely a future cleanup goal!

seqr/views/apis/report_api.py
hanars: no, in `parse_anvil_metadata` we always call ` add_row` with `FAMILY_ROW_TYPE` before we ever call `add_row` with `SUBJECT_ROW_TYPE`. There are a couple of other places where we rely on this strict prdering

jklugherz (approved):

seqr/views/utils/anvil_metadata_utils.py
jklugherz: thank you for the explanation!! From a maintainability and readability standpoint, an OOP approach might work well here with inheritance between gregor/metadata and classmethods to instantiate the reports.

PR #4140, opened 2024-06-05
VEP update UI

hanars:
Updates the seqr UI to properly display all the new VEP annotations, while also continuing to work with the current VEP as well. The only changes to the hail backend are generic changes to how pass-through behavior works and therefore allow the new VEP annotations to format properly while at the same time not actually changing anything in the current response.

I made the decision not to update the fixture data for the search unit tests in this PR, as this code will be going live before the new VEP annotations do and I would like the unit tests to continue to cover current behavior. 

New UI (@lynnpais can you take a look and let me know if anything should change for the MVP):

Screenshot 2024-06-04 at 2 45 59 PM

Transcripts modal: only  change is showing Intron/Exon
Screenshot 2024-06-04 at 11 25 35 AM

Screenshot 2024-06-04 at 11 45 39 AM

Screenshot 2024-06-04 at 5 40 44 PM

Screenshot 2024-06-04 at 2 55 34 PM

Screenshot 2024-06-04 at 11 48 50 AM


ui/shared/components/panel/variants/Transcripts.jsx
```diff
@@ -42,7 +94,62 @@ const TRANSCRIPT_LABELS = [
...
+  </div>
+)
+
+const ANNOTATION_SECTIONS = [
+  [{ title: 'Codons' }, { title: 'Amino Acids' }],
+  [
+    { title: 'Biotype' },
+    {
+      title: 'Intron/Exon',
+      getContent: c => ['intron', 'exon'].filter(f => c[f]).map(f => `${camelcaseToTitlecase(f)} ${c[f].index}/${c[f].total}`).join(', '),
```
hanars, line 143: this is the only actual change to the Transcripts content, everything else is just a refactor to share code.  We use to show `cdnaPosition` here, but that has been absent from the data for a while so theres no reason to keep it

lynnpais:
Looks great! 

Minor request - regarding the first screenshot showing UTR consequences, display the UTRAnnotator result first, then SCREEN, and regulatory feature last.

jxchong:
As you know, we have been slowly working on the implementation of this in the "old backend." We're showing UTRannotator if any of the transcripts have annotations from it, just like if a missense was only a missense on an alternative transcript but synonymous on the main transcript. It looks like you are only surfacing UTRannotator info from the main transcript, is that correct?

hanars:
>  just like if a missense was only a missense on an alternative transcript but synonymous on the main transcript.

So I'm not sure I'm understanding you correctly, but thats not really seqr's behavior. If you do a search for missense variants and the canonical transcript is synonymous, seqr does show the missense transcripts for everything. So it shows missense and not syonymous, but it also shows the hgvsc for that missense transcript, the gene associated with that transcript, etc. We do not mix and match different transcripts into the main annotation display, and are choosing to continue with that behavior for UTRAnnotator and only show the annotation associated with the selected transcript. However, just as when choosing the transcript to show we factor in what consequences are searched on, we will also factor in if utrannotator was searched on and preferentially show the transcript with UTRAnnotator if that is the reason the variant is returned from search in the first place. 

That said, I think it does make sense to show UTRannotator it in the transcript detail modal for all transcripts regardless of whether or not they are returned from search

bpblanken (approved):

PR #4150, opened 2024-06-07
Extended intronic splice region variant

hanars:
Adds the ability to filter on extended_intronic_splice_region_variant
This code is against the VEP branch and will be released to production after the updated VEP is live


hail_search/queries/snv_indel.py
```diff
@@ -25,6 +27,23 @@ class SnvIndelHailTableQuery(SnvIndelHailTableQuery37):
...
+    def _get_allowed_consequence_ids(self, annotations):
+        consequence_ids = super()._get_allowed_consequence_ids(annotations)
+        if EXTENDED_SPLICE_REGION_CONSEQUENCE in (annotations.get('extended_splice_site') or []):
+            consequence_ids.add(EXTENDED_SPLICE_REGION_CONSEQUENCE)
+        return consequence_ids
+
+    @staticmethod
+    def _get_allowed_transcripts_filter(allowed_consequence_ids):
+        has_extended_splice = EXTENDED_SPLICE_REGION_CONSEQUENCE in allowed_consequence_ids
+        allowed_consequence_ids = allowed_consequence_ids - {EXTENDED_SPLICE_REGION_CONSEQUENCE}
```
bpblanken, line 39: why do we need to remove 'extended_intronic_splice_region_variant' from the ids here?

hanars: because everything else in here is an integer ID and so if we leave this string in when it gets converted to a hail set it will have the wrong type

PR #4154, opened 2024-06-10
VEP: get mane/refseq from search backend

hanars:
Use the mane and refseq annotations directly from the VEP/ the loading pipeline instead of from a gencode download stored in a postgres database. 

Will go live after new VEP annotations are live


seqr/views/utils/variant_utils.py
```diff
@@ -239,6 +239,12 @@ def get_variant_key(xpos=None, ref=None, alt=None, genomeVersion=None, **kwargs)
     return '{}-{}-{}_{}'.format(xpos, ref, alt, genomeVersion)
 
 
+def _requires_transcript_metadata(variant):
+    if isinstance(variant, list):
+        return _requires_transcript_metadata(variant[0])
```
jklugherz, line 244: Can you just call `_requires_transcript_metadata` with variant[0] in the first place?

seqr/views/utils/variant_utils.py
```diff
@@ -239,6 +239,12 @@ def get_variant_key(xpos=None, ref=None, alt=None, genomeVersion=None, **kwargs)
...
+    return variant.get('genomeVersion') != GENOME_VERSION_GRCh38 or variant.get('chrom', '').startswith('M')
```
jklugherz, line 245: Do you also have to add a case to handle SVs? (I think SVs can be saved variants too)

seqr/views/utils/variant_utils.py
```diff
@@ -263,7 +270,7 @@ def _saved_variant_genes_transcripts(variants):
             TranscriptInfo.objects.filter(transcript_id__in=transcript_ids),
             nested_fields=[{'fields': ('refseqtranscript', 'refseq_id'), 'key': 'refseqId'}]
         )
-    }
+    } if transcript_ids else None
```
jklugherz, line 273: The behavior is changing now such that we don't set `response['transcriptsById']` if the variants don't have genome version 38 or are mito. Why can we get rid of this?

ui/shared/components/panel/variants/Transcripts.jsx
```diff
@@ -78,6 +78,10 @@ ConsequenceDetails.propTypes = {
   ensemblLink: PropTypes.object,
 }
 
+export const isManeSelect = (transcript, transcriptsById) => (
+  !!transcript.maneSelect || transcriptsById[transcript.transcriptId]?.isManeSelect
```
jklugherz, line 82: What data is stored on the `transcript` and what is stored in `transcriptsById`?

seqr/views/utils/variant_utils.py
hanars: sometimes `variant ` is a list of objects and sometimes its a single object so without the instance check `variant[0]` would throw an error

seqr/views/utils/variant_utils.py
hanars: SVs do not have Ensembl transcripts though so they do not need to be checked here

seqr/views/utils/variant_utils.py
hanars: Because the only information in `response['transcriptsById']` is the refeseq transcript ID and whether or not a transcript is MANE. For 38 SNV/Indel calls in the hail backend, that information is now coming directly from search and we do not need to add that infornmation in a post processing step

ui/shared/components/panel/variants/Transcripts.jsx
hanars: The best way to understand this is to actually look into the test fixtures, as the fields are pretty different depending on what the data source is - the new VEP annotations have different transcript annotations than old VEP. `transcriptsById` adds in the refeseq transcript ID and whether or not a transcript is MANE.With thelatest round of VEP there is nothing we need from `transcriptsById`, it is all moved into the new pipeline

jklugherz (approved):

PR #4158, opened 2024-06-11
add clingen allele id to reports

jklugherz:
Screenshot 2024-06-12 at 12 31 27 PM


seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -492,6 +492,7 @@ def _get_genetic_findings_rows(rows: list[dict], individual: Individual, partici
                 }[zygosity],
                 'allele_balance_or_heteroplasmy_percentage': heteroplasmy,
                 'variant_inheritance': _get_variant_inheritance(individual, genotypes),
+                'ClinGen_allele_ID': row.pop('CAID', None),
```
jklugherz, line 495: This line means that ClinGen_allele_ID is in all genetic findings reports, whether or not we asked for CAID in the initial query. Is this the wrong approach @hanars ?

ui/pages/Report/components/VariantMetadata.jsx
```diff
@@ -12,7 +12,7 @@ const COLUMNS = [
   { name: 'participant_id' },
   ...VARIANT_METADATA_COLUMNS.slice(0, -1),
   { name: 'allele_balance_or_heteroplasmy_percentage' },
-  { name: 'Clinvar allele ID', format: ({ clinvar }) => clinvar?.alleleId },
+  { name: 'ClinGen allele ID', format: ({ ClinGen_allele_ID }) => ClinGen_allele_ID },
```
hanars, line 15: we want to keep `Clinvar allele ID` here, we should just also add `ClinGen_allele_ID`

ui/shared/components/table/LoadReportTable.jsx
```diff
@@ -54,7 +54,7 @@ const ReportTable = React.memo((
       striped
       collapsing
       horizontalScroll
-      downloadFileName={`${viewAllPages.find(({ path }) => path === projectGuid)?.downloadName || (data?.length && data[0][PROJECT_ID_FIELD].replace(/ /g, '_'))}_${new Date().toISOString().slice(0, 10)}_${fileName}`}
+      downloadFileName={`${viewAllPages.find(({ path }) => path === projectGuid)?.downloadName || (data?.length && data[0][PROJECT_ID_FIELD] && data[0][PROJECT_ID_FIELD].replace(/ /g, '_'))}_${new Date().toISOString().slice(0, 10)}_${fileName}`}
```
jklugherz, line 57: I was getting something like 'can't use .replace on undefined' during local testing, I think because I wasn't using internal projects. We want variant metadata reports to be available for non-internal projects?

hanars, line 57: is there reason this is changing?

seqr/views/apis/report_api.py
```diff
@@ -395,6 +395,7 @@ def _add_row(row, family_id, row_type):
         individual_data_types=grouped_data_type_individuals,
         add_row=_add_row,
         format_id=_format_gregor_id,
+        variant_json_fields=['CAID'],
```
hanars, line 398: we should have this be part of the default behavior of `parse_anvil_metadata`

seqr/views/utils/anvil_metadata_utils.py
hanars: we should have this column in all the genetics findings reports, there should not be any conditional logic around it

ui/shared/components/table/LoadReportTable.jsx
hanars: > We want variant metadata reports to be available for non-internal projects?

No

seqr/views/utils/anvil_metadata_utils.py
jklugherz: The column `ClinGen_allele_ID` is in all genetics findings reports. This line stays the same, the `None` default is there in case `CAID` doesn't exist in the saved variant json.

ui/pages/Report/components/VariantMetadata.jsx
hanars: We want to add this to the `VARIANT_METADATA_COLUMNS` so it is also shared with the individual summary data

hanars (changes requested):

hanars (approved):

hanars:
approval contingent on tests passing

PR #4164, opened 2024-06-14
Gregor report airtable fixes

hanars:
Adjusts seqr to properly work with how airtable is handling multiple sample records for different sample types associated with the same participant. Also adjusts to work with foreign references for RNA SMIDs and to handle uniqueness for the `aligned_dna_short_read_set_id`


seqr/views/utils/airtable_utils.py
```diff
@@ -105,49 +104,3 @@ def _populate_records(self, record_type, records, offset=None):
 
         if response_json.get('offset'):
             self._populate_records(record_type, records, offset=response_json['offset'])
-
-
-def _get_airtable_samples_for_id_field(sample_ids, id_field, fields, session):
```
hanars, line 110: these helper functions were only used in one place and needed to be edited. In general, for utilities that are only used once it is better to keep them as private methods in the file where they are used so its easier to know what is and is not safe to change

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -543,10 +544,52 @@ def _get_variant_inheritance(individual, genotypes):
 LIST_SAMPLE_FIELDS = ['SequencingProduct', 'dbgap_submission']
 
 
-def _get_sample_airtable_metadata(sample_ids, user, fields):
-    sample_records, _ = get_airtable_samples(
-        sample_ids, user, fields=fields or SINGLE_SAMPLE_FIELDS, list_fields=None if fields else LIST_SAMPLE_FIELDS,
+def _get_airtable_samples_for_id_field(sample_ids, id_field, fields, session):
```
hanars, line 547: directly copied from airtable_utils.py unless noted otherwise

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -543,10 +544,52 @@ def _get_variant_inheritance(individual, genotypes):
...
     )
+
+    records_by_id = defaultdict(list)
+    for airtable_id, record in raw_records.items():
+        records_by_id[record[id_field]].append({**record, 'airtable_id': airtable_id})
+    return records_by_id
+
+
+def _get_sample_airtable_metadata(sample_ids, user, airtable_fields):
+    fields, list_fields = airtable_fields or [SINGLE_SAMPLE_FIELDS, LIST_SAMPLE_FIELDS]
```
hanars, line 560: changed to support specifying `list_fields` in the `airtable_fields` param

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -543,10 +544,52 @@ def _get_variant_inheritance(individual, genotypes):
...
+                    record_id, field)
+                raise ErrorsWarningsException([error])
+            if record_field:
+                parsed_record[field] = record_field.pop()
+        for field in list_fields:
+            parsed_record[field] = {} if airtable_fields else set()
+            for record in records:
+                if field in record:
+                    if airtable_fields:
+                        parsed_record[field][record['airtable_id']] = record[field]
```
hanars, line 587: this line is new, and aggregates list fields into a dictionary keyed by the original record id (so here SMID would be a dictionary of sample record IDs to the actual SMIDs. The default behavior for list fields is the below line, which remains unchanged

seqr/views/apis/report_api.py
```diff
@@ -510,23 +513,37 @@ def _get_gregor_airtable_data(participants, user):
...
+    rna_metadata_by_smid_record = {}
     for data_type in GREGOR_DATA_TYPES:
         for r in airtable_metadata_by_participant.values():
             data_type_fields = [f for f in r if f.endswith(f'_{data_type}')]
             if data_type_fields:
-                r[data_type.upper()] = {f.replace(f'_{data_type}', ''): r.pop(f) for f in data_type_fields}
+                data_type_metadata = {f.replace(f'_{data_type}', ''): r.pop(f) for f in data_type_fields}
+                r[data_type.upper()] = data_type_metadata
+                if data_type == 'rna':
+                    smid_record_id = data_type_metadata[SMID_FIELD][0]
```
hanars, line 524: SMIDs for RNA are now a foreign reference to the Sample record in airtable, they are no longer strings

seqr/views/apis/report_api.py
```diff
@@ -398,12 +401,12 @@ def _add_row(row, family_id, row_type):
         get_additional_individual_fields=_get_participant_row,
         post_process_variant=_post_process_gregor_variant,
         include_svs=False,
-        airtable_fields=[SMID_FIELD, PARTICIPANT_ID_FIELD, 'Recontactable'],
+        airtable_fields=[[PARTICIPANT_ID_FIELD, 'Recontactable'], [SMID_FIELD]],
```
bpblanken, line 404: it might be cleaner to use split this into two variables, `airtable_fields` and `airtable_list_fields` or something like that rather a list with two elements.

hanars: I thought of that, but this gets passed through about 5 different places so it made or a kind of messy signature change to a lot of unrelated functions and I thought that messiness was not worth the trade off

bpblanken: 🤷 yeah, could do a `namedtuple` or something along those lines but doesn't matter much!

bpblanken (approved):

jklugherz (approved):

PR #4171, opened 2024-06-17
show RNAseq type in sample hover

jklugherz:
Screenshot 2024-06-17 at 3 29 57 PM


seqr/views/apis/family_api.py
```diff
@@ -45,9 +45,22 @@ def family_page_data(request, family_guid):
     sample_models = Sample.objects.filter(individual__family=family)
     samples = get_json_for_samples(sample_models, project_guid=project.guid, family_guid=family_guid, skip_nested=True, is_analyst=is_analyst)
     response = {
-        'samplesByGuid': {s['sampleGuid']: s for s in samples},
+        'samplesByGuid': {s['sampleGuid']: {**s, 'rnaSeqTypes': []} for s in samples},
     }
 
+    # Add Rna Seq metadata to samples
+    rna_seq_models = [
```
jklugherz, line 52: I'm thinking this should belong in `get_json_for_samples()` because I also need to update the `project_overview` api if I also want the Family Table sample to have the same hover-over behavior as the Family Page sample.

hanars: So the Family Table sample in the compact summary view should only be a search sample - if its not thats a bug. When you expand the Family row to see the full detail it loads the same data as when you load the family page directly so there shoudl be no issue there. So you should only need to change behavior in one place

seqr/views/apis/family_api.py
```diff
@@ -45,9 +45,22 @@ def family_page_data(request, family_guid):
...
+        (RnaSeqTpm, 'TPM'),
+        (RnaSeqOutlier, 'Outlier'),
+        (RnaSeqSpliceOutlier, 'Splice Outlier'),
+    ]
+    for model, rna_seq_type in rna_seq_models:
+        rna_seq_samples = model.objects.filter(
+            sample__in=sample_models, sample__is_active=True
+        ).values('sample__guid').distinct()
```
hanars, line 60: The more readable/maintainable way to do this would be to pass the `additional_values` argument into  `get_json_for_samples` with an `ArrayAgg` expression to get this list at query time instead of needing this post processing step

jklugherz: I'm using `additional_values` instead but wasn't able to get `ArrayAgg` to work. I'm still annotating 3 fields and doing some post-processing in python to turn it into an array of strings.

jklugherz: I tried an `ArrayAgg` with 3 cases statements but was seeing `function array_agg(text, text, text) does not exist`

hanars: I think its also fine to just return 3 booleans and have the frontend handle that, the backend response does not *have* to be a single list of strings

jklugherz: I was going to do it that way but then I was like.. I'd rather write the code to parse the 3 booleans into strings in python 😆

seqr/views/apis/family_api.py
```diff
@@ -43,7 +43,20 @@ def family_page_data(request, family_guid):
     has_case_review_perm = has_case_review_permissions(project, request.user)
 
     sample_models = Sample.objects.filter(individual__family=family)
-    samples = get_json_for_samples(sample_models, project_guid=project.guid, family_guid=family_guid, skip_nested=True, is_analyst=is_analyst)
+    additional_values = {
+        'rnaSeqTpm': Case(When(Exists(RnaSeqTpm.objects.filter(sample_id=OuterRef('pk'))), then=Value('TPM')), default=None),
```
hanars, line 47: So you should be able to do this using the related field on the sample model without having to write query objects with outer refs - something like `When(rnaseqtpm__isnull=False)`

seqr/views/apis/family_api.py
hanars: I hear that, but the parsing code looks pretty messy in the end, and in general we try not to put display logic in the server side code and keep display logic in the UI. Also, doing the formatting when we are rendering skips this extra iteration we are running on the server side to add formatting, and instead does the formatting on the client as part of all the other formatting that is being done. I think the JS code would be really simple, something like - 
```
const RNA_TYPES =  [
  ('rnaSeqTpm', 'TPM'), ('rnaSeqOutlier', 'Expression Outlier'), ('rnaSeqSpliceOutlier', 'Splice Outlier),
]
...
{loadedSample.sampleType === SAMPLE_TYPE_RNA &&
   `RNAseq methods: ${RNA_TYPES.map(
       (field, title) => loadedSample[field] && title,
    ).filter(value => value).join(', ')}`}
```

seqr/views/apis/family_api.py
jklugherz: I originally had the `__isnull` syntax but the query was timing out with more than 1 rnaseq type. I could add indexes on the foreign keys to the rnaseq tables on the sample model?

hanars: There should be indices on the foreign keys already, so that sounds like a case where what django is doing is not what we want. You could try using the debug toolbar to see the actual executed SQL and figure out why its so inefficient and adjust it, but honestly that feels like overkill and its probably better not to try to use those fields in this case

That said, if you are going to make 3 distinct queries to each of the subtypes and then iterate through all the sample models and add annotations to them, using the `additional_values` does not really save us any compute and just makes the whole thing less readable in the end. In that case, you are better off with getting lists of sample Ids for each type and then adding to `rnaSeqTypes` based on that, something like

```
rna_type_samples = {
    'TPM': set(RnaSeqTpm.objects.filter(sample__in= sample_models).values_list('sample', flat=True),
    'Expression Outlier': set(RnaSeqOutlier.objects.filter(...
    'Splice Outlier': set(RnaSeqSpliceOutlier.objects.filter(...
}
...
sample['rnaSeqTypes'] = [type for type, sample_ids in rna_type_samples.items() if sample['id'] in sample_ids]
```

jklugherz: I implemented it this way ^ which does not require changes to the JS code as you suggested [here](https://github.com/broadinstitute/seqr/pull/4171#discussion_r1651231878). 

Is there a general rule for what belongs in additional_values and when it's better to query related tables separately?

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -169,7 +169,7 @@ const DataDetails = React.memo(({ loadedSamples, individual, mmeSubmission }) =>
         />
       ) : <MmeStatusLabel title="Submitted to MME" dateField="lastModifiedDate" color="violet" individual={individual} mmeSubmission={mmeSubmission} />
     )}
-    {individual.hasRnaOutlierData && (
+    {loadedSamples.some(sample => sample.rnaSeqTypes?.length > 0) && (
```
hanars, line 172: `hasRnaOutlierData` was based on whether there was `rnaseqoutlier` or `rnaseqspliceoutlier` data, for samples with only tpm data this would have been false, so we should update the behavior here to reflect that. Also it only considered active samples

ui/shared/components/panel/sample.jsx
```diff
@@ -33,9 +33,16 @@ const Sample = React.memo(({ loadedSample, isOutdated, hoverDetails }) => (
...
-    content={loadedSample ?
-      `data was${isOutdated ? ' previously ' : ''} ${hoverDetails ? `${hoverDetails} on ${new Date(loadedSample.loadedDate).toLocaleDateString()}` : 'loaded'}` :
-      'no data available'}
+    content={
+      <div>
+        {loadedSample ?
+          `data was${isOutdated ? ' previously ' : ''} ${hoverDetails ? `${hoverDetails} on ${new Date(loadedSample.loadedDate).toLocaleDateString()}` : 'loaded'}` :
+          'no data available'}
+        <VerticalSpacer height={5} />
+        {loadedSample.sampleType && loadedSample.sampleType === SAMPLE_TYPE_RNA &&
```
hanars, line 42: `rnaSeqTypes` will only have values for RNA samples so you can remove this part of the condition check

hanars (changes requested):

ui/shared/components/panel/sample.jsx
jklugherz: Without this part of the check some old(?) rna samples which do not have any matches in the 3 rnaseqtype tables will display like this:
image

There are 76 of these samples
```
WITH sample_rnaseq as (
    select s.guid, s.id,
           seqr_rnaseqtpm.id as tpm_id, seqr_rnaseqoutlier.id as outlier_id, sr.id as splice_outlier_id
    from seqr_sample s
    left join seqr_rnaseqtpm on s.id = seqr_rnaseqtpm.sample_id
    left join seqr_rnaseqoutlier on s.id = seqr_rnaseqoutlier.sample_id
    left join public.seqr_rnaseqspliceoutlier sr on s.id = sr.sample_id
    where s.sample_type in ('RNA')
)
SELECT * FROM sample_rnaseq
WHERE tpm_id is null
and outlier_id is null
and splice_outlier_id is null


```

jklugherz: they're all probably from the R0246_rna_seq_bc project, I've been using that for testing since I was certain it would have rna samples given the name

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
jklugherz: I adjusted this condition check to match the behavior to show the link only if the sample is active and has either rnaseqoutlier or rnaseqspliceoutlier data

ui/shared/components/panel/sample.jsx
hanars: Oh we should only be showing this for active samples, not inactive sample

hanars (approved):

PR #4176, opened 2024-06-18
return current user with assigned analyst options

hanars:


seqr/views/apis/users_api.py
```diff
@@ -49,9 +49,12 @@ def get_all_user_group_options(request):
 @login_and_policies_required
 def get_project_collaborator_options(request, project_guid):
     project = get_project_and_check_permissions(project_guid, request.user)
+    user_fields = {'display_name', 'username', 'email'}
     users = get_project_collaborators_by_username(
-        request.user, project, fields={'display_name', 'username', 'email'}, expand_user_groups=True,
+        request.user, project, fields=user_fields, expand_user_groups=True,
     )
+    if not users:
```
jklugherz, line 56: So if the project has anvil, users is empty if the current user doesn't have permission to get the AnVIL access control list.

Do you have to be the creator of the anvil project to view the acl?

jklugherz (approved):

seqr/views/apis/users_api.py
jklugherz: Just saw your comment in the ticket

PR #4183, opened 2024-06-20
update hail search snapshot release values file

jklugherz:


.github/workflows/hail-search-persistent-volume-snapshot-release.yaml
```diff
@@ -46,13 +46,13 @@ jobs:
         uses: mikefarah/yq@v4.22.1
         with:
           cmd: >
-            yq -i '.global.hail_search.datasetVersions.${{ inputs.reference_genome }}/${{ inputs.dataset_type }} = "${{ inputs.version }}"' charts/broad-seqr/values-${{ inputs.environment }}.yaml
+            yq -i '.global.hail_search.datasetVersions.${{ inputs.reference_genome }}/${{ inputs.dataset_type }} = "${{ inputs.version }}"' charts/${{ inputs.environment }}-broad-seqr/values.yaml
```
bpblanken, line 49: I think we need a conditional on `${{ inputs.environment }}` so that it's a prefix of `''` if its prod 😢.

bpblanken:
we also fixed the pipeline so that the action can work fine on [the dev branch](https://github.com/broadinstitute/seqr-pipeline-airflow/blob/main/dags/dependencies/operators.py#L57).  However, we do actually want this to go out to production or the production pipeline will also be broken.

.github/workflows/hail-search-persistent-volume-snapshot-release.yaml
jklugherz: 🙏

hanars (approved):

bpblanken (approved):

PR #4198, opened 2024-06-26
Aip family mapping

hanars:
AIP upload is ambiguously mapping individuals to families/ projects, so reguire explicit whitelist for projects

### DIFF
--- a/seqr/views/apis/summary_data_api.py
+++ b/seqr/views/apis/summary_data_api.py
@@ -184,17 +184,25 @@ def _load_aip_data(data: dict, user: User):
     category_map = data['metadata']['categories']
     results = data['results']
 
-    family_id_map = dict(Individual.objects.filter(
+    family_id_map = defaultdict(list)
+    for individual_id, family_id in Individual.objects.filter(
         family__project__in=get_internal_projects(), individual_id__in=results.keys(),
-    ).values_list('individual_id', 'family_id'))
+    ).values_list('individual_id', 'family_id'):
+        family_id_map[individual_id].append(family_id)
+    errors = []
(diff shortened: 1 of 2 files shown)


seqr/views/apis/summary_data_api.py
```diff
@@ -182,19 +182,31 @@ def bulk_update_family_external_analysis(request):
...
+    ).values_list('individual_id', 'family_id'):
+        family_id_map[individual_id].append(family_id)
+    errors = []
     missing_individuals = set(results.keys()) - set(family_id_map.keys())
     if missing_individuals:
-        raise ErrorsWarningsException([f'Unable to find the following individuals: {", ".join(sorted(missing_individuals))}'])
+        errors.append(f'Unable to find the following individuals: {", ".join(sorted(missing_individuals))}')
+    multi_family_individuals = {individual_id for individual_id, families in family_id_map.items() if len(families) > 1}
+    if multi_family_individuals:
+        errors.append(f'The following individuals are found in multiple families: {", ".join(sorted(multi_family_individuals))}')
```
jklugherz, line 202: Why can't we load aip data if an individual is in multiple families?

jklugherz: I see in the upload fixture that the results are just keyed by individual with no family ID. Maybe the AIP results format should change to include family ids so that we can capture all of the AIP results.

hanars: AIP results are only generated for one specific family, so if multiple families match that is ambiguous and we are currently randomly assigning the results to one, instead of having the user provide disambiguation. Family IDs are not unique across projects so having the AIP upload include that data does not really fix the issue

jklugherz (approved):

PR #4199, opened 2024-06-26
remove unused param from create analysis group body

hanars:
If a user creates an analysis group by uploading a file, there is an extra field in the POST body that should be ignored

### DIFF
--- a/seqr/views/apis/analysis_group_api.py
+++ b/seqr/views/apis/analysis_group_api.py
@@ -55,6 +55,7 @@ def update_analysis_group_handler(request, project_guid, analysis_group_guid=Non
     valid_families = set()
 
     def _validate_families(request_json):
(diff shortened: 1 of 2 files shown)


seqr/views/apis/analysis_group_api.py
```diff
@@ -55,6 +55,7 @@ def update_analysis_group_handler(request, project_guid, analysis_group_guid=Non
     valid_families = set()
 
     def _validate_families(request_json):
+        request_json.pop('uploadedFamilyIds', None)
```
jklugherz, line 58: Is it possible to remove the field from the request body when it's created instead?

hanars: not easily, its would involve a pretty messy update to a shared utility helper

jklugherz (approved):

PR #4201, opened 2024-06-26
Fix lookup error on missing entry

hanars:


hail_search/queries/mito.py
```diff
@@ -331,7 +331,7 @@ def _add_project_lookup_data(self, ht, annotation_fields, *args, **kwargs):
 
         annotation_fields.update({
             'familyGenotypes': lambda r: hl.dict(r.family_entries.map(
-                lambda entries: (entries.first().familyGuid, entries.map(self._get_sample_genotype))
+                lambda entries: (entries.first().familyGuid, entries.filter(hl.is_defined).map(self._get_sample_genotype))
```
jklugherz, line 334: Why would a family entry be None here?

jklugherz: also is it worth adding a test case for this?

hanars: If there was no call for one sample at a specific site, it would end up as missing in the entries table

And probably. We do not have a table with. missing entry but its probably worth adding

jklugherz (approved):
approved, up to you if you think now is the time to add the test case

hanars:
> approved, up to you if you think now is the time to add the test case

It is, I was just being lazy 🫤

jklugherz (approved):

PR #4204, opened 2024-06-27
Use all metadata fields

hanars:
There are 3 "metadata" reports in seqr, at the Individual, Family, and Variant level. These reports work via a server side endpoint that returns lists of row dicts and then on the client side assigns desired fields to their corresponding columns in the table. If additional fields are returned from the server side endpoint that are not used in the report this does not throw an error, so these endpoints have been returning extra fields. However, this creates a maintenance headache, as it makes it really difficult to make changes to any of the shared utility functions without knowing what is and is not okay to remove or change. 

This PR makes sure every field returned in the server side metadata endpoints are used in their corresponding UI tables. In some cases this involved changing the server side behavior to stop computing certain fields  and in other cases this involved adding useful fields to the reports in the UI instead of ignoring them. As part of this effort, I removed the `include_metadata` flag from the `parse_anvil_metadata` and broke it out into more targeted configurations that add fields only where necessary. I also updated the column configuration in the Ui from the family and individual reports to explicitly share columns that are used by both reports. 

Once this change is live, all subsequent changes to these reprots will be explicit, so anything removed or added on the server side will have a corresponding client side change.


seqr/views/apis/summary_data_api.py
```diff
@@ -286,39 +286,45 @@ def _add_row(row, family_id, row_type):
         elif row_type == DISCOVERY_ROW_TYPE:
             for i, discovery_row in enumerate(row):
                 participant_id = discovery_row.pop('participant_id')
-                parsed_row = {'{}-{}'.format(k, i + 1): v for k, v in discovery_row.items()}
+                parsed_row = {'{}-{}'.format(k, i + 1): v for k, v in discovery_row.items() if k != 'allele_balance_or_heteroplasmy_percentage'}
```
jklugherz, line 289: what's the significance of excluding `allele_balance_or_heteroplasmy_percentage` here?

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -121,7 +120,12 @@ def _format_transcript_id(transcript_id, transcript):
 }
 
 
-def _get_family_metadata(family_filter, family_fields, include_metadata, include_mondo, format_id):
+def _get_family_metadata(family_filter, family_fields, include_family_name_display, include_family_sample_metadata, include_mondo, format_id):
+    family_fields = {'analysis_groups': {
+        'value': ArrayAgg('analysisgroup__name', distinct=True, filter=Q(analysisgroup__isnull=False)),
+        'format': lambda f: '; '.join(f['analysis_groups']),
+    }} if include_family_sample_metadata else family_fields
+    include_family_name_display = include_family_name_display or include_family_sample_metadata
```
jklugherz, line 128: Why are we using `include_family_sample_metadata` to reset the value of `include_family_name_display`? Can we instead be more explicit in` setting include_family_name_display=True` when we want `FAMILY_NAME_DISPLAY_VALUES`?

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -185,16 +186,16 @@ def parse_anvil_metadata(
             sample_ids.add(sample.sample_id)
 
     saved_variants_by_family = _get_parsed_saved_discovery_variants_by_family(
-        list(family_data_by_id.keys()), include_metadata, include_svs=include_svs, variant_json_fields=variant_json_fields,
+        list(family_data_by_id.keys()), bool(mme_value), include_svs, variant_json_fields, variant_attr_fields,
```
jklugherz, line 189: Why is the value of `mme_value` responsible for whether or not we include `seqr_chosen_consequence` on the parsed saved variant?

jklugherz: You could rename `include_metadata` to `include_mme_metadata` to be more explicit

seqr/views/apis/summary_data_api.py
hanars: its not used in the report and as part of this cleanup I am trying to stop returning fields that are unused. The effort to configure the function that generates the rows to omit this field was too clunky/ difficult so this seemed clearer

seqr/views/utils/anvil_metadata_utils.py
hanars: because the "family_sample_metadata" includes the display name fields, so setting `include_family_name_display` to true implies `include_family_name_display`.  It seemed messier to require multiple function to pass a redundant extra flag than to set the behavior explicitly here

seqr/views/utils/anvil_metadata_utils.py
hanars: because those fields go together, and I was trying to reduce the number of configuration options wherever I could. I could add another boolean flag to the function and pass that as true whenever I pass mme_value, but that seemed more complex than it is worth

jklugherz (approved):

PR #4208, opened 2024-06-28
VLM configure contact

hanars:
Allows users to specify the email that VLM contact should be sent to, and to explicitly disable the option to be contacted
Screenshot 2024-06-28 at 1 17 54 PM
Screenshot 2024-06-28 at 1 18 03 PM


jklugherz (approved):

seqr/migrations/0068_project_vlm_contact_email.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations, models
+
+
+def update_vlm_contact_email(apps, schema_editor):
+    Project = apps.get_model('seqr', 'Project')
+    db_alias = schema_editor.connection.alias
+
+    projects = Project.objects.using(db_alias).all()
+    for project in projects:
+        project.vlm_contact_email = project.mme_contact_url.replace('mailto:', '').replace('matchmaker', 'vlm')
```
jklugherz, line 12: what's the reasoning for creating a new email vlm@broadinstitute.org instead of just repurposing the matchmaker email?

hanars: Because different people will have different responsibilities for monitoring and responding to emails to these different accounts, and the expected responses and turnaround times are potentially different. Also, variant level matching is not part of the matchmaker exchange, that is a separate formal network we are a part of, and it would be really misleading if we were communicating to people who know about matchmaker from the matchmaker email about something thats unrelated to matchmaker

PR #4209, opened 2024-07-01
Add PMID to finding notes

hanars:


seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -165,7 +165,7 @@ def parse_anvil_metadata(
         airtable_fields: Iterable[str] = None, mme_value: Aggregate = None, include_svs: bool = True,
         variant_json_fields: Iterable[str] = None, variant_attr_fields: Iterable[str] = None, post_process_variant: Callable[[dict, list[dict]], dict] = None,
         include_no_individual_families: bool = False, omit_airtable: bool = False, include_family_name_display: bool = False, include_family_sample_metadata: bool = False,
-        include_discovery_sample_id: bool = False, include_mondo: bool = False, include_parent_mnvs: bool = False,
+        include_discovery_sample_id: bool = False, include_mondo: bool = False, omit_parent_mnvs: bool = False,
```
hanars, line 168: changes the default behavior of this function to include parent MNVs since that is the behavior we want almost everywhere, so the one place we don't want it sets that explicitly

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -305,23 +305,17 @@ def _get_genotype_zygosity(genotype):
...
-        parent_details = [parent_mnv[key] for key in ['hgvsc', 'hgvsp'] if parent_mnv.get(key)]
-        parent = f'{parent_name} ({", ".join(parent_details)})' if parent_details else parent_name
-        mnv_names = [_get_nested_variant_name(v) for v in gene_variants]
-        nested_mnvs = sorted([v for v in mnv_names if v != parent_name])
-        discovery_notes = f'The following variants are part of the {variant_type} variant {parent}: {", ".join(nested_mnvs)}'
-    return {
-        'sv_name': _get_sv_name(v),
-        'notes': discovery_notes,
-    }
+def _get_discovery_notes(variant, gene_variants, omit_parent_mnvs):
```
hanars, line 308: functionality did not change, only indenting

jklugherz (approved):
code LGTM but wondering if you can tell me more about the relationship between mnvs (multi-nucleotide-variants I think) and pmids

PR #4220, opened 2024-07-09
add hail search backend to CSRF_TRUSTED_ORIGINS

jklugherz:
this is a strong hunch but I couldn't replicate the bug locally so need to test this in dev


hanars:
I commented on the ticket but I do not think this is it, and it is a security violation to do this as all requests to the hail backend need to go through a server side proxy

PR #4222, opened 2024-07-10
set CSRF_TRUSTED_ORIGINS to BASE_URL

jklugherz:


settings.py
```diff
@@ -303,7 +299,8 @@
 SEQR_PRIVACY_VERSION = float(os.environ.get('SEQR_PRIVACY_VERSION', 1.1))
 SEQR_TOS_VERSION = float(os.environ.get('SEQR_TOS_VERSION', 1.2))
 
-BASE_URL = os.environ.get("BASE_URL", "/")
+BASE_URL = os.environ.get("BASE_URL", "http://localhost/")
```
hanars, line 302: rather than changing this, especially since we didn;t have a CSRF issue locally, it might be better to just put the `CSRF_TRUSTED_ORIGINS` under an if statement for the presence of the base url

jklugherz: I put it under the same `if DEPLOYMENT_TYPE in {'prod', 'dev'}:` check we do earlier in the file. I could also check if BASE_URL != '/'.

settings.py
```diff
@@ -304,6 +300,8 @@
 SEQR_TOS_VERSION = float(os.environ.get('SEQR_TOS_VERSION', 1.2))
 
 BASE_URL = os.environ.get("BASE_URL", "/")
+if DEPLOYMENT_TYPE in {'prod', 'dev'}:
```
hanars, line 303: probably neater to move the BASE_URL definition up higher in the file so we can use the one conditional block for this

hanars (approved):

PR #4242, opened 2024-07-17
Rna sample refactor

hanars:
This refactors the Sample model in seqr, splitting it into a search-specific model and a separate RnaSample model for tracking RNA data. The new RNA model behaves a bit differently from the old Sample model for RNA, most crucially that it stores which RNA data type is associated with it, and when new data is loaded it creates a new model rather than replacing the existing model, so we are able to show historic data loading for RNA the way we do for other samples


seqr/management/tests/deactivate_project_search_tests.py
```diff
@@ -31,7 +31,7 @@ def test_command(self, mock_logger, mock_input):
         # Test success
         mock_input.return_value = 'y'
         call_command('deactivate_project_search', PROJECT_GUID)
-        mock_logger.info.assert_called_with('Deactivated 14 samples')
+        mock_logger.info.assert_called_with('Deactivated 11 samples')
```
hanars, line 34: this is a change in behavior, but a desired one - there is no need to reactivate RNA data in a project when we are deactivating search data, as they are unrelated

seqr/migrations/0069_remove_sample_dataset_type_and_more.py
```diff
@@ -0,0 +1,83 @@
...
+
+from django.conf import settings
+from django.db import migrations, models
+from django.db.models import F
+from django.db.models.functions import Coalesce
+import django.db.models.deletion
+import django.utils.timezone
+
+
+def split_samples(apps, schema_editor):
```
hanars, line 11: Note that this migration works by copying all the non-RNA data into a new table and leaving the RNA data in the old table. This is a bit counterintuitive as, in the end, the table named "sample" (which is the old table name) has all the non-RNA samples.  However, the non-RNA samples have no foreign key references while the RNA samples do, and migrating those was prohibitively slow so it made more sense to leave the RNA samples in their exsting table to preserve their IDs

seqr/migrations/0070_remove_rnasample_dataset_type_and_more.py
```diff
@@ -0,0 +1,123 @@
+# Generated by Django 4.2.13 on 2024-07-12 20:14
+
```
hanars, line 2: this migration is split into 2 separate migrations because due to limitations in Django it is not possible to split the data into 2 tables and then rename those tables in the same migration

seqr/models.py
```diff
@@ -740,10 +728,45 @@ def __unicode__(self):
     class Meta:
        json_fields = [
            'guid', 'created_date', 'sample_type', 'dataset_type', 'sample_id', 'is_active', 'loaded_date',
-           'elasticsearch_index',
```
hanars, line 743: this was only being used on the client side to determine whether or not a sample could be used for search, which is no longer necessary and was also kind of hacky to begin with

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -923,6 +926,14 @@ def _set_file_iter_stdout(rows):
...
         })
 
+        mapping_body = {'mappingFile': {'uploadedFileId': 'map.tsv'}}
+        body.update(mapping_body)
+        mock_subprocess.side_effect = [mock_does_file_exist, mock_file_iter]
+        response = self.client.post(url, content_type='application/json', data=json.dumps(body))
+        self.assertEqual(response.status_code, 400)
+        self.assertDictEqual(response.json(), {'error': 'Must contain 2 columns: a'})
+
+        mock_load_uploaded_file.return_value = [['NA19675_D2', 'NA19675_1']]
```
hanars, line 936: in the old model we stored the sample_id and if we tried to load data where the sample_id differed from the individual id we would look and see if there was already a sample model with that ID. We no longer store the unmapped sample_id, we just store the relationship to the individual, so now you have to explicitly provide a mapping from sampel_id to individual_id for all samples where they differ, instead of only the new ones

seqr/views/apis/family_api.py
```diff
@@ -49,15 +49,6 @@ def family_page_data(request, family_guid):
     response = {
         'samplesByGuid': {s['sampleGuid']: s for s in samples}
     }
-    rna_type_samples = {
```
hanars, line 52: this was causing a performance problem in this endpoint, which is why the performance bug is also solved by theis PR

seqr/views/apis/project_api.py
```diff
@@ -215,6 +218,7 @@ def project_families(request, project_guid):
         families_by_id[family_id].update({
             'individualGuids': sorted(id_guid_map.values()),
             'hasPhenotypePrioritization': family_id in phenotype_priority_families,
+            'hasRna': family_id in rna_families,
```
hanars, line 221: the RNA samples on the project page are only used for filtering the family table, the detailed models are never used themselves, so theres no need to return all those models as long as we add this extra aggregation

ui/pages/Project/components/FamilyTable/IndividualRow.jsx
```diff
@@ -138,27 +140,17 @@ MmeStatusLabel.propTypes = {
 const DataDetails = React.memo(({ loadedSamples, individual, mmeSubmission }) => (
   <div>
     {loadedSamples.map(
-      sample => <div key={sample.sampleGuid}><Sample loadedSample={sample} isOutdated={!sample.isActive} /></div>,
+      sample => <div key={sample.sampleGuid}><Sample {...sample} isOutdated={!sample.isActive} /></div>,
     )}
-    {mmeSubmission && (
```
hanars, line 143: moved this lower because it looked weird in this order

ui/shared/components/panel/sample.jsx
```diff
@@ -16,39 +16,40 @@ const iconColor = (loadedSample, isOutdated) => {
   return isOutdated ? 'grey' : 'green'
 }
 
-const Sample = React.memo(({ loadedSample, isOutdated, hoverDetails }) => (
+const Sample = React.memo(({ sampleType, datasetType, loadedDate, hoverContent, isOutdated, hoverDetails }) => (
```
hanars, line 19: I refactored this component to work with passing all the sample props through instead of requiring an object representation of the sample, so that for displaying phenotype priority and RNA we can explicitly pass in the fields we want instead of needing to format the data as if it were a search. sample

seqr/migrations/0070_remove_rnasample_dataset_type_and_more.py
```diff
@@ -0,0 +1,123 @@
...
+            model_name='rnasample',
+            name='data_source',
+            field=models.TextField(),
+        ),
+        migrations.AlterField(
+            model_name='rnasample',
+            name='tissue_type',
+            field=models.CharField(choices=[('WB', 'whole_blood'), ('F', 'fibroblasts'), ('M', 'muscle'), ('L', 'lymphocytes'), ('A', 'airway_cultured_epithelium')], max_length=2),
+        ),
+        migrations.RunPython(add_data_type_rna_samples, reverse_code=migrations.RunPython.noop),
```
jklugherz, line 122: why make the first migration reversible but not this one?

seqr/models.py
```diff
@@ -704,23 +704,11 @@ class Sample(ModelWithGUID):
...
-        ('A', 'airway_cultured_epithelium'),
-        (NO_TISSUE_TYPE, 'None'),
-    )
-
     individual = models.ForeignKey('Individual', on_delete=models.PROTECT)
 
     sample_type = models.CharField(max_length=10, choices=SAMPLE_TYPE_CHOICES)
     dataset_type = models.CharField(max_length=13, choices=DATASET_TYPE_CHOICES)
 
-    tissue_type = models.CharField(max_length=2, choices=TISSUE_TYPE_CHOICES)
```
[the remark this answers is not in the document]
hanars: It was only used for the RNA samples, it was hardcoded to be NO_TISSUE_TYPE for all the searchable samples. It still exists on the RnaSample model

seqr/migrations/0070_remove_rnasample_dataset_type_and_more.py
hanars: This migration is reversible, this line is just saying that there is no custom python to run in the reverse migration. Thats because `add_data_type_rna_samples` creates new sample models where necessary, but if in the reverse migration we keep those new models no harm happens, and then if we re-ran the forward migration it would not recreate them new models, it would just detect that there were no instances of multiple data types sharing a model. 
In the previous migration, if we did not write reverse code than all the non-RNA models would be deleted, which would be an irreverisible change. Here, the change is reversible with no custom handling

seqr/views/apis/project_api.py
```diff
@@ -254,17 +258,19 @@ def project_families(request, project_guid):
...
     samples_by_guid = {s['sampleGuid']: s for s in get_json_for_samples(samples, project_guid=project_guid)}
 
-    sample_load_counts = sample_models.values(
-        'sample_type', 'dataset_type', loadedDate=TruncDate('loaded_date'),
-    ).order_by('loadedDate').annotate(familyCounts=ArrayAgg('individual__family__guid'))
     grouped_sample_counts = defaultdict(list)
-    for s in sample_load_counts:
+    for s in sample_load_counts + rna_sample_load_counts:
         s['familyCounts'] = {f: s['familyCounts'].count(f) for f in s['familyCounts']}
         grouped_sample_counts[f'{s.pop("sample_type")}__{s.pop("dataset_type")}'].append(s)
```
jklugherz, line 276: Should the project_overview endpoint also return rnaSamples? I imagine we want to display this in the `RNA-seq Datasets` section of the project overview UI but I haven't gotten to the UI changes in this PR yet

seqr/views/apis/report_api.py
```diff
@@ -70,10 +74,10 @@ def seqr_stats(request):
     })
 
 
-def _get_sample_counts(sample_q):
-    samples_agg = sample_q.filter(is_active=True).values('sample_type', 'dataset_type').annotate(count=Count('*'))
+def _get_sample_counts(sample_q, data_type_key='dataset_type'):
```
jklugherz, line 77: I was about to get confused about data_type vs dataset_type - but I imagine that it was a decision you made to use data_type the RNA samples because to keep the 2 concepts separate, and I think I agree with it

seqr/views/apis/project_api.py
jklugherz: ok yeah I could be wrong but I think `sample_models` will not include any RNA samples any more so you'll have to explicitly pass them either inside `samplesByGuid` or elsewhere in the response

ui/shared/components/panel/sample.jsx
jklugherz: 🙏 thanks for doing this, I always felt like I was using this component in hackyish ways for rna and phenotype priority but refactoring was out of scope

seqr/models.py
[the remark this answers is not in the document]
jklugherz: riiiight and this is because rna expression changes depending on the tissue the sample comes from

seqr/views/apis/data_manager_api.py
```diff
@@ -317,8 +317,8 @@ def _get_sample_file_path(file_dir, sample_guid):
 
 @pm_or_data_manager_required
 def load_rna_seq_sample_data(request, sample_guid):
-    sample = Sample.objects.get(guid=sample_guid)
-    logger.info(f'Loading outlier data for {sample.sample_id}', request.user)
+    sample = RnaSample.objects.get(guid=sample_guid)
+    logger.info(f'Loading outlier data for {sample.individual.individual_id}', request.user)
 
     request_json = json.loads(request.body)
     file_name = request_json['fileName']
```
bpblanken, line 324: (Unrelated to this pr, but I'm poking around in here.).  Looks like `model_cls.bulk_create` below (line 339) is missing the `batch_size` arg that exists in `management/commands/load_rna_seq.py`.  Should we add it?

seqr/views/apis/project_api.py
hanars: so we do not want to return all the RNA models, thats an intentional choice: https://github.com/broadinstitute/seqr/pull/4242#discussion_r1681656818

seqr/views/apis/report_api.py
hanars: yeah I went back and forth a couple times but ultimately felt like having it named something different made it a tiny bit clearer that its not the same thing

seqr/views/apis/data_manager_api.py
hanars: I would say if you are making a PR to change the batch sizefor a bulk create, it would be in-scope to add it there

jklugherz (approved):

PR #4248, opened 2024-07-18
Subset pedigree updates

hanars:


seqr/views/apis/data_manager_api.py
```diff
@@ -514,16 +515,69 @@ def load_data(request):
...
         project_models, sample_type, dataset_type, request_json['filePath'], request.user, success_message,
         SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, f'ERROR triggering internal {sample_type} {dataset_type} loading',
-        is_internal=True, project_samples=project_samples if dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS else None,
+        is_internal=True, individual_ids=individual_ids, additional_project_files=additional_project_files,
     )
 
     return create_json_response({'success': True})
 
 
+def _get_valid_project_samples(project_samples, sample_type):
```
hanars, line 537: Before, we would get all the project individuals, use that list for the pedigree, and then compute and validate the subset in the `trigger_data_loading` helper. Now that we want to only use the subsetted pedigree  it makes more sense to move the subsetting logic up and then in the helper method just compute the pedigree for the already-validated subset. The subsetting logic itself is very usage-specific because is closely tied to airtable so it makes more sense to have it in this file rather than in the generic helper file

bpblanken (approved):

PR #4256, opened 2024-07-23
Require affected hpo anvil project

hanars:


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -170,10 +182,11 @@ def _convert_fam_file_rows_to_json(column_map, rows, required_columns=None):
         ValueError: if there are unexpected values or row sizes
     """
     required_columns = [JsonConstants.FAMILY_ID_COLUMN, JsonConstants.INDIVIDUAL_ID_COLUMN] + (required_columns or [])
-    missing_cols = set(required_columns) - set(column_map.values())
+    missing_cols = [_to_title_case(_to_snake_case(col)) for col in set(required_columns) - set(column_map.values())]
+    if update_features and JsonConstants.FEATURES not in column_map.values():
+        missing_cols.append('HPO Terms')
```
jklugherz, line 187: Is 'HPO Terms' the column name expected [here](https://github.com/broadinstitute/seqr/pull/4256/files#diff-a3dbd9acae63c78c7ea2e5514f0945c7068606fe9c3c9688a23c78671ef24ad0R231) in the file header? 

Can we make this a constant?

seqr/views/utils/pedigree_info_utils.py
```diff
@@ -215,6 +228,8 @@ def _parse_header_columns(header, allow_id_update):
         elif 'indiv' in key and 'previous' in key:
             if allow_id_update:
                 column = JsonConstants.PREVIOUS_INDIVIDUAL_ID_COLUMN
+        elif update_features and 'hpo' in key and 'term' in key:
```
jklugherz, line 231: why don't we enforce a strict schema for the headers in this file like checking if key == 'HPO Terms'?

hanars: Because we wanted to make it easier for people to generate these files on their own, and being flexible in what the column names  are helps avoid nitpicky errors that are hard for non-technical users to manage. For things like RNA data uploads or other data genrated programmatically we are very strict about column formatting, but for use cases like these where an average user is manually making an excel file we don;t want to trip them up because we expect "HPO Terms" and they put "Hpo term"

jklugherz: this makes sense. I guess one of the only downsides is that being flexible in the code means it's a little harder for the reader to figure out what is allowed, but that's solved by examples in the unit tests

jklugherz (approved):

PR #4259, opened 2024-07-24
External anvil project- edit family ID

hanars:


seqr/views/apis/family_api.py
```diff
@@ -269,15 +269,20 @@ def update_family_fields_handler(request, family_guid):
...
+        'display_name',
+    ] + immutable_keys)
 
     return create_json_response({
-        family.guid: _get_json_for_model(family, user=request.user)
+        family.guid: _get_json_for_model(family, user=request.user, process_result=_set_display_name)
     })
 
 
+def _set_display_name(family_json, family_model):
```
jklugherz, line 282: In `update_family_from_json` there's also a few lines that set `displayName` in the json response. Do you need both?

ui/shared/components/panel/family/Family.jsx
```diff
@@ -221,25 +223,38 @@ class Family extends React.PureComponent {
...
       const familyHeader = (
-        <InlineHeader
-          key="name"
-          size="small"
-          content={showFamilyPageLink ?
-            <Link to={`/project/${family.projectGuid}/family_page/${family.familyGuid}`}>{family.displayName}</Link> :
-            family.displayName}
+        <BaseFieldView
+          field="familyId"
+          idField="familyGuid"
```
jklugherz, line 251: does `idField` need to be set here? it's not required and I don't think it's used by the `familyHeader`

seqr/views/apis/family_api.py
hanars: yeah, so in that its converting the json to be used to update the model in the database and set the fields correctly, here its pulling the data from the database and formatting it into json. So its the reverse

ui/shared/components/panel/family/Family.jsx
hanars: Its used in `BaseFieldView` to generate a unique ID for the modal thats opened to do the editing. Its not required because if its not editable theres no need, but since this can be editable we should provide it

jklugherz (approved):

PR #4261, opened 2024-07-25
No max families for hpo search

hanars:
The HPO summary page created search links that put the projects and families for the relevant families directly in the URL. This ran into browser URL size limits if there were too many families, so we disabled search over that number. Since users want to do searches in more families than the limit, this rearchitects that link to instead save the families to the state/ local storage and put a hash in the URL (similar to the behavior we use for running searches)

### DIFF
--- a/ui/pages/Search/components/PageHeader.jsx
+++ b/ui/pages/Search/components/PageHeader.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
-import { getProjectsByGuid, getFamiliesByGuid, getAnalysisGroupsByGuid, getSearchesByHash } from 'redux/selectors'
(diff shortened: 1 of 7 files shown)


ui/pages/Search/components/PageHeader.jsx
```diff
@@ -19,7 +19,14 @@ const PAGE_CONFIGS = {
     entity: analysisGroupsByGuid[entityGuid],
     entityUrlPath: `analysis_group/${entityGuid}`,
   }),
-  families: entityGuid => ({ description: `Searching in ${entityGuid.split(/[,:]/).length} Families` }),
+  families: (entityGuid, p, f, a, s, searchFamiliesByHash) => {
```
jklugherz, line 22: what do the params `p, f, a, s` represent?

hanars: `projectsByGuid, familiesByGuid, analysisGroupsByGuid, searchesByHash`

jklugherz (approved):

PR #4262, opened 2024-07-26
Freq summary display

hanars:


bpblanken (approved):

ui/shared/components/panel/search/FrequencyFilter.jsx
```diff
@@ -145,9 +145,9 @@ const callsetChange = (onChange, initialValues) => val => onChange(
   { ...initialValues, [THIS_CALLSET_FREQUENCY]: val, [SV_CALLSET_FREQUENCY]: val },
 )
 
-const freqChange = (onChange, initialValues) => val => onChange(FREQUENCIES.filter(
-  ({ name }) => name !== THIS_CALLSET_FREQUENCY && name !== SV_CALLSET_FREQUENCY,
-).reduce((acc, { name }) => ({ ...acc, [name]: val }), initialValues || {}))
+const freqChange = (onChange, initialValues) => val => onChange(FREQUENCIES.reduce((acc, { name }) => ({
+  ...acc, [name]: name !== THIS_CALLSET_FREQUENCY && name !== SV_CALLSET_FREQUENCY ? val : initialValues[name],
+}), initialValues || {}))
```
hanars, line 150: the reported issue with the AC in the header not updating occurred when a saved search was selected that still had a freq filter applied for a deprecated population (i.e. exac). The update did nto change those populations so the AC was not the same for all populations and thus it was not showing in the header. This fixes that issue by updating the filter on change to only include the currently supported/ visible to user populations

ui/shared/components/panel/search/VariantSearchFormPanels.jsx
```diff
@@ -195,7 +195,7 @@ export const FREQUENCY_PANEL = {
   name: 'freqs',
   headerProps: {
     title: 'Frequency',
-    inputSize: 10,
+    inputSize: 12,
```
hanars, line 198: the reported issue of the filter increasing from 9 to 1 is just that the input is too narrow to show a two digit number on some screens, increasing the size of the inputs should fix the issue

PR #4270, opened 2024-07-31
Reload saved variant genome build

hanars:
Project in seqr can be manually updated to a new genome version, and when this happens we attempt to lift all the saved variants, but in some edge cases that fails and we have saved variants associated with a different genome version than the project. When that has happened, we should no longer try to sync the saved variant json with the data in the search backend, as it would sync with the wrong variant/ have invalid coordinates for that build

### DIFF
--- a/seqr/views/utils/variant_utils.py
+++ b/seqr/views/utils/variant_utils.py
@@ -66,13 +66,24 @@ def update_projects_saved_variant_json(projects, user_email, **kwargs):
     return updated_variants_by_id
 
 
-def update_project_saved_variant_json(project_id, family_guids=None, dataset_type=None, user=None, user_email=None):
-    saved_variants = SavedVariant.objects.filter(family__project_id=project_id).select_related('family')
+def get_saved_variants(project_id=None, family_guids=None, dataset_type=None, genome_version=None):
+    saved_variants = SavedVariant.objects.all()
+    if project_id:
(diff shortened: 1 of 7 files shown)


seqr/views/utils/variant_utils.py
```diff
@@ -66,13 +66,22 @@ def update_projects_saved_variant_json(projects, user_email, **kwargs):
     return updated_variants_by_id
 
 
-def update_project_saved_variant_json(project_id, family_guids=None, dataset_type=None, user=None, user_email=None):
-    saved_variants = SavedVariant.objects.filter(family__project_id=project_id).select_related('family')
+def get_saved_variants(genome_version, project_id=None, family_guids=None, dataset_type=None):
+    saved_variants = SavedVariant.objects.filter(
+        Q(saved_variant_json__genomeVersion__isnull=True) |
+        Q(saved_variant_json__genomeVersion=genome_version.replace('GRCh', ''))
```
jklugherz, line 72: you could pass `db_genome_version` to this function and use it here instead of calling `replace('GRCh', ''))` again

jklugherz (approved):

seqr/views/utils/variant_utils.py
hanars: yeah but then in the other place its used I would have to call replace, which currently is not the case

PR #4271, opened 2024-08-01
fix error for loading project context for variants when no variants present

hanars:


seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -330,6 +333,17 @@ def test_saved_variant_data(self):
...
         self.assertEqual(set(response_json['familiesByGuid'].keys()), {'F000001_1', 'F000002_2', 'F000012_12'})
 
+        # Test empty project
+        empty_project_url = url.replace(PROJECT_GUID, 'R0002_empty')
+        response = self.client.get(empty_project_url)
+        self.assertEqual(response.status_code, 200)
+        empty_response = {k: {} for k in VARIANT_TAG_RESPONSE_KEYS}
+        self.assertDictEqual(response.json(), empty_response)
+
+        response = self.client.get(f'{empty_project_url}?loadProjectTagTypes=true&loadFamilyContext=true')
```
hanars, line 343: this test covers the error, without the change this test returns the same 500 issue

jklugherz: just to confirm, by 'empty' do you mean that the project has no saved variants?

jklugherz (approved):

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -330,6 +333,17 @@ def test_saved_variant_data(self):
         self.assertListEqual(variants['SV0000002_1248367227_r0390_100']['familyGuids'], ['F000002_2'])
         self.assertEqual(set(response_json['familiesByGuid'].keys()), {'F000001_1', 'F000002_2', 'F000012_12'})
 
+        # Test empty project
```
jklugherz, line 336:
```suggestion
        # Test project with no saved variants
```

jklugherz: this feels silly but if it's true it just makes it really clear what empty means and what this block of code is really testing

bpblanken (approved):

seqr/views/apis/saved_variant_api_tests.py
hanars: It has no saved variants or families. Its referred to as the empty project in several test files

seqr/views/apis/saved_variant_api_tests.py
hanars: The name of the project is empty project and this is what its called in several other places, so I am going to leave it as-is for consistency with all the other test files

PR #4274, opened 2024-08-02
search uses family and project tables in sample_type directory

jklugherz:
This PR won't change the behavior of current of search or lookup functionality. It's not a very pretty solution but it's temporary code.


hail_search/queries/base.py
```diff
@@ -296,7 +296,8 @@ def _parse_sample_data(self, sample_data):
     def _load_filtered_project_hts(self, project_samples, skip_all_missing=False, n_partitions=MAX_PARTITIONS, **kwargs):
         if len(project_samples) == 1:
             project_guid = list(project_samples.keys())[0]
-            project_ht = self._read_table(f'projects/{project_guid}.ht', use_ssd_dir=True)
+            sample_type = list(project_samples[project_guid].values())[0][0]['sample_type']
```
jklugherz, line 299: This is a gross way to get the sample_type of the first sample from the first family, and it will change in the near future when we have multiple sample types in one family.

hail_search/queries/base.py
```diff
@@ -296,7 +300,19 @@ def _parse_sample_data(self, sample_data):
     def _load_filtered_project_hts(self, project_samples, skip_all_missing=False, n_partitions=MAX_PARTITIONS, **kwargs):
         if len(project_samples) == 1:
             project_guid = list(project_samples.keys())[0]
-            project_ht = self._read_table(f'projects/{project_guid}.ht', use_ssd_dir=True)
+            # for variant lookup, project_samples looks like
+            #   {<project_guid>: {<family_guid>: True, <family_guid_2>: True}, <project_guid_2>: ...}
+            # for variant search, project_samples looks like
+            #   {<project_guid>: {<family_guid>: [<sample_data>, <sample_data>, ...], <family_guid_2>: ...}, <project_guid_2>: ...}
```
hanars, line 306: I think it would be neater to change `project_samples` to include some sort of sample type - maybe change the key from project_guid to a project/sample_type tuple. The you could just do the `hfs.exists` call in the variant lookup explicitly instead of running it for everything even when we deterministically know the sample type

jklugherz:
TODO: update v03 references in a few places

hail_search/queries/base.py
jklugherz: I agree about adding `sample_type` to the `project_samples` dict. I'm going to leave the variant lookup `project_samples` structure as is because it will change when the lookup table also has sample_type!

hanars: So I am not comfortable with the change to the `_read_table` behavior in this PR as it is now, I think it adds an inefficiency on every search to check whether or not the table exists, which is unnecessary since for all non-lookup searches we explicitly set the sample type. I think the cleanest way to solve this is to  change the `project_samples` to have the sample_type set for lookup as well, but if you want to leave it as is for now then I think you should move the `hfs.exists` call into the `if isinstance(first_family_samples, bool)` block to explicitly set the sample type, with a comment indicating that this is a temporary change until the lookup table can be refactored

hanars (changes requested):

hail_search/queries/base.py
jklugherz: You're right about the extra inefficiency of `hfs.exists`. I took your advice and changed `project_samples` to include `sample_type` in the lookup flow

hail_search/queries/mito.py
```diff
@@ -325,6 +326,15 @@ def _add_project_lookup_data(self, ht, annotation_fields, *args, **kwargs):
                 hl.dict(family_indices.map(lambda j: (lookup_ht.project_families[project_guid][j], True))),
             ))), 1),
         )[0]
+
+        for project_guid, families in variant_projects.items():
+            if hfs.exists(self._get_table_path(f'projects/WES/{project_guid}.ht', use_ssd_dir=True)):
```
bpblanken, line 331: Is there a need for `hfs` here?  `hfs` is for cloud-agnosticism but I don't think we'll ever be serving the project tables from cloud (much slower!). `os.path.exists` is probably fine?

Just for kicks:

```
In [3]: %timeit os.path.exists('a_file')
983 ns ± 10.8 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit hfs.exists('a_file')
194 µs ± 566 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

hail_search/queries/base.py
```diff
@@ -393,6 +399,14 @@ def _merge_project_hts(project_hts, n_partitions, include_all_globals=False):
     def _filter_entries_table(self, ht, sample_data, inheritance_filter=None, quality_filter=None, **kwargs):
         ht = self._prefilter_entries_table(ht, **kwargs)
 
+        # Temporarily reset sample_data
```
hanars, line 402: it might be helpful to include an indicator of how temporary it is (i.e "until full blended eS/GS support is added" or "until lookup table supports sample type")

hanars (approved):

PR #4282, opened 2024-08-09
Large gene list search timeout

hanars:
Parsing the intervals for very large gene lists is very slow in the hail backend, and for gene searches we do a post-annotation filtering step where we filter to ensure the variants is annotated in a transcript in the specified genes, and not just in the interval of said gene so we added an "optimization" where for large gene lists we do not pre-filter the entries table by interval and instead just filter for the gene IDs after the join. This PR updates that optimization to still do some interval pre-filtering without needing to parse all the thousands of intervals, by instead computing one interval per chromosome that covers all the possible genes in that chromosome. While this does still include regions between genes that are not part of the filtered genes, its still better than no interval filtering


hail_search/queries/base.py
```diff
@@ -644,25 +644,28 @@ def _parse_intervals(self, intervals, gene_ids=None, **kwargs):
...
-            reference_genome = hl.get_reference(self.GENOME_VERSION)
-            intervals = (intervals or []) + [reference_genome.x_contigs[0]]
+            intervals = [[f'chr{interval[0]}', *interval[1:]] for interval in (intervals or [])]
 
         if len(intervals) > MAX_GENE_INTERVALS and len(intervals) == len(gene_ids or []):
-            return []
+            super_intervals = defaultdict(lambda: (1e9, 0))
+            for chrom, start, end in intervals:
+                super_intervals[chrom] = (min(super_intervals[chrom][0], start), max(super_intervals[chrom][1], end))
+            intervals = [(chrom, start, end) for chrom, (start, end) in super_intervals.items()]
```
hanars, line 653: this is the actual performance change in this PR

hail_search/queries/base.py
```diff
@@ -259,7 +259,7 @@ def _load_filtered_table(self, sample_data, intervals=None, annotations=None, an
         parsed_intervals = self._parse_intervals(intervals, **kwargs)
         parsed_annotations = self._parse_annotations(annotations, annotations_secondary, **kwargs)
         self.import_filtered_table(
```
jklugherz, line 261: possibly out of scope for this PR but I noticed that `import_filtered_table` takes a parameter 'intervals' but I don't see it used anywhere inside that function, so perhaps you could delete it in this PR too

hail_search/queries/base.py
jklugherz: this is clever I like it!

jklugherz (approved):

PR #4283, opened 2024-08-09
cluster intervals

hanars:
This attempts to create a better set of intervals to use for pre-filtering large gene lists. My first attempt created one interval per chromosome, this instead merges intervals together if they aren't too far apart, increasing the allowed distance until we get a manageable number of intervals for filtration. I feel like this a very ugly algorithm and there has t be a better approach - @bpblanken @jklugherz any ideas?


hail_search/queries/base.py
```diff
@@ -647,10 +647,18 @@ def _parse_intervals(self, intervals, gene_ids=None, **kwargs):
             intervals = [[f'chr{interval[0]}', *interval[1:]] for interval in (intervals or [])]
 
         if len(intervals) > MAX_GENE_INTERVALS and len(intervals) == len(gene_ids or []):
-            super_intervals = defaultdict(lambda: (1e9, 0))
-            for chrom, start, end in intervals:
-                super_intervals[chrom] = (min(super_intervals[chrom][0], start), max(super_intervals[chrom][1], end))
-            intervals = [(chrom, start, end) for chrom, (start, end) in super_intervals.items()]
+            intervals = sorted(intervals)
```
bpblanken, line 650: one stupid question here, can we just precompute the correct intervals for all of our gene lists?

hanars: Not easily, we have thousands of gene lists and some are auto updated weekly and some are manually updated whenever a user feels like it. This code takes under a second to run, so its not performance I am worried about here so much as maintainability/stability/readability

hanars: Also,  the user is not necessarily choosing a gene list, they can also type free text lists of genes into the search box

bpblanken: Ok... I think the one maintainability thing here would be to turn this into a helper function with unit tests that demonstrate the expected behavior!

bpblanken (approved):

PR #4312, opened 2024-08-16
Gregor SV report

hanars:


seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -484,17 +469,21 @@ def _get_genetic_findings_rows(rows: list[dict], individual: Individual, family_
     for row in (rows or []):
         genotypes = row['genotypes']
         individual_genotype = genotypes.get(individual.guid) or {}
-        zygosity = _get_genotype_zygosity(individual_genotype)
+        zygosity = _get_genotype_zygosity(individual_genotype, individual, row)
+        copy_number = individual_genotype.get('cn') or -1
```
jklugherz, line 473: instead of setting `copy_number` to -1 if it doesn't exist and then to None inside `parsed_row`, can't you just do `individual_genotype.get('cn')`?

jklugherz (approved):

seqr/views/utils/anvil_metadata_utils.py
hanars: We also want to fset it to None if the value is actually -1 in the database, and this seemed cleaner than writing a bunch of special caes around `cn is None or cn < 0`

PR #4318, opened 2024-08-20
External AnVIL users - edit IGV

hanars:


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -424,9 +424,26 @@ def test_validate_anvil_vcf(self, mock_subprocess, mock_file_logger, mock_utils_
 
     @mock.patch('seqr.utils.file_utils.logger')
     @mock.patch('seqr.utils.file_utils.subprocess.Popen')
-    def test_get_anvil_vcf_list(self, mock_subprocess, mock_file_logger, mock_utils_logger):
+    def test_get_anvil_igv_options(self, *args):
+        # Requesting to load data from a workspace without an existing project
```
jklugherz, line 428:
```suggestion
```

jklugherz: this comment is in this file twice, does it apply in both spots?

seqr/views/apis/igv_api_tests.py
```diff
@@ -292,8 +293,22 @@ def test_add_alignment_sample(self, mock_local_file_exists, mock_subprocess):
...
+            'filePath': '/readviz/NA21987.cram',
+        }))
+        self.assertEqual(response.status_code, 403)
+
+        self.login_manager()
+        response = self.client.post(ext_anvil_edit_url, content_type='application/json', data=json.dumps({
+            'filePath': '/readviz/NA21987.cram',
+        }))
+        self.assertEqual(response.status_code, 200)
+
```
jklugherz, line 309: if it's not already covered, it would be nice if there was a test case where the requesting user isn't a pm or manager but is a collaborator that passes the `external_anvil_project_can_edit` check

seqr/views/apis/anvil_workspace_api_tests.py
hanars: nope good catch!

seqr/views/apis/igv_api_tests.py
hanars: This test is testing that a non-PM user is able to successfully access this endpoint for an external anvil project. Are there any other edge cases you would want to make sure are covered?

jklugherz: ah never mind then, I mistook the manager_user to be the more privileged data_manager_user.

jklugherz (approved):

PR #4319, opened 2024-08-20
Only search in genes with searched genome build

hanars:
Some genes in seqr's database only have coordinates for one genome build but both. Currently, if you do a search on those genes seqr will see that the gene is "real" and then try to do a search on an interval with null coordinates. This updates the logic to instead filter to genes with coordinates defined for the desired reference build. While this issue is most noticeable in search, I also updated any other gene queries where we have a specific genome build in the given context


seqr/utils/gene_utils_tests.py
```diff
@@ -1,60 +0,0 @@
-from django.contrib.auth.models import User
-from django.test import TestCase
-
-from seqr.utils.gene_utils import get_gene, get_genes, get_genes_for_variant_display, get_genes_for_variants, \
-    get_genes_with_detail
-from seqr.views.utils.test_utils import GENE_FIELDS, GENE_DETAIL_FIELDS, GENE_VARIANT_FIELDS, GENE_VARIANT_DISPLAY_FIELDS
-
-GENE_ID = 'ENSG00000223972'
-
-class GeneUtilsTest(TestCase):
```
hanars, line 10: I have always found it frustrating to test utility functions in isolation instead of where they are used, particularly with tests such as this one that are pretty basic. While for very complex helper functions it can be good to directly test them to ensure every case is covered, for straightforward helpers like these I would rather test that the api responses are correct, and not worry about having to keep these tests up to date whenever the function signature changes. I have ensured that everything in this fils is fully tested elsewhere (including adding some test cases elsewhere)

jklugherz (approved):

PR #4330, opened 2024-08-26
wes and wgs samples in hail search 

jklugherz:


jklugherz:
@hanars This is still in progress and some tests are failing, but can you take a look and give me some feedback as I continue working on it? I am for sure missing some test cases and probably some edge case handling in the code. It would be helpful if you could point out test cases that I need to add.

Also, I'm changing the variant response structure slightly, namely the `genotypes` dictionary, so do I have to update the elasticsearch code so that the variant response there matches what's returned from the hail search service?

hanars:
>  do I have to update the elasticsearch code so that the variant response there matches what's returned from the hail search service?

So a change to the format of the variant is not just a problem for new searches, its a problem for every saved variant ever created. In general, migrating saved variant json is a bad idea and we do not do it. So in fact changing the response of elasticearch will not even solve the problem. I think it would be better to have the UI handle this instead, and work with either the existing structure or the new structure, and only use the new structure for these multi-sample type searches and leave everything lese the same

hail_search/fixtures/GRCh38/SNV_INDEL/annotations.ht/README.txt
```diff
@@ -1,3 +1,3 @@
 This folder comprises a Hail (www.hail.is) native Table or MatrixTable.
-  Written with version 0.2.128-eead8100a1c1
-  Created at 2024/06/14 15:14:52
\ No newline at end of file
+  Written with version 0.2.130-bea04d9c79b5
+  Created at 2024/08/30 15:20:20
```
hanars, line 3: why is the annotations table changing?

hail_search/queries/base.py
```diff
@@ -360,15 +315,147 @@ def import_filtered_table(self, project_samples, num_families, **kwargs):
...
+            sample_type, entries_hts = list(entries_hts_map.items())[0]
+            ht, sorted_family_sample_data = self._add_entry_sample_families(
+                ht=entries_hts[0][0], sample_data=entries_hts[0][1], sample_type=sample_type,
+            )
+            return self._filter_entries_table(ht, sorted_family_sample_data, **kwargs)
+
+        # Project has both sample types
+        unmerged_hts, unmerged_comp_het_hts = self.filter_entries_table_both_sample_types(entries_hts_map, **kwargs)
+        merged_ht = self._merge_project_hts(unmerged_hts, n_partitions)
+        merged_comp_het_ht = self._merge_project_hts(unmerged_comp_het_hts, n_partitions)
```
hanars, line 343: `filter_entries_table_both_sample_types` is only used twice and both times its followed by the same lines of code for merging, so it would be more readable if the merge calls where part of the `filter_entries_table_both_sample_types` helper

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+                ))
+        else:
+            ht = ht.annotate(wgs_passes=ht.wgs_family_entries)
+        ht = ht.filter(ht.wes_passes.any(hl.is_defined) | ht.wgs_passes.any(hl.is_defined))
+        return ht
+
+    @staticmethod
+    def _merge_both_sample_types_project_ht(ht: hl.Table):
+        ht = ht.drop('wes_passes', 'wgs_passes')
+        ht = ht.transmute(
```
hanars, line 63: instead of a drop and a transmute you could do this as a single select

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+            family_entries=hl.coalesce(ht.wes_family_entries, hl.empty_array(ht.wes_family_entries[0].dtype)).extend(
```
hanars, line 64: I think you can use `ht.wes_family_entries.dtype.element_type` to get the element type for an array field

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+                hl.coalesce(ht.wgs_family_entries, hl.empty_array(ht.wgs_family_entries[0].dtype))
+            ),
+            filters=hl.coalesce(ht.wes_filters, hl.empty_set(hl.tstr)).union(
+                hl.coalesce(ht.wgs_filters, hl.empty_set(hl.tstr))
+            )
+        )
+        return ht.transmute_globals(family_guids=ht.family_guids.extend(ht.family_guids_1))
```
hanars, line 71: so I do not think this is going to work, I think you are going to end up with a `family_guids` global array with duplicates in it that will have the length of  <#WES families> + <#WGS families> and then for each row `family_entries` will have length of either <#WES families> + <#WGS families>, just <#WES families>, or just <#WGS families>. The `_merge_project_hts` has logic so that if a row is missing `family_entries` for a certain project it puts in an empty array of the desired length in, you may need somthing like that here

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+    def _filter_inheritance_both_sample_types(
+        self, ht, inheritance_filter, sorted_wes_family_sample_data, sorted_wgs_family_sample_data
+    ):
+        # At least one family member must have non-ref gt
+        any_valid_entry, is_any_affected = self._get_any_family_member_gt_has_alt_filter()
+        ht = ht.annotate(
+            wes_passes=ht.wes_family_entries.map(
```
hanars, line 80: shouldn't this be `ht.wes_passes.map` so we avoid failed quality families

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+            wgs_passes=ht.wgs_family_entries.map(
+                lambda entries: hl.or_missing(entries.any(any_valid_entry), entries)
+            )
+        )
+
+        comp_het_ht = None
+        if self._has_comp_het_search:
+            comp_het_ht = self._annotate_families_inheritance(
+                ht, COMPOUND_HET, inheritance_filter, sorted_wes_family_sample_data,
+                sample_type='WES',
```
hanars, line 92: we should use constants for the sample types

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+                ht.wes_passes.any(hl.is_defined) | ht.wgs_passes.any(hl.is_defined)
+            ).select_globals('family_guids', 'family_guids_1')
+
+        return ht, comp_het_ht
+
+    def _post_process_inheritance_both_sample_types(self, ht: hl.Table):
+        if self._inheritance_mode == DE_NOVO:
+            def _validate_de_novo(sample_type_1_passes, sample_type_2_passes):
+                # Remove a variant if it was filtered out due to inheritance in one sample type
+                # and is de novo in the other sample type but there is only one affected sample.
```
hanars, line 129: Is there a reason we need to special case "de novo in the other sample type but there is only one affected sample" - shouldn't the behavior just be if it passes the filter in either return it in both, if it fails in both don't return it?

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+                return hl.enumerate(sample_type_1_passes).map(
+                    lambda family_entries: hl.if_else(  # family_entries: (idx, [<family_1_sample_1>, <family_1_sample_2>, ...]) or (idx, NA)
+                        hl.is_missing(sample_type_2_passes[family_entries[0]]) & hl.is_defined(family_entries[1]),
```
hanars, line 132: so we can't assume that the entries arrays are properly sorted for both sample types such that the index of a family in one is the same as the index of a family in the other (i.e WES might be `['F1', 'F3', 'F5']` and WGS might be `['F2', 'F3']`, and theres no reason to assume that the entries within a family are the same as there could be different numbers of samples per sample type. I think you either need to do the normalizing/ merging logic before the filtering, or you need to be using the sample index data to explicitly compare the correct samples

hail_search/fixtures/GRCh38/SNV_INDEL/annotations.ht/README.txt
jklugherz: I added some variants for my new test cases

hail_search/queries/base.py
```diff
@@ -820,7 +922,7 @@ def _filter_by_annotations(self, ht, is_comp_het=False, consequence_ids=None, an
             self._has_allowed_transcript_filter(ht, field) if field in transcript_filter_fields else ht[field]
             for field in filter_fields
         ]
-        return ht.filter(hl.any(all_filters))
+        return ht if not all_filters else ht.filter(hl.any(all_filters))
```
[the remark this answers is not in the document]
jklugherz: I don't think there is one, I must've added this when a test was failing but that points to my code being wrong so I'm undoing this change.

hail_search/queries/base.py
```diff
@@ -66,7 +68,9 @@ class BaseHailTableQuery(object):
     TRANSCRIPTS_FIELD = None
     CORE_FIELDS = [XPOS]
     BASE_ANNOTATION_FIELDS = {
-        FAMILY_GUID_FIELD: lambda r: r.family_entries.filter(hl.is_defined).map(lambda entries: entries.first().familyGuid),
+        FAMILY_GUID_FIELD: lambda r: hl.array(hl.set(
+            r.family_entries.filter(hl.is_defined).map(lambda entries: entries.first().familyGuid))
+        ).filter(hl.is_defined),
```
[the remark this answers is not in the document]
jklugherz: I also don't remember why I changed this so I will remove it..

hail_search/queries/base.py
```diff
@@ -104,9 +108,10 @@ def annotation_fields(self, include_genotype_overrides=True):
         annotation_fields = {
             GENOTYPES_FIELD: lambda r: r.family_entries.flatmap(lambda x: x).filter(
                 lambda gt: hl.is_defined(gt.individualGuid)
-            ).group_by(lambda x: x.individualGuid).map_values(lambda x: self._get_sample_genotype(
-                x[0], r, select_fields=['individualGuid'], include_genotype_overrides=include_genotype_overrides,
-            )),
+            ).group_by(lambda x: x.individualGuid).map_values(lambda x: x.map(
```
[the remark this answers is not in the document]
jklugherz: that makes sense

I plan to switch this back to how it was and override the GENOTYPES_FIELD annotation to this in the mito class

hail_search/queries/both_sample_types_mixin.py
```diff
@@ -0,0 +1,141 @@
+import hail as hl
+
+from hail_search.constants import COMPOUND_HET, DE_NOVO
+
+
+class BothSampleTypesMixin:
```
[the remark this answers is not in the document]
jklugherz: I wanted to keep the both sample type stuff in a separate file but now it feels right that it should live in mito.py

hail_search/queries/base.py
```diff
@@ -360,15 +315,147 @@ def import_filtered_table(self, project_samples, num_families, **kwargs):
             self._ht = self._query_table_annotations(families_ht, self._get_table_path('annotations.ht'))
             self._ht = self._filter_annotated_table(self._ht, **kwargs)
 
-    def _filter_merged_project_hts(self, project_hts, sample_data, filtered_project_hts, filtered_comp_het_project_hts, n_partitions, **kwargs):
+    def _import_and_filter_entries_ht(self, project_samples: dict, num_families: int,  n_partitions=MAX_PARTITIONS, **kwargs):
```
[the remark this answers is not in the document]
jklugherz: I moved a few things around and hopefully the diff is more readable

hail_search/queries/base.py
```diff
@@ -360,15 +315,147 @@ def import_filtered_table(self, project_samples, num_families, **kwargs):
...
+
+    def _load_prefiltered_project_ht(self, project_guid, sample_types, entries_hts_map, project_sample_type_data, **kwargs):
+        for sample_type in sample_types:
+            path = f'projects/{sample_type}/{project_guid}.ht'
+            ht = self._read_table(path, use_ssd_dir=True)
+            entries_hts_map[sample_type] = [
+                (self._prefilter_entries_table(ht, **kwargs), project_sample_type_data[sample_type])
+            ]
+
+    def import_and_filter_multiple_project_hts(self, project_samples: dict, n_partitions=MAX_PARTITIONS, **kwargs):
```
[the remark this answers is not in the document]
jklugherz: no, it's private now

hail_search/queries/base.py
```diff
@@ -424,7 +506,12 @@ def _add_entry_sample_families(self, ht, sample_data):
             samples = sample_data[family_guid]
             if samples is True:
                 samples = ht_family_samples
-                get_sample_data = lambda s: {'sampleId': s}
+                get_sample_data = lambda s: {
+                    'sampleId': s,
+                    'individualGuid': hl.missing(hl.tstr),
+                    'affected_id': hl.missing(hl.tint),
+                    'is_male': hl.missing(hl.tbool)
```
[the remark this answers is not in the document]
jklugherz: I removed it because I don't remember 💀

hail_search/queries/base.py
```diff
@@ -467,43 +554,49 @@ def _sample_entry_data(cls, sample):
...
             # No sample-specific inheritance filtering needed
             sorted_family_sample_data = []
 
-        ht = None if self._inheritance_mode == COMPOUND_HET else self._filter_families_inheritance(
+        ht = None if self._inheritance_mode == COMPOUND_HET else self._annotate_families_inheritance(
             ht, self._inheritance_mode, inheritance_filter, sorted_family_sample_data,
         )
-
+        if ht:
+            ht = ht.filter(ht.family_entries.any(hl.is_defined)).select_globals('family_guids')
```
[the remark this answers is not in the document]
jklugherz: I did this!

hail_search/queries/base.py
```diff
@@ -529,27 +622,36 @@ def _filter_families_inheritance(self, ht, inheritance_mode, inheritance_filter,
...
                 lambda i: cls.GENOTYPE_QUERY_MAP[REF_REF](entries[i].GT)
             )
             if min_unaffected < 2:
-                unaffected_filter |= gentoype_entry_indices.size() < 2
+                unaffected_filter |= genotype_entry_indices.size() < 2
             is_valid &= unaffected_filter
         return hl.or_missing(is_valid, entries)
 
-    def _get_family_passes_quality_filter(self, quality_filter, ht, **kwargs):
+    def _get_family_passes_quality_filter(self, quality_filter, ht, filters_field_name, **kwargs):
```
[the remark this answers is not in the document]
jklugherz: I removed the argument when I refactored but this is a good point in general for keyword arg defaults

seqr/migrations/0074_alter_saved_variant_json_genotypes.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.15 on 2024-08-26 15:27
+from tqdm import tqdm
+
+from django.db import migrations
+
+def update_genotype_for_individual_to_array(apps, schema_editor):
```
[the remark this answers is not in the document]
jklugherz: removed in favor of handling different genotype structures in saved variant json on the UI!

hail_search/queries/base.py
```diff
@@ -33,7 +35,7 @@ def _to_camel_case(snake_case_str):
     return converted[0].lower() + converted[1:]
 
 
-class BaseHailTableQuery(object):
+class BaseHailTableQuery(BothSampleTypesMixin, object):
```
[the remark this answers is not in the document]
jklugherz: > Its probably better to add a direct implementation for this support to the MITO base class

this is what I'm doing now!

hail_search/queries/base.py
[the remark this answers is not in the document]
jklugherz: I did this by adding a `_get_genotypes_annotation` helper to both mito and base query classes. let me know if that's different than what you were thinking

hail_search/queries/base.py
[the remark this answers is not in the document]
jklugherz: okay, the reason I added this because I am seeing this error during my recessive test case:

TypeError: any: parameter 'collection': expected (expression of type set<any> or array<any> or expression of type bool), found list: []

And it's not returning early because is_comp_het is true.

hail_search/queries/both_sample_types_mixin.py
jklugherz: This handles case 2 [here](https://docs.google.com/presentation/d/1hqDV8ulhviUcR5C4PtNUqkCLXKDsc6pccgFVlFmWUAU/edit#slide=id.p) - and the noted outcome is to not return either variant. This is the only case that violates the OR logic that the other cases follow (if it passes the filter in either return it in both) but it would be much simpler to not have to handle this explicitly and just return it in both here.

I didn't re-implement this when I moved the both sample types code into the mito query class, but I can. 

The case is unit tested in test_both_sample_types_search (family 5). Without this special logic, a de novo search will return the variant. With the logic, it will be filtered out.

hail_search/queries/both_sample_types_mixin.py
jklugherz: This makes sense. If I don't need to handle this special case, I don't need to keep track of sample indices.

hail_search/queries/base.py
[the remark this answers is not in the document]
jklugherz: I'm not sure why this didn't happen in the existing flow, but family_entries can contain an entry that is [NA] (family has 1 sample and didn't pass checks so it will be filtered out) which is defined in hail world.

hail_search/queries/both_sample_types_mixin.py
jklugherz: I think we want to check quality and inheritance independent of each other passing or failing in this case

hail_search/queries/base.py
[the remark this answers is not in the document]
hanars: A missing entry within a family can also be a valid represention that a given sample in the family was not called in the VCF for that particular variant. In that case it would be missing in the project/family table directly, regardless of inheirtance/quality filtering

hail_search/queries/base.py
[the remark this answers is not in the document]
hanars: yes,`all_filters ` should always be populated at this point as if this is not a compound het search we should already have returned (line 804 in master), and if it is a compound het search there have to be consequence filters applied (you can't even submit a comp het search without annotation filters)

hail_search/queries/base.py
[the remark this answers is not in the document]
hanars: > I'm not sure why this didn't happen in the existing flow,

Can we confirm that this actually did happen in the existing flow? It seems unlikely that our inheritance filtering would let this through

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+
+            entries_hts_map[sample_type] = prefiltered_project_hts
+        return entries_hts_map
+
+    def _get_genotypes_annotation(self, include_genotype_overrides):
+        if not self._has_both_sample_types:
+            return super()._get_genotypes_annotation(include_genotype_overrides)
+
+        return {
+            GENOTYPES_FIELD: lambda r: r.family_entries.flatmap(lambda x: x).filter(
```
hanars, line 322: this looks very close to the base `_get_genotypes_annotation`, in fact the only difference appears to be the behavior in `map_values`. It would probably be better to not override this entire annotation, and instead add a helper for `self._get_sample_genotypes` thats passed to `map_values` which you can make conditionally different based on class

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                lambda gt: hl.is_defined(gt.individualGuid)
+            ).group_by(lambda x: x.individualGuid).map_values(lambda x: x.map(
```
hanars, line 324: it would probably be better to only return the list if there are actually multiple sample types for that individuals (or at least for that family), so adding conditionality based on the length of `x`

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+
+        self._has_both_sample_types = True
+        entries_hts_map: dict[str, list[tuple[hl.Table, dict]]] = {}
+        for sample_type in sample_types:
+            if num_families == 1:
+                family_guid = list(list(project_sample_type_data.values())[0].keys())[0]
+                ht, sample_data = self._load_prefiltered_family_ht(family_guid, sample_type, project_sample_type_data, **kwargs)
+            else:
+                ht, sample_data = self._load_prefiltered_project_ht(project_guid, sample_type, project_sample_type_data, **kwargs)
+            entries_hts_map[sample_type] = [(ht, sample_data)]
```
hanars, line 136: why make this a list of length 1? If you are only ever considering 1 project here it would be much clearer to not have a list at all

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+    def _filter_entries_ht_both_sample_types(
+        self, entries_hts_map: dict[str, list[tuple[hl.Table, dict]]], n_partitions=MAX_PARTITIONS,
+        inheritance_filter=None, quality_filter=None, **kwargs
+    ):
+        filtered_hts = []
+        filtered_comp_het_hts = []
+        wes_entries, wgs_entries = entries_hts_map[SAMPLE_TYPE_WES], entries_hts_map[SAMPLE_TYPE_WGS]  # entries_hts_map: {<sample_type>: [(ht, project_samples), ...]}
+
+        # Iterate over each chunk of project tables
+        for (wes_ht, wes_project_samples), (wgs_ht, wgs_project_samples) in zip(wes_entries, wgs_entries):
```
hanars, line 149: as noted above, it makes more sense to not have this as a list as you are only ever considering one table for each sample type in this function

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+            # Filter by inheritance
+            any_valid_entry, is_any_affected = self._get_any_family_member_gt_has_alt_filter()
+            ht = ht.annotate(wes_passes_inheritance=ht[WES_FAMILY_ENTRIES_FIELD].map(
+                lambda entries: hl.or_missing(entries.any(any_valid_entry), entries)
+            ))
+            ht = ht.annotate(wgs_passes_inheritance=ht[WGS_FAMILY_ENTRIES_FIELD].map(
+                lambda entries: hl.or_missing(entries.any(any_valid_entry), entries)
+            ))
+
+            if self._has_comp_het_search:
```
hanars, line 170: the control flow of this function looks nearly identical to that of the original `_filter_inheritance`. Instead of needing to copy-past that function here and then replicate that logic, its probably better to leave the control flow implemented once in that function and change the behavior of the helpers called by that function (like `_filter_families_inheritance`). I think the rough implementation in this class would look something like this:

```
def _filter_families_inheritance(self, ht, inheritance_mode, inheritance_filter, sorted_family_sample_data):
    if len(sorted_family_sample_data) == 1:
        return super()._filter_families_inheritance(ht, inheritance_mode, inheritance_filter, sorted_family_sample_data[0])

     ht = self._annotate_families_inheritance(ht, inheritance_mode, inheritance_filter, sorted_family_sample_data[0], SAMPLE_TYPE_WES)
     ht = self._annotate_families_inheritance(ht, inheritance_mode, inheritance_filter, sorted_family_sample_data[1], SAMPLE_TYPE_WES)
     return self._apply_filters(ht, family_idx_map)
```

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                # No sample-specific inheritance filtering needed
+                sorted_wes_family_sample_data = sorted_wgs_family_sample_data = []
+
+            if self._inheritance_mode != COMPOUND_HET:
+                ht = self._annotate_families_inheritance(ht, self._inheritance_mode, inheritance_filter, sorted_wes_family_sample_data, SAMPLE_TYPE_WES)
+                ht = self._annotate_families_inheritance(ht, self._inheritance_mode, inheritance_filter, sorted_wgs_family_sample_data, SAMPLE_TYPE_WGS)
+                ht = self._apply_filters(ht, family_idx_map)
+                filtered_hts.append(ht)
+
+        merged_ht = self._merge_project_hts(filtered_hts, n_partitions)
```
hanars, line 186: I don't think it makes sense to change the overall flow for how multiple project hts are merged as part of this PR - if at the end of the day you are generating each project ht with the correct sample types and filtering, why would the logic of combining multiple projects need to change to support this functionality?

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+
+        for family_idx, samples in enumerate(sorted_wes_family_sample_data):
+            family_guid = samples[0]['familyGuid']
+            family_guid_idx_map[family_guid][SAMPLE_TYPE_WES] = family_idx
+
+        for family_idx, samples in enumerate(sorted_wgs_family_sample_data):
+            family_guid = samples[0]['familyGuid']
+            family_guid_idx_map[family_guid][SAMPLE_TYPE_WGS] = family_idx
+
+        return hl.literal(family_guid_idx_map)
```
hanars, line 202: Is there a reason to use a generic `literal` instead of making it a dictionary?

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+
+        wgs_passes_quality_filter = self._get_family_passes_quality_filter(quality_filter, ht, WGS_FILTERS_FIELD, **kwargs)
+        wgs_passes = ht[WGS_FAMILY_ENTRIES_FIELD] if wgs_passes_quality_filter is None else (
+            ht[WGS_FAMILY_ENTRIES_FIELD].map(lambda entries: hl.or_missing(wgs_passes_quality_filter(entries), entries))
+        )
+        ht = ht.annotate(wgs_passes_quality=wgs_passes)
+        return ht
+
+    @staticmethod
+    def _apply_filters(ht, family_idx_map) -> hl.Table:
```
hanars, line 219: we should probably use a more descriptive function name, like `_apply_multi_sample_type_entry_filters`

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                        family_idx_map.get(x[1][0]['familyGuid'], hl.empty_dict(hl.tstr, hl.tint32)).contains(SAMPLE_TYPE_WGS),
+                        hl.is_defined(ht.wgs_passes_quality[family_idx_map[x[1][0]['familyGuid']][SAMPLE_TYPE_WGS]]),
+                        False
+                    )) & (hl.is_defined(ht.wes_passes_inheritance[x[0]]) | hl.if_else(
+                        family_idx_map.get(x[1][0]['familyGuid'], hl.empty_dict(hl.tstr, hl.tint32)).contains(SAMPLE_TYPE_WGS),
+                        hl.is_defined(ht.wgs_passes_inheritance[family_idx_map[x[1][0]['familyGuid']][SAMPLE_TYPE_WGS]]),
+                        False
+                    )), x[1]
+            )),
+            wgs_family_entries=hl.enumerate(ht[WGS_FAMILY_ENTRIES_FIELD]).map(
```
hanars, line 234: this looks very similar to the above `wes_family_entries` annotation, might be nice to have a shared helper function

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        wgs_passes_quality_filter = self._get_family_passes_quality_filter(quality_filter, ht, WGS_FILTERS_FIELD, **kwargs)
+        wgs_passes = ht[WGS_FAMILY_ENTRIES_FIELD] if wgs_passes_quality_filter is None else (
+            ht[WGS_FAMILY_ENTRIES_FIELD].map(lambda entries: hl.or_missing(wgs_passes_quality_filter(entries), entries))
+        )
+        ht = ht.annotate(wgs_passes_quality=wgs_passes)
+        return ht
+
+    @staticmethod
+    def _apply_filters(ht, family_idx_map) -> hl.Table:
+        # Keep family from both sample types if either passes quality and either passes inheritance
```
hanars, line 220: this does not sound right - if WES passes quality and fails inheritance while WGS fails quality and passes inheritance then it should not be returned

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        ht = ht.annotate(
+            wes_family_entries=hl.enumerate(ht[WES_FAMILY_ENTRIES_FIELD]).map(
+                lambda x: hl.or_missing(  # x[0] is family index, x[1] is list of samples
+                    (hl.is_defined(ht.wes_passes_quality[x[0]]) | hl.if_else(
```
hanars, line 224: theres no need to call `hl.if_else` with boolean expressions as hail allows you to combine boolean expressions so
```
hl.if_else(a, b, False)
```
is the exact same thing as 
```
a & b
```

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                        family_idx_map.get(x[1][0]['familyGuid'], hl.empty_dict(hl.tstr, hl.tint32)).contains(SAMPLE_TYPE_WGS),
+                        hl.is_defined(ht.wgs_passes_quality[family_idx_map[x[1][0]['familyGuid']][SAMPLE_TYPE_WGS]]),
+                        False
+                    )) & (hl.is_defined(ht.wes_passes_inheritance[x[0]]) | hl.if_else(
+                        family_idx_map.get(x[1][0]['familyGuid'], hl.empty_dict(hl.tstr, hl.tint32)).contains(SAMPLE_TYPE_WGS),
+                        hl.is_defined(ht.wgs_passes_inheritance[family_idx_map[x[1][0]['familyGuid']][SAMPLE_TYPE_WGS]]),
+                        False
+                    )), x[1]
```
hanars, line 232: this would be more readable if you use  a helper method because then the arguments would be clearer. One possible approach:
```
hl.or_missing(
    self._family_passes_quality(ht, x[0], family_idx_map.get(x[1][0]['familyGuid']), 
    x[1],
)
...
def _family_passes_quality(self, ht, wes_index, wgs_index):
    hl.is_defined(ht.wes_passes_quality[wes_index]) | (
        hl.is_defined(wgs_index) & hl.is_defined(ht.wgs_passes_quality[wgs_index]))
```

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                    )) & ( hl.is_defined(ht.wgs_passes_inheritance[x[0]]) | hl.if_else(
+                        family_idx_map.get(x[1][0]['familyGuid'], hl.empty_dict(hl.tstr, hl.tint32)).contains(SAMPLE_TYPE_WES),
+                        hl.is_defined(ht.wes_passes_inheritance[family_idx_map[x[1][0]['familyGuid']][SAMPLE_TYPE_WES]]),
+                        False
+                    )), x[1]
+            ))
+        )
+        ht = ht.transmute(
+            family_entries=hl.coalesce(
+                ht.wes_family_entries, hl.empty_array(ht.wes_family_entries.dtype.element_type)).extend(
```
hanars, line 249: Is there a reason you can't just do `ht.wes_family_entries.extend(ht.wgs_family_entries)`

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                hl.coalesce(ht.wgs_family_entries, hl.empty_array(ht.wgs_family_entries.dtype.element_type))
+            ),
+            filters=hl.coalesce(ht.wes_filters, hl.empty_set(hl.tstr)).union(
```
hanars, line 252: we should check with Lynn to see if we want to show all the WES and WGS filters merged together on the  variant or if we want to keep them separated by sample type

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        return self._filter_entries_ht_both_sample_types(entries_hts_map, n_partitions, **kwargs)
+
+
+    def _load_prefiltered_project_hts_both_sample_types(
+        self, project_samples: dict, n_partitions=MAX_PARTITIONS, **kwargs
+    ) -> dict[str, list[tuple[hl.Table, dict]]]:
+        entries_hts_map = {}
+        chunk_size = 64
+        project_guid_chunks: list[list[str]] = []
+        current_project_guid_chunk: list[str] = []
```
hanars, line 279: I commented this above, but I do not understand why the logic for merging the project tables needs to be different depending on how many sample types there are. As long as the result of reading in the multi sample type tables ins a single filtered table for the project, the existing logic should just work

hail_search/queries/base.py
```diff
@@ -360,18 +410,15 @@ def import_filtered_table(self, project_samples, num_families, **kwargs):
...
-        if ht is not None:
-            filtered_project_hts.append(ht)
-        if comp_het_ht is not None:
-            filtered_comp_het_project_hts.append(comp_het_ht)
+        ht = self._prefilter_entries_table(ht, **kwargs)
+        prefiltered_project_hts.append((ht, project_families))
 
     @staticmethod
-    def _merge_project_hts(project_hts, n_partitions, include_all_globals=False):
+    def _merge_project_hts(project_hts, n_partitions=MAX_PARTITIONS, include_all_globals=False):
```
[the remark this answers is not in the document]
jklugherz: it shouldn't be, I'm removing this change.

hail_search/queries/base.py
```diff
@@ -293,63 +309,97 @@ def _parse_sample_data(self, sample_data):
...
-        if len(project_samples) == 1:
-            project_guid = list(project_samples.keys())[0]
-            # for variant lookup, project_samples looks like
-            #   {<project_guid>: {<sample_type>: {<family_guid>: True}, <sample_type_2>: {<family_guid_2>: True}}, <project_guid_2>: ...}
-            # for variant search, project_samples looks like
-            #   {<project_guid>: {<sample_type>: {<family_guid>: [<sample_data>, <sample_data>, ...]}, <sample_type_2>: {<family_guid_2>: []} ...}, <project_guid_2>: ...}
-            sample_type = list(project_samples[project_guid].keys())[0]
-            project_ht = self._read_table(f'projects/{sample_type}/{project_guid}.ht', use_ssd_dir=True)
-            return self._filter_entries_table(project_ht, project_samples[project_guid][sample_type], **kwargs)
+    def _load_prefiltered_family_ht(
```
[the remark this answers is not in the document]
jklugherz: I think this is a good idea

hail_search/queries/mito.py
jklugherz: Ah for some reason I thought that the original logic would not ensure that the same project would be in each chunk across both sample types, but it actually does and you're right I don't need to change that logic.

hail_search/queries/mito.py
jklugherz: I'm trying this but running into issues with the hail conditional expression restriction that the if and else expressions must have same type. I also tried case() and this happened.

hail_search/queries/mito.py
jklugherz: I wanted to use the same  function `_filter_entries_ht_both_sample_types` as is used in the code path for multiple project tables, but a cleaner way to do this would be to convert them to lists when I'm calling `_filter_entries_ht_both_sample_types`.

hail_search/queries/mito.py
jklugherz: image
If that family doesn't have an entry at all for a variant in ht, it is NA after I join the two tables together.

hail_search/queries/mito.py
jklugherz: Does this sound right? 
# Keep family from both sample types if either passes quality AND inheritance

hail_search/queries/mito.py
jklugherz: I added a shared helper to reduce some code duplication

hail_search/queries/mito.py
jklugherz: good catch, using hl.dict instead here now.

hail_search/queries/mito.py
jklugherz: this function is used in the flow (called above) where there is only one table per sample type as well as for when there may be multiple tables per sample type

hail_search/queries/base.py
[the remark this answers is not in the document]
jklugherz: ah. my issues was that I created a recessive test case without including filters, so I updated my test.

hail_search/queries/base.py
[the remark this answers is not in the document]
jklugherz: I removed the filter I added here and moved this to where I filter quality/inheritance and merge wes/wgs entries. 

I believe that the filter is necessary because in the case where there is no call for a family in wgs but a passing family in wes, we will keep the wgs call even though it's an array of None, and this needs to be explicitly handled when we merge wes/wgs.

hail_search/queries/mito.py
```diff
@@ -112,6 +118,215 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
     SORTS[PATHOGENICTY_HGMD_SORT_KEY] = SORTS[PATHOGENICTY_SORT_KEY]
 
+    def _import_and_filter_entries_ht(
+        self, project_guid: str, num_families: int, project_sample_type_data, **kwargs
+    ) -> tuple[hl.Table, hl.Table]:
+        sample_types = set(project_sample_type_data.keys())
+        if len(sample_types) == 1:
+            return super()._import_and_filter_entries_ht(project_guid, num_families, project_sample_type_data, **kwargs)
+
+        self._has_both_sample_types = True
```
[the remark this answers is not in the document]
jklugherz: I don't think so, the behavior for a variant that's present in one sample type only for a family should be the same as it is in searches today, even though it will follow the 'both sample types' flow instead of the original flow - [this line](https://github.com/broadinstitute/seqr/pull/4330/files#diff-71932dea9e59d8d1ce70d88444658acffce2f3865c9afb11b57f93daf6610f66R304) will treat the missing sample type entries as failing checks for that sample type and only keep the variant if it passes in the one sample type.

hail_search/queries/base.py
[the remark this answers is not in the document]
hanars: > the case where there is no call for a family in wgs but a passing family in wes

In this case we would want to keep the WES only, not keep the missing call for WGS

hail_search/queries/mito.py
jklugherz: Yeah the entire flow for both sample types is veeery similar to the one in base.py but it's hard to make the base.py functions generic enough to handle both sample types without explicitly referencing sample types.

The divergence from the base.py filtering occurs [here](https://github.com/broadinstitute/seqr/pull/4330/files#diff-71932dea9e59d8d1ce70d88444658acffce2f3865c9afb11b57f93daf6610f66R155-R166) in `_import_and_filter_multiple_project_hts`. After that, all of the functions are similarly named and have similar behavior but the both sample types ones are suffixed "both_sample_types".

I will take another stab at this to see if I can share code between functions that have very similar logic.

hail_search/queries/mito.py
```diff
@@ -117,6 +124,236 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                ht.wgs_family_entries, hl.empty_array(ht.wgs_family_entries.dtype.element_type)
+            )).filter(lambda entries: entries.any(hl.is_defined)),
+            filters=hl.coalesce(ht.wes_filters, hl.empty_set(hl.tstr)).union(
+                hl.coalesce(ht.wgs_filters, hl.empty_set(hl.tstr))
+            )
+        )
+
+    @staticmethod
+    def _family_passes_quality_inheritance(ht, sample_type, sample_type_family_idx, family_guid, family_idx_map):
+        if sample_type == SAMPLE_TYPE_WES:
```
jklugherz, line 318: I'm thinking about creating a SampleType enum that has these fields as properties instead of a big if/else block

jklugherz:
@hanars I'm adding test cases and fixing the failing test but in the meantime this PR is ready for your eyes again

hail_search/queries/mito.py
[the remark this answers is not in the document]
jklugherz: This is also covered in the first test case in `test_both_sample_types_search` - the search has multiple projects and only one project has samples from both sample types.

seqr/views/utils/test_utils.py
```diff
@@ -914,22 +914,24 @@ def _get_list_param(call, param):
             'gnomad_genomes': {'af': 0.007},
         },
         'genotypes': {
-            'NA19675': {
+            'NA19675': [{
```
hanars, line 917: i would change these back since they should not be changed when theres only one sample type. We actually import  the fixture variants from hail_search/test_utils in a lot of our seqr app tests, so I feel like to move would be to update those tests to also include a mutli-sample variant from the hail search tests as well

hanars (commented):
This is looking really good!

hail_search/queries/gcnv.py
```diff
@@ -61,6 +61,10 @@ class GcnvHailTableQuery(SvHailTableQuery):
 
     POPULATIONS = {k: v for k, v in SvHailTableQuery.POPULATIONS.items() if k != 'gnomad_svs'}
 
+    def _read_project_table(self, project_guid: str, sample_type: str):
```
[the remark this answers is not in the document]
jklugherz: ah yes removing it from the gcnv class.

hail_search/queries/base.py
```diff
@@ -299,63 +316,98 @@ def _parse_sample_data(self, sample_data):
...
+            if len(project_hts) >= HT_CHUNK_SIZE:
+                self._prefilter_merged_project_hts(
+                    project_hts, sample_data, prefiltered_project_hts, n_partitions, **kwargs
                 )
                 project_hts = []
                 sample_data = {}
 
-        self._filter_merged_project_hts(
-            project_hts, sample_data, filtered_project_hts, filtered_comp_het_project_hts, n_partitions, **kwargs,
+        self._prefilter_merged_project_hts(
```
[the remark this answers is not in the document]
jklugherz: The intention was keep the flow close to what's happening now for a single sample type, and everything up to applying the prefilters was the same, but applying quality and inheritance filters is where we needed to start special handling for both sample types. 

I agree that it can be confusing, so I will remove 'prefilter' from most of functions that have prefilter in the name, so `_load_prefiltered_project_hts` becomes `_load_project_hts`. However `_prefilter_merged_project_hts` is the only fn for which I'll maintain the prefilter rename.

seqr/utils/search/hail_search_utils.py
```diff
@@ -243,7 +243,12 @@ def validate_hail_backend_no_location_search(samples):
         project_count=Count('individual__family__project_id', distinct=True),
     )
     from seqr.utils.search.utils import InvalidSearchException
-    if sample_counts and (len(sample_counts) > 1 or sample_counts[0]['project_count'] > 1):
-        raise InvalidSearchException('Location must be specified to search across multiple projects')
-    if sample_counts and sample_counts[0]['family_count'] > MAX_FAMILY_COUNTS[sample_counts[0]['sample_type']]:
-        raise InvalidSearchException('Location must be specified to search across multiple families in large projects')
+    if sample_counts:
```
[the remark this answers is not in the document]
jklugherz: The original check `len(sample_counts) > 1` fails for projects with multiple sample types - do we want to allow searches without location filters to be made for multiple projects, some having both sample types?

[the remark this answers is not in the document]
hanars: Location should be required to search across multiple projects always, regardless of the number of sample types per project. We do want to allow searches in a single project with multiple sample types to run without a location filter

hail_search/queries/mito.py
```diff
@@ -117,6 +121,217 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+    def _filter_inheritance_both_sample_types(self, ht, inheritance_filter, sorted_wes_family_sample_data, sorted_wgs_family_sample_data):
+        family_idx_map = self._build_family_index_map(sorted_wes_family_sample_data, sorted_wgs_family_sample_data)
+        any_valid_entry, is_any_affected = self._get_any_family_member_gt_has_alt_filter()
+        for sample_type in SampleType:
+            ht = ht.annotate(**{
+                sample_type.passes_inheritance_field: ht[sample_type.family_entries_field].map(
+                    lambda entries: hl.or_missing(entries.any(any_valid_entry), entries)
+            )})
+
+        comp_het_ht = None
```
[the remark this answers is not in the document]
jklugherz: I did some refactoring, now `_filter_inheritance_both_sample_types` calls `super()._filter_inheritance` and I've abstracted out the logic in `_filter_inheritance` into methods that are overriden by the mito class.

hail_search/queries/mito.py
```diff
@@ -117,6 +121,217 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+    @staticmethod
+    def _family_passes_quality_inheritance(ht, sample_type, sample_type_family_idx, family_guid, family_idx_map):
+        return (
+            hl.is_defined(ht[sample_type.passes_quality_field][sample_type_family_idx]) &
+            hl.is_defined(ht[sample_type.passes_inheritance_field][sample_type_family_idx])
+        ) | (
+            (hl.is_defined(family_guid) & family_idx_map.get(family_guid).contains(sample_type.other_sample_type.value)) &
+            hl.bind(lambda other_sample_type_family_idx: (
+                hl.is_defined(ht[sample_type.other_sample_type.passes_quality_field][other_sample_type_family_idx]) &
+                hl.is_defined(ht[sample_type.other_sample_type.passes_inheritance_field][other_sample_type_family_idx])
```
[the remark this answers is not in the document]
jklugherz: sounds good!

hail_search/queries/mito.py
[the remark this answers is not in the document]
jklugherz: ok I did even more refactoring and simplified the inheritance filtering and quality filtering so `_filter_inheritance_both_sample_types` doesn't even exist any more

seqr/views/utils/test_utils.py
jklugherz: so I want to just add a mutli-sample variant to the expected api response in `VariantSearchAPITest` - do you know if I'll have to update any of the django fitxures?

hail_search/queries/mito.py
```diff
@@ -117,6 +121,217 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        self, project_guid: str, num_families: int, project_sample_type_data, **kwargs
+    ) -> tuple[hl.Table, hl.Table]:
+        sample_types = set(project_sample_type_data.keys())
+        if len(sample_types) == 1:
+            # Import and filter entry table for one family or one project with one sample type
+            return super()._import_and_filter_entries_ht(project_guid, num_families, project_sample_type_data, **kwargs)
+
+        self._has_both_sample_types = True
+        entries = {}
+        for sample_type in sample_types:
```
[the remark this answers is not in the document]
jklugherz: extracted shared code into `_load_family_or_project_ht`

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -262,10 +262,13 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
   if (!variant.genotypes) {
     return null
   }
-  const genotype = variant.genotypes[individual.individualGuid]
+
+  let genotype = variant.genotypes[individual.individualGuid]
   if (!genotype) {
     return null
   }
+  // Temporarily use the first genotype for an individual until blended es/gs are supported in UI
```
hanars, line 270: If we have a ticket for this, put the ticket number in this comment

hail_search/test_utils.py
```diff
@@ -418,8 +423,534 @@
...
+    "transcripts": {},
+    "sortedMotifFeatureConsequences": None,
+    "sortedRegulatoryFeatureConsequences": None,
+    "_sort": [None],
+    "genomeVersion": "38",
+}
+
+FAMILY_5_SAMPLE_DATA = {
+    'SNV_INDEL': [
+        {'sample_id': 'BON_B15-95_1_D1', 'individual_guid': 'I00001_bon_b15_95_1_d1', 'family_guid': 'F000005_5',
```
hanars, line 937: As above, do we really need a new family for this or could we add additional sample type data for one of the existing families?

hail_search/test_search.py
```diff
@@ -341,6 +348,63 @@ async def test_multi_project_search(self):
...
+            if 'I000015_na20885' in v['genotypes']:
+                v['genotypes']['I000015_na20885'].append({**v['genotypes']['I000015_na20885'][0], 'sampleType': 'WGS'})
+            expected_results.append(v)
+
+        await self._assert_expected_search(
+            expected_results, gene_counts=GENE_COUNTS, sample_data=MULTI_PROJECT_SAMPLE_TYPES_SAMPLE_DATA,
+        )
+
+        # Variant in family_4 is de novo in exome but maternally inherited in genome.
+        # Genome passes quality and inheritance, show genotypes for both sample types.
```
hanars, line 368: I think these tests would be more maintainable if we used the existing families. Since we do not have WGS data for family 2 you could add a WGS table for that family with variants that cover you edge cases with the otherwise existing fixtures. `VARIANT1` is already de novo in exome, so adding the same variant as maternally inherited in genome seems easier that adding and remembering about these new variants

hail_search/test_search.py
```diff
@@ -341,6 +348,63 @@ async def test_multi_project_search(self):
...
+        )
+        # Exome passes quality and inheritance, show genotypes for both sample types.
+        inheritance_mode = 'de_novo'
+        await self._assert_expected_search(
+            [FAMILY_4_VARIANT], sample_data=FAMILY_4_SAMPLE_DATA, inheritance_mode=inheritance_mode,
+        )
+
+        # Variant 1 in family_5 is inherited in exome and there is no parental data in genome.
+        # Variant 2 in family_5 is inherited in exome with low parental GQ.
+        # Variant 1 genome passes quality and inheritance (since there is no parental data), show genotypes for both sample types.
```
hanars, line 382: You can test families with no paternal data without creating new fixtures by changing the sample_data in the request to not have the paternal sample - seqr will not consider samples with no sample data in the request body in search even if they are in the table and actually removes them in `_add_entry_sample_families`

hanars (changes requested):

hail_search/queries/base.py
```diff
@@ -303,63 +319,65 @@ def _parse_sample_data(self, sample_data):
...
+        project_ht = self._read_project_table(project_guid, sample_type)
+        if project_ht is not None:
+            project_ht = project_ht.select_globals('sample_type', 'family_guids', 'family_samples')
+            project_hts.append(project_ht)
+            sample_data.update(family_sample_data)
+
+        return project_ht
+
+    def import_filtered_table(self, project_samples: dict, num_families: int, **kwargs):
+        if num_families == 1 or len(project_samples) == 1:
```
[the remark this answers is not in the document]
jklugherz: yes, if there's multiple families in a search for a single project, we still only want to follow the code path that loads 1 table

hail_search/queries/mito.py
```diff
@@ -118,6 +122,177 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+
+    @staticmethod
+    def _family_passes_quality_inheritance(ht, sample_type, sample_type_family_idx, family_guid, family_idx_map):
+        # Note: This logic does not sufficiently handle case 2 here https://docs.google.com/presentation/d/1hqDV8ulhviUcR5C4PtNUqkCLXKDsc6pccgFVlFmWUAU/edit?usp=sharing
+        # and will need to be changed to support it.
+        return (
+            hl.is_defined(ht[sample_type.passes_quality_field][sample_type_family_idx]) &
+            hl.is_defined(ht[sample_type.passes_inheritance_field][sample_type_family_idx])
+        ) | (
+            (hl.is_defined(family_guid) & family_idx_map.get(family_guid).contains(sample_type.other_sample_type.value)) &
```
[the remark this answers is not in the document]
jklugherz: `family_guid` is not defined if there are no calls for the family for a variant. 
image

hail_search/queries/mito.py
```diff
@@ -118,6 +122,177 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        for project_guid, project_sample_type_data in project_samples.items():
+            for sample_type, family_sample_data in project_sample_type_data.items():
+                project_ht = self._read_project_data(
+                    family_sample_data, project_guid, sample_type, project_hts[sample_type], sample_data[sample_type]
+                )
+                if project_ht is None:
+                    continue
+
+            # Merge both WES and WGS project_hts when either of their lengths reaches the chunk size
+            if len(project_hts[SampleType.WES.value]) >= HT_CHUNK_SIZE or len(project_hts[SampleType.WGS.value]) >= HT_CHUNK_SIZE:
```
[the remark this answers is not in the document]
hanars: I mean the issue with this is it could be pretty unbalanced - you could have 64 WES projects and only 2 WGS projects but you are still going to do the merge operation on those 2 WGS, when in fact maybe you only have a small number of WGS tables total. 

I think what would make this a bit neater is if you actually differentiated between projects with multiple sample types and those with one. So before you call this function in `_import_and_filter_multiple_project_hts` when you are iterating through `project_samples` put all the projects with only 1 sample type in one list and all those with 2 sample types in the other. Then use `super()` to filter and merge all the single sample type projects, and call this helper only with projects that have both sample types. This means that here the number of WES and WGS tables will always be identical, so when you do this merge step it will be as balanced as possible and so you will minimize theoverall number of merges

hail_search/queries/mito.py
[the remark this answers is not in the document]
hanars: That makes sense, but then it seems like if there was no call in the first sample but calls in the second sample this behavior would be wrong. I think rather than checking `family_guid` here it would be better to only call this helper above if theres a valid sample, and for getting the family_guid use `hl.coalesce(samples)[0]['familyGuid']`:
```
hl.or_missing(samples.any(hl.is_defined) &  self._family_passes_quality_inheritance(...
```

hail_search/queries/mito.py
```diff
@@ -118,6 +122,177 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        if ht is None:
+            return ht
+
+        # Keep family from both sample types if either passes quality AND inheritance
+        for sample_type in SampleType:
+            ht = ht.annotate(**{
+                sample_type.family_entries_field: hl.enumerate(ht[sample_type.family_entries_field]).starmap(
+                    lambda i, samples: hl.or_missing(
+                        self._family_passes_quality_inheritance(
+                            ht, sample_type, i, samples[0]['familyGuid'], family_idx_map
```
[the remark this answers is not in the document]
jklugherz: so I moved this code here but it still feels kind of messy, lmk if you have suggestions for making the new stuff more readable

hail_search/queries/mito.py
```diff
@@ -118,6 +122,178 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+        if ht is None:
+            return ht
+
+        # Keep family from both sample types if either passes quality AND inheritance
+        for sample_type in SampleType:
+            ht = ht.annotate(**{
+                sample_type.family_entries_field: hl.enumerate(ht[sample_type.family_entries_field]).starmap(
+                    lambda i, samples: hl.or_missing(
+                        (hl.case()
+                        .when(hl.is_defined(samples[0]['familyGuid']),
```
hanars, line 256: So a case statement with exactly one clause is unnecessary. Also, I think this would actually be more readable if you don't have the current helper function and instead move more of the logic inline here and just have a helper function for the shared code of evaluating whether a family passes for a given sample type. Then here use `hl.bind` to evaluate the other sample type index. Also, per my comment on the last review, we should handle the case where the first sample is missing a call but other samples in the family are not. You should be able to rewrite this as 
```
 hl.or_missing(
    hl.bind(
        lambda other_sample_type_idx: (
            self._family_has_valid_sample_type_entries(ht, sample_type, i) | 
            self._family_has_valid_sample_type_entries(ht, sample_type.other_sample_type, other_sample_type_idx)
        ),
        family_idx_map.get(samples.find(hl.is_defined['familyGuid']).get(sample_type.other_sample_type.value),
   ), samples)

...

@staticmethod
def _family_has_valid_sample_type_entries(ht, sample_type, sample_type_family_idx):
    return (
        hl.is_defined(sample_type_family_idx) &
        hl.is_defined(ht[sample_type.passes_quality_field][sample_type_family_idx]) &
        hl.is_defined(ht[sample_type.passes_inheritance_field][sample_type_family_idx])
    )
```

jklugherz: This is much cleaner, thank you for helping me write it. I'm just tweaking `samples.find(hl.is_defined['familyGuid'])` to `hl.coalesce(samples)[0]['familyGuid']`.

hail_search/queries/mito.py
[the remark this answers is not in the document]
jklugherz: I did this. I also refactored `_load_project_hts()` on the base class as best I could so that we aren't duplicating the chunk size checking logic. Let me know what you think.

hail_search/queries/base.py
```diff
@@ -21,6 +21,11 @@
 # https://github.com/broadinstitute/seqr-private/issues/1283#issuecomment-1973392719
 MAX_PARTITIONS = 12
 
+# Need to chunk tables or else evaluating table globals throws LineTooLong exception
+# However, minimizing number of chunks minimizes number of aggregations/ evals and improves performance
+# Adapted from https://discuss.hail.is/t/importing-many-sample-specific-vcfs/2002/8
+HT_CHUNK_SIZE = 64
```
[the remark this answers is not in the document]
jklugherz: In my latest refactor I'm only checking chunk size once so I put it back inside that function.

hail_search/queries/mito.py
```diff
@@ -118,6 +121,179 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+                )
+                if ht is not None:
+                    filtered_project_hts.append(ht)
+                if comp_het_ht is not None:
+                    filtered_comp_het_project_hts.append(comp_het_ht)
+
+        return self._merge_filtered_hts(filtered_comp_het_project_hts, filtered_project_hts, n_partitions)
+
+    def _load_project_hts_both_sample_types(self, project_samples, n_partitions, **kwargs):
+        def initialize_project_hts():
```
jklugherz, line 184: I can't tell if this is an antipattern. I didn't want to make these class-level functions because I wanted them to be called only when _load_project_hts_both_sample_types on the mito class is called, not from [this](https://github.com/broadinstitute/seqr/pull/4330/files#diff-71932dea9e59d8d1ce70d88444658acffce2f3865c9afb11b57f93daf6610f66R160) super() call.

hanars: Alternatively, maybe you could leave the data structures alone and just call `_load_project_hts` once per sample type? There might be an edge case I am missing, but if not I think it would be relatively straightforward to change `_load_project_hts ` to return the hts and the sample data and then just do something like this in `_import_and_filter_multiple_project_hts`:

```
self._has_both_sample_types = True
wes_entries =  self._load_project_hts(
    both_sample_type_project_samples[SampleType.WES.value], n_partitions, **kwargs,
)
wgs_entries = self._load_project_hts(
    both_sample_type_project_samples[SampleType.WGS.value], n_partitions, **kwargs,
)
for i, (wes_ht, wes_project_samples) in enumerate(wes_entries):
    wgs_ht, wgs_project_samples = wgs_entries[i]
    ...
```

jklugherz: ok, I made a few changes to simply this section and added some post-processing in the mito `_load_project_hts`

jklugherz: oh I just saw your second comment that is even better

hail_search/queries/mito.py
```diff
@@ -118,6 +121,156 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+            # Import and filter entry table for one family or one project with one sample type
+            return super()._import_and_filter_entries_ht(project_guid, num_families, project_sample_type_data, **kwargs)
+
+        self._has_both_sample_types = True
+        entries = {}
+        for sample_type in sample_types:
+            ht, sample_data = self._load_family_or_project_ht(
+                num_families, project_guid, project_sample_type_data, sample_type, **kwargs
+            )
+            entries[sample_type] = (ht, sample_data)
```
hanars, line 138: nitpciky, but slighlty neater to do this in one line: 
```suggestion
            entries[sample_type] = self._load_family_or_project_ht(
                num_families, project_guid, project_sample_type_data, sample_type, **kwargs
            )
```

hail_search/queries/mito.py
```diff
@@ -118,6 +121,156 @@ class MitoHailTableQuery(BaseHailTableQuery):
...
+    def _import_and_filter_multiple_project_hts(
+        self, project_samples: dict, n_partitions=MAX_PARTITIONS, **kwargs
+    ) -> tuple[hl.Table, hl.Table]:
+        single_sample_type_project_samples = {}
+        both_sample_type_project_samples = defaultdict(dict)
+        for project_guid, sample_dict in project_samples.items():
+            if len(sample_dict) == 1:
+                single_sample_type_project_samples[project_guid] = sample_dict
+            else:
+                both_sample_type_project_samples[SampleType.WES][project_guid] = {SampleType.WES.value: sample_dict[SampleType.WES]}
```
hanars, line 153: Is there a reason the new dictionary uses `SampleType.WES.value` as the key but `sample_dict` and `both_sample_type_project_samples` just use `SampleType.WES` without the `.value`

hail_search/queries/mito.py
```diff
@@ -354,9 +508,10 @@ def _add_project_lookup_data(self, ht, annotation_fields, *args, **kwargs):
         if not variant_projects:
             raise HTTPNotFound()
 
+        self._has_both_sample_types = True
```
hanars, line 511: is this always true?

hail_search/test_search.py
```diff
@@ -189,6 +196,16 @@
...
 
+MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS = [
+    deepcopy(v) for v in
+    [PROJECT_2_VARIANT, MULTI_PROJECT_VARIANT1, MULTI_PROJECT_VARIANT2, VARIANT3, VARIANT4]
+]
+for v in MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS:
+    for indiv_id, gts in v['genotypes'].items():
+        v['genotypes'][indiv_id] = [gts]
+    if 'I000015_na20885' in v['genotypes']:
+        v['genotypes']['I000015_na20885'].append({**v['genotypes']['I000015_na20885'][0], 'sampleType': 'WGS'})
```
hanars, line 207: rather than having an if statement in the loop, it would be better to explicitly use `PROJECT_2_VARIANT_BOTH_SAMPLE_TYPES` 
```suggestion
MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS = [PROJECT_2_VARIANT_BOTH_SAMPLE_TYPES] + [
    deepcopy(v) for v in
    [MULTI_PROJECT_VARIANT1, MULTI_PROJECT_VARIANT2, VARIANT3, VARIANT4]
]
for v in MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS[1:]:
    for indiv_id, gts in v['genotypes'].items():
        v['genotypes'][indiv_id] = [gts]
```

ui/shared/components/panel/fixtures.js
```diff
@@ -22,7 +22,7 @@ export const VARIANT = {
   genomeVersion: "37",
   geneIds: ['ENSG00000228198'],
   genotypes: {
-    NA19675: {
+    NA19675: [{
```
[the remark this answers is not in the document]
hanars: bump

hanars (commented):
This is really close to done! Just a couple nitpck little comments, and then a couple high level test things:
- I think we should be able to remove the Project_4.ht fixtures
- In variant_search_api_tests in the main seqr app we import `HAIL_BACKEND_SINGLE_FAMILY_VARIANTS` from the hail search utils files, to make sure that the seqr app can handle the actual variant schema returned from the real hail backend. I think updating that list so instead of `VARIANT2` it has `VARIANT2_BOTH_SAMPLE_TYPES` would be a good idea, as it means if tere are any seqr app bugs related to the new genotypes schema we would catch them here

hail_search/queries/mito.py
jklugherz: This is set to True because it forces `_get_sample_genotype` to return a list of familyGenotypes instead of getting the genotype for the first sample. 

A few commits ago I had an additional flag `all_samples`  but decided to use _has_both_sample_types instead

hail_search/test_search.py
jklugherz: this didn't quite work, but I added a second loop through `MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS[:-2]` (because variants 3 and 4 don't have gts for that individual) to add the WGS gt

jklugherz: I realized that variants 3 and 4 probably shouldn't have genotype arrays for individuals, but since `_has_both_sample_types` gets set for an entire request, as the code is written right now there will be an array with a single element.

I remember trying to conditionally return a genotype array or dictionary based on # of unique sampleTypes there were for an individual, but I was running into hail's rule of conditional expressions that the return value must be the same type.

hanars: yeah I think it will be okay to have everything returned as an array instead of a dict to get the hail type checking to pass here. We just need to remember in the UI that the presence of a list does not always guarantee multiple genotypes

hanars (approved):
Approved 🎉

PR #4331, opened 2024-08-26
Better airtable enabled checks

hanars:
While working to make the data loading interface work for local installs I discovered we are not robustly checking and error handling when airtable requests are made while airtable is not configured. This adds a validation when starting up airtable sessions to sanity check that airtable is configured, and then adds explicit checks where needed to avoid running into that case

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -31,8 +31,8 @@ logger = SeqrLogger(__name__)
 MONDO_BASE_URL = 'https://monarchinitiative.org/v3/api/entity'
 
 
-anvil_enabled_analyst_required = active_user_has_policies_and_passes_test(
-    lambda user: user_is_analyst(user) and anvil_enabled())
+airtable_enabled_analyst_required = active_user_has_policies_and_passes_test(
+    lambda user: user_is_analyst(user) and AirtableSession.is_airtable_enabled())
 
 
 @pm_or_analyst_required
@@ -115,7 +115,7 @@ PHENOTYPE_PROJECT_CATEGORIES = [
 ]
 
 
-@anvil_enabled_analyst_required
+@airtable_enabled_analyst_required
 def anvil_export(request, project_guid):
     project = get_project_and_check_permissions(project_guid, request.user)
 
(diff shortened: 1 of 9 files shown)


seqr/views/apis/report_api.py
```diff
@@ -111,7 +115,7 @@ def _get_sample_counts(sample_q, data_type_key='dataset_type'):
 ]
 
 
-@analyst_required
+@airtable_enabled_analyst_required
```
hanars, line 118: The 2 endpoints with the added airtable check in this file are already not available in the UI for local users so they would not be hitting these endpoints anyways, but its better to be explicit here

seqr/views/apis/summary_data_api.py
```diff
@@ -274,7 +274,7 @@ def _search_new_saved_variants(family_variant_ids: list[FamilyVariantKey], user:
 def _get_metadata_projects(request, project_guid):
     is_analyst = user_is_analyst(request.user)
     is_all_projects = project_guid == ALL_PROJECTS
-    include_airtable = 'true' in request.GET.get('includeAirtable', '') and is_analyst and not is_all_projects
+    include_airtable = 'true' in request.GET.get('includeAirtable', '') and AirtableSession.is_airtable_enabled() and is_analyst and not is_all_projects
```
hanars, line 277: the checkbox to add this param is already hidden for local installs so this would never actually be in the URL without airtable, but its still better to have explicit server side filtering

seqr/views/apis/report_api.py
```diff
@@ -31,6 +31,10 @@
 MONDO_BASE_URL = 'https://monarchinitiative.org/v3/api/entity'
 
 
+airtable_enabled_analyst_required = active_user_has_policies_and_passes_test(
```
jklugherz, line 34: should this live in `permissions_utils.py`?

jklugherz (approved):

seqr/views/apis/report_api.py
hanars: For permissions that are specific to a single file we tend to put them directly in the file, like for `anvil_workspace_access_required` which is only in the anvil_workspace_api file. So it could go either way

PR #4332, opened 2024-08-26
Es data added notifications

hanars:
Fixes the error reported https://github.com/broadinstitute/seqr/discussions/4085#discussioncomment-10456344
Also since ES is only used for local installs, simplifies and removes all Broad-specific notification when ES data is loaded


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -111,8 +112,9 @@ def handle(self, *args, **options):
         session = AirtableSession(user=None, no_auth=True)
         for project, sample_ids in samples_by_project.items():
             project_sample_data = update_sample_data_by_project[project.id]
-            is_internal = notify_search_data_loaded(
-                project, dataset_type, sample_type, inactivated_sample_guids,
+            is_internal = not project_has_anvil(project) or is_internal_anvil_project(project)
```
hanars, line 115: This was actually recommended on my split PDO PR but I opted against it because of the chnges to shared functionality. Since the function is no longer shared, I agree that it makes more sense to split it out here

hanars: @jklugherz any feedback blocking your approval?

jklugherz (approved):

PR #4336, opened 2024-08-28
Local install loading UI

hanars:
The loading Ui was already being shown to local users, it would just be very broken if they tried to use it. This changes the behavior to trigger the pipeline runner API for local installs and bypass all calls to airtable


seqr/views/utils/export_utils.py
```diff
@@ -97,9 +98,14 @@ def export_multiple_files(files, zip_filename, **kwargs):
         return response
 
 
-def write_multiple_files_to_gs(files, gs_path, user, **kwargs):
+def write_multiple_files(files, file_path, user, **kwargs):
+    is_gs_path = is_google_bucket_file_path(file_path)
+    if not is_gs_path:
+        os.makedirs(file_path, exist_ok=True)
     with TemporaryDirectory() as temp_dir_name:
+        dir_name = temp_dir_name if is_gs_path else file_path
```
jklugherz, line 106: You could conditionally create the temporary directory based on is_gs_path, but the resource usage of creating it and not using it is probably negligible.

jklugherz (approved):
exciting stuff!!

PR #4337, opened 2024-08-29
better interval filtering and partitioning

hanars:
Improves large gene search performance in a couple ways:
 - Allows for a larger number of clustered intervals (meaning far fewer variants outside of the true intervals are included)
 - Reduces the number of partitions. Hail does not allow for specifying both intervals and number of partitions at table read time, but the number of partitions based on the large number of intervals is too high and reduces performance. While a full `repartiton` is far too slow to be helpful, `naive_coalesce` reduces the number of paritions quickly and does not seem to introduce any problems with how balanced the partitions are

### DIFF
--- a/hail_search/queries/base.py
+++ b/hail_search/queries/base.py
@@ -6,7 +6,7 @@ import os
 
 from hail_search.constants import AFFECTED_ID, ALT_ALT, ANNOTATION_OVERRIDE_FIELDS, ANY_AFFECTED, COMP_HET_ALT, \
     COMPOUND_HET, GENOME_VERSION_GRCh38, GROUPED_VARIANTS_FIELD, ALLOWED_TRANSCRIPTS, ALLOWED_SECONDARY_TRANSCRIPTS,  HAS_ANNOTATION_OVERRIDE, \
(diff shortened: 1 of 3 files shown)


hail_search/queries/base.py
```diff
@@ -237,7 +237,8 @@ def __init__(self, sample_data, sort=XPOS, sort_metadata=None, num_results=100,
         self._has_secondary_annotations = False
         self._is_multi_data_type_comp_het = False
         self.max_unaffected_samples = None
-        self._load_table_kwargs = {'_n_partitions': min(MAX_PARTITIONS, (os.cpu_count() or 2)-1)}
+        self._n_partitions = min(MAX_PARTITIONS, (os.cpu_count() or 2)-1)
+        self._load_table_kwargs = {'_n_partitions': self._n_partitions}
```
jklugherz, line 241: Why did you keep `self._load_table_kwargs` here? To me it seems superfluous since it's referencing the same thing as ` self._n_partitions`.

hanars: So when any table is read in `_load_table_kwargs` is passed into the read function. Under certain search conditions and data types certain searches use different kwargs, so essentially this is setting the default value and then it may be updated elsewhere. It would be impossible to replicate that behavior cleanly without leaving this as is

jklugherz (approved):

PR #4350, opened 2024-09-04
allow service to start with missing annotations tables

bpblanken:


hail_search/queries/base.py
```diff
@@ -89,9 +89,12 @@ class BaseHailTableQuery(object):
 
     @classmethod
     def load_globals(cls):
-        ht_path = cls._get_table_path('annotations.ht')
-        ht_globals = hl.eval(hl.read_table(ht_path).globals.select(*cls.GLOBALS))
-        cls.LOADED_GLOBALS = {k: ht_globals[k] for k in cls.GLOBALS}
+        try:
```
bpblanken, line 92: This is probably not the best long term solution (I think we discussed doing a "dataset type/reference genomes" enabled flag), but the service flails in `helm` right now until the pipeline runs and that's not ideal; this at least get the service up.

jklugherz, line 92: do we need to call this function again once the pipeline runs and annotations table exists / or will it be called again when the hail search service restarts after the pipeline runs?

bpblanken: In our environment this isn't an issue though.

jklugherz (approved):

hail_search/queries/base.py
hanars: I don't feel great about putting this in prod - `In our environment this isn't an issue though` is not necessarily true, as in the past if there was a problem with how the disk mounted or if the data got corrupted somehow the pod start up would fail and kubernetes would either make a new pod or at least it would not take down the old pods until there were successful new pods. Now it might silently fail to read any data but the pod would start up and the old pods would be deleted and then search would be down

hanars: I think it would be better to decide on a plan for explicitly configuring data types rather than have this as the interim solution

hanars (changes requested):
I think it would be better to decide on a plan for explicitly configuring data types rather than have this as the interim solution

bpblanken:
Closing in deference of https://github.com/broadinstitute/seqr/issues/4360

PR #4351, opened 2024-09-04
Enable Azure Entra authentication 

gregsmi:
This PR refactors the login and authentication code to allow AzureAD (Entra) to serve as an OAuth authentication provider.
- configured `social_core.backends.azuread_tenant.AzureADV2TenantOAuth2` as auth backend option
- `settings.py`: added `SOCIAL_AUTH_AZUREAD_V2_TENANT_OAUTH2_*` settings to mirror `SOCIAL_AUTH_GOOGLE_OAUTH2_*` settings and updated `SOCIAL_AUTH` pipeline to account for possibility of AzureAD auth
- configured `SOCIAL_AUTH_PROVIDER` variable to reflect active provider and passed to `redux` init as `oauthLoginProvider` for use in `react` state store
- replaced `google_auth_enabled` with `oauth_enabled`
- replaced `is_google_authenticated` with `is_cloud_authenticated`
- replaced `getGoogleLoginEnabled` selector with `getOauthLoginEnabled` and added `getOauthLoginProvider`
- updated login `jsx` components to account for multiple providers
- removed hard-coded login urls
- updated tests to account for new auth variables 
- note 2 tests failing (on master too): `test_update_gencode_command*`


seqr/utils/social_auth_pipeline.py
```diff
@@ -37,3 +37,9 @@ def log_signed_in(backend, response, is_new=False, *args, **kwargs):
     logger.info('Logged in {} ({})'.format(response['email'], backend.name), extra={'user_email': response['email']})
     if is_new:
         logger.info('Created user {} ({})'.format(response['email'], backend.name), extra={'user_email': response['email']})
+
+
+def log_azure_signed_in(backend, details, is_new=False, *args, **kwargs):
```
hanars, line 42: this appears identical to `log_signed_in`, is there a reason this needs to be its own function?

seqr/views/apis/awesomebar_api_tests.py
```diff
@@ -152,6 +152,7 @@ class LocalAwesomebarAPITest(AuthenticationTestCase, AwesomebarAPITest):
 class AnvilAwesomebarAPITest(AnvilAuthenticationTestCase, AwesomebarAPITest):
     fixtures = ['users', 'social_auth', '1kg_project', 'reference_data']
 
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 155: instead of updating this for every test case,  we should update the `AnvilAuthenticationTestCase` to include this mock

settings.py
```diff
@@ -453,12 +472,12 @@
...
                            SOCIAL_AUTH_PIPELINE_BASE + \
-                           ('social_core.pipeline.user.get_username',
-                            'social_core.pipeline.user.create_user') + \
-                           SOCIAL_AUTH_PIPELINE_ASSOCIATE_USER + \
-                           ('social_core.pipeline.social_auth.load_extra_data',
-                            'social_core.pipeline.user.user_details') + \
+                           SOCIAL_AUTH_PIPELINE_CLOUD_BASE + \
                            SOCIAL_AUTH_PIPELINE_LOG
+elif SOCIAL_AUTH_AZUREAD_V2_TENANT_OAUTH2_KEY:
+    SOCIAL_AUTH_PIPELINE = SOCIAL_AUTH_PIPELINE_BASE + \
```
hanars, line 478: So this auth approach allows anyone in the world with a valid azure account to log into your seqr instance, which I don't really feel comfortable making the defautl behavior for all seqr Azure instances.  Currently we have 2 ways of validating users when google auth is enabled: 
1) If terra/anvil integration is enabled, confirm that the google account is registered with anvil
2) If the terra integration is not enabled, require the account to have been explicitly added to at least one seqr project (the `SOCIAL_AUTH_PIPELINE_USER_EXIST` step)

I think approach 2 (pipeline defined in the "else" block) would be a better and more secure default behavior

hanars (changes requested):

settings.py
gregsmi: no, actually the `AzureADV2TenantOAuth2` provider is tenant-specific (see `SOCIAL_AUTH_AZUREAD_V2_TENANT_OAUTH2_TENANT_ID`), so the approach is to create a SEQR instance-specific tenant (which is what we've done with our various test instances) - only users defined directly by that tenant or explicitly guested in from a federated auth provider have access...

seqr/utils/social_auth_pipeline.py
gregsmi: the Azure provider returns `email` in `details`, not `response` so `log_signed_in` crashes against Azure - it could be updated to handle both cases but I didn't want to alter the existing behavior in case you were relying on it...

hanars: Ah I see. It would probably be a bit neater to have a `_log_signed_in` helper that both `log_signed_in` and `log_azure_signed_in` call to encapsulate that it is identical behavior, but this is fine as it is

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1418,6 +1418,7 @@ def test_validate_callset(self, mock_subprocess, mock_glob, mock_os_isfile):
         response = self.client.post(url, content_type='application/json', data=json.dumps(body))
         self.assertEqual(response.status_code, 200)
 
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 1421: If this is properly mocked in the bas class it should not be need to be added in any of the test files

hanars (changes requested):

seqr/views/apis/project_api_tests.py
```diff
@@ -186,6 +186,7 @@ def test_create_project_no_pm(self):
         self.assertSetEqual(set(response.json()['projectsByGuid'].keys()), {new_project.guid})
 
     @mock.patch('seqr.views.utils.permissions_utils.PM_USER_GROUP', 'project-managers')
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 189: this should not be needed

seqr/views/apis/saved_variant_api_tests.py
```diff
@@ -124,6 +124,7 @@
 class SavedVariantAPITest(object):
 
     @mock.patch('seqr.views.utils.variant_utils.OMIM_GENOME_VERSION', '37')
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 127: this should not be needed

seqr/views/apis/superuser_api_tests.py
```diff
@@ -17,6 +17,7 @@
 class SuperusersAPITest(object):
 
     @mock.patch('seqr.views.utils.permissions_utils.PM_USER_GROUP')
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 20: this should not be needed

seqr/views/utils/orm_to_json_utils_tests.py
```diff
@@ -28,6 +28,7 @@ def test_json_for_user(self):
                 {'firstName', 'email', 'displayName', 'isActive', 'isDataManager'},
             )
 
+    @mock.patch('seqr.views.utils.terra_api_utils.SOCIAL_AUTH_PROVIDER', TEST_OAUTH2_PROVIDER)
```
hanars, line 31: this should not be needed

hanars (changes requested):

hanars (approved):

hanars:
Thanks so much for contributing these updates!

PR #4354, opened 2024-09-09
Change CRDQ path.

bpblanken:
Idea here is for local installations to not have to copy these in any capacity and instead use the mounted disk with the rest of the reference data.

I've updated airflow and helm to [copy to the new location](https://github.com/broadinstitute/seqr-pipeline-airflow/pull/285)


hanars (approved):

hail_search/queries/base.py
```diff
@@ -262,11 +263,13 @@ def _load_filtered_table(self, sample_data, intervals=None, annotations=None, an
             *self._parse_sample_data(sample_data), parsed_intervals=parsed_intervals, raw_intervals=intervals, parsed_annotations=parsed_annotations, **kwargs)
 
     @classmethod
-    def _get_table_path(cls, path, use_ssd_dir=False):
+    def _get_table_path(cls, path, use_ssd_dir=False, is_prefilter_table=False):
+        if is_prefilter_table:
+            return f'{REFERENCE_DATA_DIR}/{cls.GENOME_VERSION}/{cls.DATA_TYPE}/cached_reference_dataset_queries/{path}'
```
hanars, line 268: I do not think we want `REFERENCE_DATA_DIR` to be a configurable location, I think it adds uneeccesary complexity. I think it would be much better to use either `SSD_DATASETS_DIR` or `DATASETS_DIR`, the same as for everything else
```
if is_prefilter_table:
    path = f'cached_reference_dataset_queries/{path}'
    use_ssd_dir = True
```

hanars (changes requested):

hail_search/queries/base.py
bpblanken: let's discuss in the engineering sync.  There's some context from the helm persistent volume work that's necessary (and been very unstable over the past few days as I've tried to get it working) to explain and I think we also might want to change `SSD_DATASETS_DIR` and `DATASETS_DIR` too.

bpblanken: Closing this ticket in favor of deferring to [env var renaming](https://github.com/broadinstitute/seqr/issues/4359) ticket.

PR #4357, opened 2024-09-10
skip proxy for all tracks

jklugherz:
our proxy is likely not necessary any more for IVG track requests!

use new sources for all ivg tracks from [here](https://igv.org/genomes/genomes.json)

verified locally that IGV loads for 37 and 38 variants

### DIFF
--- a/ui/shared/components/panel/family/constants.js
+++ b/ui/shared/components/panel/family/constants.js
@@ -77,15 +77,13 @@ export const IGV_OPTIONS = {
   showCommandBar: true,
 }
 
-const BASE_REFERENCE_URL = '/api/igv_genomes'
-
 const REFERENCE_URLS = [
   {
     key: 'fastaURL',
-    baseUrl: BASE_REFERENCE_URL,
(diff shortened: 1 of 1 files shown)


ui/shared/components/panel/family/constants.js
```diff
@@ -137,11 +135,11 @@ export const REFERENCE_LOOKUP = ['37', '38'].reduce((acc, genome) => ({
   [genome]: {
     id: GENOME_VERSION_DISPLAY_LOOKUP[GENOME_VERSION_LOOKUP[genome]],
     tracks: REFERENCE_TRACKS.map(({ baseUrl, path, indexPostfix, ...track }) => ({
-      url: `${baseUrl}/${path[genome]}`,
+      url: baseUrl ? `${baseUrl}/${path[genome]}` : path[genome],
```
hanars, line 138: This change is not needed, `baseUrl` is defiend for all the `REFERENCE_TRACKS` its only `REFERENCE_URLS` where that changed

hanars (approved):

PR #4369, opened 2024-09-11
Add reload globals route

bpblanken:


hail_search/web_app.py
```diff
@@ -116,6 +121,7 @@ async def init_web_app():
     app = web.Application(middlewares=[error_middleware], client_max_size=(1024**2)*10)
     app.add_routes([
         web.get('/status', status),
+        web.post('/reload_globals', reload_globals),
```
bpblanken, line 124: The internet convinced me that `POST` was more correct than `GET` for this, since we're potentially modifying server state.  It's a gray area though.

hail_search/test_search.py
```diff
@@ -219,6 +219,12 @@ async def test_status(self):
             resp_json = await resp.json()
         self.assertDictEqual(resp_json, {'success': True})
 
+    async def test_reload_globals(self):
```
hanars, line 222: Do we want to test that the globals were actually reloaded? It seems like that would be hard to test but I feel like we do want to be sure that something happened. Maybe we mock something?

bpblanken: Updated to test with a set of mock globals, and also updated the route to return the globals themselves.

hail_search/web_app.py
```diff
@@ -97,6 +98,17 @@ async def multi_lookup(request: web.Request) -> web.Response:
     return web.json_response({'results': result}, dumps=hl_json_dumps)
 
 
+async def reload_globals(request: web.Request) -> web.Response:
+    await sync_to_async_hail_query(request, lambda _: load_globals())
+    return web.json_response(
```
hanars, line 103: I feel like it would be neater to have `load_globals` return this, rather than need to separately compute it

hanars (approved):

PR #4394, opened 2024-09-26
Convert check new samples command to cron job

hanars:


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -22,7 +23,10 @@
 
 logger = logging.getLogger(__name__)
 
-GS_PATH_TEMPLATE = 'gs://seqr-hail-search-data/v3.1/{path}/runs/{version}/'
+GS_PATH_TEMPLATE = 'gs://seqr-hail-search-data/v3.1/{genome_version}/{dataset_type}/runs/{run_version}/_SUCCESS'
```
bpblanken, line 26: Can we make `gs://seqr-hail-search-data` a reference to `HAIL_SEARCH_DATA_DIR`?

bpblanken: @jklugherz we should probably update the luigi piece to write a `_SUCCESS` file so the local pipeline has that functionality as well!

hanars: `HAIL_SEARCH_DATA_DIR` is the relative local path for the mounted volume not the gs path

bpblanken: We wanted this to work for local users though right?  So saved variants would be updated?  I can’t remember exactly.

hanars: We did and we do, but we also need this to work for us. `HAIL_SEARCH_DATA_DIR` has to be a relative local path for the hail backend to work for anyone, and for local users thats where run metadata will be for seqr as well (assuming we add a volume mount, as currently the seqr service is not accessing this data at all) but using that local path can not work for our seqr instance, which needs run metadata to be read from GCP directly. 

We have 2 options here that I can think of:
1) For our seqr deployment have `HAIL_SEARCH_DATA_DIR` be an env variable for both the hail backend and seqr and have it be different in both places (local path in hail backend, gs path in seqr).
2) Add an additional env variable to seqr like `HAIL_SEARCH_RUN_METADATA_DIR` and then here use that if defined and if not fall back to `HAIL_SEARCH_DATA_DIR`
In both cases the local installs would work with only the one `HAIL_SEARCH_DATA_DIR` variable, so that should be straightforward for them at least

bpblanken: What if we modify the `rsync` in `airflow` to also sync the run directories?  I think the original motivation for not syncing them was because they contained snapshots of the hail search data, but that's no longer true.  

The `runs` directory currently has the `new_variants.ht`, which at least internally, we probably don't want or need on the ssd.  We could exclude that from the sync or make the path of that file rooted in the`loading-temp` directory where it probably belongs.

hanars: That doesn't really help - the hail-search pod and the seqr pod are totally isolated from each other and the seqr pod has no access to the mounted disk with all the data, nor should it. Having the seqr pod also connect to a mounted volume with all the search data feels like the wrong direction to take this

bpblanken: Hrm, for local users the seqr pod _will_ have access to the mounted disk, which is the only way this would work.  I don't particularly love the idea of having the env var be different across different services (confusing...).

I agree that `seqr` shouldn't have access to the disk, but I think that logic also applies to the `gs://seqr-hail-search-data` bucket.  If properly permission-ed with a service account, I don't think we'd want the seqr pod to have read access there _except_ for this file.  

What if we did something like add an endpoint to the hail search service to `json` dump the run metdata served from local disk?

hanars: That would work and is probably a bette rlong term solution but this feels like a huge amount of overkill for a solution that would work just as well by adding 1 extra environment variable and otherwise leaving all of our data and infrastructure alone. The seqr pod currently has access to this data in GCP and has been reading the metadata from there since the hail backend went live and locking that down feels really out of scope for this issue.

bpblanken: Ok... I really don't love doubling down on weird behavior in our system to avoid a couple of hours of work (especially if we're going to do it anyway in the future).  

Of the options presented I think defining `HAIL_SEARCH_DATA_DIR` differently in `hail-search` and `seqr` makes the most sense.

hanars: I agree that not doubling down on weird behavior to have a few hours of work is a good engineering goal, but I also think avoiding scope creep is an important engineering goal. At the start of this PR, seqr was reading the metadata from GCP. This scope of this work said nothing about changing where the data was read from. While it may be "only a couple of hours of work" to change the open source helm chart, the tgg helm chart, airflow/pipeline to change how run directories are copied, and this spot of the code, those hours will have to be spread out over the course of a few days and lots of iterative discussion and PR review. In the meantime this PR stays open and I accrue merge conflicts. 

When in the course of building something we discover that there is a better way to do something that requires more work I think it is better for team velocity to do a reasonable engineering solution that keeps the work within the original scope and then create tickets to track the new work and prioritize those tickets accordingly as part of our normal planning process. In this case, I also think we actually would want to have an engineering meeting or at least design plan for this if we are going to change what data the seqr pod does and does not have access to, as I am really not comfortable with the idea of mounting the full hail search data disk to the seqr pod

bpblanken: talked offline, on the same page now!

jklugherz: what did y'all decide to do here?

bpblanken: We're gonna do the `HAIL_SEARCH_DATA_DIR` being defined differently in `seqr` and `hail-search` environments.  I made a [ticket](https://github.com/broadinstitute/seqr-loading-pipelines/issues/914) to re-think the delivery mechanism between the pipeline and seqr.

settings.py
```diff
@@ -156,6 +156,7 @@
     MEDIA_URL = '/media/'
 
 LOADING_DATASETS_DIR = os.environ.get('LOADING_DATASETS_DIR')
+HAIL_SEARCH_DATA_DIR = os.environ.get('HAIL_SEARCH_DATA_DIR')
```
bpblanken, line 159: 👍 

Do we want to have defaults for these here or just have them as defaults in `sear-helm`?

hanars: lets just have the defaulted in seqr helm

bpblanken: 👍

bpblanken (approved):

PR #4395, opened 2024-09-26
Genotype filters per sample instead of per variant

hanars:
Currently, we combine them all variant filters and have a singe value per variant, but this is a bit weird since we display them on the individual genotypes and it adds a lot of complexity for merging tables, exacerbated in the multi sample type search. This instead assigns them to the samples


hail_search/queries/base.py
```diff
@@ -461,9 +462,17 @@ def _add_entry_sample_families(self, ht, sample_data):
         ht = ht.transmute(family_entries=family_sample_index_data.map(lambda family_tuple: family_tuple[1].map(
             lambda sample_tuple: ht.family_entries[family_tuple[0]][sample_tuple[0]].annotate(**sample_tuple[1])
         )))
+        if not is_merged_ht:
+            ht = self._annotate_entry_filters(ht)
 
         return ht, sorted_family_sample_data
 
+    @staticmethod
+    def _annotate_entry_filters(ht):
```
jklugherz, line 471: do we need to use the `family_sample_index_data` here to make sure filters are in the correct sample order?

seqr/views/apis/variant_search_api.py
```diff
@@ -240,6 +239,7 @@ def _get_variant_main_transcript_field_val(parsed_variant):
 VARIANT_SAMPLE_DATA = [
     {'header': 'sample', 'value_path': 'sampleId'},
     {'header': 'num_alt_alleles', 'value_path': 'numAlt'},
+    {'header': 'filters', 'process': lambda val: ';'.join(val or []), 'variant_value_path': '[genotypeFilters][*]'},
```
jklugherz, line 242: what does `variant_value_path` represent here? is this code change for backwards compatibility?

hanars: yes, for local installs running elasticsearch the UI needs to support both

hanars: and `value_path ` here represents the path to the field within the genotype, so `variant_value_path` represents the oath to the field within the variant

jklugherz (approved):

PR #4398, opened 2024-09-27
allow hail backend to run without all data types

hanars:


hail_search/queries/base.py
```diff
@@ -90,7 +90,12 @@ class BaseHailTableQuery(object):
     @classmethod
     def load_globals(cls):
         ht_path = cls._get_table_path('annotations.ht')
-        ht_globals = hl.eval(hl.read_table(ht_path).globals.select(*cls.GLOBALS))
+        try:
+            ht = hl.read_table(ht_path)
+        except Exception:
```
bpblanken, line 95: Do we want to be more precise and catch only `hail.utils.java.FatalError` here?  We can even check that `No file or directory found` is in the error message to make sure it's only a "no directory" failure and not a "table is there but broken" error.

hanars: I thought about that, but the `FatalError` felt a little fragile to me - its not like the most specific error here so what if it changes in a future hail release? Also, the table would have to be really incredibly mangled to fail to read and I think anything that would cause read to fail is also likely to be the generic FatalError. I did make sure that selecting the globals is outside the try/except so if the table is there without the expected data structrue it will still raise an exception

bpblanken: 👍 

Cool, that actually aligns with [what I did here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/io.py#L30), catching a generic `Exception`.

bpblanken (approved):

PR #4405, opened 2024-10-01
Improve consistency of `HAIL_BACKEND_SERVICE_HOSTNAME` with other `_HOSTNAME` env vars.

bpblanken:
There was inconsistency in how we've been using this.  I believe `hostname` refers to the name of the server.  

Paired with: https://github.com/broadinstitute/tgg-helm/pull/106/files


settings.py
```diff
@@ -347,7 +347,7 @@
 KIBANA_ELASTICSEARCH_USER = os.environ.get('KIBANA_ELASTICSEARCH_USER', 'kibana')
 KIBANA_ELASTICSEARCH_PASSWORD = os.environ.get('KIBANA_ES_PASSWORD')
 
-HAIL_BACKEND_SERVICE_HOSTNAME = os.environ.get('HAIL_BACKEND_SERVICE_HOSTNAME')
+HAIL_BACKEND_SERVICE_HOSTNAME = os.environ.get('HAIL_BACKEND_SERVICE_HOSTNAME', 'hail-search')
```
bpblanken, line 350: Yet another question here, is it better practice to default to `localhost` for both this and the pipeline runner server?

hanars: so my understanding is that using docker-compose things are named the service name by default not localhost

hanars (approved):

settings.py
bpblanken: 👍

PR #4410, opened 2024-10-09
Subset loading samples from airtable for SVs and MITO

hanars:


jklugherz (approved):

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1654,12 +1666,18 @@ def _assert_write_pedigree_error(self, response):
 class AnvilDataManagerAPITest(AirflowTestCase, DataManagerAPITest):
     fixtures = ['users', 'social_auth', '1kg_project', 'reference_data']
 
+    ADDITIONAL_REQUEST_COUNT = 1
```
jklugherz, line 1669: why is this needed?

hanars: The `AirflowTestCase` has some class level attributes it uses in its helper methods for validating that the calls to airflow are as expected. This is neccessary so that when it counts that there are the expected number of airflow calls it also knows that there is 1 additional request made thats not to airflow (namely the dag update call)

PR #4411, opened 2024-10-09
handle failing wes-wgs search case

jklugherz:


hail_search/queries/mito.py
```diff
@@ -263,6 +269,31 @@ def _family_has_valid_sample_type_entries(ht, sample_type, sample_type_family_id
             hl.is_defined(ht[sample_type.passes_inheritance_field][sample_type_family_idx])
         )
 
+    def _post_process_inheritance_filter(self, ht, family_idx_map):
+        if ht is None:
+            return ht
+
+        # If a family has variants in both sample types but one sample type is proband-only,
+        # keep the family in the proband-only sample type only if the proband-only call passes the inheritance filtering
+        # AND the same family with multiple samples in the other sample type also passes inheritance filtering.
```
hanars, line 278: So unfortunately I think you applied this case too narrowly - the point of this edge case is not to specifically handle proband only cases, its to handle any case where the number available samples differs - for example in there was a proband and Mom in one sample type and a full trio in another, or if there was a sibling available in one sample type but not the other.

Unfortunately I think this actually requires a refactor of `passes_inheritance_field` to more granularly track which samples failed inheritance, so you can determine whether the failed samples in the other sample type passed this sample type or are missing from this sample type.

PR #4414, opened 2024-10-09
add command to bulk update IGV paths

hanars:


seqr/management/commands/update_igv_location.py
```diff
@@ -0,0 +1,26 @@
...
+        parser.add_argument('new_prefix')
+
+    def handle(self, *args, **options):
+        old_prefix = options['old_prefix']
+        new_prefix = options['new_prefix']
+        igv_samples = IgvSample.objects.filter(file_path__startswith=old_prefix)
+        if not igv_samples:
+            raise CommandError(f'No IGV samples found with file prefix "{old_prefix}"')
+        logger.info(f'Updating {len(igv_samples)} IGV samples')
+        for sample in tqdm(igv_samples):
```
bpblanken, line 23: is it better to use `bulk_update()` or whatever method can push this into a single call to the db rather than hitting the db for each sample?

hanars: 🤦‍♀️ yes, I used an old manage command as the template for this and didn't think too hard about it but that would be a much better approach

seqr/management/commands/update_igv_location.py
```diff
@@ -0,0 +1,26 @@
...
+from seqr.models import IgvSample
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('old_prefix')
+        parser.add_argument('new_prefix')
```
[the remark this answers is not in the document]
hanars: I'd rather be explicit, as who knows how weirdly folks have configured things, what nested directories they are using etc

bpblanken (approved):

seqr/management/tests/update_igv_location_tests.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management import call_command, CommandError
+from django.test import TestCase
+import mock
+
+from seqr.models import IgvSample
+
+
+class TUpdateIgvPathsTest(TestCase):
```
bpblanken, line 8: I think there's an extra `T` in `Update` here 🙉 !

bpblanken (approved):
small thing but otherwise good!

PR #4415, opened 2024-10-09
show all vep consequence terms

hanars:


ui/shared/components/panel/variants/Annotations.jsx
```diff
@@ -542,6 +542,10 @@ const Annotations = React.memo(({ variant, mainGeneId, showMainGene, transcripts
     hoverable: true,
   }
 
+  const nonMajorConsequences = (mainTranscript.consequenceTerms || []).filter(
+    c => c !== mainTranscript.majorConsequence,
```
jklugherz, line 546: Is there a reason you chose to do the filtering in JS as opposed to in the backend where we build the response? (A pro of doing it in the backend is reducing the size of the response but a single string is probably be negligible)

hanars: Mostly because consequence_terms is already a field on the transcript object in the search backend that we were dropping, so keeping it as-is was as straightforward as no longer dropping that field, whereas needing to filter it meant adding a more complex annotation/ post processing. Also if you look in the Transcripts ui code it was actually simpler to use the full consequenceTerms array and not filter it so theres a slight advantage there as well.  But ultimately it would not be terrible to do the filtering in the backend either, this just was a little easier toimplement

bpblanken (approved):

PR #4418, opened 2024-10-11
Sex aneuploidy

hanars:


seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -411,9 +411,18 @@ def _get_transcript_field(field, config, transcript):
...
+    if sex in Individual.MALE_ANEUPLOIDIES:
+        sex_detail = sex
+        sex = Individual.SEX_MALE
+    elif sex in Individual.FEMALE_ANEUPLOIDIES:
+        sex_detail = sex
+        sex = Individual.SEX_FEMALE
     subject_row = {
         'participant_id': format_id(individual.individual_id),
-        'sex': Individual.SEX_LOOKUP[individual.sex],
+        'sex': Individual.SEX_LOOKUP[sex],
```
bpblanken, line 424: Is the lookup within `SEX_LOOKUP` necessary here if `sex` has already been reassigned to either `SEX_MALE` or `SEX_FEMALE` in the above `if`/`else` block?

bpblanken: is there a need to do the reassignment then, since Individual.SEX_LOOKUP contains the aneuploidies now?

hanars: the lookup  value for an aneuploidy is the aneuploidy itself, which is an invalid sex value for the reports we generate, which is what this code is for, but everywhere else that is the desired lookup behavior so its better to have this one special casing in our reporting code for aneuploidies than need to add complicated logic everywhere else sex is referenced

bpblanken: ah... ok.  I think I was getting confused by the `SEX_LOOKUP` on `MatchmakerSubmission` being present on this diff, where the lookup of the aneuploidy is not the aneuploidy, but `Individual.SEX_LOOKUP` being unchanged.  

A comment about this being a special case would be helpful!

bpblanken (approved):

PR #4419, opened 2024-10-11
Luigi UI

hanars:
Screenshot 2024-10-11 at 3 51 50 PM


settings.py
```diff
@@ -80,7 +80,8 @@
 CSP_CONNECT_SRC = ("'self'", 'https://gtexportal.org', 'https://www.google-analytics.com', 'https://igv.org',
                    'https://storage.googleapis.com', 'https://s3.amazonaws.com', 'https://igv-genepattern-org.s3.amazonaws.com', 'https://hgdownload.soe.ucsc.edu',  # used by IGV
                    'https://reg.genome.network')
-CSP_SCRIPT_SRC = ("'self'", "'unsafe-eval'", 'https://www.googletagmanager.com')
+CSP_SCRIPT_SRC = ("'self'", "'unsafe-eval'", 'https://www.googletagmanager.com',
+                  "'sha256-A16xkExeIj9J9P70pOQ+HDBXdFYcm2Cn3J/phdqk9hc='")  # luigi UI script
```
jklugherz, line 84: where does this value come from?

hanars: This is the hashed encoding of the actual script that the luigi UI is executing. The way CSP blocking works (at least in our local mode) it will not execute a script not allowed by the csp script policy, but the error it generates actually says what you would need to change to allow it, including generating what the hash would be if we wanted to use that so I just copied it from there

jklugherz (approved):

hanars:
@bpblanken I will want your eyes on this one even with Julia's approval just in case I'm missing something obvious

bpblanken:
👍 yep, looks good to me.  I went ahead and added the [`luigi env vars`](https://github.com/broadinstitute/seqr-helm/commit/2e06e90df49dae2cd1cae9e1441cb5db1c27ec81#diff-4b26aa7393c315cfedebde22c2890cfbcd5efb1f6b51c9b61fcf077ea2ef3f8eR64-R65) to helm as we didn't provide a default here which is ok.

bpblanken (approved):

hanars:
> 👍 yep, looks good to me. I went ahead and added the [`luigi env vars`](https://github.com/broadinstitute/seqr-helm/commit/2e06e90df49dae2cd1cae9e1441cb5db1c27ec81#diff-4b26aa7393c315cfedebde22c2890cfbcd5efb1f6b51c9b61fcf077ea2ef3f8eR64-R65) to helm as we didn't provide a default here which is ok.

Sounds good, we should just make sure to turn that off in tgg-helm, since we won't have the luigi UI enabled for Broad seqr

PR #4437, opened 2024-10-16
update genotype display for multiple sample types

jklugherz:
image

Screenshot 2024-10-21 at 5 37 17 PM


ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -159,40 +161,110 @@ const svGenotype = (genotype, isHemiX) => {
...
+
+AllelesHeaderContent.propTypes = {
+  genotypeInfo: PropTypes.object,
+  variant: PropTypes.object,
+  useStyledAllele: PropTypes.bool,
+}
+
+AllelesHeaderContent.defaultProps = { useStyledAllele: true }
+
+export const Alleles = React.memo(({ variant, genotypes }) => {
```
hanars, line 196: This class is exported and used elsewhere and I do not think we want to change the behavior for everywhere it is used, as much of this new logic is not relevant. For example, the fact that this component could take any "warning" but now explicitly computes warning is a loss of functionality for the component. 

Keep the exported `Alleles` class the same as much as possible. The allele display itself should not change, but if there is anything you really need to change then add a nw prop to the class to conditionally change that. Then add a `MultiSampleTypeAlleles` or a `DetailedAlleles` component to change the behavior that needs to change just within this component

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +397,93 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    filters,
+  }
+}
+
+const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById }) => {
+  if (!variant.genotypes) {
+    return null
+  }
+
+  let genotypes = variant.genotypes[individual.individualGuid]
```
hanars, line 419: in general, using `let` and then changing the value as you go is more bug prone than using `const`. Unless there is a really compelling reason why you need to change the value it is better to define new constants -
for example: 
```
const individualGenotypes = variant.genotypes[individual.individualGuid]
const genotypes =  (Array.isArray(individualGenotypes) ? individualGenotypes : [individualGenotypes]).sort(
  ...).map(...)
```

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +397,93 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+  if (!genotypes) {
+    return null
+  }
+  genotypes = Array.isArray(genotypes) ? genotypes : [genotypes]
+
+  const sampleTypeOrder = [SAMPLE_TYPE_GENOME, SAMPLE_TYPE_EXOME]
```
hanars, line 425: constants like this should be defined at the top of the file in all caps

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +397,93 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    if (index > 0) {
+      acc.push(<Divider />)
+    }
+    return acc.concat(details)
+  }, [])
+
+  const firstGenotype = genotypes[0]
+  const legacyOtherSampleGenotypes = useMemo(() => firstGenotype.genotype?.otherSample && [
+    { genotype: firstGenotype.genotype.otherSample, isHemiX: firstGenotype.isHemiX },
+  ], [firstGenotype])
```
hanars, line 441: `useMemo` is an antipattern here - rather than creating a new dictionary as a prop it seems like it would be much cleaner to just pass in `firstGenotype.genotype` as the prop and add conditional logic in the component to use `otherSample` instead if it is present. Or add a `useOtherSampleType` prop

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +397,93 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+
   const content = (
     <span>
-      {genotype.otherSample && (
+      {firstGenotype.genotype?.otherSample && (
         <Popup
           header="Additional Sample Type"
-          trigger={<Icon name="plus circle" color={hasConflictingNumAlt ? 'red' : 'green'} />}
+          trigger={<Icon name="plus circle" color={firstGenotype.hasConflictingNumAlt ? 'red' : 'green'} />}
```
hanars, line 448: can you add a screenshot for this

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +397,93 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
-      {showSecondaryQuality && `, ${secondaryQuality ? secondaryQuality.toPrecision(2) : '-'}`}
-      {filters && (
-        <small>
-          <br />
-          {filters}
-        </small>
-      )}
+      <Alleles variant={variant} genotypes={genotypes} />
+      {genotypes.map(({ genotype, quality, showSecondaryQuality, secondaryQuality, previousCall, filters }) => (
+        <div key={genotype.sampleType}>
```
hanars, line 465: for elaticsearch and old saved variants theres not necessarily a `sampleType` for the genotype - use an `|| sampleId` or something to ensure the key is always defined

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: These are great points, I added a new `MultiSampleTypeAlleles` component instead of modifying the original `Alleles` to support this case

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: yeah that's way clearer, thank you

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -336,46 +438,98 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
   const quality = Number.isInteger(genotype.gq) ? genotype.gq : genotype.qs
   const filters = genotype.filters?.join(', ') || variant.genotypeFilters
 
+  return {
+    genotype,
+    isHemiX,
```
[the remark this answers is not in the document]
jklugherz: I refactored and went with this approach to use helper classes instead!

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -160,39 +164,42 @@ const svGenotype = (genotype, isHemiX) => {
...
+  isHemiX: PropTypes.bool,
+}
+
 export const Alleles = React.memo(({ genotype, variant, isHemiX, warning }) => (
   <AlleleContainer>
     {warning && (
       <Popup
         wide
         trigger={<WarningIcon />}
-        content={
```
hanars, line 169: so as mentioned on a previous PR, this is an exported component so its better to its behavior alone. I also think it actually better for us to say "warning" one time in the popup, not once per warningas repeating the word multiple time reduces readability

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -258,18 +371,7 @@ const genotypeDetails = (genotype, variant, genesById) => {
...
-  if (!genotype) {
-    return null
-  }
-  // Temporarily use the first genotype for an individual until blended es/gs are supported in UI - https://github.com/broadinstitute/seqr/issues/4269
-  genotype = Array.isArray(genotype) ? genotype[0] : genotype
-
+const getAllGenotypeDetails = (genotype, variant, individual, isCompoundHet, genesById) => {
   const hasCnCall = isCalled(genotype.cn)
   if (!hasCnCall && !isCalled(genotype.numAlt)) {
     return <b>NO CALL</b>
```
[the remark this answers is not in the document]
hanars: this is still a problem

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +363,166 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    </div>
+  )
+}
+
+GenotypeQuality.propTypes = {
+  genotype: PropTypes.object,
+  variant: PropTypes.object,
+}
+
+const LegacyAdditionalSampleTypePopup = ({ genotype, variant, isHemiX, genesById }) => {
```
hanars, line 375: We should never be maintaining a special legacy UI unless there is an overwhelming reason not to use the new UI. You should be transforming the `otherSample` into an array of genotypes with both samples and using the new, supported UI

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +363,166 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+            <div key={sampleType}>
+              <b>
+                {sampleType}
+                Warning:
+              </b>
+              {warning}
+            </div>
+          ))}
+          {hasDifferentNumAlt && (
+            <span>
```
hanars, line 437: This is going to look weird if you have a sample-type specific warnings and also different number of genotypes - it will be something like 
```
WES: Potential UPD/ Hemizygosity
WGS: Potential UPD/ Hemizygosity. Variant absent in parents
Genotypes differ across sample types
WES:
  G/C
WGS:
  C/C
```

I think the differing genotypes warning belongs at the top, and also that rather than combining all the warnings per sample type, we should show each warning on each own row, with an indicator if its only foor some sample types:

```
Genotypes differ across sample types
WES:
  G/C
WGS:
  C/C
Potential UPD/ Hemizygosity
Variant absent in parents (WGS)
```

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +363,166 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+              <b>Warning: </b>
+              Genotypes differ across sample types
+              {genotypes.map(genotype => (
+                <div key={genotype.sampleType}>
+                  {genotype.sampleType}
+                  :
+                  <AlleleContainer>
```
hanars, line 444: you can just use the `Alleles` class here, as without a warning prop or an SV (which can't have multiple sample types) this is just the same code as that class

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +363,166 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+            genesById={genesById}
+          />
+          <Alleles
+            genotype={genotypes[0]}
+            variant={variant}
+            isHemiX={isHemiX}
+            warning={
+              <div>
+                <b>Warning: </b>
+                {getWarningsForGenotype(genotypes[0], variant, individual, isHemiX, isCompoundHet)}
```
hanars, line 502: As mentioned above warning formatting needs to be left the same in the Alleles class instead of moving it here. One key problem with this current approach is that even if `getWarningsForGenotype` returns nothing this will still show a warning icon on every genotype. Whenever I change the variant UI based on changes from search, I always make sure to also open up a saved variant page for an unchanged family, to make sure that legacy saved variants still look right

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +363,166 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+              </div>
+            }
+          />
+        </span>
+      ) : (
+        <MultiSampleTypeAlleles
```
hanars, line 508: So in the end `MultiSampleTypeAlleles` returns `Alleles` using the same props as for a single-sample genotype - the only difference is the warning. Rather than have a helper class and a ternary here, it would be much neater to have a `getWarningsForGenotypes` method that handles constructing the correct warning for all the genotypes and leave this rendering much simpler

ui/shared/components/panel/variants/VariantIndividuals.jsx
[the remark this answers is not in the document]
jklugherz: Would making it a functional component instead of a helper solve the problem?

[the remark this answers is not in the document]
jklugherz: oh I see why it's a problem now.. `return <b>NO CALL</b>` is supposed to be a return value of `Genotypes`

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: I added otherSample to the genotypes array, screenshot incoming.

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: Screenshot 2024-10-23 at 4 22 43 PM

I agree here!

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: I removed `MultiSampleTypeAlleles` and conditionally set the `warning` prop on `Alleles` based on genotypes length.

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,10 +358,106 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+  const individualGenotypes = variant.genotypes[individual.individualGuid]
+  if (!individualGenotypes) {
+    return null
+  }
+  const genotypes = (Array.isArray(individualGenotypes) ? individualGenotypes : [individualGenotypes]).sort(
+    (a, b) => SAMPLE_TYPE_DISPLAY_ORDER.indexOf(a.sampleType) - SAMPLE_TYPE_DISPLAY_ORDER.indexOf(b.sampleType),
+  )
+
+  if (genotypes.every(genotype => !isCalled(genotype.cn) && !isCalled(genotype.numAlt))) {
+    return <b>NO CALL</b>
```
jklugherz, line 412: Do you think there will be a problem with the way I'm handling this NO CALL code path?

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: `otherSample` is not part of any saved variant json, it is [added](https://github.com/broadinstitute/seqr/blob/master/seqr/utils/search/elasticsearch/es_search.py#L1171) to the genotype inside of the es_utils code path. Is this something I can test?

hanars: Hmm okay so not easy to test but we should keep support for it, but we do not need to keep the current UI for it. I think changing the code to move it into the the genotypes array is sufficient here, no screenshot needed

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,6 +397,65 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById }) => {
+  const individualGenotypes = variant.genotypes[individual.individualGuid]
+  if (!individualGenotypes) {
+    return null
+  }
+  const genotypes = (Array.isArray(individualGenotypes) ? individualGenotypes : [individualGenotypes]).sort(
+    (a, b) => SAMPLE_TYPE_DISPLAY_ORDER.indexOf(a.sampleType) - SAMPLE_TYPE_DISPLAY_ORDER.indexOf(b.sampleType),
+  )
+
+  if (!isCalled(genotypes[0].cn) && !isCalled(genotypes[0].numAlt)) {
```
hanars, line 419: nitpcik, but shouldn't you check that all the genotypes are not called for this, not just the first one

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,6 +397,65 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    return <b>NO CALL</b>
+  }
+
+  if (genotypes[0].otherSample) {
```
hanars, line 423: I would add a comment stating that this is to support legacy elasticsearch formatting

ui/shared/components/panel/variants/VariantIndividuals.jsx
```diff
@@ -377,6 +397,65 @@ const Genotype = React.memo(({ variant, individual, isCompoundHet, genesById })
...
+    genotypes.push(genotypes[0].otherSample)
+  }
+
+  const isHemiX = isHemiXVariant(variant, individual)
+
+  const sampleTypeWarnings = genotypes.reduce((acc, genotype) => {
+    const warnings = getWarningsForGenotype(genotype, variant, individual, isHemiX, isCompoundHet)
+    warnings.forEach((warning) => {
```
hanars, line 431: this reduction is not used in the case where theres only one sample type, so it would be neater to just do a `genotypes.map` here to get an array of warnings. Then for the single sample case your warnings would just be `genotypeWarnings[0].join('. ')` and in `getWarningsForGenotypes` you could do this grouping, 
something like
```
`genotypeWarnings.reduce((acc, [warnings, index]) => {
  const { sampleType } = genotypes[index]
  warnings.forEach((warning) => {
    ...
}, {})
```

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: I was thinking that since cn is an SV-only annotation and there will only be one genotype per SV, there would only be one genotype in the list here but I could be wrong.

hanars: its true for now but if this ever changes its unlikely we would think to go back to this spot in the UI code and fix it. Building the UI to generically work with multiple sample types is better as it means we are unlikely to introduce bugs in the future, as long as it doesn't make the code to complex/ unreadable for now

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: agree, done!

ui/shared/components/panel/variants/VariantIndividuals.jsx
jklugherz: that makes sense, I updated the code to return no call if all genotypes fail that check

hanars (approved):

PR #4440, opened 2024-10-18
handle wes/wgs inheritance edge case 

jklugherz:


hail_search/queries/mito.py
```diff
@@ -206,37 +206,70 @@ def _filter_entries_ht_both_sample_types(
...
+        ht = ht.annotate(**{annotation: hl.empty_dict(hl.tint32, hl.tarray(hl.tint32))})
+        # print(annotation, ht[annotation].collect())
+
+        for genotype, entry_indices in entry_indices_by_gt.items():
+            if not entry_indices:
+                continue
+            # print(genotype, entry_indices)
+            entry_indices = hl.dict(entry_indices)
+            ht = ht.annotate(
+                **{annotation: hl.dict(
```
jklugherz, line 251: This isn't quite working how I want it.

hail_search/queries/mito.py
```diff
@@ -252,15 +273,49 @@ def _apply_multi_sample_type_entry_filters(self, ht, family_idx_map):
...
+            **{sample_type.family_entries_field: hl.enumerate(ht[sample_type.family_entries_field]).starmap(
+                lambda family_idx, family_samples: hl.or_missing(
+                    hl.bind(lambda other_sample_type_family_idx: (
+                        hl.bind(
+                            lambda other_sample_type_pass_samples, sample_type_pass_samples: (
+                                ht[sample_type.failed_family_sample_field][family_idx].all(
+                                    other_sample_type_pass_samples.contains
+                                ) & ht[sample_type.other_sample_type.failed_family_sample_field][other_sample_type_family_idx].all(
+                                    sample_type_pass_samples.contains
+                                )),
```
jklugherz, line 306: these binds are a little wild, but the logic is that for all families, keep family if for all samples, keep a failed sample if sample failed in one sample type it passed in the other

hanars (commented):
Looking at this, I think we may also need extra logic for comp hets, but I would do that as a separate follow up PR after this is done to aooid scope creep

hail_search/queries/mito.py
```diff
@@ -253,13 +284,20 @@ def _apply_multi_sample_type_entry_filters(self, ht, family_idx_map):
...
-            hl.is_defined(ht[sample_type.passes_quality_field][sample_type_family_idx]) &
-            hl.is_defined(ht[sample_type.passes_inheritance_field][sample_type_family_idx])
+            hl.is_defined(ht[sample_type.passes_quality_field][sample_type_family_idx])
+        )
+
+    @staticmethod
+    def _family_has_valid_inheritance(ht, sample_type, family_idx, other_sample_type_family_idx):
+        return hl.bind(
+            lambda other_sample_type_pass_samples: (
+                ht[sample_type.failed_family_sample_field][family_idx].all(other_sample_type_pass_samples.contains)
```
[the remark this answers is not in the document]
jklugherz: `ht[sample_type.failed_family_sample_field][family_idx]` should also be a list of sample ids

hail_search/queries/mito.py
```diff
@@ -253,13 +284,20 @@ def _apply_multi_sample_type_entry_filters(self, ht, family_idx_map):
...
+            ), ht[sample_type.other_sample_type.family_entries_field][other_sample_type_family_idx].filter(
+                lambda s: ~ht[sample_type.other_sample_type.failed_family_sample_field][other_sample_type_family_idx].contains(s['sampleId'])
```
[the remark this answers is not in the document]
jklugherz: I think this won't work since `family_entries_field` and `failed_family_sample_field` for a given sample type are not necessarily the same length. ` ht[sample_type.other_sample_type.failed_family_sample_field][other_sample_type_family_idx]` is a list of sample_ids of variable length - it only contains failed samples - so the [i] access won't work. I originally had `failed_family_sample_field` as an array of failing indices but switched it to an array of failing sample IDs because it was easier for me to come up with a solution to compare the same samples from different sample types. 

Maybe a better way to structure `failed_family_sample_field` is an array of booleans that matches the shape of `family_entries_field` where the value at each sample index is true or false for failed. It's just more complicated to compare samples individually that way - I'd need to reintroduce that map of families to samples to sample types to indices inside of their respective family_entries field.

hail_search/queries/mito.py
```diff
@@ -205,37 +205,68 @@ def _filter_entries_ht_both_sample_types(
...
-                family_guid_idx_map[family_guid][sample_type.value] = family_idx
+                family_idx_map[family_guid][sample_type.value] = family_idx
 
-        family_idx_map = hl.dict(family_guid_idx_map)
+        family_idx_map = hl.dict(family_idx_map)
         ht = self._apply_multi_sample_type_entry_filters(ht, family_idx_map)
         ch_ht = self._apply_multi_sample_type_entry_filters(ch_ht, family_idx_map)
         return ht, ch_ht
 
+    def _annotate_failed_family_samples_inheritance(
```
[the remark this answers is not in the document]
jklugherz: I implemented this, but because I don't want the single sample type families that come originally through the mito class code path to call the mito 'family_passes_inheritance_filter' function, I'm still passing a `family_passes_inheritance_filter` function to `_filter_inheritance` instead of using python inheritance here. Do you know if there's a cleaner way to do this?

hail_search/test_search.py
```diff
@@ -388,10 +388,11 @@ async def test_both_sample_types_search(self):
             [VARIANT2_BOTH_SAMPLE_TYPES], sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA_MISSING_PARENTAL_WGS,
             inheritance_mode=inheritance_mode, **COMP_HET_ALL_PASS_FILTERS, intervals=[variant2_interval]
         )
-        # Genome passes quality and inheritance exome fails inheritance (parental data shows variant is inherited).
+        # Genome passes quality and inheritance but exome fails inheritance (parental data shows variant is inherited).
+        # Variant is excluded from search results.
```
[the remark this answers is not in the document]
jklugherz: I think this case is covered by an above test -

        # Variant1 in family_2 is de novo in exome but maternally inherited in genome.
        # Genome passes quality and inheritance, show genotypes for both sample types.

hail_search/queries/mito.py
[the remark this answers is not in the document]
hanars: >  I'd need to reintroduce that map of families to samples to sample types to indices inside of their respective family_entries field.

if `failed_family_sample_field` is a list of booleans then you can easily get the list of passing sample ids from the main family_entries field without maintaining an entry map, and then your logic for using `other_sample_type_pass_samples` would not need to change
```
hl.enumerate(ht[family_entries_field][family_idx]).starmap(
  lambda sample_idx, sample: hl.or_missing(
    ~ht[failed_family_sample_field][family_idx][sample_idx]),
    sample['sampleId'],
  )
).filter(hl.is_defined)
```

hail_search/test_search.py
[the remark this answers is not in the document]
hanars: The search criteria changes between these 2 test cases. We really need back-to-back tests that the exact same search returns different results when the paternal data is present/absent. Instead of running a dominant and recessive search each filtered to variant 1 and variant 2, it would probably be better to not use an interval filter at all and test that dominant and recessive search return the expected combo of variants with and without paternal data

hanars (changes requested):

hail_search/queries/mito.py
```diff
@@ -205,37 +205,61 @@ def _filter_entries_ht_both_sample_types(
...
+
+    def _get_family_passes_inheritance_filter_both_sample_types(
+        self, entry_indices, family_idx, genotype, family_samples, ht, annotation
+    ):
+        return hl.bind(
+            lambda failed_samples: ht[annotation][family_idx].extend(failed_samples),
+            entry_indices.get(family_idx).filter(
+                lambda sample_idx: ~self.GENOTYPE_QUERY_MAP[genotype](family_samples[sample_idx].GT)
+            ).map(lambda sample_idx: family_samples[sample_idx]['sampleId'])
+        )
```
hanars, line 245: Bind is actually less readable and does not have any performance boost if the bound variable is used exactly once and not within a loop, like in this case
```suggestion
        return ht[annotation][family_idx].extend(
            entry_indices.get(family_idx).filter(
                lambda sample_idx: ~self.GENOTYPE_QUERY_MAP[genotype](family_samples[sample_idx].GT)
            ).map(lambda sample_idx: family_samples[sample_idx]['sampleId'])
        ))
```

hail_search/queries/mito.py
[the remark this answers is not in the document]
jklugherz: ok, I refactored the `passes_inheritance_field` to be a list of booleans corresponding to the samples in family entries, ^ and used this more performant code to get a list of passing sample IDs

hail_search/queries/mito.py
jklugherz: resolving this because I changed the structure of the annotation so this comment no longer applies. I also didn't use a bind here in my most recent refactor.

hail_search/fixtures/GRCh38/SNV_INDEL/families/WGS/F000002_2.ht/README.txt
```diff
@@ -1,3 +1,3 @@
 This folder comprises a Hail (www.hail.is) native Table or MatrixTable.
   Written with version 0.2.130-bea04d9c79b5
-  Created at 2024/10/02 14:46:35
\ No newline at end of file
+  Created at 2024/11/04 13:45:23
```
jklugherz, line 3: I updated this table to make variant2 de novo in WGS.

hail_search/test_search.py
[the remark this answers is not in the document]
jklugherz: I did this^!

hail_search/queries/mito.py
```diff
@@ -205,37 +205,62 @@ def _filter_entries_ht_both_sample_types(
...
-                            ),
-                            family_idx_map.get(hl.coalesce(family_samples)[0]['familyGuid']).get(sample_type.other_sample_type.value),
-                       ), family_samples)
+                    lambda family_idx, family_samples: hl.or_missing(
+                        hl.bind(lambda other_sample_type_family_idx: ((
+                            self._family_has_valid_quality(ht, sample_type, family_idx) |
+                            self._family_has_valid_quality(ht, sample_type.other_sample_type, other_sample_type_family_idx)
+                             ) &
+                            self._family_has_valid_inheritance(ht, sample_type, family_idx, other_sample_type_family_idx) &
+                            self._family_has_valid_inheritance(ht, sample_type.other_sample_type, other_sample_type_family_idx, family_idx)
```
hanars, line 261: Commenting for posterity that I'm not 100% sure that this is the final logic we want, because it would allow us to return variants that pass quality and fail inheritance in one sample type and fail quality and pass inheritance in the other, meaning theres no sample type that clearly passes both inheritance and quality. However, I think we maybe do want to return these, the logic gets kind of confusing and I can't quite be sure these would not be helpful. I think being overly permissive here is better, if the analysts are seeing a bunch of cases where they ultimately think that the returned variants are not helpful and should be filtered out we can always go back later and make this a stricter criteria, so we should leave this as is for now

hail_search/test_search.py
```diff
@@ -365,34 +368,35 @@ async def test_both_sample_types_search(self):
...
+            [VARIANT1_BOTH_SAMPLE_TYPES, VARIANT2_BOTH_SAMPLE_TYPES, VARIANT4_BOTH_SAMPLE_TYPES],
+            sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA, inheritance_mode=inheritance_mode,
+            **COMP_HET_ALL_PASS_FILTERS
         )
 
-        # Variant 2 in family_2 is inherited in exome and there is no parental data in genome.
-        # Genome and exome pass quality and inheritance, show genotypes for both sample types.
-        variant2_interval = ['1', 38724418, 38724420]
+        # Same variants, but genome data is proband-only.
         inheritance_mode = 'recessive'
```
hanars, line 388: I think this test file would be more readable if instead of toggling the inheritance back and forth you run both recessive searches back to back and then both de novo searches, so its clearer that nothings changing other than the parental data

hail_search/test_search.py
```diff
@@ -365,34 +368,35 @@ async def test_both_sample_types_search(self):
...
+            **COMP_HET_ALL_PASS_FILTERS
         )
-        # Exome passes quality and inheritance, show genotypes for both sample types.
         inheritance_mode = 'de_novo'
         await self._assert_expected_search(
-            [VARIANT1_BOTH_SAMPLE_TYPES], sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA, inheritance_mode=inheritance_mode,
-            intervals=[variant1_interval]
+            [VARIANT1_BOTH_SAMPLE_TYPES, VARIANT2_BOTH_SAMPLE_TYPES, VARIANT4_BOTH_SAMPLE_TYPES],
+            sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA, inheritance_mode=inheritance_mode,
+            **COMP_HET_ALL_PASS_FILTERS
```
hanars, line 384: we shouldn't need `COMP_HET_ALL_PASS_FILTERS` for de novo searches

hail_search/test_search.py
```diff
@@ -365,34 +368,35 @@ async def test_both_sample_types_search(self):
             MULTI_PROJECT_BOTH_SAMPLE_TYPE_VARIANTS, gene_counts=GENE_COUNTS, sample_data=MULTI_PROJECT_SAMPLE_TYPES_SAMPLE_DATA,
         )
 
-        # Variant1 in family_2 is de novo in exome but maternally inherited in genome.
-        # Genome passes quality and inheritance, show genotypes for both sample types.
-        variant1_interval = ['1', 10438, 10440]
+        # Variant 1 is de novo in exome but maternally inherited in genome.
+        # Variant 2 is inherited in exome and de novo in genome.
```
hanars, line 372: I would clarify in the comment that inherited means "inherited and homozygous" as usually maternally inherited means mom has one alt allele and proband inherited that and has 1 alt allele

hail_search/queries/mito.py
jklugherz: This makes sense, we may want to consider quality and inheritance passing together and handle it differently (the `&` seems like it could be too simple) but it's not clear to me what the logic/change would be. I agree that trying this out and getting feedback from analysts before we do that is the way to go.

hail_search/test_search.py
hanars: this line is unneeded

hail_search/test_search.py
```diff
@@ -365,34 +368,32 @@ async def test_both_sample_types_search(self):
...
-        inheritance_mode = 'recessive'
+        inheritance_mode = 'de_novo'
         await self._assert_expected_search(
-            [VARIANT2_BOTH_SAMPLE_TYPES], sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA_MISSING_PARENTAL_WGS,
-            inheritance_mode=inheritance_mode, **COMP_HET_ALL_PASS_FILTERS, intervals=[variant2_interval]
+            [VARIANT1_BOTH_SAMPLE_TYPES, VARIANT2_BOTH_SAMPLE_TYPES, VARIANT4_BOTH_SAMPLE_TYPES],
+            sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA, inheritance_mode=inheritance_mode,
         )
-        # Genome passes quality and inheritance exome fails inheritance (parental data shows variant is inherited).
         inheritance_mode = 'de_novo'
```
hanars, line 393: this line is unneeded

hail_search/test_search.py
hanars: You updated the comment for variant 1 but not variant 2, de novo is usually a het call so its still confusing

hail_search/test_search.py
```diff
@@ -365,34 +368,32 @@ async def test_both_sample_types_search(self):
...
-        variant2_interval = ['1', 38724418, 38724420]
-        inheritance_mode = 'recessive'
+        inheritance_mode = 'de_novo'
         await self._assert_expected_search(
-            [VARIANT2_BOTH_SAMPLE_TYPES], sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA_MISSING_PARENTAL_WGS,
-            inheritance_mode=inheritance_mode, **COMP_HET_ALL_PASS_FILTERS, intervals=[variant2_interval]
+            [VARIANT1_BOTH_SAMPLE_TYPES, VARIANT2_BOTH_SAMPLE_TYPES, VARIANT4_BOTH_SAMPLE_TYPES],
+            sample_data=FAMILY_2_BOTH_SAMPLE_TYPE_SAMPLE_DATA, inheritance_mode=inheritance_mode,
         )
-        # Genome passes quality and inheritance exome fails inheritance (parental data shows variant is inherited).
```
hanars, line 391: I still think a comment here explaining whats being tested is helpful. Maybe something like "Variant 2 fails inheritance when parental data is present"

hail_search/test_search.py
hanars: I think having the ability to have at least one variant passing for homozygous recessive search is valuable for testing and I like the coverage we get from this variant configuration.  And we do sometimes see de novo homozygotes, we just don't refer to it as plain "de novo" which is why I think a comment update is sufficient

hail_search/test_search.py
jklugherz: updated!

hanars (approved):

PR #4456, opened 2024-10-28
add relatedness check tsv to failure message

jklugherz:
PR to add the file path to metadata: https://github.com/broadinstitute/seqr-loading-pipelines/pull/935


seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -169,7 +169,8 @@
             'F000002_2': {'reasons': ["Missing samples: {'HG00732', 'HG00733'}"]},
             'F000003_3': {'reasons': ["Missing samples: {'NA20870'}"]},
         },
-    }
+    },
+    'relatedness_check_file_path': 'test.tsv',
```
hanars, line 173: It would be better to test this in a representative way, so if `relatedness_check_file_path` is expected to be a gs path we should test it as such

hanars (approved):

PR #4477, opened 2024-11-08
Reference data handler cleanup

hanars:
I statrted this when I though I needed to refactor the update reference data job. i don;t for now, but this is still better than what we currently have so worth merging

### DIFF
--- a/reference_data/management/commands/utils/update_utils.py
+++ b/reference_data/management/commands/utils/update_utils.py
@@ -4,6 +4,8 @@ import gzip
 from tqdm import tqdm
 import traceback
 from django.core.management.base import BaseCommand, CommandError
+from django.db import models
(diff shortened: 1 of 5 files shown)


reference_data/management/commands/utils/update_utils.py
```diff
@@ -55,6 +55,61 @@ def get_gene_for_record(self, record):
             raise ValueError('Gene "{}" not found in the GeneInfo table'.format(gene_id or gene_symbol))
         return gene
 
+    def update_records(self, file_path=None):
```
hanars, line 58: this is just copy paste from below, the only difference is the `reference_data_handler` arg has been changed to `self`

bpblanken (approved):

PR #4487, opened 2024-11-18
Better gregor missing airtable warnings

hanars:
Adds better errors and warnings when data is missing from airtable for the gregor reports

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -416,19 +416,43 @@ def gregor_export(request):
     airtable_rows = {table: [] for table in AIRTABLE_TABLE_COLUMNS.keys()}
     experiment_lookup_rows = []
     experiment_ids_by_participant = {}
+    missing_participant_ids = []
+    missing_airtable_data_types = defaultdict(list)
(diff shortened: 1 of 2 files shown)


seqr/views/apis/report_api.py
```diff
@@ -416,21 +416,57 @@ def _add_row(row, family_id, row_type):
...
+            f'{", ".join(sorted(missing_participant_ids))}'
+        )
+    if missing_airtable:
+        errors.append(
+            f'The following entries are missing airtable metadata: '
+            f'{", ".join(sorted(missing_airtable))}'
+        )
+    warnings = [
+        f'The following entries are missing {data_type} airtable data: {", ".join(participants)}'
+        for data_type, participants in sorted(missing_airtable_data_types.items())
```
jklugherz, line 462: are entries missing airtable data types also omitted / should this warning reflect that?

hanars: so in that case where the data is not in airtable there is no data to be omitted. The data is in airtable so if its not there we aren't omitting anything

jklugherz (approved):

PR #4495, opened 2024-11-21
Require affected family members

hanars:
Updates seqr in a few places to try and reduce instances of families with no affected individuals:
- when  bulk editing individuals, validate at least one individual in family is affected
- disable ability to create empty families (families are already created by creating individuals for them)
- add additional check when generating loading pipeline to fail loading request when attempted with an unaffected family

### DIFF
--- a/seqr/views/utils/pedigree_info_utils.py
+++ b/seqr/views/utils/pedigree_info_utils.py
@@ -279,6 +279,7 @@ def validate_fam_file_records(project, records, fail_on_warnings=False, errors=N
     errors = errors or []
     warnings = []
     individual_id_counts = defaultdict(int)
+    affected_status_by_family = defaultdict(list)
     for r in records:
         individual_id = r[JsonConstants.INDIVIDUAL_ID_COLUMN]
         individual_id_counts[individual_id] += 1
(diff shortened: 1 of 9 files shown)


seqr/views/utils/pedigree_info_utils.py
```diff
@@ -323,11 +324,20 @@ def validate_fam_file_records(project, records, fail_on_warnings=False, errors=N
...
                 errors.append(f'{individual_id} has invalid HPO terms: {", ".join(sorted(invalid_features))}')
 
+        affected_status_by_family[family_id].append(r.get(JsonConstants.AFFECTED_COLUMN))
+
     errors += [
         f'{individual_id} is included as {count} separate records, but must be unique within the project'
         for individual_id, count in individual_id_counts.items() if count > 1
     ]
 
+    no_affected_families = [
```
jklugherz, line 334: This logic is duplicated inside of `_upload_data_loading_files` [here](https://github.com/broadinstitute/seqr/pull/4495/files#diff-d97edf947abfdd9ae291360399589f1b801ad8ca0586ea531ca2f60cf04927ceR159), can it be shared?

jklugherz (approved):

PR #4501, opened 2024-11-26
make vcf message for loading data shared

jklugherz:


ui/shared/components/form/FormHelpers.jsx
```diff
@@ -64,6 +64,22 @@ const removeField = (fields, i) => (e) => {
...
+    {isAnvil ? ' have a joint called VCF available in your workspace. ' : ' use a joint called VCF.'}
+    &nbsp;
+    For more information about generating and validating this file,
+    see &nbsp;
+    <b><a href={documentationUrl} target="_blank" rel="noreferrer">this documentation</a></b>
+  </Message>
+))
+
+LoadDataVCFMessage.propTypes = {
+  documentationUrl: PropTypes.string,
```
hanars, line 79: documenation url should be the same in both places so it would be clearer not to allow this to be configurable and instead just use the correct url directly here

hanars (changes requested):

ui/shared/components/form/FormHelpers.jsx
jklugherz: I was getting a circular dependency using `VCF_DOCUMENTATION_URL` inside of this file so I'll just move the component

hanars: Yes, this does not belong in form helpers as this is not a generic form component. I think having this directly in `shared/utils/constants` is fine

hanars (approved):

PR #4506, opened 2024-12-02
Vlm match endpoint

hanars:
Adds the main matc endpoint for the VLM

### DIFF
--- a/.github/workflows/vlm-unit-tests.yaml
+++ b/.github/workflows/vlm-unit-tests.yaml
@@ -30,4 +30,4 @@ jobs:
       - name: Run coverage tests
         run: |
           coverage run --source="./vlm" --omit="./vlm/__main__.py" -m pytest vlm/
-          coverage report --fail-under=90
+          coverage report --fail-under=99
--- a/.github/workflows/vlm-unit-tests.yaml
+++ b/.github/workflows/vlm-unit-tests.yaml
@@ -32,5 +32,7 @@ jobs:
           export VLM_DATA_DIR=./hail_search/fixtures
           export SEQR_BASE_URL=https://test-seqr.org/
           export NODE_ID=TestVLM
+          export MACHINE_MEM=24
+          export JAVA_OPTS_XSS=16M
           coverage run --source="./vlm" --omit="./vlm/__main__.py" -m pytest vlm/
-          coverage report --fail-under=99
+          coverage report --fail-under=95
--- a/.github/workflows/vlm-unit-tests.yaml
+++ b/.github/workflows/vlm-unit-tests.yaml
@@ -29,5 +29,8 @@ jobs:
           pip install -r hail_search/requirements-test.txt
       - name: Run coverage tests
         run: |
+          export VLM_DATA_DIR=./hail_search/fixtures
+          export SEQR_BASE_URL=https://test-seqr.org/
+          export NODE_ID=TestVLM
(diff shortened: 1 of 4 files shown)


vlm/match.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+    ht = ht.filter(ht.alleles==hl.array([ref, alt]))
+    counts = ht.aggregate(hl.agg.take(ht.gt_stats, 1))
+
+    return _format_results(counts, genome_build, f'{chrom}-{pos}-{ref}-{alt}')
+
+
+def _parse_match_query(query: dict) -> tuple[str, int, str, str, str]:
+    missing_params = [key for key in QUERY_PARAMS if key not in query]
+    if missing_params:
```
bpblanken, line 52: If there's not a huge rush on getting this in, now might be a good time to explore [pydantic validation](https://docs.pydantic.dev/latest/concepts/validators/#annotated-validators).

hanars: Theres not a rush to get this specific code out, but I have other things on my plate that I think are higher priority for me. So its less about getting this code released and more about freeing up my time for other feature work

vlm/web_app.py
```diff
@@ -25,9 +33,23 @@ async def status(request: web.Request) -> web.Response:
...
 
+async def match(request: web.Request) -> web.Response:
+    return web.json_response(get_variant_match(request.query))
+
+
 async def init_web_app():
+    spark_conf = {}
+    # memory limits adapted from https://github.com/hail-is/hail/blob/main/hail/python/hailtop/hailctl/dataproc/start.py#L321C17-L321C36
+    if MACHINE_MEM:
+        spark_conf['spark.driver.memory'] = f'{int((int(MACHINE_MEM) - 11) * JVM_MEMORY_FRACTION)}g'
```
jklugherz, line 44: The code to build `spark_conf` here looks identical to how it's done in the `hail_search` web_app, can this be put in a helper function somewhere?

.github/workflows/vlm-unit-tests.yaml
```diff
@@ -29,5 +29,10 @@ jobs:
...
       - name: Run coverage tests
         run: |
+          export VLM_DATA_DIR=./hail_search/fixtures
+          export SEQR_BASE_URL=https://test-seqr.org/
+          export NODE_ID=TestVLM
+          export MACHINE_MEM=24
+          export JAVA_OPTS_XSS=16M
           coverage run --source="./vlm" --omit="./vlm/__main__.py" -m pytest vlm/
-          coverage report --fail-under=90
+          coverage report --fail-under=95
```
jklugherz, line 38: 😎

vlm/web_app.py
hanars: So ideally yes. However, the docker builds for hail_search and the vlm each only include the single directory for their respective apps, and then the unit tests and releases are only triggered on change to those specific directories. Creating a shared directory that is included in thh build for both apps and then updating the github triggers to include the additional shared  directory is possible, but it seemed like more work tahn it was worth to avoid this relatively small amout of copy-paste. However, if you and/or @bpblanken feel like its worth doing than I can do so. However, in the interest of keeping scope sane I would recommend that this gets merged first and then I do the shared code move in a subsequent PR

jklugherz: Considering the amount of work it would take to make this code shared, I think it's fine if it's duplicated.

jklugherz (approved):

PR #4509, opened 2024-12-03
Updates for new reference data structure

bpblanken:
Tests are passing!

I made two massive mistakes when starting this:

1) Fetching globals from the production tables... this helped find the `hl.Struct()` vs `None` issue we encountered yesterday but our enums/versions are out of sync enough that this led to test failures.
2) Annotating the `dbnsfp` values with:
```
ht = ht.annotate(dbnsfp=ht.dbnsfp.annotate(PrimateAI_score=ht.primate_ai.score, MPC_score=ht.mpc.MPC, CADD_phred=ht.cadd.PHRED))
```
`dbnsfp` was missing for several variants, but `CADD` was **not**.   The outer missing swallowed up the inner non-missing.


hanars (approved):
this looks good to me!!

hanars:
We should probably hold off on merging until we are ready to release the updated annotation tables in prod

PR #4516, opened 2024-12-04
update local install test

hanars:
https://github.com/broadinstitute/seqr/pull/4505 failed to update the local install test so it is failing, this will fix the issue

### DIFF
--- a/test_local_deployment.sh
+++ b/test_local_deployment.sh
@@ -7,6 +7,8 @@ set -ex
 docker compose up -d elasticsearch
 docker compose exec -T elasticsearch chmod 777 ./data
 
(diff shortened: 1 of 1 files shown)


bpblanken (approved):
Is there a way for us to run this test in the exact way we have in the docs?

So, entering a folder without the rest of the `seqr` repo and just:

```
wget https://raw.githubusercontent.com/broadinstitute/seqr/master/docker-compose.yml
 wget https://raw.githubusercontent.com/broadinstitute/seqr/master/deploy/postgres/initdb.sql
```

hanars:
probably, but I feel like thats out of scope for fixing a failing test thats blocking deployment. I would have to do research into getting github to run something in a clean environment since by default it runs it with the repo

PR #4524, opened 2024-12-06
Variant Lookup multi genome build build

hanars:
Includes both genome builds on the variant lookup page


hanars:
This is necessary code to do variant lookup across genome builds, but requires fixture updates and new unit tests

jklugherz (approved):

PR #4528, opened 2024-12-10
wes/wgs: stricter comp het inheritance filtering

jklugherz:


hail_search/fixtures/GRCh38/SNV_INDEL/families/WES/F012345_bon_uc499.ht/README.txt
```diff
@@ -0,0 +1,3 @@
+This folder comprises a Hail (www.hail.is) native Table or MatrixTable.
```
jklugherz, line 1: updated variant 4 here to be hom-alt for an unaffected family member (would this be a valid call in real life?) to illustrate how comp het search would behave if one sample type is invalid

Screenshot 2024-12-10 at 4 15 16 PM

hanars: yes this is possible although not super common. But, for example, sometimes if a person has a variant on one strand and a deletion on the other the caller calls it as hom alt, so like this could happen if a parent has an alt and a deletion and the child inherits the deletion and not the alt allele. Its a weird edge case but okay to include in a test

PR #4534, opened 2024-12-11
update action to increment seqr and hail-search chart versions

jklugherz:


.github/workflows/prod-hail-search-release.yaml
```diff
@@ -45,11 +45,14 @@ jobs:
...
           fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
 
-      - name: Update appVersion in hail-search Chart file
+      - name: Update appVersion and Increment hail-search Chart version
         uses: mikefarah/yq@v4.22.1
         with:
           cmd: >
-            yq -i '.appVersion = "${{ github.event.workflow_run.head_sha }}"' charts/hail-search/Chart.yaml
+            yq -i '
+            .appVersion = "${{ github.event.workflow_run.head_sha }}" |
```
bpblanken, line 53: What do you think of using a GitHub action to do this? (https://github.com/marketplace/actions/semver-bump-action)

So the flow might be, 1) fetch the current version, 2) incr with the action 3) set the new version.

jklugherz: Yes, it's more maintainable if the steps are explicit.

I updated the action to follow this flow, using cbrgm/semver-bump-action@main to bump the version. A difference in my solution that I use `yq` to read the version from the file instead of using the github tag. The tag approach didn't feel right because we publish multiple charts and multiple tags from the same repo.

PR #4539, opened 2024-12-12
Improve status designation for in-progress DAG

bpblanken:
We encountered a bug in this connection between the UI and airflow resulting in an in-progress DAG being over-written because it was in a "queued"/"scheduled" state rather than "running".  This expands the check to include all non-completed airflow statuses.

### DIFF
--- a/seqr/views/utils/airflow_utils.py
+++ b/seqr/views/utils/airflow_utils.py
@@ -68,7 +68,7 @@ def _check_dag_running_state():
     endpoint = f'dags/{DAG_NAME}/dagRuns'
     resp = _make_airflow_api_request(endpoint, method='GET')
     dag_runs = resp['dag_runs']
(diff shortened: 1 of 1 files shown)


seqr/views/utils/airflow_utils.py
```diff
@@ -68,7 +68,7 @@ def _check_dag_running_state():
     endpoint = f'dags/{DAG_NAME}/dagRuns'
     resp = _make_airflow_api_request(endpoint, method='GET')
     dag_runs = resp['dag_runs']
-    if dag_runs and dag_runs[-1]['state'] == 'running':
+    if dag_runs and dag_runs[-1]['state'] not in {'success', 'failed'}:
```
bpblanken, line 71: Screenshot 2024-12-12 at 9 56 10 AM

jklugherz (approved):

hanars (approved):

PR #4542, opened 2024-12-13
make relatedness check file downloadable

jklugherz:


seqr/utils/communication_utils.py
```diff
@@ -29,6 +34,49 @@ def _post_to_slack(channel, message):
...
+        _upload_gcs_file_to_slack(channel_id, gcs_file_path, uploaded_filename, message)
+    except Exception as e:
+        logger.error('Slack error: {}: Original message in channel ({}) - {}'.format(e, channel_id, message))
+
+
+def _upload_gcs_file_to_slack(channel_id, gcs_file_path, uploaded_filename, message):
+    file_content = download_gcs_file_as_bytes(gcs_file_path)
+
+    # https://api.slack-gov.com/messaging/files#uploading_files
+    # Note: slackbot needs files:read and files:write scopes
```
jklugherz, line 48: I made a new bot for testing, but we can reuse the existing bot as long as it has these scopes. 

The message _will_ come from the 'Slack API Tester' (the name of the bot app) so maybe we rename it when the scopes are added (which I believe @hanars has to do because only the slack app creator can change these values, or I can be added as an authorized app member and possibly do that).

jklugherz:
> At a high level, this seems like a lot of work to support - is there a strong reason why the file has to be attached in the slack link instead of providing the downloadable link in the slack message like https://storage.cloud.google.com/seqr-loading-temp/v3.1/GRCh38/SNV_INDEL/relatedness_check/e1ffb2bc91f041634068d3c0c0a9f99ed84dd58590238cdf4d6990c0c1480591.tsv

I did it this way because I assumed that the downloadable link might not work for everyone since seqr-loading-temp is private, but I haven't confirmed that and it's possible to give them bucket access.

hanars:
> I did it this way because I assumed that the downloadable link might not work for everyone since seqr-loading-temp is private, but I haven't confirmed that and it's possible to give them bucket access.

So I got the url here from the "Authenticated URL" shown in GCP for this object: 
https://console.cloud.google.com/storage/browser/_details/seqr-loading-temp/v3.1/GRCh38/SNV_INDEL/relatedness_check/e1ffb2bc91f041634068d3c0c0a9f99ed84dd58590238cdf4d6990c0c1480591.tsv;tab=live_object?project=seqr-project

When you hover over the help icon on "Authenticated URL" it states "Only users granted permission can access the object with this link", which implies that users who do not have permission can't access it and those who have access can. But I feel like granting the PMs read access to this bucket is still easier than us having to copy a file and attack it in slack

jklugherz:
> But I feel like granting the PMs read access to this bucket is still easier than us having to copy a file and attack it in slack

I agree. Do we grant one-off access to certain users or is there a PM service acct?

hanars:
All access to buckets is controlled via terraform. We should add user groups and not individual accounts

bpblanken:
@jklugherz i made a pr two days in terraform that added access for a Google group to a bucket.  I would bet there’s already a Google group for the pms, if not the request goes through BITS/Pengo.

hanars (approved):

PR #4556, opened 2024-12-19
make gene order in search deterministic

hanars:


hail_search/test_utils.py
```diff
@@ -1034,7 +1034,7 @@
 }
 
 LOCATION_SEARCH = {
-    'gene_ids': ['ENSG00000177000', 'ENSG00000097046'],
+    'gene_ids': ['ENSG00000097046', 'ENSG00000177000'],
     'intervals': [['2', 1234, 5678], ['7', 1, 11100], ['1', 11785723, 11806455], ['1', 91500851, 91525764]],
```
bpblanken, line 1038: Do the intervals also need to be flipped? `91500851` -> `91525764` belongs with `ENSG00000097046` I think?

hanars: no there does not need to be a relationship between the gene ids and the intervals, and in fact often there are not when users provide a mix of genes and intervals

bpblanken (approved):

PR #4565, opened 2025-01-03
trigger DELETE_FAMILIES dag on transfer families command 

jklugherz:


seqr/views/utils/test_utils.py
```diff
@@ -562,12 +562,32 @@ def assert_no_extra_anvil_calls(self):
         self.mock_get_group_members.assert_not_called()
 
 
+class AirflowTestCase(TestCase):
```
hanars, line 565: Airflow and Anvil are both Broad-only features. There is no case where one is enabled but not the other or vice versa. Therefore, we should not be testing behavior for them when they are decoupled so we should leave the test cases alone and just use the existing `AirflowTestCase`

seqr/management/commands/transfer_families_to_different_project.py
```diff
@@ -1,21 +1,28 @@
...
     search_samples = Sample.objects.filter(is_active=True, individual__family__in=families)
     if search_samples:
         updated_families = search_samples.values_list("individual__family__family_id", flat=True).distinct()
+        updated_family_dataset_types = list(search_samples.values_list('dataset_type', 'individual__family__family_id').distinct())
         family_summary = ", ".join(sorted(updated_families))
         num_updated = search_samples.update(is_active=False)
         logger.info(
             f'Disabled search for {num_updated} samples in the following {len(updated_families)} families: {family_summary}'
         )
+        return updated_family_dataset_types
```
hanars, line 25: It makes more sense to call `trigger_delete_families_dags` directly here, rather than to return and than have another conditional block farther down

seqr/views/utils/airflow_utils.py
```diff
@@ -19,86 +21,138 @@ class DagRunningException(Exception):
...
+        success_message=success_message,
+        success_slack_channel=success_slack_channel,
+        error_message=error_message,
+        upload_info=upload_info
+    )
+
+
+def trigger_airflow_delete_families(
+    dataset_type: str, genome_version: str, error_message: str, families: list[Family],
+    projects: list[Project], success_message: str, success_slack_channel: str
```
hanars, line 46: The `trigger_airflow_data_loading ` function has this level of configurability because it is called in 2 different places with 2 different sets of parameters. However, this function is only ever called once, so including this level of configurability just adds an unnecessary level of complexity. In general it is best to make helper functions as simple as possible and then refactor them to be more configurable later when there is a specific need for that flexibility, rather than building in configurability at the beginning that may never be used at all

In this case, `error_message`, `success_message` and `success_slack_channel` should all be hardcoded in this function. And `projects` is a list you are constructing when instead you could have the function take a singe `project_guid`. Also, if you took `family_guids` instead of `families` you could have the orm query use `values_list` to just fetch the guids from the database instead of fetching the full objects and computing the guids here

seqr/views/utils/airflow_utils.py
```diff
@@ -19,86 +21,138 @@ class DagRunningException(Exception):
...
+):
+    variables = {
+        'projects_to_run': sorted([p.guid for p in projects]),
+        'family_guids': sorted([f.guid for f in families]),
+        'reference_genome': reference_genome_version(genome_version),
+        'dataset_type': dag_dataset_type(dataset_type),
```
hanars, line 52: I'm like 80% sure deleting families is not supported for SVs (and maybe not for MITO) so we need to handle checking the data type before this helper is called and add human-level error handling instead. We should probably actually cause the job to fail entirely if these data types are loaded. Unless I'm wrong and it is supported, in which case cool!

seqr/management/commands/transfer_families_to_different_project.py
```diff
@@ -52,7 +59,30 @@ def handle(self, *args, **options):
...
+    @staticmethod
+    def trigger_delete_families_dags(from_project, updated_family_dataset_types):
+        updated_families_by_dataset_type = defaultdict(list)
+        for dataset_type, family_id in updated_family_dataset_types:
+            updated_families_by_dataset_type[dataset_type].append(family_id)
+
+        for dataset_type, family_ids in updated_families_by_dataset_type.items():
+            families = Family.objects.filter(project=from_project, family_id__in=family_ids)
+            success_message = f'Successfully deleted {len(families)} from {from_project.name} {dataset_type}'
+            error_message = f'ERROR triggering delete families from {from_project.name} {dataset_type}'
```
hanars, line 78: I don't think slack messages are a helpful way to manage notifications here- when non-data managers are triggering loading from within the seqr app it makes sense to report everything in slack so the data managers are kept in the loop about whats happening with loading. But when a privileged user is executing a manage command directly it makes much more sense to show success and error information directly to that user in the logger output. In the event of success just logging it should be sufficient similar to how we log that search iis disabled. In the event of an error logging is also sufficient, but we would want to make sure that log is shown at the end of the job so that the user does not miss it

It probably makes sense to *not* share the error handling behaviors for triggering the different types of dags and instead just have a try-except block here around the `trigger_airflow_delete_families` helper

seqr/views/utils/test_utils.py
jklugherz: My intention here was to create a base test case class for the shared setUp() behavior between airflow test cases. I'll rename it to `BaseAirflowTestCase` and have it inherit `AnvilAuthenticationTestCase` instead of `TestCase` to keep the coupling between both features.

seqr/views/utils/airflow_utils.py
jklugherz: Alright I simplified the `trigger_airflow_delete_families` function following your suggestions here and kept the control flow and error handling separate from what's happening in `trigger_airflow_data_loading`.

seqr/management/commands/transfer_families_to_different_project.py
jklugherz: In addition to making the error handling behavior separate for `trigger_airflow_delete_families`, success and error message are now just logs.

seqr/views/utils/airflow_utils.py
jklugherz: I don't see anything in the pipeline code that would make deleting families not possible in any dataset types, @bpblanken can you confirm that deleting families works for all dataset types?

bpblanken: Yes, "delete families" does work for all types!

hanars: nevermind then, disregard my comment!

seqr/utils/search/add_data_utils.py
```diff
@@ -124,8 +124,8 @@ def prepare_data_loading_request(projects: list[Project], sample_type: str, data
         'projects_to_run': project_guids,
         'callset_path': data_path,
         'sample_type': sample_type,
-        'dataset_type': _dag_dataset_type(sample_type, dataset_type),
-        'reference_genome': GENOME_VERSION_LOOKUP[genome_version],
+        'dataset_type': dag_dataset_type(dataset_type, sample_type),
+        'reference_genome': reference_genome_version(genome_version),
     }
```
hanars, line 129: rather than exporting the formatting for these variables, it might be better to have a `format_loading_pipeline_variables` helper  that could be used both here and in `trigger_airflow_delete_families` - something like 
```
def format_loading_pipeline_variables(projects: list[Project], sample_type: str, dataset_type: str, genome_version: str, **kwargs):
    project_guids = sorted([p.guid for p in projects])
    variables = {
        'projects_to_run': project_guids,
        'sample_type': sample_type,
        'dataset_type': _dag_dataset_type(sample_type, dataset_type),
        'reference_genome': GENOME_VERSION_LOOKUP[genome_version],
        **kwargs
    }
```

seqr/views/utils/airflow_utils.py
```diff
@@ -30,75 +33,112 @@ def trigger_airflow_data_loading(*args, user: User, individual_ids: list[int], s
...
+        'family_guids': sorted(family_guids),
+        'reference_genome': reference_genome_version(from_project.genome_version),
+        'dataset_type': dag_dataset_type(dataset_type),
+    }
+    try:
+        _check_dag_running_state(DELETE_FAMILIES_DAG_NAME)
+        _update_variables(variables, DELETE_FAMILIES_DAG_NAME)
+        _wait_for_dag_variable_update(variables, DELETE_FAMILIES_DAG_NAME)
+        _trigger_dag(DELETE_FAMILIES_DAG_NAME)
+    except Exception as e:
```
hanars, line 68: error handling for this should be handled directly in the manage command, because if this command *was* ever used elsewhere it would probably need different error handling

seqr/views/utils/airflow_utils.py
```diff
@@ -30,75 +33,112 @@ def trigger_airflow_data_loading(*args, user: User, individual_ids: list[int], s
...
+        logger_call = logger.warning if isinstance(e, DagRunningException) else logger.error
+        logger_call(str(e))
+        success = False
+
+    if success:
+        logger.info(f'Successfully triggered DELETE_FAMILIES DAG for {len(family_guids)} family from {from_project.name}/{dataset_type}')
```
hanars, line 74: rather than setting a success variable, just move this to the bottom of the try block - if an exception does happen it won't be called

hanars (changes requested):

seqr/views/utils/test_utils.py
```diff
@@ -562,12 +562,31 @@ def assert_no_extra_anvil_calls(self):
         self.mock_get_group_members.assert_not_called()
 
 
+class BaseAirflowTestCase(AnvilAuthenticationTestCase):
```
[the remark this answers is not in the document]
jklugherz: That makes sense, but I am unsure of what to do with the loading-specific response setup that happens in `AirflowTestCase.setUp` and how I'd adapt the class/setup to make it generalizable for the delete families dag.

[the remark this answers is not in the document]
hanars: So as I said above I would change the existing helper function `assert_airflow_calls` to `assert_airflow_loading_calls` and then add another helper method that would work here for you, such as `assert_ariflow_dag_trigger_calls` and have all shared functionality between the function you need and the existing helper abstracted into shared methods on the class

seqr/views/utils/airflow_utils.py
jklugherz: 👍 I moved this error handling inside of the `_disable_search` method called from manage command

seqr/views/utils/test_utils.py
```diff
@@ -562,17 +562,24 @@ def assert_no_extra_anvil_calls(self):
...
-
-    def setUp(self):
-        self._dag_url = f'{MOCK_AIRFLOW_URL}/api/v1/dags/{DAG_NAME}'
-
+    DAG_RUNS_KWARGS = [{}]
+    DAG_NAME = 'LOADING_PIPELINE'
+
+    def setUp(self, **kwargs):
+        self._dag_url = f'{self.MOCK_AIRFLOW_URL}/api/v1/dags/{self.DAG_NAME}'
+        for kwargs in self.DAG_RUNS_KWARGS:
```
jklugherz, line 576: For the `delete families` test, the response setup needs to happen once for each dag, and there is one dag for each dataset type. This was the solution I came up with so that we can set up responses multiple times and include additional kwargs in some of the responses mocks. It feels kind of clunky though.

seqr/management/tests/transfer_families_to_different_project_tests.py
```diff
@@ -1,12 +1,86 @@
...
     fixtures = ['users', '1kg_project']
+    LOADING_PROJECT_GUID = 'R0001_1kg'  # from-project
+
+    DAG_VARIABLES = {
+        'projects_to_run': [LOADING_PROJECT_GUID],
+        'family_guids': ['F000002_2'],
+        'reference_genome': 'GRCh37',
+    }
+
+    def setUp(self):
```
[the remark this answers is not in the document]
jklugherz: Since I adapted `AirflowTestCase` such that the `setUp` can be used by inheriting classes, I moved the hail backend test outside of the `class TransferFamiliesTest(TestCase)` and into a new `class TransferFamiliesAirflowTest(TransferFamiliesTest, AirflowTestCase)` that inherits `TransferFamiliesTest` so that we're only testing the airflow path in the hail backend test and not the es test.

seqr/views/utils/test_utils.py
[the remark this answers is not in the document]
jklugherz: Thank you for your suggestions. I made a bunch of changes to the `AirflowTestCase` to reduce code duplication and modularize the test setup so that it works for different dags.

seqr/utils/search/add_data_utils.py
```diff
@@ -116,17 +116,30 @@ def format_email(sample_summary, project_link, num_new_samples):
...
-    variables = {
-        'projects_to_run': project_guids,
-        'callset_path': data_path,
-        'sample_type': sample_type,
+def format_loading_pipeline_variables(
+    projects: list[Project], genome_version: str, dataset_type: str, sample_type: str = None, **kwargs
+):
+    keyword_args = dict(**kwargs)
+    if sample_type:
+        keyword_args['sample_type'] = sample_type
```
hanars, line 124: it would be a little neater not to cast `kwargs` to. dict - you could do that by doing 
```
variables = {
 'projects_to_run': ...,
  ...
  **kwargs,
}
if sample_type:
  variables['sample_type'] = sample_type
return variables
```

seqr/views/utils/airflow_utils.py
```diff
@@ -30,26 +31,44 @@ def trigger_airflow_data_loading(*args, user: User, individual_ids: list[int], s
...
+        _trigger_dag(LOADING_PIPELINE_DAG_NAME)
     except Exception as e:
         logger_call = logger.warning if isinstance(e, DagRunningException) else logger.error
         logger_call(str(e), user)
         _send_slack_msg_on_failure_trigger(e, updated_variables, error_message)
         success = False
 
     if success or success_slack_channel != SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL:
-        _send_load_data_slack_msg([success_message] + upload_info, success_slack_channel, updated_variables)
+        _send_load_data_slack_msg(
```
hanars, line 45: this is only used once so I don't think the signature needs to change

seqr/views/utils/airflow_utils.py
```diff
@@ -30,26 +31,44 @@ def trigger_airflow_data_loading(*args, user: User, individual_ids: list[int], s
...
+        [from_project],
+        from_project.genome_version,
+        dataset_type,
+        family_guids=sorted(family_guids)
+    )
+    _check_dag_running_state(DELETE_FAMILIES_DAG_NAME)
+    _update_variables(variables, DELETE_FAMILIES_DAG_NAME)
+    _wait_for_dag_variable_update(variables, DELETE_FAMILIES_DAG_NAME)
+    _trigger_dag(DELETE_FAMILIES_DAG_NAME)
+    logger.info(f'Successfully triggered DELETE_FAMILIES DAG for {len(family_guids)} family from {from_project.name}/{dataset_type}')
```
hanars, line 64: It would be better to move this log either into `_trigger_dag` so all dag events are logged here, or move it to manage command, but its a bit inconsistant where it is

seqr/views/utils/test_utils.py
```diff
@@ -655,54 +663,62 @@ def assert_airflow_calls(self, trigger_error=False, additional_tasks_check=False
...
             'reference_genome': dag_variable_overrides.get('reference_genome', 'GRCh38'),
+            'callset_path': f'gs://test_bucket/{dag_variable_overrides["callset_path"]}',
+            'sample_type': dag_variable_overrides['sample_type'],
         }
         if dag_variable_overrides.get('skip_validation'):
             dag_variables['skip_validation'] = True
         dag_variables['sample_source'] = dag_variable_overrides['sample_source']
-        self._assert_airflow_calls(dag_variables, call_count, offset=offset)
+        self._dag_variables = dag_variables
+        self._assert_airflow_calls(call_count, offset=offset)
```
hanars, line 675: rather than having `self._dag_variables` being constantly  overwritten it would be better for  `_assert_airflow_calls` to take `dag_variables` as an argument

seqr/views/utils/test_utils.py
hanars: Yeah this is a tricky one, theres no amazing solution. I think it would be more in the pattern of how this test case is used elsewhere to have the `setUp` just call the setup once, and then have a public method that can be called by the class to explicitly add additional responses, similar to the existing `add_dag_tasks_response`

seqr/views/utils/test_utils.py
```diff
@@ -655,54 +663,62 @@ def assert_airflow_calls(self, trigger_error=False, additional_tasks_check=False
...
+        )
+
+    def _assert_update_check_airflow_calls(self, call_count, offset, check_updated_path='tasks'):
+        self.assertEqual(responses.calls[offset + 2].request.url, f'{self._dag_url}/{check_updated_path}')
+        self.assertEqual(responses.calls[offset + 2].request.method, 'GET')
+
+        self.assertEqual(responses.calls[offset + 3].request.url, f'{self._dag_url}/{check_updated_path}')
+        self.assertEqual(responses.calls[offset + 3].request.method, 'GET')
+
+        call_cnt = call_count - 1
```
hanars, line 715: this is not actually used in this helper function so instead of computing it here and returning it you can just compute it directly above

seqr/views/utils/test_utils.py
jklugherz: This was perfect! I added a method `set_up_one_dag` that I call two more times in the delete families test setup.

seqr/views/utils/test_utils.py
```diff
@@ -562,17 +562,37 @@ def assert_no_extra_anvil_calls(self):
...
+PROJECT_GUID = 'R0001_1kg'
 
 class AirflowTestCase(AnvilAuthenticationTestCase):
+    MOCK_AIRFLOW_URL = 'http://testairflowserver'
     ADDITIONAL_REQUEST_COUNT = 0
+    DAG_NAME = 'LOADING_PIPELINE'
 
-    def setUp(self):
-        self._dag_url = f'{MOCK_AIRFLOW_URL}/api/v1/dags/{DAG_NAME}'
+    def setUp(self, **kwargs):
```
hanars, line 573: `setUp` should not take keyword args, since you can definitionally set default values for kwargs in the functions that use them just make sure to set one such that a call to `setUp` with no kwargs does the first initialization you want

seqr/management/commands/transfer_families_to_different_project.py
```diff
@@ -1,21 +1,39 @@
...
+            logger_call = logger.warning if isinstance(e, DagRunningException) else logger.error
+            logger_call(str(e))
+
+def _trigger_delete_families_dags(from_project, updated_family_dataset_types):
+    updated_families_by_dataset_type = defaultdict(list)
+    for dataset_type, family_guid in updated_family_dataset_types:
+        updated_families_by_dataset_type[dataset_type].append(family_guid)
+
+    for dataset_type, family_guids in sorted(updated_families_by_dataset_type.items()):
+        trigger_airflow_delete_families(dataset_type, family_guids, from_project)
```
hanars, line 35: you want the try/except block here, not around the whole helper function. If one job fails you will still want to try the other jobs

seqr/management/commands/transfer_families_to_different_project.py
```diff
@@ -1,21 +1,39 @@
...
+        logger.info(f'Successfully triggered DELETE_FAMILIES DAG for {len(family_guids)} family from {from_project.name}/{dataset_type}')
```
hanars, line 36: this command can only be run on a single project so including the project name in the log is a bit redundant/ makes it harder to read. It would be neater as f'Successfully triggered DELETE_FAMILIES DAG for {len(family_guids)} {dataset_type} families'

hanars (approved):

PR #4566, opened 2025-01-03
S3 support added

monkollek:
Here is the code added to seqr to support access for bams/crams living on S3 buckets. This feature has only been tested in regards to viewing them using IGV but is likely to work in other scenarios. 

The implementation supports scenarios where the EC2 instance is running in a different AWS project to where the S3 buckets are stored. The permissions across AWS projects are granted using `assume_role`. The credentials are stored in `S3_ARN_ASSUME_ROLE` and stored as an environment variable. If cross project permissions are not needed leave `S3_ARN_ASSUME_ROLE` undefined and it's assumed the standard aws credentials are setup correctly for `boto3` to use.

The python requirements files may need to be modified to include:
```
urllib.parse
boto3
```


monkollek:
@hanars as I guessed the unit test failed as I didn't put boto3 in the requirements. Can you please make suggestion on how best to do it as I'm not familiar with how the various python requirements files are organized. Thanks!

hanars:
`urllib3` is already in the seqr requirmenets - do we really need to add `urllib.parse` as well or could this be refactored to use the existing requirements?
For adding requirements, add the dependecy to requirements.in and run pip-compile requirements.in

settings.py
```diff
@@ -486,3 +486,6 @@
 else:
     SOCIAL_AUTH_PIPELINE = SOCIAL_AUTH_PIPELINE_BASE + SOCIAL_AUTH_PIPELINE_USER_EXIST + \
                            SOCIAL_AUTH_PIPELINE_ASSOCIATE_USER + SOCIAL_AUTH_PIPELINE_LOG
+
+#S3 specific settings
+S3_ARN_ASSUME_ROLE = os.environ.get('S3_ARN_ASSUME_ROLE', 'none')
```
hanars, line 491: Using a magic string for `'none'` is not great practice as it is really confusing for developers. It would be better if this did not have a default value, and then if there are cases where you actually need this magic string value in calls to s3 you would use something like `S3_ARN_ASSUME_ROLE  or 'none'` to make it clear that its intentional

hanars (commented):
Thanks so much for submitting! Would you mind adding some unit test for this new code? The best way to do this would be to update the tests in `igv_api_tests.py` - there is a test called `test_proxy_google_to_igv` and we would a similar test for s3. Also, `test_add_alignment_sample` has tests for adding IGV samples on both gs and the local file sytsem - we would want to update that test to also include the test case for an s3 file as well

Also, out of curiosity, did you also do any sort of auth integration for AWS, or just file system?

PR #4567, opened 2025-01-03
"Loading Failed" analysis status

hanars:


seqr/migrations/0079_alter_family_analysis_status.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.2.16 on 2025-01-03 19:55
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('seqr', '0078_rename_submit_to_clinvar_variantnote_report'),
```
bpblanken, line 9: oof, I think the wrong migration name snuck through 🙈!

hanars: I mean, thats the name of the last migration file in source control so this is correct...

bpblanken: 😮 I definitely misunderstood how this worked!

bpblanken (approved):

PR #4570, opened 2025-01-06
add selectedMOIs to saved search

jklugherz:
it works! 

https://github.com/user-attachments/assets/77148ccb-b035-45f5-8467-2bb8c9ef4354


hanars:
I can pair on this with you if you want, but at a high level I think changing the behavior for onChange the way you did is probably not the right direction to get it to populate properly. For the issue with initial form population you mention,  I _think_ you need to change the `componentDidUpdate` callback on the `LocusListSelector` thats called when the loading of the locus list completes, and/or change `formatPanelAppItems` to handle the MOI filtering . And for the actual `onChange` behavior itself I think you just want to move the MOI selector up a level so its not nested under panelapp anymore

jklugherz:
@hanars I implemented the `PaLocusListSelector` using the `shouldComponentUpdate` and `componentDidUpdate` approach (and removed the new listener component) but I'm still seeing the values reset, after the page loads.

https://github.com/user-attachments/assets/9614f2d6-9871-48f2-b135-01857708a2cc

I logged the value inside of the `PaMoiSelector` render() and the first time it's logged it is accurately set (as `[XD]`) but on subsequent renders it's replaced with undefined. Do you have any ideas as to what's happening here?

jklugherz:
@hanars I got this to work!! had to really understand what was happening during the lifecycle methods to see when and where the `locus` state was being updated.

ui/shared/utils/panelAppUtils.ts
```diff
@@ -74,15 +75,27 @@ export type PanelAppItem = {
...
-  return items.reduce((acc, item) => {
+  const filteredItems = items.filter((item) => {
+    if (!selectedMOIs || selectedMOIs.length === 0) {
+      return true
+    }
+    const initials = moiToMoiInitials(item.pagene?.modeOfInheritance, false)
+    return selectedMOIs.some((moi) => initials.includes(moi))
+  })
+
+  return filteredItems.reduce((acc, item) => {
```
hanars, line 98: this would be slightly neater as a single `reduce` with a helper method
```suggestion
  const hasSelectedMoi = (item, selectedMOIs) => {
    if (!selectedMOIs || selectedMOIs.length === 0) {
      return true
    }
    const initials = moiToMoiInitials(item.pagene?.modeOfInheritance, false)
    return selectedMOIs.some((moi) => initials.includes(moi))
  }
 ...
 
  return items.reduce((acc, item) => {
    if (!hasSelectedMoi(item, selectedMOIs) {
      return acc
    }
    ...
```

hanars (approved):
yay so excited this is working! One intr suggestion on syntax but otherwise looks awesome!

PR #4572, opened 2025-01-07
shared notification url formatting

hanars:
email notification for RNA loading did not include the project name/URL because we were sending it with a url formatted for slack instead of regular html. Since most notification include links to projects and this is an easy mistake to make I updated the `send_project_notification` helper function to handle slack as well as email notification and to handle constructing the URL for each in the helper instead of having the URL passed in by the calling function. This fixes the bug and also means that this is less likely to happen again in the future. In the process of unifying the functionality there are some slight differences to some of the slack/email messages being generatdd by this helper but they should all be either improvements or at worst neutral changes

### DIFF
--- a/seqr/utils/communication_utils.py
+++ b/seqr/utils/communication_utils.py
@@ -7,6 +7,7 @@ from django.utils.html import strip_tags
 from notifications.signals import notify
 
 BASE_EMAIL_TEMPLATE = 'Dear seqr user,\n\n{}\n\nAll the best,\nThe seqr team'
+EMAIL_MESSAGE_TEMPLATE = 'This is to notify you that data for {notification} has been loaded in seqr project {project_link}'
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1376,17 +1376,17 @@ def test_load_phenotype_prioritization_data(self, mock_subprocess, mock_send_ema
...
     @staticmethod
-    def _assert_expected_notifications(mock_send_email, expected_notifs: list[dict], has_html=False):
+    def _assert_expected_notifications(mock_send_email, expected_notifs: list[dict]):
         calls = []
         for notif_dict in expected_notifs:
             project_guid = notif_dict.get('project_guid', PROJECT_GUID)
             project_name = notif_dict.get('project_name', '1kg project nåme with uniçøde')
             url = f'https://test-seqr.org/project/{project_guid}/project_page'
-            project_link = f'<a href={url}>{project_name}</a>' if has_html else f'<{url}|{project_name}>'
+            project_link = f'<a href={url}>{project_name}</a>'
```
hanars, line 1389: this test change reflects the actual bug fix - we should never be sending an email with the URL formatted in anything other than html

seqr/utils/communication_utils.py
```diff
@@ -55,9 +56,16 @@ def send_html_email(email_body, process_message=None, **kwargs):
     email_message.send()
 
 
-def send_project_notification(project, notification, email, subject):
+def send_project_notification(project, notification, subject, notification_prefix='Loaded ', email_template=None, slack_channel=None, slack_detail=None):
     users = project.subscribers.user_set.all()
-    notify.send(project, recipient=users, verb=notification)
+    notify.send(project, recipient=users, verb=f'{notification_prefix}{notification}')
```
jklugherz, line 61: I don't see `notification_prefix` being overridden in any calls to `send_project_notification`, can we remove it and make the `verb` `f'Loaded {notification}')`?

jklugherz (approved):
everything else lgtm

PR #4576, opened 2025-01-08
Local vcf option dropdown

hanars:


seqr/utils/file_utils_tests.py
```diff
@@ -30,32 +30,3 @@ def test_mv_file_to_gs(self, mock_logger, mock_subproc):
         mock_subproc.Popen.assert_called_with('gsutil mv /temp_path gs://bucket/target_path', stdout=mock_subproc.PIPE, stderr=mock_subproc.STDOUT, shell=True)  # nosec
         mock_logger.info.assert_called_with('==> gsutil mv /temp_path gs://bucket/target_path', None)
         process.wait.assert_called_with()
-
-    @mock.patch('seqr.utils.file_utils.subprocess')
-    @mock.patch('seqr.utils.file_utils.logger')
-    def test_get_gs_file_list(self, mock_logger, mock_subproc):
```
hanars, line 36: In general in the seqr repo we try not to directly unit test helper functions and instead test the places code is called so changes to function names/signatures etc do not need test updates. However, this test for the helper did exist and since as part of this refactor it is not longer a public function I opted to remove the test for the helper function here instead of updating it. There was one edge case that was only covered by this test so I updated the anvil api tests that call this helper to test that edge case

bpblanken (approved):

PR #4581, opened 2025-01-14
report pipeline validation errors from seqr

jklugherz:
draft pr while I work on tests!

### DIFF
--- a/seqr/management/commands/check_for_new_samples_from_pipeline.py
+++ b/seqr/management/commands/check_for_new_samples_from_pipeline.py
@@ -13,7 +13,7 @@ import re
 from reference_data.models import GENOME_VERSION_LOOKUP
 from seqr.models import Family, Sample, SavedVariant
 from seqr.utils.communication_utils import safe_post_to_slack
-from seqr.utils.file_utils import file_iter, list_files
+from seqr.utils.file_utils import file_iter, list_files, is_google_bucket_file_path
 from seqr.utils.search.add_data_utils import notify_search_data_loaded
 from seqr.utils.search.utils import parse_valid_variant_id
 from seqr.utils.search.hail_search_utils import hail_variant_multi_lookup, search_data_type
@@ -140,17 +140,18 @@ class Command(BaseCommand):
                     f'Dataset Type: {run_details["dataset_type"]}',
                     f'Run ID: {run_details["run_version"]}',
                     f'Validation Errors: {error_summary["error_messages"]}',
-                    f'See more at https://storage.cloud.google.com{file_path}'
                 ]
-                messages.append('/n'.join(summary))
+                if is_google_bucket_file_path(file_path):
+                    summary.append(f'See more at https://storage.cloud.google.com/{file_path[5:]}')
+                messages.append('\n'.join(summary))
                 reported_runs.add(run_dir)
 
         if messages:
             safe_post_to_slack(
                 SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, '\n\n'.join(messages),
             )
-        for run_dir in reported_runs:
-            write_multiple_files([(ERRORS_REPORTED_FILE_NAME, [], [])], run_dir, user=None, file_format=None)
+        # for run_dir in reported_runs:
+        #     write_multiple_files([(ERRORS_REPORTED_FILE_NAME, [], [])], run_dir, user=None, file_format=None)
 
     @classmethod
     def _load_new_samples(cls, metadata_path, genome_version, dataset_type, run_version):
--- a/seqr/management/commands/check_for_new_samples_from_pipeline.py
+++ b/seqr/management/commands/check_for_new_samples_from_pipeline.py
@@ -1,6 +1,5 @@
 import os
 from collections import defaultdict
-from tempfile import TemporaryDirectory
 
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.core.management.base import BaseCommand, CommandError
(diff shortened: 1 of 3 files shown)


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -87,9 +94,49 @@ def handle(self, *args, **options):
...
+                f'See more at https://storage.cloud.google.com/{path}'
+            ]
+            messages.append('/n'.join(summary))
+            reported_run_paths.add(path)
+
+        if messages:
+            safe_post_to_slack(
+                SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL, '\n\n'.join(messages),
+            )
+            with TemporaryDirectory() as temp_dir_name:
```
hanars, line 125: while it is only writing one file, I think using the `write_multiple_files` helper function in export utils would allow you to reuse most of that code and not need to reimplement so much boiler plate

hanars: I think the usage would just be
```
write_multiple_files([(ERRORS_REPORTED_FILE_NAME, [], [])], run_directory, user)
```

hanars:
At a high level it looks like you are running 3 separate `ls` commands to check the runs - one to get the runs with validation errors, one to get the runs with the errors already reported and one to get the success runs. I think at this point its probably more effective to run a single `ls` with the `file_name` arg as a wildcard and then group the results by `run_version` to get which files exist for which runs in a single operation

seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -324,29 +333,41 @@ def test_command(self, mock_email, mock_airtable_utils):
...
 
         call_command('check_for_new_samples_from_pipeline')
         self.mock_logger.info.assert_called_with('No loaded data available')
         self.mock_subprocess.assert_not_called()
         mock_email.assert_not_called()
         self.mock_send_slack.assert_not_called()
+        mock_mv_file_to_gs.assert_not_called()
 
         local_files = [
+            '/seqr/seqr-hail-search-data/GRCh38/SNV_INDEL/runs/manual__2025-01-13/_ERRORS_REPORTED',
```
jklugherz, line 347: I added 3 new files in 2 run directories. The first directory has both `_ERRORS_REPORTED` and `validation_errors.json` and the second has just `validation_errors.json`.

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -87,13 +96,63 @@ def handle(self, *args, **options):
 
         logger.info('DONE')
 
+    def _get_runs(self, **kwargs):
+        """ Returns two dictionaries:
+            - run_files, a mapping of the run directory to a set of filenames inside the directory
+            - run_args, a mapping of the run directory to a dict of args for that run
```
jklugherz, line 102: There's probably a better data structure to capture these mappings, but we're definitely only making one `list_files`/`gsutil ls` call now.

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -1,4 +1,7 @@
+import os
 from collections import defaultdict
+from tempfile import TemporaryDirectory
```
hanars, line 3: this is no longer used

seqr/management/commands/check_for_new_samples_from_pipeline.py
hanars: I think you could use a single structure with `runs = defaultdict(lambda: {'files': set()})`. Then where you are currently updating `run_files` and `run_args` you would do 
```
runs[run_dirname]['files'].add(file_name)
runs[run_dirname].update(match_dict)
```

And then for example in `_report_validation_errors` the way you would use it is
```
for run_dir, run_details in runs.items():
    files = run_details['files']
```

seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
hanars: for consistency we should update `RUN_PATHS` to align with these changes

seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -282,10 +287,14 @@ def _test_call(self, error_logs, reload_annotations_logs=None, run_loading_logs=
     @mock.patch('seqr.management.commands.check_for_new_samples_from_pipeline.MAX_LOOKUP_VARIANTS', 1)
     @mock.patch('seqr.views.utils.airtable_utils.BASE_URL', 'https://test-seqr.org/')
     @mock.patch('seqr.views.utils.airtable_utils.MAX_UPDATE_RECORDS', 2)
+    @mock.patch('seqr.views.utils.export_utils.os.makedirs')
+    @mock.patch('seqr.views.utils.export_utils.mv_file_to_gs')
```
hanars, line 291: since we are already mocking the subprocess calls used for interacting with `gs` in this test case it would be better not to mock this helper function and instead assert that the subprocess calls all look right including the `mv` calls

seqr/management/commands/check_for_new_samples_from_pipeline.py
jklugherz: this is perfect, thanks!

hanars (approved):

PR #4603, opened 2025-01-17
standardize pipeline options in Load Data UI

jklugherz:
adds 'skip_check_sex_and_relatedness' as a dag variable

### DIFF
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -117,11 +117,9 @@ def format_loading_pipeline_variables(
         variables['sample_type'] = sample_type
     return variables
 
-def prepare_data_loading_request(
-    projects: list[Project], sample_type: str, dataset_type: str, genome_version: str,
-    data_path: str, user: User, pedigree_dir: str, raise_pedigree_error: bool = False, individual_ids: list[int] = None,
-    skip_validation: bool = False, skip_check_sex_and_relatedness: bool = False, ignore_missing_samples_when_remapping: bool = False,
-):
+def prepare_data_loading_request(projects: list[Project], sample_type: str, dataset_type: str, genome_version: str,
+                                 data_path: str, user: User, pedigree_dir: str,  raise_pedigree_error: bool = False,
+                                 individual_ids: list[int] = None, skip_validation: bool = False, skip_check_sex_and_relatedness: bool = False):
     variables = format_loading_pipeline_variables(
         projects,
         genome_version,
@@ -129,14 +127,13 @@ def prepare_data_loading_request(
         sample_type,
         callset_path=data_path,
     )
-    config_params = {
-        'skip_validation': skip_validation,
-        'skip_check_sex_and_relatedness': skip_check_sex_and_relatedness,
-        'ignore_missing_samples_when_remapping': ignore_missing_samples_when_remapping,
-    }
+    if skip_validation:
+        variables['skip_validation'] = True
+    if skip_check_sex_and_relatedness:
+        variables['skip_check_sex_and_relatedness'] = True
     file_path = _get_pedigree_path(pedigree_dir, genome_version, sample_type, dataset_type)
     _upload_data_loading_files(projects, user, file_path, individual_ids, raise_pedigree_error)
-    return variables, file_path, config_params
+    return variables, file_path
 
 
 def _dag_dataset_type(sample_type: str, dataset_type: str):
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -117,9 +117,11 @@ def format_loading_pipeline_variables(
         variables['sample_type'] = sample_type
     return variables
 
-def prepare_data_loading_request(projects: list[Project], sample_type: str, dataset_type: str, genome_version: str,
-                                 data_path: str, user: User, pedigree_dir: str,  raise_pedigree_error: bool = False,
-                                 individual_ids: list[int] = None, skip_validation: bool = False):
+def prepare_data_loading_request(
+    projects: list[Project], sample_type: str, dataset_type: str, genome_version: str,
+    data_path: str, user: User, pedigree_dir: str, raise_pedigree_error: bool = False, individual_ids: list[int] = None,
+    skip_validation: bool = False, skip_check_sex_and_relatedness: bool = False, ignore_missing_samples_when_remapping: bool = False,
+):
     variables = format_loading_pipeline_variables(
         projects,
         genome_version,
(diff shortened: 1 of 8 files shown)


seqr/utils/search/add_data_utils.py
```diff
@@ -117,21 +117,26 @@ def format_loading_pipeline_variables(
         variables['sample_type'] = sample_type
     return variables
 
-def prepare_data_loading_request(projects: list[Project], sample_type: str, dataset_type: str, genome_version: str,
-                                 data_path: str, user: User, pedigree_dir: str,  raise_pedigree_error: bool = False,
-                                 individual_ids: list[int] = None, skip_validation: bool = False):
+def prepare_data_loading_request(
```
hanars, line 120: So this function is used twice, once for triggering the airflow DAG and once for triggering the pipeline runner service. Rather than putting the complex airflow logic into the shared helper function and then having to update the call to the pipeline runner to have everything, I think it would be better to leave this shared function alone and just directly implement the airflow-only logic in the one place it is used

seqr/views/apis/data_manager_api.py
```diff
@@ -537,7 +536,12 @@ def load_data(request):
     loading_args = (
         project_models, sample_type, dataset_type, request_json['genomeVersion'], _callset_path(request_json),
     )
-    loading_kwargs = {'user': request.user, 'skip_validation': skip_validation}
+    loading_kwargs = {
+        'user': request.user,
+        'skip_validation': request_json.get('skipValidation', False),
+        'skip_check_sex_and_relatedness': request_json.get('skipSRChecks', False),
+        'ignore_missing_samples_when_remapping': request_json.get('ignoreMissingRemapSamples', False),
```
hanars, line 543: I do not think we should add support  for`ignore_missing_samples_when_remapping` in the UI. I think when remap fails we should be fixing the remap files, not just perpetually kicking the can down the road with increasingly broken remap files. In like an emergency situation we can always manually kick off the DAG with this argument if we really have to

seqr/views/apis/data_manager_api.py
```diff
@@ -537,7 +536,12 @@ def load_data(request):
     loading_args = (
         project_models, sample_type, dataset_type, request_json['genomeVersion'], _callset_path(request_json),
     )
-    loading_kwargs = {'user': request.user, 'skip_validation': skip_validation}
+    loading_kwargs = {
+        'user': request.user,
+        'skip_validation': request_json.get('skipValidation', False),
+        'skip_check_sex_and_relatedness': request_json.get('skipSRChecks', False),
```
hanars, line 542: is this parmateer supported in both the pipeline runner service and airflow?

seqr/utils/search/add_data_utils.py
```diff
@@ -117,21 +117,26 @@ def format_loading_pipeline_variables(
...
+):
     variables = format_loading_pipeline_variables(
         projects,
         genome_version,
         dataset_type,
         sample_type,
         callset_path=data_path,
     )
-    if skip_validation:
-        variables['skip_validation'] = True
```
hanars, line 131: I like the current behavior of only be including this argument when its true, is there a reason to change this behvior?

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1708,9 +1712,9 @@ def _has_expected_ped_files(self, mock_open, mock_mkdir, dataset_type, *args, sa
         super()._has_expected_ped_files(mock_open, mock_mkdir, dataset_type,  *args, sample_type, **kwargs)
         mock_mkdir.assert_called_once_with(self._local_pedigree_path(dataset_type, sample_type), exist_ok=True)
 
-    def _assert_success_notification(self, dag_json):
+    def _assert_success_notification(self, dag_json, config_params):
```
hanars, line 1715: it appears in both calls to `_assert_success_notification` you merge `dag_json` and `config_params` so it would be clearer to leave the helper function alone and instead change the call to it

ui/pages/DataManagement/components/LoadData.jsx
```diff
@@ -60,6 +60,18 @@ const CALLSET_PAGE_FIELDS = [
...
     asFormInput: true,
   },
+  {
+    name: 'skipSRChecks',
+    label: 'Skip Sex and Relatedness Checks',
+    component: InlineToggle,
+    asFormInput: true,
+  },
+  {
+    name: 'ignoreMissingRemapSamples',
```
hanars, line 70: per my above comment, this should not be added to the UI

seqr/views/apis/data_manager_api.py
bpblanken: Yes, Julia just added this to the pipeline runner service.   It won't actually do anything though, since local users are not given instructions for enabling s&r in their environment.

seqr/views/apis/data_manager_api.py
bpblanken: Yeah, I'm on board with removing this as an option from the pipeline as soon as possible.

seqr/views/apis/data_manager_api.py
jklugherz: @bpblanken should it not be in the pipeline-runner service then?

bpblanken: I think we still need it there, it'll just default to false.

seqr/utils/search/add_data_utils.py
```diff
@@ -117,21 +117,26 @@ def format_loading_pipeline_variables(
...
+    config_params = {
```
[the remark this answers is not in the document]
hanars: For some loads we will start out with validation disabled which is what thi use case is for, but I feel like for most loading we start out with validation enabled and then if validation fails we re-trigger the DAG with validation off. In that scenario I feel like we will still be re-triggering loading directly in aiflow, not going through the whole seqr UI again, as its much easier to re-launch a job with almost the exact same arguments in airflow then going through the whole seqr workflow, right?

seqr/utils/search/add_data_utils.py
jklugherz: I think it's ok to update this function because ideally we want behavior to match between the airflow and pipeline-runner services, so it's they should take the same arguments.

seqr/utils/search/add_data_utils.py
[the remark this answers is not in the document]
jklugherz: We (Dan) will have the ability to re-trigger the pipeline with validation enabled by submitting another seqr UI Load Data request once this PR is merged, right? 

Then it might make sense to make these dag args again.

seqr/utils/search/add_data_utils.py
jklugherz: I had updated the defaulting behavior but I changed it back and made the `skip_check_sex_and_relatedness` defaulting match.

seqr/views/apis/data_manager_api.py
jklugherz: > so we should make sure to only add this for the Broad users
this is done

seqr/utils/search/add_data_utils.py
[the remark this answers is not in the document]
bpblanken: (sorry I missed this yesterday)

I think we want to move towards triggering the re-runs with validation disabled after a validation failure through the UI since we're trying to deprecate airflow and have Dan (or anyone who knows how airflow works) removed from loading.  It _is_ easier in airflow, but we're effectively giving up that simplicity to avoid having airflow at all.

ui/pages/DataManagement/components/LoadData.jsx
```diff
@@ -97,7 +97,14 @@ const MULTI_DATA_TYPE_CALLSET_PAGE = {
       placeholder: 'gs://',
       ...FILE_PATH_FIELD,
     },
-    ...CALLSET_PAGE_FIELDS,
+    CALLSET_PAGE_FIELDS[0],
```
hanars, line 100: this change seems incorrect to me - why would we be removing fields from the UI as part of this change?

hanars, line 100: this would be much more maintainable if you just put the SR toggle before the other validation toggle instead of splitting the fields

PR #4611, opened 2025-01-23
chore: bump hail search and vlm hail & python versions

bpblanken:


hail_search/definitions.py
```diff
@@ -1,7 +1,7 @@
-from enum import Enum
+from enum import StrEnum
 
 
-class SampleType(str, Enum):
+class SampleType(StrEnum):
```
bpblanken, line 4: There was a behavior change between 3.9 and 3.11 here:

in 3.9:
```
In [5]: str(SampleType.WGS)
Out[5]: 'SampleType.WGS'

In [6]: f'/{SampleType.WGS}/'
Out[6]: '/WGS/'
```

in 3.11:

```
In [2]: f'/{SampleType.WGS}/'
Out[2]: '/SampleType.WGS/'

In [3]: str(SampleType.WGS)
Out[3]: 'SampleType.WGS'
```

using the new `StrEnum` that was released with `3.11` resolves. the issue.

PR #4615, opened 2025-01-23
properly handle regex for directory name

hanars:
The latest changes to the check samples job to include validation results are failing because the new `ls` command output includes the directory name as well as all the files in the directory and the regex no longer works


jklugherz (approved):
thanks for fixing this!!!

bpblanken:
It would be great to release this today if possible!  There's a successful pipeline run that's hanging out in intermediate state.

PR #4631, opened 2025-02-03
make multi-sample type comp-het search more permissive

jklugherz:


hail_search/queries/mito.py
```diff
@@ -298,6 +298,25 @@ def _family_has_valid_inheritance(ht, sample_type, family_idx, other_sample_type
...
+                    self._is_valid_comp_het_family(v.v1, v.v2, i), family_samples[0]['familyGuid']
+                )
+            ).filter(hl.is_defined)
+        ))
+        # Get indices of valid comp het families
+        return variants.map(lambda v: v.annotate(
+            valid_family_indices=hl.enumerate(v.v1.family_entries).map(lambda x: x[0]).filter(
+                lambda i: v.valid_family_guids.contains(v.v1.family_entries[i][0]['familyGuid'])
+            )
+        ))
```
hanars, line 318: Here you are enumerating `v.v1.family_entries`, mapping it to get just the index `i`, and then using the index to access `v.v1.family_entries[i]`. It would be much neater if you skipped the enumeration entirely and did
```
valid_family_indices=v.v1.family_entries.filter(
    lambda family_samples: v.valid_family_guids.contains(family_samples[0]['familyGuid'])
)
```

hail_search/queries/mito.py
```diff
@@ -298,6 +298,25 @@ def _family_has_valid_inheritance(ht, sample_type, family_idx, other_sample_type
...
+            return super()._annotate_comp_het_valid_family_indices(variants)
+
+        # Get familyGuids that are valid comp het families
+        variants = variants.map(lambda v: v.annotate(
+            valid_family_guids=hl.enumerate(v.v1.family_entries).starmap(
+                lambda i, family_samples: hl.or_missing(
+                    self._is_valid_comp_het_family(v.v1, v.v2, i), family_samples[0]['familyGuid']
+                )
+            ).filter(hl.is_defined)
+        ))
```
hanars, line 312: I think you could reduce some of the copy paste/ logic duplication you have implemented here and actually use the `super` method to annotate `valid_family_indices`, and then use the indices to get the guids. So something like
```
variants = super()._annotate_comp_het_valid_family_indices(variants)
if not self._has_both_sample_types:
    return variants

variants = variants.map(lambda v: v.annotate(
    valid_family_guids=v.valid_family_indices.map(
        lambda i:  v.v1.family_entries[i][0]['familyGuid']
    )
))
...
```

hail_search/queries/mito.py
jklugherz: I believe we still need enumeration here to get the valid indices, I can use the 'family_samples' directly from the tuple created by hl.enumerate, but this is almost the same as what I had before:

```
  valid_family_indices=hl.enumerate(v.v1.family_entries).filter(
      lambda x: v.valid_family_guids.contains(x[1][0]['familyGuid'])
  ).map(lambda x: x[0])
```

hail_search/queries/mito.py
jklugherz: I am now using the indices to get the guids then re-annotating the indices based on the guids

hanars (approved):

Issue #4580, opened 2025-01-14
Validate expected samples present in internal VCF before loading


lynnpais:
**Is your feature request related to a problem? Please describe.**
Improv efficiency of data loading workflow
 
**Describe the solution you'd like**
Validate expected samples present in internal VCF before loading

PR #4636, opened 2025-02-05
Pedigree info validation clean up

hanars:
The helper functions we use to validate pedigrees involve a lot of code duplication, in particular in the way that they look up potential existing individuals in the database. In order to implement https://github.com/broadinstitute/seqr/issues/4580 we need better shared utilities for validating samples which was hard to do with the existing mess of validation. This cleans up the validation code and also abstracts out the logic used for validating the search sample subset into a single helper function.
This PR contains essentially no behavior changes- the only exception is that previously we did 2 steps to validate AnVIL loading requests and thats now done in one step, so certain requests now include more validation falures on the first pass instead of some failures and then more failures on the subsequent request after those are fixed

### DIFF
--- a/seqr/views/apis/anvil_workspace_api.py
+++ b/seqr/views/apis/anvil_workspace_api.py
@@ -4,6 +4,7 @@ import time
 from datetime import datetime
 from functools import wraps
 from collections import defaultdict
+from random import sample
 
 from django.contrib.auth.decorators import user_passes_test
 from django.contrib.auth.views import redirect_to_login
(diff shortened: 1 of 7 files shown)


seqr/views/apis/anvil_workspace_api.py
```diff
@@ -225,61 +224,72 @@ def add_workspace_data(request, project_guid):
...
-    # Parse families/individuals in the uploaded pedigree file
+def _parse_uploaded_pedigree(request_json, project=None, search_dataset_type=None):
+    loaded_sample_type = None
+    loaded_individual_ids = []
+    def validate_expected_samples(record_family_ids, affected_status_by_family, previous_loaded_individuals, sample_type):
+        errors, loaded_ids = _validate_expected_samples(
+            request_json['vcfSamples'], search_dataset_type,
+            record_family_ids, affected_status_by_family, previous_loaded_individuals, sample_type,
+        )
+        nonlocal loaded_individual_ids
```
jklugherz, line 244: I hadn't seen `nonlocal` before, I feel like personally it would be preferable to return these values from the nested function and do the updating of the `loaded_sample_type` and `loaded_individual_ids` via the returned values, but that's just style preference I guess and also I wouldn't request you change it here in this PR since you're just moving stuff around in here.

jklugherz (approved):

PR #4649, opened 2025-02-12
Internal callset validation

hanars:
Shares the VCF validation we run for AnVIL data loading with our internal data loading. This also adds that validation for local installs as well as the Broad installation

### DIFF
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -183,7 +183,7 @@ def _get_pedigree_path(pedigree_dir: str, genome_version: str, sample_type: str,
 
 
 def get_loading_samples_validator(vcf_samples: list[str], loaded_individual_ids: list[int], sample_source: str,
-                                  missing_family_samples_error: str, missing_family_samples_template: str = '{family} ({samples})',
+                                  missing_family_samples_error: str, missing_family_samples_template: str = '{family_id} ({samples})',
                                   missing_family_samples_divider: str ='; ', loaded_sample_types: list[str] = None,
                                   fetch_missing_loaded_samples: Callable = None) -> Callable:
 
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -181,11 +181,16 @@ def _get_pedigree_path(pedigree_dir: str, genome_version: str, sample_type: str,
     return f'{pedigree_dir}/{GENOME_VERSION_LOOKUP[genome_version]}/{dag_dataset_type}/pedigrees/{sample_type}'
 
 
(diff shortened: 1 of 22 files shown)


seqr/utils/vcf_utils.py
```diff
@@ -60,7 +60,12 @@ def _get_vcf_meta_info(line):
     return None
 
 
-def validate_vcf_and_get_samples(vcf_filename, genome_version):
+def validate_vcf_and_get_samples(data_path, user, genome_version, path_name=None, allowed_exts=None):
+    vcf_filename = _validate_vcf_exists(data_path, user, path_name, allowed_exts)
+
+    if allowed_exts and vcf_filename.endswith(allowed_exts):
+        return None
```
jklugherz, line 67: does this chunk mean that we don't validate MITO or SV files?

seqr/utils/search/add_data_utils.py
```diff
@@ -178,3 +172,65 @@ def _upload_data_loading_files(projects: list[Project], user: User, file_path: s
 def _get_pedigree_path(pedigree_dir: str, genome_version: str, sample_type: str, dataset_type: str):
     dag_dataset_type = _dag_dataset_type(sample_type, dataset_type)
     return f'{pedigree_dir}/{GENOME_VERSION_LOOKUP[genome_version]}/{dag_dataset_type}/pedigrees/{sample_type}'
+
+
+def get_loading_samples_validator(vcf_samples: list[str], loaded_individual_ids: list[int], sample_source: str,
+                                  missing_family_samples_error: str, loaded_sample_types: list[str] = None,
+                                  fetch_missing_loaded_samples: Callable = None) -> Callable:
+
+    def validate_expected_samples(record_family_ids, previous_loaded_individuals, sample_type):
```
jklugherz, line 181: why not use `validate_expected_samples` directly instead of returning it inside of `get_loading_samples_validator` and then using the returned callable?

jklugherz: ah I see the callable is used inside of `parse_basic_pedigree_table` - can we just call `validate_expected_samples` inside of this function instead of passing the callable? I guess the purpose of this pattern is to provide the appropriate messages based on where `parse_basic_pedigree_table` is called, and I'm not sure there's a better way to do that.

seqr/utils/vcf_utils.py
hanars: We validate VCFs only, so MITO MTs and GCNV BED files are not validated, by SV VCFs are

seqr/utils/search/add_data_utils.py
hanars: because in the anvil case we are running this inside the parse pedigree file validator as part of a broader set of pedigree file validation that is run before any databse models are created, and in the internal loading case we are runnng this on the already created dtaabse models with no other pedigree validation

jklugherz (approved):

PR #4653, opened 2025-02-13
Sv gene search

hanars:
SV search is no longer working for an all-project single gene search. This is because unlike with other data types we can not prefilter the project tables to a single gene. However, the full SV_WGS is only ~500,000 rows (roughly the same size as a SNV_INDEL WGS project table) and can be fully read into hail. This therefore refactors SV gene searches to first read in the annotations table and prefilter the project tables on that, instead of vice-versa

### DIFF
--- a/hail_search/queries/base.py
+++ b/hail_search/queries/base.py
@@ -369,14 +369,17 @@ class BaseHailTableQuery(object):
             all_project_hts.append((ht, sample_data))
         return all_project_hts
 
-    def import_filtered_table(self, project_samples: dict, num_families: int, **kwargs):
+    def _import_families_tables(self, project_samples: dict, num_families: int, **kwargs):
         if num_families == 1 or len(project_samples) == 1:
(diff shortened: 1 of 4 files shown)


hail_search/queries/base.py
```diff
@@ -879,7 +889,7 @@ def _parse_annotations(self, annotations, annotations_secondary, **kwargs):
                 secondary_annotation_overrides = self._get_annotation_override_fields(
                     annotations_secondary, override_fields=self.SECONDARY_ANNOTATION_OVERRIDE_FIELDS, **kwargs)
                 has_data_type_secondary_annotations |= bool(secondary_annotation_overrides)
-                has_different_secondary &= secondary_annotation_overrides != annotation_overrides
+                has_different_secondary |= secondary_annotation_overrides != annotation_overrides
```
hanars, line 892: this is an unrelated bug I bumped into when I added a test case

PR #4654, opened 2025-02-14
Check sample reload only bug

hanars:
We ran into a bug in the check samples job when there are no new samples being loaded. This PR does the following:
- Fixes that bug
- Changes the behavior for reloading saved variants to also reload previously loaded families (when adding a unit test to cover this case I discovered this was not happening and we want it to)
- Splits out the logic into more localized try/except blocks with clearer errors so that future errors will be clearer about where they are happening, and also where possible will allow other unrelated steps to run even where some upstream steps fail

### DIFF
--- a/seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
+++ b/seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
@@ -26,7 +26,7 @@ EXISTING_WGS_SAMPLE_GUID = 'S000144_na20888'
 EXISTING_SV_SAMPLE_GUID = 'S000147_na21234'
 SAMPLE_GUIDS = [ACTIVE_SAMPLE_GUID, REPLACED_SAMPLE_GUID, NEW_SAMPLE_GUID_P3, NEW_SAMPLE_GUID_P4]
 GCNV_SAMPLE_GUID = f'S00000{GCNV_GUID_ID}_na20889'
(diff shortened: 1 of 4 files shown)


seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -427,28 +429,31 @@ def test_command(self, mock_email, mock_temp_dir):
         })
 
         # Tests Sample models created/updated
-        updated_sample_models = Sample.objects.filter(guid__in=SAMPLE_GUIDS+GCNV_SAMPLE_GUIDS)
-        self.assertEqual(len(updated_sample_models), len(SAMPLE_GUIDS+GCNV_SAMPLE_GUIDS))
+        snv_indel_samples = Sample.objects.filter(data_source='auto__2023-08-09')
+        gcnv_samples = Sample.objects.filter(data_source='auto__2024-09-14')
+        updated_sample_models = snv_indel_samples | gcnv_samples
```
hanars, line 434: I changed the logic in this part of the test because the way it was written if we alter the test so that more samples are created the test will succeed without testing those new samples. Changing the order here now causes the test to fail if you are creating more samples that are not added to `SAMPLE_GUIDS` or `GCNV_SAMPLE_GUIDS`

jklugherz (approved):

PR #4663, opened 2025-02-20
update individuals with sample qc

jklugherz:
much of the logic is based on [this](https://github.com/broadinstitute/seqr/blob/master/seqr/views/apis/data_manager_api.py#L193) function, but I tried to make fewer db queries


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -350,6 +359,59 @@ def _update_pdos(session, project_guid, sample_ids):
...
+                else:
+                    unknown_pop_filter_flags.add(flag)
+
+            individual.filter_flags.update(filter_flags)
+            individual.pop_platform_filters.update(pop_platform_filters)
+            individual.population.update(record['qc_pop'].upper())
+            updated_individuals.append(individual)
+
+        if updated_individuals:
+            Individual.objects.bulk_update(
```
hanars, line 400: Its better to use the `Individual.bulk_update_models` helper as it includes audit logging by default

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -350,6 +359,59 @@ def _update_pdos(session, project_guid, sample_ids):
...
+                updated_individuals,
+                ['filter_flags', 'pop_platform_filters', 'population'],
+            )
+
+        if unknown_filter_flags:
+            message = 'The following filter flags have no known corresponding value and were not saved: {}'.format(
+                ', '.join(unknown_filter_flags))
+            logger.warning(message)
```
hanars, line 408: warnings are useful when something is triggered in the UI and the person who does so will be shown the warnings, but they are essentially useless in a job that runs automatically in the background. 
I think we should hard fail updating QC if there are unexpected values, as we now have more control over how these are generated and can choose ot to include things we don't care about in our metadata

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -350,6 +359,59 @@ def _update_pdos(session, project_guid, sample_ids):
 
         return sorted(pdos_to_create.keys())
 
+    @classmethod
+    def update_individuals_sample_qc(cls, sample_type, projects, sample_qc_map):
+        sample_individual_map = {
+            i.sample.sample_id: i for i in Individual.objects.filter(
+                family__project__in=projects,
+                sample__sample_id__in=sample_qc_map.keys(),
+            ).select_related('sample')
```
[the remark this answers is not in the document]
jklugherz: I was considering using `updated_samples`! glad to hear I was on the right track, thanks for the suggestion.

seqr/management/commands/check_for_new_samples_from_pipeline.py
jklugherz: I agree here, this was a copy paste without critical thinking. There's actually no way we'd have erroneous filter flags, and for the hail qc_metrics_filters, which are set by hail's sample qc, we can trust that the output has the same flags every time, so I'm leaning towards just taking the flag validation out of the new seqr sample qc flow.

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -242,6 +249,10 @@ def _load_new_samples(cls, metadata_path, genome_version, dataset_type, run_vers
         except Exception as e:
             logger.error(f'Error reporting loading failure for {run_version}: {e}')
 
+        # Update sample qc
+        if 'sample_qc' in metadata:
+            cls.update_individuals_sample_qc(sample_type, updated_samples, metadata['sample_qc'])
```
hanars, line 254: we should wrap this in a try/except block that logs an error because if something goes wrong somehow we would still want the rest of the job to proceed

seqr/management/commands/update_individuals_sample_qc.py
```diff
@@ -0,0 +1,46 @@
+import json
```
jklugherz, line 1: this and 'update_individuals_sample_qc_tests' have been approced

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -352,6 +367,36 @@ def _update_pdos(session, project_guid, sample_ids):
...
 
+    @classmethod
+    def _update_individuals_sample_qc(cls, sample_type, individuals, sample_qc_map):
+        sample_individual_map = {i.individual_id: i for i in individuals}
+        updated_individuals = []
+        for individual_id, record in sample_qc_map.items():
+            individual = sample_individual_map[individual_id]
+            filter_flags = {}
+            for flag in record['filter_flags']:
+                flag = '{}_{}'.format(flag, 'exome' if sample_type.lower() == 'wes' else 'genome') if flag == 'coverage' else flag
```
jklugherz, line 378: small change on this line since 'DATA_TYPE_MAP' constant was removed

hanars (approved):
looks great! Pending fixes to unit tests :)

PR #4664, opened 2025-02-20
Require filtered annotations in filtered genes

hanars:


hail_search/queries/base.py
```diff
@@ -958,7 +964,7 @@ def _get_allowed_consequence_ids(self, annotations):
 
     def _get_allowed_transcripts(self, ht, allowed_consequence_ids):
         transcript_filter = self._get_allowed_transcripts_filter(allowed_consequence_ids)
-        return ht[self.TRANSCRIPTS_FIELD].filter(transcript_filter)
+        return getattr(ht, FILTERED_GENE_TRANSCRIPTS, ht[self.TRANSCRIPTS_FIELD]).filter(transcript_filter)
```
hanars, line 967: this is the main change for the desired functionality in this PR

hail_search/queries/base.py
```diff
@@ -1001,7 +1007,7 @@ def _filter_compound_hets(self):
         def key(v):
             ks = [v[k] for k in self.KEY_FIELD]
             return ks[0] if len(self.KEY_FIELD) == 1 else hl.tuple(ks)
-        ch_ht = ch_ht.annotate(key_=key(ch_ht.row), gene_ids=self._gene_ids_expr(ch_ht))
+        ch_ht = ch_ht.annotate(key_=key(ch_ht.row), gene_ids=self._gene_ids_expr(ch_ht, filtered_genes_only=True))
```
hanars, line 1010: this change means that if a gene filter has been applied we will only compute possible comp het pairs in those genes

hail_search/queries/gcnv.py
```diff
@@ -84,11 +85,44 @@ def get_allowed_sv_type_ids(self, sv_types):
             type.replace(self.SV_TYPE_PREFIX, '') for type in sv_types if type.startswith(self.SV_TYPE_PREFIX)
         ])
 
+    def _annotate_families_table_annotations(self, families_ht, annotations_ht, is_comp_het=False):
```
hanars, line 88: For gCNV data, we have an entry-level gene_ids annotation, as the breakpoints for the SV in the annotation table can be different than for an individual call. The changes in this file ensure that gene and annotation filters only consider the genes present in the entries, and not all the gene. This function is added because we *also* recently added the logic to do an annotation-first search if genes are preent, which means at the time we do the gene and annotation filtering we don;t have the entries available, so we need to add a post-processing step to re-filter based on the filtered genes in that case

hail_search/queries/mito.py
```diff
@@ -329,31 +329,16 @@ def _get_sample_genotype(self, samples, r=None, include_genotype_overrides=False
...
-        else:
-            matched_transcript = main_transcript
+            return hl.coalesce(*[
+                ht[field].find(lambda t: comp_het_gene_ids.contains(t.gene_id)) for field in [
+                    ALLOWED_TRANSCRIPTS, ALLOWED_SECONDARY_TRANSCRIPTS, 'sorted_transcript_consequences',
+                ] if hasattr(ht, field)
+            ])
 
-        return hl.or_else(matched_transcript, main_transcript)
+        main_transcript = getattr(ht, FILTERED_GENE_TRANSCRIPTS, ht.sorted_transcript_consequences).first()
```
jklugherz, line 338: I like how the addition of filtering transcripts by genes earlier on makes selecting the main transcript a little simpler

jklugherz (approved):

PR #4669, opened 2025-02-21
Combine search pages

hanars:
Removed the separate "custom search"page and moves the all project search functionality from that page to all search pages


ui/pages/Report/components/CustomSearch.jsx
jklugherz, line 1: <3

jklugherz (approved):

PR #4670, opened 2025-02-22
DB migration for PanelApp AU domain name change

tommyli:
Reference discussion: https://github.com/broadinstitute/seqr/discussions/4623

This script bulk replaces domain name in table field `panelapp_palocuslist.url` from `https://panelapp.agha.umccr.org` to `https://panelapp-aus.org`.  This allows subsequent runs of [import_all_panels.py](https://github.com/broadinstitute/seqr/blob/master/panelapp/management/commands/import_all_panels.py) script to not create new panels.

To run this script: `python manage.py migrate panelapp`

The Django admin script [import_all_panels.py](https://github.com/broadinstitute/seqr/blob/master/panelapp/management/commands/import_all_panels.py) should be run with new url from Tue, 25th Feb onwards.

`python manage.py import_all_panels --label AU https://panelapp-aus.org/api/v1`

### DIFF
--- a/panelapp/migrations/0003_panelapp_aus_domain_migration.py
+++ b/panelapp/migrations/0003_panelapp_aus_domain_migration.py
@@ -1,0 +1,20 @@
+from django.db import migrations
+
+
+def update_urls(apps, schema_editor):
+    PaLocusList = apps.get_model('panelapp', 'PaLocusList')
+
+    # Replace the old domain with the new domain in the url field
+    for entry in PaLocusList.objects.filter(url__startswith='https://panelapp.agha.umccr.org'):
+        entry.url = entry.url.replace('https://panelapp.agha.umccr.org', 'https://panelapp-aus.org')
+        entry.save()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('panelapp', '0002_auto_20210915_1049'),
+    ]
+
+    operations = [
+        migrations.RunPython(update_urls),
+    ]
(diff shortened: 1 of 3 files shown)


panelapp/migrations/0003_panelapp_aus_domain_migration.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def update_urls(apps, schema_editor):
+    PaLocusList = apps.get_model('panelapp', 'PaLocusList')
+
+    # Replace the old domain with the new domain in the url field
+    for entry in PaLocusList.objects.filter(url__startswith='https://panelapp.agha.umccr.org'):
+        entry.url = entry.url.replace('https://panelapp.agha.umccr.org', 'https://panelapp-aus.org')
+        entry.save()
```
hanars, line 10: how long does this migration take to run? We tend to use bulk updates for large data migrations like this as our application startup times out after 5 minutes if pending migrations can not complete in that time.
Here is an example migration that has simialr behavior that uses the bulk_update functionality as an example: https://github.com/broadinstitute/seqr/blob/master/seqr/migrations/0068_project_vlm_contact_email.py#L6

hanars:
Thanks so much for your hep getting this migrated!

panelapp/migrations/0003_panelapp_aus_domain_migration.py
tommyli: The DB migration script should only take a few seconds at most.  The table is tiny in our instance, only 310 rows.  For your instance, I'm guessing around 1000 rows which should still take a few seconds only.

PR #4679, opened 2025-02-26
Exclude search section

hanars:
Adds a new "exclude" section to search. Refactors the behavior for excluding locations to use this new section, and adds functionality for excluding variants based on clinvar status

Screenshot 2025-02-26 at 3 42 23 PM


hail_search/queries/mito.py
```diff
@@ -448,6 +448,13 @@ def _get_allowed_transcripts_filter(allowed_consequence_ids):
              if canonical_consequences else allowed_consequence_ids
         ).contains)
 
+    def _filter_excluded(self, ht, exclude):
+        for key in self.PATHOGENICITY_FILTERS.keys():
+            path_terms = (exclude or {}).get(key)
```
jklugherz, line 453: is there an advantage to looping through `self.PATHOGENICITY_FILTERS.keys()` first as opposed to starting with `for key, value in enumerate(exclude or {})`

seqr/utils/search/utils.py
```diff
@@ -259,10 +265,16 @@ def _query_variants(search_model, user, previous_search_results, sort=None, num_
     if variant_ids:
         num_results = len(variant_ids)
 
+    exclude.pop('rawItems', None)
+    if exclude.get('clinvar') and (search.get('pathogenicity') or {}).get('clinvar'):
+        duplicates = set(search['pathogenicity']['clinvar']).intersection(exclude['clinvar'])
+        if duplicates:
+            raise InvalidSearchException(f'ClinVar pathogenicity {", ".join(sorted(duplicates))} is both included and excluded')
```
jklugherz, line 272: why is clinvar exclude treated differently from locus in that we allow both pathogenicity selection and exclusion, and only error out on duplicates, while for locus, we only allow a locus selection or exclusion?

jklugherz: is it that location exclusion was already implemented in a binary way (due to `exclude_locations`)?

hail_search/queries/mito.py
hanars: Most of the exclude filters are not going to be pathogenicity, currently theres only 2 possible filters,  one of pathogenicity and one not, but all subsequent new filters will not be path filters so to me it makes more sense to check for the small number of possible path filters than go through the potentially long list of non-pathogenicity excluded filters
Ultimately though either is fine

seqr/utils/search/utils.py
hanars: because pathpgenicity works as an OR filter with annotations so its possible to get back variants that do not pass the included filter - for example its easy if you filter for pathogenic to also get back benign variants that passed a different annotation filter, and you might want to exclude those. For location filter theres no OR behavior avaiable, so if you include a specific list of genes to include every single variant is in those genes and therefore its redundant to specify a different list of genes to exclude

jklugherz (approved):

PR #4684, opened 2025-02-28
fix gcnsv transcript filter error

hanars:


hail_search/queries/gcnv.py
```diff
@@ -87,21 +87,26 @@ def get_allowed_sv_type_ids(self, sv_types):
 
     def _annotate_families_table_annotations(self, families_ht, annotations_ht, is_comp_het=False):
         ht = super()._annotate_families_table_annotations(families_ht, annotations_ht)
-        gene_ids = self._get_genotype_override_field(ht, 'gene_ids')
-        transcript_has_gene = lambda t: hl.is_missing(gene_ids) | gene_ids.contains(t.gene_id)
```
hanars, line 91: this expression incorporates `ht`, so below if we reassign `ht` with the transcript annotations then any subsequent use of this variable to further re-annotate a different transcript field means you. are writing a hail expression that is using different tables (the old ht and the updated ht) which throws an error. By annotating the gene_ids onto the table we can safely use them even after the table updates

jklugherz (approved):

PR #4692, opened 2025-03-05
SV gene search improvement

hanars:
Fixes a bug when an empty structural type list is passed in the search body (which occurs when a user checks and then unchecks an option). Also improves performance for multi project SV gene search

### DIFF
--- a/hail_search/queries/base.py
+++ b/hail_search/queries/base.py
@@ -979,7 +979,7 @@ class BaseHailTableQuery(object):
 
     def _get_annotation_override_fields(self, annotations, override_fields=None, **kwargs):
         override_fields = override_fields or self.ANNOTATION_OVERRIDE_FIELDS
-        return {k: annotations[k] for k in override_fields if k in annotations}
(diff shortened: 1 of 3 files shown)


hail_search/queries/base.py
```diff
@@ -979,7 +985,7 @@ def _get_allowed_transcripts_filter(allowed_consequence_ids):
 
     def _get_annotation_override_fields(self, annotations, override_fields=None, **kwargs):
         override_fields = override_fields or self.ANNOTATION_OVERRIDE_FIELDS
-        return {k: annotations[k] for k in override_fields if k in annotations}
+        return {k: annotations[k] for k in override_fields if annotations.get(k)}
```
jklugherz, line 988: is this the bugfix? (and the `skip_missing_field` the performance improvement?)

jklugherz (approved):

hail_search/queries/base.py
hanars: correct

PR #4718, opened 2025-03-24
Retry multi lookup

hanars:
This should fix the flappiness in the reload annotation jobs


bpblanken (approved):
Out of scope for now, but we should consider [the requests Retry adapter](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/requests.py) as this idea might show up again.

hanars:
I looked into it but it was a bit overly complex for our use case, and also seems like it by default works best with connection errors and timeouts and requires some extra tweaking to get it to work for actual error responses

PR #4725, opened 2025-03-27
SV multi project comp het performance

hanars:
We are having trouble running all project comp het searches with combined SVs and SNV_INDEL data. This change splits those queries under the hood into 2 searches instead of 1, running all of the WGS sample in one search and all of the WES samples in another. Since we will not have a project with both WES and WGS for SVs in the same project on the hail backend this will only impact multi project searches with SVs, and since those searches are currently timing out any inefficiency introduced by waiting on 2 searches is better than being unable to run the search at all.

The reason this is a logical way to speed up the searches is that the SV data is stored separately for WES and WGS so the queries are running sequentially, and then the joins with SNV_INDEL data for the comp hets are relatively sparse as the only roughly half of the project are even possible to have a comp het hit in. Only querying SNV_INDEL for those projects which are possible to pair with a given sample type speeds up both the SNV_INDEL query and the join and comp het computation


bpblanken (approved):
logic seems sound to me!

PR #4728, opened 2025-03-27
do not run sv search when annotations are empty

hanars:


seqr/utils/search/search_utils_tests.py
```diff
@@ -395,11 +399,12 @@ def _mock_get_variants(families, search, user, previous_search_results, genome_v
         )
 
         del self.search_model.search['exclude']['rawItems']
-        self.search_model.search.update({'pathogenicity': {'clinvar': ['pathogenic', 'likely_pathogenic']}})
+        self.search_model.search.update({'pathogenicity': {'clinvar': ['pathogenic', 'likely_pathogenic']},
+                                         'annotations': {'frameshift': [], 'structural': []}})
```
hanars, line 403: this is the test for the edge case being changed here - the presence of empty list in the annotations does not change the actual executed search `dataset_type` and samples

bpblanken (approved):

PR #4792, opened 2025-05-15
bugfix: handle gzipped dd lines

bpblanken:
Address most recent reported issue on: https://github.com/broadinstitute/seqr-helm/issues/104


seqr/utils/file_utils_tests.py
```diff
@@ -30,3 +32,30 @@ def test_mv_file_to_gs(self, mock_logger, mock_subproc):
         mock_subproc.Popen.assert_called_with('gsutil mv /temp_path gs://bucket/target_path', stdout=mock_subproc.PIPE, stderr=mock_subproc.STDOUT, shell=True)  # nosec
         mock_logger.info.assert_called_with('==> gsutil mv /temp_path gs://bucket/target_path', None)
         process.wait.assert_called_with()
+
+
+    def test_file_iter_byte_range(self):
```
bpblanken, line 37: A few notes:

- I made an attempt to test this in the broader integration test, but it `file_iter` is mocked and harder to get to.
- `dd` actually gives slightly different results on my laptop and GitHub actions.

hanars: Unless its the only way to get a test to work, the convention in this repo is to test for helper functions via the apis where they are actually used, not writing utility-specific unit tests. We have some old test files like this one still because I have not had the bandwidth to deprecate them, but the overhead of maintaining tests for helper functions thats totally divorced from their real world usage is a pain. 
The error this si solving was reported when trying to trigger data loading, so the better option here is to adjust the unit tests for the `validate_callset` endpoint in data manager api - currently it only tests unzipped files but if you change the part of the test thats testing a sharded vcf to make that one gzipped, then you would be able to test both zipped and unzipped

hanars (changes requested):

seqr/utils/file_utils_tests.py
bpblanken: I've got _a_ solution to this, but struggled to get it fully correct/precise.  I tried to preserve the actual `dd` call by passing in a temporary file (which isn't really a pattern in these tests) but that failed because `/local_datasets` isn't writeable on newer MacBooks.  Instead the `dd` subprocess is mocked, the `file_iter` mock is copied, and I just checked that it was called with the right arguments

PR #4801, opened 2025-05-21
Clickhouse location filter

hanars:
Adds location filter to clickhouse. 

One change from the hail search backend is the removal of an optimization we implemented that clusters intervals for large gene lists (so instead of 1000+ intervals the in the interval pre-filter it created one interval per chromosome and pre-filtered to that). The clustering logic is particularly helpful in the hail backend as we need to call `hl.eval` on every interval passed to the read table pre-filter, whic is very slow when called hundreds of times but is not a concern in the clickhouse backend. I tested  a search with a gene list with over 2000 genes and it ran in ~20 seconds, but a) I was using annotations_disk and not annotations_memory and b) I have not yet implemented the gnomad af pre-filter on the entries table so I feel confident that the overall performance for a large gene list search will be acceptable when all is said and done and there is no need to maintain the clustering logic here.


clickhouse_search/search_tests.py
```diff
@@ -403,17 +404,14 @@ def test_quality_filter(self):
         #     annotations=annotations, pathogenicity={'clinvar': ['pathogenic']},
         # )
 #
-#     def test_location_search(self):
-#         self._assert_expected_search(
-#             [MULTI_FAMILY_VARIANT, VARIANT4], omit_data_type='SV_WES', **LOCATION_SEARCH,
-#         )
-#
-#         # Test "large" gene list search
```
hanars, line 411: removing this test as the clustering logic for large gene lists has been removed

seqr/utils/search/hail_search_utils.py
```diff
@@ -203,44 +203,6 @@ def _get_sort_metadata(sort, samples):
     return sort_metadata
 
 
-def _parse_location_search(search):
```
hanars, line 206: since parsing intervals from location search is now shared between hail search and clickhouse search, I moved the logic into the shared search utils file

PR #4802, opened 2025-05-21
Clickhouse in silico filter

hanars:


clickhouse_search/models.py
```diff
@@ -65,6 +65,111 @@ def __init__(self, name, select='*', order_by=None):
         self.order_by = order_by
 
 
+class BaseAnnotationsSnvIndel(models.ClickhouseModel):
```
hanars, line 68: I moved the Annotations tables up in the files so that the EntiresManager would have access to the class constants but I did not really change anything, the only change was to define `PREDICTION_FIELDS` as a class-level constant instead of directly defining it in the prediction column definition

bpblanken (approved):

PR #4819, opened 2025-06-03
Update ClickHouse default data directory.

bpblanken:
The new helm chart major version provided an option to set the data directory (which was previously not provided).  This hopefully lets us follow the pattern we set previously with the `/var/seqr/...` directories.

### DIFF
--- a/clickhouse_search/models.py
+++ b/clickhouse_search/models.py
@@ -612,7 +612,7 @@ class TranscriptsSnvIndel(models.ClickhouseModel):
 
     class Meta:
         db_table = 'GRCh38/SNV_INDEL/transcripts'
-        engine = EmbeddedRocksDB(primary_key='key', flatten_nested=0)
+        engine = EmbeddedRocksDB(0, f'{CLICKHOUSE_IN_MEMORY_DIR}/GRCh38/SNV_INDEL/transcripts', primary_key='key', flatten_nested=0)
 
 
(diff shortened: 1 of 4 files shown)


clickhouse_search/models.py
```diff
@@ -612,7 +612,7 @@ class TranscriptsSnvIndel(models.ClickhouseModel):
 
     class Meta:
         db_table = 'GRCh38/SNV_INDEL/transcripts'
-        engine = EmbeddedRocksDB(primary_key='key', flatten_nested=0)
+        engine = EmbeddedRocksDB(0, f'{CLICKHOUSE_DATA_DIR}/GRCh38/SNV_INDEL/transcripts', primary_key='key', flatten_nested=0)
```
bpblanken, line 615: The provide backup functionality for Clickhouse databases doesn't work for rocksdb tables, so we will want a hard-coded path that we can use to back up in another way.

hanars: changes to models require corresponding changes to migrations. Usually you would run `manage.py makemigrations` but for clickhouse we've just been editing the migratiion file manually, so you would edit this line: https://github.com/broadinstitute/seqr/blob/master/clickhouse_search/migrations/0001_initial.py#L135

hanars (changes requested):

hanars (approved):

PR #4838, opened 2025-06-11
chore: rename models for grch37 split

bpblanken:
I split out this piece because there are a few places where the naming convention doesn't necessarily make sense right now, and wanted to get feedback!


bpblanken:
ok!  that make sense.

PR #4839, opened 2025-06-11
feat: make grch37 tables.

bpblanken:


clickhouse_search/models.py
```diff
@@ -651,25 +666,105 @@ def _save_table(
...
+class EntriesGRCh38SnvIndel(BaseEntriesSnvIndel):
+    objects = EntriesManager()
+
+    # primary_key is not enforced by clickhouse, but setting it here prevents django adding an id column
+    key = ForeignKey('AnnotationsGRCh38SnvIndel', db_column='key', primary_key=True, on_delete=CASCADE)
+
+    class Meta:
+        db_table = 'GRCh38/SNV_INDEL/entries'
+
+class BaseTranscripts(models.ClickhouseModel):
```
hanars, line 770: I feel like splitting this out for different data types might be a premature optimization - given how small the SV data is and also that fact that SV "transcripts" have so few fields (instead of the tons of strings) I do not think we will want a transcript table for SVs at all, and I think we would actually have to look at the data size for MITO to see if we want it for that

clickhouse_search/models.py
```diff
@@ -651,25 +666,105 @@ def _save_table(
...
+
+    class Meta:
+        abstract = True
+
+
+class TranscriptsGRCh37SnvIndel(BaseTranscriptsSnvIndel):
+    key = OneToOneField('AnnotationsGRCh37SnvIndel', db_column='key', primary_key=True, on_delete=CASCADE)
+    transcripts = NestedField(
+        sorted([
+        ('isLofNagnag', models.BoolField(null=True, blank=True)),
```
hanars: Can we format loftee the same for 37 and 38 by changing it around in the parquet export? Or is that really hard?

clickhouse_search/models.py
bpblanken: ah.  I think you're right that it's in over-optimization, but I thought that we'd decided to just keep the table for the other dataset types as it's simpler to maintain consistency.

The fields are completely redundant though, so perhaps not.

clickhouse_search/models.py
bpblanken: I can look!

clickhouse_search/models.py
bpblanken: Just thought through this.  I think it's easier than I expected to not produce the transcripts parquets for SV/GCNV/MITO, so we should go with only having SNV_INDEL transcripts tables.  The MITO annotations table is currently < 1MB so we are fine to avoid the optimization.  I will go do that.

clickhouse_search/models.py
bpblanken: Yeah, it's doable!  It'll be a bit weird but is ok for now.

clickhouse_search/migrations/0001_initial.py
```diff
@@ -23,17 +23,17 @@ class Migration(migrations.Migration):
...
                 ('pos', clickhouse_backend.models.UInt32Field()),
                 ('ref', clickhouse_backend.models.StringField()),
                 ('alt', clickhouse_backend.models.StringField()),
                 ('variant_id', clickhouse_backend.models.StringField(db_column='variantId')),
                 ('rsid', clickhouse_backend.models.StringField(blank=True, null=True)),
                 ('caid', clickhouse_backend.models.StringField(blank=True, db_column='CAID', null=True)),
-                ('lifted_over_chrom', clickhouse_backend.models.StringField(blank=True, db_column='liftedOverChrom', low_cardinality=True, null=True)),
+                ('lifted_over_chrom', clickhouse_backend.models.Enum8Field(return_int=False, db_column='liftedOverChrom', blank=True, null=True, choices=[(1, '1'), (2, '2'), (3, '3'), (4, '4'), (5, '5'), (6, '6'), (7, '7'), (8, '8'), (9, '9'), (10, '10'), (11, '11'), (12, '12'), (13, '13'), (14, '14'), (15, '15'), (16, '16'), (17, '17'), (18, '18'), (19, '19'), (20, '20'), (21, '21'), (22, '22'), (23, 'X'), (24, 'Y'), (25, 'M')])),
```
bpblanken: I added `M` here as it was missing and also made lifted over chrom an enum.

clickhouse_search/migrations/0001_initial.py
```diff
@@ -23,17 +23,17 @@ class Migration(migrations.Migration):
...
                 ('lifted_over_pos', clickhouse_backend.models.UInt32Field(blank=True, db_column='liftedOverPos', null=True)),
                 ('hgmd', clickhouse_search.backend.fields.NamedTupleField(base_fields=[('accession', clickhouse_backend.models.StringField(blank=True, null=True)), ('classification', clickhouse_backend.models.Enum8Field(blank=True, return_int=False, choices=[(0, 'DM'), (1, 'DM?'), (2, 'DP'), (3, 'DFP'), (4, 'FP'), (5, 'R')], null=True))], null_if_empty=True)),
-                ('screen_region_type', clickhouse_search.backend.fields.Enum8Field(blank=True, return_int=False, choices=[(0, 'CTCF-bound'), (1, 'CTCF-only'), (2, 'DNase-H3K4me3'), (3, 'PLS'), (4, 'dELS'), (5, 'pELS'), (6, 'DNase-only'), (7, 'low-DNase')], db_column='screenRegionType', null=True)),
+                ('screen_region_type', clickhouse_backend.models.Enum8Field(blank=True, return_int=False, choices=[(0, 'CTCF-bound'), (1, 'CTCF-only'), (2, 'DNase-H3K4me3'), (3, 'PLS'), (4, 'dELS'), (5, 'pELS'), (6, 'DNase-only'), (7, 'low-DNase')], db_column='screenRegionType', null=True)),
```
bpblanken: this changed because I added a `models` prefix [here](https://github.com/broadinstitute/seqr/pull/4839/files#diff-c587af7ae1ffb9696d3c7db3719aaa177eccf4f8e75f4731698b3bf8ae7b1e86L448)

clickhouse_search/migrations/0001_initial.py
```diff
@@ -23,17 +23,17 @@ class Migration(migrations.Migration):
             fields=[
                 ('key', clickhouse_backend.models.UInt32Field(primary_key=True, serialize=False)),
                 ('xpos', clickhouse_backend.models.UInt64Field()),
-                ('chrom', clickhouse_search.backend.fields.Enum8Field(return_int=False, choices=[(1, '1'), (2, '2'), (3, '3'), (4, '4'), (5, '5'), (6, '6'), (7, '7'), (8, '8'), (9, '9'), (10, '10'), (11, '11'), (12, '12'), (13, '13'), (14, '14'), (15, '15'), (16, '16'), (17, '17'), (18, '18'), (19, '19'), (20, '20'), (21, '21'), (22, '22'), (23, 'X'), (24, 'Y')])),
+                ('chrom', clickhouse_backend.models.Enum8Field(return_int=False, choices=[(1, '1'), (2, '2'), (3, '3'), (4, '4'), (5, '5'), (6, '6'), (7, '7'), (8, '8'), (9, '9'), (10, '10'), (11, '11'), (12, '12'), (13, '13'), (14, '14'), (15, '15'), (16, '16'), (17, '17'), (18, '18'), (19, '19'), (20, '20'), (21, '21'), (22, '22'), (23, 'X'), (24, 'Y'), (25, 'M')])),
```
bpblanken: Ah, I did this wrong and didn't see the overridden Enum8Field.  Fixing!

clickhouse_search/migrations/0002_annotationsdiskgrch37snvindel_and_more.py
```diff
@@ -0,0 +1,136 @@
...
+            ],
+            options={
+                'db_table': 'GRCh37/SNV_INDEL/annotations_disk',
+                'engine': clickhouse_search.backend.engines.EmbeddedRocksDB(0, '/var/seqr/clickhouse-data/GRCh37/SNV_INDEL/annotations', flatten_nested=0, primary_key='key'),
+            },
```
hanars, line 42: you will want to modify this file to use env vars for the table directories isntead of the computed strings

clickhouse_search/models.py
```diff
@@ -427,35 +438,57 @@ class BaseAnnotationsSnvIndel(models.ClickhouseModel):
...
-        ('canonical', models.UInt8Field(null=True, blank=True)),
-        ('consequenceTerms', models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=CONSEQUENCE_TERMS))),
         ('extendedIntronicSpliceRegionVariant', models.BoolField(null=True, blank=True)),
         ('fiveutrConsequence', models.Enum8Field(null=True, blank=True, return_int=False, choices=[(1, '5_prime_UTR_premature_start_codon_gain_variant'), (2, '5_prime_UTR_premature_start_codon_loss_variant'), (3, '5_prime_UTR_stop_codon_gain_variant'), (4, '5_prime_UTR_stop_codon_loss_variant'), (5, '5_prime_UTR_uORF_frameshift_variant')])),
-        ('geneId', models.StringField(null=True, blank=True)),
-    ], db_column='sortedTranscriptConsequences')
+        *BaseAnnotationsGRCh37SnvIndel.SORTED_TRANSCRIPT_CONSQUENCES_FIELDS,
+    ])
+
+    objects = AnnotationsQuerySet.as_manager()
```
hanars: this belongs in the base class too

clickhouse_search/models.py
```diff
@@ -707,42 +722,104 @@ def _save_table(
...
+
+class BaseEntriesSnvIndel(BaseEntries):
+    CALL_FIELDS = [
+        ('sampleId', models.StringField()),
+        ('gt', models.Enum8Field(null=True, blank=True, choices=[(0, 'REF'), (1, 'HET'), (2, 'HOM')])),
+        ('gq', models.UInt8Field(null=True, blank=True)),
+        ('ab', models.DecimalField(max_digits=9, decimal_places=5, null=True, blank=True)),
+        ('dp', models.UInt16Field(null=True, blank=True)),
+    ]
+    is_gnomad_gt_5_percent = models.BoolField()
```
hanars, line 740: nit, but we usually put a blank line between class constants and model fields

clickhouse_search/models.py
```diff
@@ -707,42 +722,104 @@ def _save_table(
...
+            order_by=('project_guid', 'family_guid', 'is_gnomad_gt_5_percent', 'key'),
+            partition_by='project_guid',
+            deduplicate_merge_projection_mode='rebuild',
+            index_granularity=8192,
+        )
+        projection = Projection('xpos_projection', order_by='xpos, is_gnomad_gt_5_percent')
+
+class EntriesGRCh37SnvIndel(BaseEntriesSnvIndel):
+
+    objects = EntriesManager()
```
hanars: you can put this on the base class instead of putting it explicitly on both inherited classes

PR #4840, opened 2025-06-12
feat: make MITO tables.

bpblanken:


clickhouse_search/models.py
```diff
@@ -383,10 +383,40 @@ class BaseAnnotations(models.ClickhouseModel):
     lifted_over_chrom = Enum8Field(db_column='liftedOverChrom', return_int=False, null=True, blank=True, choices=[(i+1, chrom) for i, chrom in enumerate(CHROMOSOMES)])
     lifted_over_pos = models.UInt32Field(db_column='liftedOverPos', null=True, blank=True)
 
+    objects = AnnotationsQuerySet.as_manager()
```
bpblanken, line 384: I lifted this so that it's inherited by all subclasses.  Not sure if that's correct.

clickhouse_search/models.py
```diff
@@ -514,6 +537,67 @@ class Meta:
...
+        ('helix_heteroplasmy', NamedTupleField([
+            ('ac', models.UInt32Field()),
+            ('af', models.DecimalField(max_digits=9, decimal_places=5)),
+            ('an', models.UInt32Field()),
+            ('max_hl', models.DecimalField(max_digits=9, decimal_places=5)),
+        ])),
+    ]
+    PREDICTION_FIELDS = [
+        ('apogee', models.DecimalField(max_digits=9, decimal_places=5, null=True, blank=True)),
+        ('haplogroup_defining', models.Enum8Field(null=True, blank=True, return_int=False, choices=[(0, 'Y')])),
```
bpblanken: this field is an enum with a single value.

bpblanken:
Updated to include redundant field removal + `Y` enum change.

clickhouse_search/models.py
hanars: thats correct - i will adjust the manager to work with the different models as needed/ I can always override it on a specific model if I really need to

clickhouse_search/models.py
```diff
@@ -724,7 +813,14 @@ def _save_table(
...
         abstract = True
-
+        engine = CollapsingMergeTree(
+            'sign',
+            order_by=('project_guid', 'family_guid', 'key'),
+            partition_by='project_guid',
+            deduplicate_merge_projection_mode='rebuild',
+            index_granularity=8192,
+        )
+        projection = Projection('xpos_projection', order_by='xpos')
```
hanars, line 823: Do we want an xpos projection for mito? I'm not sure it will really improve performance all that much given how small the table is and how close all the xpos values are

hanars (approved):

clickhouse_search/models.py
bpblanken: not sure... I think it won't have any performance benefit, but it also shouldn't be a maintenance or cost burden.

PR #4841, opened 2025-06-12
bugfix: consequence terms string

bpblanken:


clickhouse_search/models.py
```diff
@@ -370,7 +373,7 @@ class BaseAnnotationsSnvIndel(models.ClickhouseModel):
...
     ]
     HGMD_CLASSES = [(0, 'DM'), (1, 'DM?'), (2, 'DP'), (3, 'DFP'), (4, 'FP'), (5, 'R')]
-    CONSEQUENCE_TERMS = [(1, 'transcript_ablation'), (2, 'splice_acceptor_variant'), (3, 'splice_donor_variant'), (4, 'stop_gained'), (5, 'frameshift_variant'), (6, 'stop_lost'), (7, 'start_lost'), (8, 'inframe_insertion'), (9, 'inframe_deletion'), (10, 'missense_variant'), (11, 'protein_altering_variant'), (12, 'splice_donor_5th_base_variant'), (13, 'splice_region_variant'), (14, 'splice_donor_region_variant'), (15, 'splice_polypyrimidine_tract_variant'), (16, 'incomplete_terminal_codon_variant'), (17, 'start_retained_variant'), (18, 'stop_retained_variant'), (19, 'synonymous_variant'), (20, 'coding_sequence_variant'), (21, 'mature_miRNA_variant'), (22, '5_prime_UTR_variant'), (23, '3_prime_UTR_variant'), (24, 'non_coding_transcript_exon_variant'), (25, 'intron_variant'), (26, 'NMD_transcript_variant'), (27, 'non_coding_transcript_variant'), (28, 'coding_transcript_variant'), (29, 'upstream_gene_variant'), (30, 'downstream_gene_variant'), (31, 'intergenic_variant'), (32, 'sequence_variant')]
+    CONSEQUENCE_TERMS = CONSEQUENCE_TERMS
```
bpblanken: I wasn't super sure what the best approach was here.  There's no clear inheritance pattern between the Annotations and Transcripts, so just lifting to a global made sense?  And I left `CONSEQUENCE_TERMS` as a property so it could be used externally.  

But there's several other potentially cleaner options.

hanars: Are the `CONSEQUENCE_TERMS` different between the annotations model and transcripts model? I feel ike they should not be, and then if not I think its fine to just do `choices=BaseAnnotationsSnvIndel.CONSEQUENCE_TERMS` in the transcript tabel column definition

bpblanken: no they should not be.  that's one of the other solutions I considered so I will go with that!

clickhouse_search/models.py
```diff
@@ -678,7 +680,7 @@ class TranscriptsSnvIndel(models.ClickhouseModel):
             ('isLofNagnag', models.BoolField(null=True, blank=True)),
             ('lofFilters', models.ArrayField(models.StringField(null=True, blank=True))),
         ], null_empty_arrays=True)),
-        ('majorConsequence', models.StringField(null=True, blank=True)),
+        ('majorConsequence', models.Enum8Field(null=True, blank=True, return_int=False, choices=BaseAnnotationsSnvIndel.CONSEQUENCE_TERMS)),
```
bpblanken, line 683: note I also fixed this!

hanars (approved):

PR #4844, opened 2025-06-13
feat: gcnv models

bpblanken:


clickhouse_search/models.py
```diff
@@ -1054,6 +1083,33 @@ class EntriesSv(BaseEntries):
...
+        ('start', models.UInt32Field(null=True, blank=True)),
+        ('end', models.UInt32Field(null=True, blank=True)),
+        ('numExon', models.UInt8Field(null=True, blank=True)),
+        ('geneIds',  models.ArrayField(models.StringField(null=True, blank=True))),
+        ('newCall', models.BoolField(null=True, blank=True)),
+        ('prevCall', models.BoolField(null=True, blank=True)),
+        ('prevOverlap', models.BoolField(null=True, blank=True)),
+    ]
+
+    objects = EntriesManager()
```
hanars: this is in the base class and is not needed here

PR #4845, opened 2025-06-13
chore: skip migration runs in check new samples job.

bpblanken:


seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -366,26 +374,32 @@ def set_up(self):
...
             ('GRCh38/SNV_INDEL', 'auto__2023-08-09'), ('GRCh37/SNV_INDEL', 'manual__2023-11-02'),
             ('GRCh38/MITO', 'auto__2024-08-12'), ('GRCh38/SV', 'auto__2024-09-14'),
+            ('GRCh38/SNV_INDEL', 'hail_search_to_clickhouse_PROJECT_ENTRIES_WGS_R0877_neptune'),
         ]
         if single_call:
             runs = runs[:1]
         for data_type, version in runs:
-            logs.append((f'Loading new samples from {data_type}: {version}', None))
             logs += self._additional_loading_logs(data_type, version)
+            if 'hail_search_to_clickhouse' in version:
```
bpblanken: one alternative to this, uh, hack, would be to refactor `run_loading_logs` to be keyed by run rather than dataset type.  that's probably cleaner?

bpblanken: I did this.  It's maybe _slightly_ better!

hanars:
As a potentially different approach, could migration runs write a different success file/ an additional file to the run directory instead of adding arbitrary keys to the metadata? Then we could ski migrations without having to read in the whole file using the logic we already have for filtering for successful runs - we could change 
```
success_run_dirs = [run_dir for run_dir, run_details in runs.items() if SUCCESS_FILE_NAME in run_details['files']]
```
to something like
```
success_run_dirs = [run_dir for run_dir, run_details in runs.items() 
    if SUCCESS_FILE_NAME in run_details['files'] and MIGRATION_FILE_NAME not inrun_details['files']]
```

hanars (approved):

PR #4856, opened 2025-06-23
feat: gt stats tables

bpblanken:
Helps address: https://github.com/broadinstitute/seqr-private/issues/1663


hanars (approved):
To clarify, this is the PR to create the `SummingMergeTree` table, but then we will need a separate PR to create the dictionary that seqr uses, correct?

bpblanken:
@hanars.  yes, the dictionary and also the materialized views that map one table to another.

PR #4857, opened 2025-06-23
Move managers to own file

hanars:


clickhouse_search/managers.py
```diff
@@ -0,0 +1,696 @@
+from clickhouse_backend import models
```
hanars, line 1: except where noted otherwise this is all copied from models.py

clickhouse_search/managers.py
```diff
@@ -0,0 +1,696 @@
...
+        return OrderedDict({
+            'family_guid': ('familyGuid', models.StringField()),
+            'sample_type': ('sampleType', models.StringField()),
+            'filters': ('filters', models.ArrayField(models.StringField())),
+            'x.gt::Nullable(Int8)': ('numAlt', models.Int8Field(null=True, blank=True)),
+            **{f'x.{column[0]}': column for column in self.model.CALL_FIELDS if column[0] != 'gt'}
+        })
+
+    @property
+    def clinvar_fields(self):
```
hanars, line 355: this property was added, previously there was a constant defined in the same way that directly referenced the `ClinvarSnvIndel` class instead of the `clinvar_model` used here

hanars:
will hold off on merging until base pr is merged

PR #4859, opened 2025-06-24
add materialized views

bpblanken:


clickhouse_search/migrations/0008_gtstatsgrch37snvindel_gtstatsmito_gtstatssnvindel_and_more.py
```diff
@@ -74,7 +74,7 @@ class Migration(migrations.Migration):
             name='GtStatsSv',
             fields=[
                 ('key', models.OneToOneField(db_column='key', on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='clickhouse_search.annotationssv')),
-                ('het_wgs', clickhouse_backend.models.UInt32Field()),
+                ('ac_wgs', clickhouse_backend.models.UInt32Field()),
```
bpblanken, line 77: this was incorrect.

bpblanken: well, maybe not incorrect but inconsistent with `SNV_INDEL` in a way that was unnecessary.

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -0,0 +1,219 @@
...
+                ])
+            ),
+            hints={'clickhouse': True},
+        ),
+        migrations.RunSQL(
+            GT_STATS_DICT.substitute(
+                reference_genome='GRCh38',
+                dataset_type='MITO',
+                columns= ",\n    ".join([
+                    'het_wes UInt32',
```
hanars: Is there a reason you updates SVs to be consistent and use AC instead of HET but not MITO?

hanars (approved):

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
bpblanken: I think just a bit of uncertainty about what the right name would be.  `ac_het_wes`, `ac_het_wgs`, `ac_hom_wgs`, ... make sense?

hanars: yeah I like adding the ac prefix because I think it disambiguates between heteroplasmy/homoplasmy and heterozygous/homoyygous

bpblanken: Updated to include mito ac_ prefixes.

hanars (approved):

PR #4864, opened 2025-06-24
Clickhouse 37

hanars:


clickhouse_search/test_utils.py
```diff
@@ -6,14 +6,29 @@
...
+GRCH37_VARIANT = {
+    **deepcopy(HAIL_GRCH37_VARIANT),
+    'key': 11,
+    'liftedOverGenomeVersion': '38',
+    'liftedOverChrom': '7',
+    'liftedOverPos': 143271368,
+}
+for genotype in GRCH37_VARIANT['genotypes'].values():
+    genotype['sampleType'] = 'WES'
+GRCH37_VARIANT['predictions'].update({'fathmm': None, 'mut_pred': None, 'vest': None})
```
hanars, line 26: @bpblanken or hail backend fixture data for 37 does not have fathmm, mut_pred or vest (they are 38 only). Just confirming that we do in fact have data for these in 37 in clickhouse

bpblanken: Yes, we should.

PR #4868, opened 2025-06-26
chore: update view to exclude projects

bpblanken:


clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -28,6 +41,7 @@
     key,
     $columns
 FROM `$reference_genome/$dataset_type/project_gt_stats`
+WHERE project_guid NOT IN %s
```
bpblanken: I tried to use the standard sql string formatting for this (first with a named param) because I thought it would be easier.  It wasn't particularly easy, and now we have two different methods of string formatting.  

Template substituting is probably better for consistency?

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -9,6 +9,19 @@
 CLICKHOUSE_USER = os.environ.get('CLICKHOUSE_USER', 'clickhouse')
 CLICKHOUSE_PASSWORD = os.environ.get('CLICKHOUSE_PASSWORD', 'clickhouse_test')
 
+EXCLUDED_PROJECT_GUIDS = (
```
hanars: So we should not be hardcoding the into a migration, these need to come from an env variable

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
hanars: Sure, but maybe we should do this the way we do with the credentials for the dictionaries so we only need to template them in once?

bpblanken: yes, that also makes sense!

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -1,11 +1,16 @@
 # Generated manually by the seqr team.
+import json
 import os
 from string import Template
 
 from django.db import migrations
 
 from settings import DATABASES
 
+CLICKHOUSE_AC_EXCLUDED_PROJECT_GUIDS  = json.loads(os.environ.get(
```
bpblanken: We can put this into our private helm chart:

```
export CLICKHOUSE_AC_EXCLUDED_PROJECT_GUIDS='["R0555_seqr_demo", "R0607_gregor_training_project_"]'
```

hanars: I've found just using a comma delimited string as the env variable and then calling `split` on  it is easier to maintain and edit than a properly json encoded list

hanars: also, the list isn't really specific to clickhouse and AC feels a bit ambiguous - maybe call this soemthing like `GLOBAL_AC_EXCLUDED_PROJECT_GUIDS`

hanars (approved):

PR #4869, opened 2025-06-26
Clickhouse mito

hanars:


clickhouse_search/managers.py
```diff
@@ -192,7 +216,9 @@ def _filter_frequency(self, results, freqs=None, pathogenicity=None, **kwargs):
                         f'populations__{population}__{af_field}__lte': pop_filter['af'],
                     })
                     if clinvar_override_q and pop_filter['af'] < PATH_FREQ_OVERRIDE_CUTOFF:
-                        af_q |= clinvar_override_q
+                        af_q |= (clinvar_override_q & Q(**{
```
hanars, line 219: this change was unrelated to MITO, but a newly enabled test showed that this was actually not implemented properly (clinvar is only supposed to overrride frequency up to `PATH_FREQ_OVERRIDE_CUTOFF`, not just override frequency entriely)

clickhouse_search/managers.py
```diff
@@ -507,6 +549,12 @@ def _search_call_data(self, entries, sample_data, inheritance_mode=None, inherit
             if call_q:
                 entries = entries.filter(call_q)
 
+            if quality_filter.get('vcf_filter'):
```
hanars, line 555: not related to mito, but realized from a new test that clinvar was not properly overriding the pass filters filter

clickhouse_search/models.py
```diff
@@ -69,9 +70,14 @@ def __init__(self, name, select='*', order_by=None):
 
 class BaseAnnotations(models.ClickhouseModel):
 
-    GENOME_VERSION = GENOME_VERSION_GRCh38
-    LIFTED_OVER_GENOME_VERSION = GENOME_VERSION_GRCh37
     CHROMOSOME_CHOICES = [(i+1, chrom) for i, chrom in enumerate(CHROMOSOMES)]
+    SEQR_POPULATIONS = [
```
hanars, line 74: I know I had said that I wanted to make models/classes for the dictionaries so the available fields are encoded, but I ended up defining these constants which I do not think will be too bad to maintain separate from the dictionaries. Thoughts?

clickhouse_search/search_tests.py
```diff
@@ -158,9 +179,12 @@ def test_single_project_search(self):
             'ENSG00000097046': {'total': 3, 'families': {'F000002_2': 2, 'F000003_3': 1}},
             'ENSG00000177000': {'total': 3, 'families': {'F000002_2': 2, 'F000003_3': 1}},
             'ENSG00000277258': {'total': 1, 'families': {'F000002_2': 1}},
+            'ENSG00000210112': {'total': 1, 'families': {'F000002_2': 1}},
+            'ENSG00000198886': {'total': 1, 'families': {'F000002_2': 1}},
+            'ENSG00000198727': {'total': 1, 'families': {'F000002_2': 1}},
         }
         self._assert_expected_search(
```
hanars, line 186: In general I have been keeping the tests the same as with hail search, but one big difference is in those test we explicitly set what sample data we are using and in these tests we use whatever is in sample fixture data. Crucially, this means that in the hail tests mito is excluded by default but in these tests its included by default. Since I can;t see a reason why adding better coverage for mito searches is a bad thing, I mostly just added the mito variants to the expected result, instead of excluding the mito samples to leave the results the same

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -46,7 +46,7 @@
     # double substitution these shared values
     clickhouse_user=CLICKHOUSE_USER,
     clickhouse_password=CLICKHOUSE_PASSWORD,
-    clickhouse_database=DATABASES['clickhouse']['NAME'],
+    clickhouse_database=DATABASES.get('clickhouse', {}).get('NAME'),
```
hanars, line 49: migrations are loaded for all the databases and then only executed on the relevent one(s). This means for local installations that have not yet enables clickhouse, this is a breaking line even though in any case where this would actually be executed this would be defined

ui/shared/utils/constants.js
```diff
@@ -1537,7 +1537,7 @@ export const ORDERED_PREDICTOR_FIELDS = [
     thresholds: [undefined, undefined, 2.18, 4, undefined],
     requiresCitation: true,
   },
-  { field: 'haplogroup_defining', indicatorMap: { Y: { color: 'green', value: '' } } },
+  { field: 'haplogroup_defining', indicatorMap: { Y: { color: 'green', value: '' }, true: { color: 'green', value: '' } } },
```
hanars, line 1540: I though it was better to update the UI here than to continue to do a weird mapping for the boolean

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
bpblanken: yes, I ran into this! thank you for fixing 🙇 !

ui/shared/components/panel/variants/Frequencies.jsx
```diff
@@ -373,13 +384,15 @@ const getValueDisplay = (pop, valueField, precision) => (valueField === 'ac' ?
...
+  const { populations = {}, svType } = variant
   const callsetHetPop = populations.callset_heteroplasmy || populations.seqr_heteroplasmy
   const isMito = callsetHetPop && callsetHetPop.af !== null && callsetHetPop.af !== undefined
   const popConfigs = isMito ? MITO_POPULATIONS : POPULATIONS
+  let datasetType = isMito ? DATASET_TYPE_MITO_CALLS : DATASET_TYPE_SNV_INDEL_CALLS
+  datasetType = svType ? DATASET_TYPE_SV_CALLS : datasetType
   const seqrAcSection = {
     name: 'seqr Global ACs',
-    details: (!isMito && populations[SEQR_POP.field]) ? [<GlobalAcPopup />].filter(s => s) : [],
+    details: populations[SEQR_POP.field] ? [<GlobalAcPopup datasetType={datasetType} />].filter(s => s) : [],
```
hanars, line 395: when I did the initial global AF deprecation I only did it for SNV_INDEL - this updates the UI to work with any dataset type

seqr/views/utils/variant_utils.py
```diff
@@ -472,8 +472,9 @@ def _set_response_gene_scores(response, family_genes, gene_ids):
 
 def _add_sample_count_stats(response):
     sample_counts = Sample.objects.filter(
```
hanars, line 474: when I did the initial global AF deprecation I only did it for SNV_INDEL - this updates the UI to work with any dataset type

bpblanken (approved):

clickhouse_search/models.py
bpblanken: yeah, I think this is fine.  The fields themselves should be the only structure that's actually different.  

This is probably conjecture, but I'd hope that the dictionary syntactical weirdness and support will improve over the next few years such that putting in more work wouldn't be worth the effort.

PR #4870, opened 2025-06-27
chore: clarify loading documentation

bpblanken:


deploy/LOCAL_INSTALL_HELM.md
```diff
@@ -78,8 +78,8 @@ loading_pipeline_queue  test.vcf.gz
...
-- Select your VCF from the dropdown and select the appropriate Sample Type (WES/WGS) and Genome Version (GRCh37/GRCh38) for your project.  The pipeline includes a sequence of validation steps to insure the validity of your VCF, but these may be skipped by enabling the **Skip Callset Validation**option.  We strongly recommend leaving validation enabled to ensure the quality of your analysis.
-- Click through to the next page and select your project from the **Projects to Load** dropdown, then click **Submit**.
+- Select your VCF from the dropdown and select the appropriate Sample Type (WES/WGS) and Genome Version (GRCh37/GRCh38) for your project.  The pipeline includes a sequence of validation steps to insure the validity of your VCF, but these may be skipped by enabling the **Skip Callset Validation** option.  We strongly recommend leaving validation enabled to ensure the quality of your analysis.
+- Click through to the next page and select your project from the **Projects to Load** dropdown, then click **Submit**.  Families, Individuals, and their associated metadata must be created within the project via the user interface prior to submitting a loading request.
```
bpblanken: I verified on a local install that it's possible to request loading for a project that is created but with no individuals or families.  Having the pipeline fail is likely less clear than having the UI prevent that (and if the UI prevented it maybe this documentation is redundant).

hanars: I would actually make this its own step at the beginning of the list, I think thats clearer than in step 4 saying there was something else you have to do first
Also, not sure what "associated metadata" means here, I am not sure what else other than families and individuals need to be created for loading to work, but if there is something we should list it explicitly

deploy/LOCAL_INSTALL_HELM.md
```diff
@@ -76,10 +76,11 @@ pipeline-runner:
 ls -h /var/seqr/seqr-loading-temp/
 loading_pipeline_queue  test.vcf.gz
 ```
+- Ensure Families and Individuals are created in the project(s) of interest.
```
bpblanken, line 79: The associated metadata that's required here is Affected Status.  The UI is clear enough on that though.

PR #4876, opened 2025-06-30
Use computed gt stats fixtures in unit tests

hanars:
Instead of hardcoding gt stats dictionaries in tests, use the actual materialized views and dictionaries. While this does change the values in the test results it does not reduce functionality coverage, and it is more accurate and maintainable to use the dictionaries as they are created by the migrations

### DIFF
--- a/clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
+++ b/clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
@@ -45,7 +45,7 @@ CREATE DICTIONARY `$reference_genome/$dataset_type/gt_stats_dict`
     $columns
 )
 PRIMARY KEY key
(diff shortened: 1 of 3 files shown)


clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -45,15 +45,14 @@
     $columns
 )
 PRIMARY KEY key
-SOURCE(CLICKHOUSE(USER $clickhouse_user PASSWORD $clickhouse_password DB $clickhouse_database TABLE `$reference_genome/$dataset_type/gt_stats`))
+SOURCE(CLICKHOUSE(USER $clickhouse_user PASSWORD $clickhouse_password TABLE `$reference_genome/$dataset_type/gt_stats`))
```
hanars, line 48: explicitly setting the database here breaks the tests, as it tries to create the dictionary sourced from the seqr database (which does not exist) instead of the test_seqr database. Since we should only ever have one seqr database in a deployment, I do not think there is any risk to removing this parameter

PR #4878, opened 2025-07-01
Clickhouse sv

hanars:


seqr/fixtures/1kg_project.json
```diff
@@ -873,6 +873,21 @@
         "sex": "M",
         "affected": "A"
     }
+}, {
```
hanars, line 876: The individual and sample data for WGS SV search in the hail backend was not based on the actual fixture data we had. To make this more maintainable I added the fixture data we needed but it did have some small changes to unrelated tests as a result

PR #4879, opened 2025-07-03
set different clickhouse credentials for migrate and app use

hanars:
this is a very basic implementation of multiple users for clickhouse - let me know if you have any objections to the approach.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -60,6 +60,8 @@ jobs:
         env:
           CLICKHOUSE_USER: clickhouse_test
           CLICKHOUSE_PASSWORD: clickhouse_test
+          CLICKHOUSE_ADMIN_USER: clickhouse_admin_test
+          CLICKHOUSE_ADMIN_PASSWORD: clickhouse_admin_test
           ALLOW_EMPTY_PASSWORD: no
 
     steps:
@@ -91,7 +93,8 @@ jobs:
       - name: Run coverage tests
         run: |
           export CLICKHOUSE_USER=clickhouse_test
-          export CLICKHOUSE_ADMIN_USER=clickhouse_test
+          export CLICKHOUSE_ADMIN_USER=clickhouse_admin_test
+          export CLICKHOUSE_ADMIN_PASSWORD=clickhouse_admin_test
           export CLICKHOUSE_SERVICE_HOSTNAME=localhost
           coverage run --source="./matchmaker","./seqr","./reference_data","./panelapp","./clickhouse_search" --omit="*/migrations/*","*/apps.py","*/admin.py" manage.py test -p '*_tests.py' -v 2 reference_data seqr matchmaker panelapp clickhouse_search
           coverage report -m --fail-under=99
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -91,6 +91,7 @@ jobs:
       - name: Run coverage tests
         run: |
           export CLICKHOUSE_USER=clickhouse_test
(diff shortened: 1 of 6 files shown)


deploy/docker/seqr/entrypoint.sh
```diff
@@ -45,7 +45,11 @@ python -u manage.py migrate --database=reference_data
 python -u manage.py loaddata variant_searches
 python -u manage.py loaddata variant_tag_types
 if [ "$CLICKHOUSE_SERVICE_HOSTNAME" ]; then
+    export CLICKHOUSE_USER="$CLICKHOUSE_ADMIN_USER"
```
bpblanken: this looks ok to me!  the only potential point of concern is potential the `delete projects` operation, which very likely will need some form of write permissions.

hanars:
This updated approach sets up 2 connections to clickhouse from the seqr app with 2 different sets of credentials, which is done essentially by setting up 2 near-identical "DATABASES". I have configured the migrations such that they will only run on the "writable" database connection, but all other model interactions use the read only connection by default. If you try to execute any write operations using the django models (`queryset.update`, `queryset.delete` etc) these will execute the sql against clickhouse which will fail due to inadequate user permissions. If however you explicitly set the connection for your queryset to the writable connections (`queryset.using('clickhouse_write').update(...`) then it will succeed. I think this achieves our goals of primarily interacting in a read only state but will allow us to intentionally write to the database in the limited cases where this is desired (i.e. deleting families)

.github/workflows/unit-tests.yml
```diff
@@ -66,7 +66,9 @@ jobs:
       - uses: actions/checkout@v2
       - name: Set up Clickhouse Settings
         run: |
-          docker exec clickhouse clickhouse-client --query "CREATE SETTINGS PROFILE clickhouse_settings SETTINGS flatten_nested=0, join_use_nulls=1 TO clickhouse_test"
+          docker exec clickhouse clickhouse-client --query "CREATE USER clickhouse_read_only IDENTIFIED WITH plaintext_password BY 'clickhouse_test'"
+          docker exec clickhouse clickhouse-client --query "CREATE SETTINGS PROFILE clickhouse_settings SETTINGS flatten_nested=0, join_use_nulls=1 TO clickhouse_test, clickhouse_read_only"
+          docker exec clickhouse clickhouse-client --query "GRANT SELECT, SYSTEM VIEWS, dictGet ON *.* TO clickhouse_read_only"
```
hanars, line 71: this is the minimal set of permissions our read-only user requires

settings.py
```diff
@@ -246,11 +246,16 @@
...
+    DATABASES['clickhouse_write'] = {
         'ENGINE': 'clickhouse_search.backend',
         'NAME': 'seqr',
         'HOST': CLICKHOUSE_SERVICE_HOSTNAME,
         'PORT': int(os.environ.get('CLICKHOUSE_SERVICE_PORT', '9000')),
+        'USER': os.environ.get('CLICKHOUSE_ADMIN_USER', 'clickhouse'),
+        'PASSWORD': os.environ.get('CLICKHOUSE_ADMIN_PASSWORD', 'clickhouse_test'),
+    }
+    DATABASES['clickhouse'] = {
+        **DATABASES['clickhouse_write'],
```
bpblanken, line 258: Do we need to change the `hints={'clickhouse': True},`  we have in the migrations to use the `clickhouse_write` database?

hanars: nope, I did not change the router logic for hints: 
https://github.com/broadinstitute/seqr/blob/cdaa29a7e80376c1050df6c9c930a8aacbcc0325/clickhouse_search/models.py#L59

bpblanken (approved):

PR #4886, opened 2025-07-15
feat: clinvar all variants job

bpblanken:


clickhouse_search/migrations/0003_annotationsdiskmito_annotationsmito_entriesmito_and_more.py
```diff
@@ -97,11 +97,11 @@ class Migration(migrations.Migration):
...
                 ('submitters', clickhouse_backend.models.ArrayField(base_field=clickhouse_backend.models.StringField())),
                 ('conditions', clickhouse_backend.models.ArrayField(base_field=clickhouse_backend.models.StringField())),
-                ('assertions', clickhouse_backend.models.ArrayField(base_field=clickhouse_backend.models.Enum8Field(choices=[(0, 'Affects'), (1, 'association'), (2, 'association_not_found'), (3, 'confers_sensitivity'), (4, 'drug_response'), (5, 'low_penetrance'), (6, 'not_provided'), (7, 'other'), (8, 'protective'), (9, 'risk_factor'), (10, 'no_classification_for_the_single_variant'), (11, 'no_classifications_from_unflagged_records')]))),
+                ('assertions', clickhouse_backend.models.ArrayField(base_field=clickhouse_backend.models.Enum8Field(choices=[(0, 'Affects'), (1, 'association'), (2, 'association_not_found'), (3, 'confers_sensitivity'), (4, 'drug_response'), (5, 'low_penetrance'), (6, 'not_provided'), (7, 'other'), (8, 'protective'), (9, 'risk_factor'), (10, 'no_classifications_from_unflagged_records'), (11, 'no_classification_for_the_single_variant'), (12, 'conflicting_data_from_submitters')]))),
```
bpblanken: I encountered the value `conflicting_data_from_submitters` value in the XML which we have not seen before.  This will be parsed into a default pathogenicity + 'conflicting_data_from_submitters' as an assertion.

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
```diff
@@ -52,7 +50,7 @@
     # Note the nested Template-ing that allows
     # double substitution these shared values
     clickhouse_user=CLICKHOUSE_USER,
-    clickhouse_password=CLICKHOUSE_PASSWORD,
+    clickhouse_password=CLICKHOUSE_PASSWORD if CLICKHOUSE_PASSWORD else "''",
```
bpblanken: I've been running into this locally, as my local admin has no password.  Is it better to support empty password use cases or not?

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+            return [(t[0].replace(' ', '_'), t[1]) for t in value]
+        return [s.replace(' ', '_') for s in value]
+    raise TypeError("Expected str or list[str]") # pragma: no cover
+
+BATCH_SIZE = 1000
+CLINVAR_ASSERTIONS = replace_underscores_with_spaces(ClinvarAllVariantsSnvIndel.CLINVAR_ASSERTIONS)
+CLINVAR_CONFLICTING_CLASSICATIONS_OF_PATHOGENICITY = replace_underscores_with_spaces(ClinvarAllVariantsSnvIndel.CLINVAR_CONFLICTING_CLASSICATIONS_OF_PATHOGENICITY)
+CLINVAR_DEFAULT_PATHOGENICITY = replace_underscores_with_spaces(ClinvarAllVariantsSnvIndel.CLINVAR_DEFAULT_PATHOGENICITY)
+CLINVAR_PATHOGENICITIES = replace_underscores_with_spaces(ClinvarAllVariantsSnvIndel.CLINVAR_PATHOGENICITIES)
+CLINVAR_GOLD_STARS_LOOKUP = {
```
bpblanken, line 43: no changes from the pipeline here, just underscores to spaces.

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+
+        for batch, model in zip(
+            (GRCh37SnvIndel_batch, SnvIndel_batch, Mito_batch),
+            (ClinvarAllVariantsGRCh37SnvIndel, ClinvarAllVariantsSnvIndel, ClinvarAllVariantsMito)
+        ):
+            if batch:
+                model.objects.using('clickhouse_write').bulk_create(batch)
+
+        # Delete previous version & refresh the view.
+        if existing_version:
```
bpblanken: Opted to `DROP PARTITION` rather than `.delete()` on the model, which _does_ work.  This is faster and safer.

clickhouse_search/migrations/0003_annotationsdiskmito_annotationsmito_entriesmito_and_more.py
hanars: Is this adding any additional information from the pathogenic status of conflicting and the detailed conflicting pathogenicities list we already have?

clickhouse_search/migrations/0009_materializedviews_and_dictionaries.py
hanars: No, I think setting up a password locally is pretty trivial and we should require that

clickhouse_search/models.py
```diff
@@ -425,24 +425,86 @@ class Meta:
...
+        'confers_sensitivity',
+        'drug_response',
+        'low_penetrance',
+        'not_provided',
+        'other',
+        'protective',
+        'risk_factor',
+        'no_classifications_from_unflagged_records',
+        'no_classification_for_the_single_variant',
+        'conflicting_data_from_submitters',  # NOTE THIS IS NEW
```
hanars: thanks for the note, but we should remove before merging

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
+import gzip
+import logging
+import requests
+import xml
+import defusedxml.ElementTree as ET
```
hanars, line 5: does this not require changes to our requirements.txt?

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+from seqr.utils.communication_utils import safe_post_to_slack
+
+logger = logging.getLogger(__name__)
+
+def replace_underscores_with_spaces(value: Union[str, list[str]]) -> Union[str, list[str]]:
+    if isinstance(value, str):
+        return value.replace('_', ' ')
+    elif isinstance(value, list):
+        return [s.replace('_', ' ') for s in value]
+    raise TypeError("Expected str or list[str]") # pragma: no cover
```
hanars: Do we use pragma for coverage in this repo?

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+                model.objects.using('clickhouse_write').bulk_create(batch)
+
+        # Delete previous version & refresh the view.
+        if existing_version:
+            clinvar_run_sql(Template(f"ALTER TABLE `$reference_genome/$dataset_type/clinvar_all_variants` DROP PARTITION '{existing_version}';"))
+        clinvar_run_sql(Template('SYSTEM REFRESH VIEW `$reference_genome/$dataset_type/clinvar_all_variants_to_clinvar`;'))
+        clinvar_run_sql(Template('SYSTEM WAIT VIEW `$reference_genome/$dataset_type/clinvar_all_variants_to_clinvar`;'))
+
+        # Save the new version in Postgres
+        DataVersions('Clinvar', new_version).save()
```
hanars: this will create duplicate records in the table, there should only ever be one record per `data_model_name`. Since you are already querying above for the existing record, I would just do that in a way where you assign that to a real variable and update it here if it exists. Also, for creating new objects it is better to explicitly use `DataVersions.objects.create`

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+                            batch.append(obj)
+                        if len(batch) == BATCH_SIZE:
+                            model.objects.using('clickhouse_write').bulk_create(batch)
+                            batch.clear()
+                    elem.clear()
+
+        for batch, model in zip(
+            (GRCh37SnvIndel_batch, SnvIndel_batch, Mito_batch),
+            (ClinvarAllVariantsGRCh37SnvIndel, ClinvarAllVariantsSnvIndel, ClinvarAllVariantsMito)
+        ):
```
hanars: redefining the `zip` in multiple places is a bit hard to read/ understand that these are the same set of relationships - I feel like defining a top level constant for the configs and then using it here as `for _, batch, model in CONFIGS` would be much more readable

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+                # Handle parsing the current date.
+                if event == 'start' and elem.tag == 'ClinVarVariationRelease':
+                    new_version = elem.attrib['ReleaseDate']
+                    existing_version = (obj := DataVersions.objects.filter(data_model_name='Clinvar').first()) and obj.version
+                    if new_version == existing_version:
+                        logger.info(f'Clinvar ClickHouse tables already successfully updated to {new_version}, gracefully exiting.')
+                        return
+                    logger.info(f'Updating Clinvar ClickHouse tables to {new_version} from {existing_version}.')
+                    clinvar_run_sql(
+                        Template(f"ALTER TABLE `$reference_genome/$dataset_type/clinvar_all_variants` DROP PARTITION '{new_version}';")
```
hanars, line 264: why would a partition with the new version name exist if we have not already loaded the new version?

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+                    for obj, batch, model in zip(
+                        (GRCh37SnvIndel, SnvIndel, Mito),
+                        (GRCh37SnvIndel_batch, SnvIndel_batch, Mito_batch),
+                        (ClinvarAllVariantsGRCh37SnvIndel, ClinvarAllVariantsSnvIndel, ClinvarAllVariantsMito)
+                    ):
+                        if obj:
+                            batch.append(obj)
+                        if len(batch) == BATCH_SIZE:
+                            model.objects.using('clickhouse_write').bulk_create(batch)
+                            batch.clear()
```
hanars: `bulk_create` takes a `batch_size` argument that I think would handle this for you if you prefer: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+        ) if grch37 and grch37['chrom'] != 'MT' else None,
+        ClinvarAllVariantsSnvIndel(
+            variant_id=f"{grch38['chrom']}-{grch38['pos']}-{grch38['ref']}-{grch38['alt']}",
+            **props,
+        ) if grch38 and grch38['chrom'] != 'MT'else None,
+        ClinvarAllVariantsMito(
+            variant_id=f"M-{grch38['pos']}-{grch38['ref']}-{grch38['alt']}",
+            **props,
+        ) if grch38 and grch38['chrom'] == 'MT' else None,
+    )
```
hanars: its kind of confusing to return 3 models only one of which is ever defined - I think it would be clearer to just retun one value from this function and then use `isinstance` checking to assign it to the appropriate list below - 
```
if isinstance(obj, model):
  batch.append(obj)
```

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+        for reference_genome, dataset_type in [
+            ('GRCh37', 'SNV_INDEL'),
+            ('GRCh38', 'SNV_INDEL'),
+            ('GRCh38', 'MITO'),
+        ]:
+            cursor.execute(sql.substitute(reference_genome=reference_genome, dataset_type=dataset_type))
+
+
+def parse_and_merge_classification_counts(text):
+    #
```
hanars, line 69: can you update the comment to be a bit clearer that this is an example of the text param

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,282 @@
...
+    # Pathogenic(18); Likely pathogenic(9); Pathogenic, low penetrance(1); Established risk allele(1); Likely risk allele(1); Uncertain significance(1)
+    #
+    counts = {}
+    for part in text.split(";"):
+        part = part.strip()
+        label, count = part.rsplit("(", 1)
+        label = label.strip()
+        count = int(count.strip(")"))
```
hanars: I think this would be much neater with regular expressions:
```
counts = defaultdict(int)
for label, count in re.findall(r'([^;]+)\((\d+)\);', text):
  label = label.strip().replace(', low penetrance', '')
  counts[label] += int(count)
```

seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -0,0 +1,310 @@
...
+    @responses.activate
+    def test_parse_variants_all_types(self, mock_logger, mock_safe_post_to_slack):
+        KeyLookupSnvIndel.objects.using('clickhouse_write').create(
+            variant_id='22-28695219-G-A',
+            key_id=12,
+        )
+        KeyLookupMito.objects.using('clickhouse_write').create(
+            variant_id='M-123-GA',
+            key_id=8,
+        )
```
hanars: it would be better for you to create json fixture data for these tests and use that instead of creating all the models at test setup like this

seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -0,0 +1,310 @@
...
+                ]
+            },
+        )
+
+        seqr_clinvar_grch37_snv_indel_models = ClinvarGRCh37SnvIndel.objects.all()
+        self.assertEqual(seqr_clinvar_grch37_snv_indel_models.count(), 0)
+        clinvar_all_variants_grch37_snv_indel_models = ClinvarAllVariantsGRCh37SnvIndel.objects.all()
+        self.assertEqual(clinvar_all_variants_grch37_snv_indel_models.count(), 1)
+        clinvar_all_variants_mito_models = ClinvarAllVariantsMito.objects.all()
+        self.assertEqual(
```
hanars: `assertListEqual` would be better here, if this ever breaks the diff/error message is better the more specific of an equality function you use

seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -0,0 +1,310 @@
...
+                    </SimpleAllele>
+                    <Classifications>
+                        <GermlineClassification>
+                            <Description>Pathogenic</Description>
+                        </GermlineClassification>
+                    </Classifications>
+                </ClassifiedRecord>
+            </VariationArchive>
+            '''
+        data += '</ClinVarVariationRelease>'
```
hanars, line 193: I think it would be slightly more readable if you defined this as a top level constant and just referenced it here as like `BATCHED_XML_DATA`

seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -0,0 +1,310 @@
...
+            responses.add(
+                responses.GET,
+                WEEKLY_XML_RELEASE,
+                status=200,
+                body=gzip.compress(data.encode()),
+                stream=True,
+            )
+
+            with self.assertRaises(CommandError):
+                call_command('reload_clinvar_all_variants')
```
hanars, line 240: we should actually test that the expected error is raised - this test would succeed if every single one of these failed on the first error and many edge cases could be missed

seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -0,0 +1,310 @@
...
+            </ClinVarVariationRelease>'''
+            responses.add(
+                responses.GET,
+                WEEKLY_XML_RELEASE,
+                status=200,
+                body=gzip.compress(data.encode()),
+                stream=True,
+            )
+            call_command('reload_clinvar_all_variants')
+
```
hanars, line 322: we should actually test. that the results of this edge case are as expected, not just that it succeeds. I think just asserting that the model count is as expected would be sufficient

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: both `GRCh38` and `GRCh37` coordinates may be defined within the same variant, so in theory the function may return 0, 1, or 2 models for SNV_INDEL variants.  MITO doesn't particularly fit that pattern though.

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: Ah!  If the script runs but fails halfway through.  Since we're inserting in batches as we go, we might have a partial run that isn't marked as completed in Postgres.

clickhouse_search/migrations/0003_annotationsdiskmito_annotationsmito_entriesmito_and_more.py
bpblanken: Not new information no, it's coming in the XML as a different pathogenic status and there is no defined conflicting pathogenicities list.

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: It does not.  It's already included.  I did not choose it by default, it was suggested by Codacy.

seqr/management/commands/reload_clinvar_all_variants.py
hanars: a comment here would be helpful!

clickhouse_search/migrations/0003_annotationsdiskmito_annotationsmito_entriesmito_and_more.py
hanars: If there is no new information I would not add this as an assertion as it provides no value. The fact that the information is represented in multiple ways in the XML does not mean we should duplicate information in seqr

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: cleaned this up!

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: I removed these

seqr/management/tests/reload_clinvar_all_variants_tests.py
bpblanken: This is updated, and tests all 6 places where we throw an exception.

seqr/fixtures/clinvar_all_variants.json
```diff
@@ -0,0 +1,37 @@
+[
```
hanars: a bit nitpicky but this file should be in `clickhouse_search/fixtures`, not `seqr/fixtures`. Moving the file will not change how the fixture setup is done for tests

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -0,0 +1,298 @@
...
+            for event, elem in ET.iterparse(gzip.GzipFile(fileobj=r.raw), events=('start', 'end',)):
+                # Handle parsing the current date.
+                if event == 'start' and elem.tag == 'ClinVarVariationRelease':
+                    new_version = elem.attrib['ReleaseDate']
+                    existing_version_obj = DataVersions.objects.filter(data_model_name='Clinvar').first()
+                    if existing_version_obj:
+                        if existing_version_obj.version == new_version:
+                            logger.info(f'Clinvar ClickHouse tables already successfully updated to {new_version}, gracefully exiting.')
+                            return
+                    logger.info(f'Updating Clinvar ClickHouse tables to {new_version} from {existing_version_obj and existing_version_obj.version}.')
```
hanars, line 260: if there is no `existing_version_obj` then this log will read `'Updating Clinvar ClickHouse tables to VERSION from.` which is not a great message, it would be better if in the absence of `existing_version_obj` it logged something that makes it explicit that it was populating for the first time

seqr/management/tests/reload_clinvar_all_variants_tests.py
hanars: I still feel this way but would not block approval on it

hanars (approved):

seqr/fixtures/clinvar_all_variants.json
bpblanken: 👍

seqr/management/commands/reload_clinvar_all_variants.py
bpblanken: It should log [`from None`](https://github.com/broadinstitute/seqr/pull/4886/files/829be1f9812664e31706efe388fecea3fabc2f3f#diff-e5cd801c4e3efc398177fd6ecaa849b81550f6a4b54218f67860d67d0d3d8fb6R95), which was intentional.  I thought that that was explicit enough.

seqr/management/tests/reload_clinvar_all_variants_tests.py
bpblanken: yeah, I left it because the pattern is local to the test and there's an embedded variable, which I didn't feel was worth handling as a global.

PR #4890, opened 2025-07-23
Clickhouse saved variants

hanars:
This updates how saved variants work in clickhouse. Big changes include:
- adding clickhouse-specific columns to the saved variant table, including to store the key in clickhouse
- a manage command to map existing saved variants to their keys
- updates to everywhere that was previously querying the saved variant json to now reference clickhouse directly
- updates to our unit tests to properly handle transactions. This includes restricting which databases are used in a test wherever possible


seqr/views/apis/summary_data_api.py
```diff
@@ -229,43 +253,6 @@ def _load_aip_data(data: dict, user: User):
     })
 
 
-FamilyVariantKey = tuple[int, str]
-
-
-def _search_new_saved_variants(family_variant_ids: list[FamilyVariantKey], user: User):
```
hanars, line 235: this was mostly copy-pasted into the `bulk_create_tagged_variants` helper directly

seqr/views/utils/orm_to_json_utils_tests.py
```diff
@@ -107,23 +107,6 @@ def test_json_for_igv_sample(self):
 
         self.assertSetEqual(set(json.keys()), IGV_SAMPLE_FIELDS)
 
-    def test_json_for_saved_variant(self):
```
hanars, line 110: I generally do not like testing helper methods directly because its very hard to maintain the tests. In this case, the behavior of this helper is now backend specific, which is hard to test here. I confirmed that all the edge cases covered by this test are covered by other tests elsewhere

seqr/views/utils/test_utils.py
```diff
@@ -516,11 +514,43 @@ def get_group_members_side_effect(user, group, use_sa_credentials=False):
     return {}
 
 
-class AnvilAuthenticationTestCase(AuthenticationTestCase):
+class DifferentDbTransactionSupportMixin(object):
```
hanars, line 517: this mixin allows us to use transactions for the postgres database but not for clickhouse - django by default forces an either/or, and transactions in postgres greatly speed up our tests

seqr/views/utils/variant_utils.py
```diff
@@ -224,6 +252,92 @@ def _set_updated_tags(key: tuple[int, str], metadata: dict[str, dict], support_v
...
+    qs = get_annotations_queryset(genome_version, Sample.DATASET_TYPE_VARIANT_CALLS, key_id_map.keys())
+    variant_genes_by_id.update({
+        key_id_map[key]: set(gene_ids) for key, gene_ids in qs.values_list(
+            'key', ArrayDistinct(ArrayMap(qs.transcript_field, mapped_expression='x.geneId'), output_field=ArrayField(StringField())),
+        )
+    })
+    return variant_genes_by_id
+
+
+def _search_new_saved_variants(family_variant_ids: set[tuple[int, str]], user: User, genome_version: str) -> dict[tuple[int, str], dict]:
```
hanars, line 276: this is adapted from the ihelper copied in from individual_api

settings.py
```diff
@@ -253,6 +253,11 @@
         'PORT': int(os.environ.get('CLICKHOUSE_SERVICE_PORT', '9000')),
         'USER': os.environ.get('CLICKHOUSE_WRITER_USER', 'clickhouse'),
         'PASSWORD': os.environ.get('CLICKHOUSE_WRITER_PASSWORD', 'clickhouse_test'),
+        'OPTIONS': {
+            'settings': {
+                'use_client_time_zone': False,
```
hanars: I was running into weird errors when this is not set because it was trying to update the settings using the read-only credentials and failing. Essentially, we are telling django to leave this setting in the db alone, instead of updating it to django's default

bpblanken: I also ran into this and was able to resolve by just using the writer credentials.  What you've done is the suggested fix from Stack Overflow, but another potential solution is being more permissive [here](https://github.com/broadinstitute/tgg-helm/blob/main/charts/lib-broad-seqr/templates/_clickhouse-config-maps.yaml#L97), letting the reader user modify client settings.

clickhouse_search/search.py
```diff
@@ -451,3 +458,52 @@ def _add_liftover_genotypes(variant, data_type, variant_id):
...
+
+def get_clickhouse_keys_for_gene(gene_id, genome_version, dataset_type, keys):
+    results = get_annotations_queryset(genome_version, dataset_type, keys)
+    return list(results.filter(
+        **{f'{results.transcript_field}__array_exists': {'geneId': (f"'{gene_id}'",)}},
+    ).values_list('key', flat=True))
+
+
+def get_clickhouse_key_lookup(genome_version, dataset_type, variants_ids, reverse=False):
+    key_lookup_class = KEY_LOOKUP_CLASS_MAP[genome_version][dataset_type]
```
bpblanken, line 519: this might be cleaner like:


```
key_lookup_class = KEY_LOOKUP_CLASS_MAP[genome_version][dataset_type]
fields = ('variant_id', 'key') if not reverse else ('key', 'variant_id')

lookup = {}
batch_size = 10_000

for i in range(0, len(variant_ids), batch_size):
    batch = variant_ids[i:i + batch_size]
    rows = key_lookup_class.objects.filter(variant_id__in=batch).values_list(*fields)
    lookup.update(rows)
```

clickhouse_search/search_tests.py
```diff
@@ -20,18 +21,25 @@
...
     def setUp(self):
         super().set_up()
-        with connections['clickhouse'].cursor() as cursor:
+
+    @classmethod
+    def setUpTestData(cls):
+        with connections['clickhouse_write'].cursor() as cursor:
             for table_base in ['GRCh38/SNV_INDEL', 'GRCh38/MITO', 'GRCh38/SV', 'GRCh37/SNV_INDEL']:
                 cursor.execute(f'SYSTEM REFRESH VIEW "{table_base}/project_gt_stats_to_gt_stats_mv"')
+                # Need to wait for refresh to succeed before reloading the dictionary
```
bpblanken: I think you need `SYSTEM WAIT VIEW` here instead of sleep.  `REFRESH` is async and returns immediately.

clickhouse_search/managers.py
```diff
@@ -158,25 +214,55 @@ def search(self, parsed_locus=None, **kwargs):
...
+            fields.append('clinvar')
+
+        fields = [field for field in fields if field not in values]
         return self.values(*fields, **initial_values).annotate(
             **{k: values[k] for k in override_model_annotations if k in values},
         )
 
+    def join_seqr_pop(self):
+        results = self
+        seqr_popualtions = self.model.SEQR_POPULATIONS
```
bpblanken: populations is misspelled here.

clickhouse_search/managers.py
```diff
@@ -19,7 +19,40 @@
...
+        })
+
+    def _clinvar_tuple(self):
+        return Tuple(
+            *self.clinvar_fields.keys(),
+            output_field=NamedTupleField(list(self.clinvar_fields.values()), null_if_empty=True, null_empty_arrays=True),
+        )
+
+    def _seqr_pop_fields(self, seqr_popualtions):
+        sample_types = [self.single_sample_type.lower()] if self.single_sample_type else ['wes', 'wgs']
```
bpblanken: ['wes', 'wgs'] has to be an enumerable set of constants somewhere right?

seqr/views/utils/anvil_metadata_utils.py
```diff
@@ -359,23 +366,25 @@ def _get_parsed_saved_discovery_variants_by_family(
...
             'partial_contribution_explained': partial_hpo_terms.replace(', ', '|'),
             'notes': variant.notes,
             'sv_type': sv_type,
-            'sv_name': (variant_json.get('svName') or '{svType}:chr{chrom}:{pos}-{end}'.format(**variant_json)) if sv_type else None,
+            'sv_name': (variant_json.get('svName') or f'{sv_type}:chr{chrom}:{pos}-{variant_json["end"]}') if sv_type else None,
             'variant_type': variant_type,
             'validated_name': variant.validated_name[0] if variant.validated_name else None,
             **{k: _get_transcript_field(k, config, main_transcript) for k, config in TRANSCRIPT_FIELDS.items()},
-            **{k: variant_json.get(k) for k in ['genotypes'] + (variant_json_fields or [])},
+            **{k: variant_json.get(k) for k in ['genotypes'] + (['clinvar'] if include_clinvar else [])},
```
bpblanken: nit, but I think cleaner and more performant as:

```
**{
    k: variant_json.get(k)
    for k in (['genotypes', 'clinvar'] if include_clinvar else ['genotypes'])
}
```

settings.py
hanars: I'm just not sure *why* we would want to change this setting. The only reason its happening at all is that the implementers of the django-clickhosue package explicitly set this to default to True, and this does not seem necessary in our case. I would rather us have the database settings be what we want them to be all the time, rather than update them quietly in the background when performing searches

> was able to resolve by just using the writer credentials

I was running into this error when trying to just execute a search query against the database, so short of getting rid of the read-only credentials altogether this was not an option for that case

bpblanken: Ah I think we wouldn't.  Yours is a better solution.  

The case where we might want a different setting is to support something like:

```
SELECT * FROM my_table SETTINGS max_threads = 8;
```

which is effectively a read only query with a setting that modifies how the data is read.

bpblanken: Looks like there's a better way to accomplish that even, the [changeable_in_readonly](https://clickhouse.com/docs/operations/settings/constraints-on-settings#:~:text=The%20changeable_in_readonly%20constraint%20type%20allows,set%20to%201%20%2C%20otherwise) value.  Looks like a pain to configure though.

hanars: I think since we have no real world use cases right now where we actually want the read only credentials to be changing settings its actually better to keep the default behavior where that is disallowed. We can always revisit in the future if we actually need to be able to have query-level settings changes

bpblanken (approved):

PR #4899, opened 2025-07-31
Single variant search/Backend specific call cleanup

hanars:
This adds the final remaining clickhouse functionality that was missed in previous PRs, searching for a single variant. As part of this work, I deprecated old manage commands that could lift elasticsearch projects from GRCh37 to GRCh38. This functionality has not been used in years, is not well supported, and we had to add some special handling to the hail backend to support it that is not worth adding to clickhouse. 
Additionally, I did the final cleanup wherever there is a backend specific call and explicitly set the clickhouse behavior, instead of defaulting to raising an error.


clickhouse_search/search_tests.py
```diff
@@ -1083,27 +1083,6 @@ def test_in_silico_filter(self):
         self._set_grch37_search()
         self._assert_expected_search([GRCH37_VARIANT], in_silico=main_in_silico)
 
-#     def test_search_errors(self):
```
hanars, line 1086: these are hail specific and have no parallel in clickhouse

seqr/utils/search/utils.py
```diff
@@ -157,9 +172,9 @@ def _get_variants_for_variant_ids(families, variant_ids, user, user_email=None,
         }
     dataset_type = _variant_ids_dataset_type(parsed_variant_ids.values())
 
-    return backend_specific_call(get_es_variants_for_variant_ids, get_hail_variants_for_variant_ids)(
+    return backend_specific_call(get_es_variants_for_variant_ids, get_hail_variants_for_variant_ids, _raise_clickhouse_not_implemented)(
```
hanars, line 175: this function is only used for reloading saved variant json which is not needed in the clickhouse backend, so no point in implementing it

bpblanken (approved):

PR #4904, opened 2025-08-01
Return hom alt variants in trans with deletion in multi data type comp hets

hanars:
This was missed when I implemented comp het search, in part because it was not actually unit tested it. This adds back this behavior and tests it


clickhouse_search/search.py
```diff
@@ -94,7 +94,7 @@ def _get_multi_data_type_comp_het_results_queryset(genome_version, sample_data_b
             continue
         entries = entry_cls.objects.search([
             s for s in sample_data_by_dataset_type[Sample.DATASET_TYPE_VARIANT_CALLS] if s['family_guid'] in families
-        ], comp_het_search, annotate_carriers=True)
+        ], {**comp_het_search, 'inheritance_mode': COMPOUND_HET_ALLOW_HOM_ALTS}, annotate_carriers=True, annotate_hom_alts=True)
```
bpblanken: should this be:
```
**{**comp_het_search, 'inheritance_mode': COMPOUND_HET_ALLOW_HOM_ALTS}
```

`comp_het_search` is exploded elsewhere when passed to `search()` (though it's also missing a few lines below).

hanars: yeah its very weird because it should require the unpacking but also the search is working fine itn the tests. I will update all the calls for constancy though

hanars: Actually thinking about this more it feel like this is too bug prone. Instead of passing around dictionaries and then updating and deconstructing them I am just going to update this to deconstruct the dicts into keyword args from the get-go so that this issue can not happen in the future

PR #4907, opened 2025-08-04
Sv search clinvar bug

hanars:
Fixes the bug and adds tests for all the standard searches. While they do not return that many variants (our synthetic data for various reasons fails our standard filters) it is important that the combination of filters we regularly use are tested

### DIFF
--- a/clickhouse_search/managers.py
+++ b/clickhouse_search/managers.py
@@ -347,7 +347,7 @@ class AnnotationsQuerySet(SearchQuerySet):
 
     def _filter_frequency(self, results, freqs=None, pathogenicity=None, **kwargs):
         frequencies =  freqs or {}
-        clinvar_override_q = self._clinvar_path_q(pathogenicity)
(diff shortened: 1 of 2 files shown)


clickhouse_search/managers.py
```diff
@@ -347,7 +347,7 @@ def populations(self):
 
     def _filter_frequency(self, results, freqs=None, pathogenicity=None, **kwargs):
         frequencies =  freqs or {}
-        clinvar_override_q = self._clinvar_path_q(pathogenicity)
+        clinvar_override_q = self._clinvar_path_q(pathogenicity) if self.has_annotation(CLINVAR_KEY) else None
```
bpblanken, line 350: what's the difference between `self._has_clinvar()` and `self.has_annotation(CLINVAR_KEY)`?

hanars: One is from the EntryManager and the other is the AnnotationManager. `_has_clinvar` is actually checking whether there exists a foreign key relationship between the entry table and a clinvar table, `self.has_annotation(CLINVAR_KEY)` is checking whether the join with the entry table included a clinvar annotation or not

PR #4909, opened 2025-08-04
add allele validation before querying VLM

hanars:
Since the variant lookup page allows users to put in a free text variant ID, we need better validation that the ID we are proxying to the VLM is valid. Looking through the logs I found a case where MyGene2 was returning an error because we had passed along a totally invalid variant. Adding this validation will prevent undo alert fatigue from the VLM (any time a request does not return a 200 we get alerted)

### DIFF
--- a/seqr/views/apis/variant_search_api.py
+++ b/seqr/views/apis/variant_search_api.py
@@ -10,6 +10,7 @@ from django.db.models import Q, F, Value
 from django.db.models.functions import JSONObject
 from django.shortcuts import redirect
 from math import ceil
+import re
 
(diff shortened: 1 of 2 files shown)


seqr/views/apis/variant_search_api.py
```diff
@@ -634,6 +635,10 @@ def _update_lookup_variant(variant, response):
 @login_and_policies_required
 def vlm_lookup_handler(request):
     parsed_variant_id, _, kwargs = _parse_lookup_request(request)
+    if parsed_variant_id:
```
hanars, line 638: the id parsing helper function already includes validation that the chromosome and position are valid so we just need to validate the alleles

PR #4917, opened 2025-08-06
Clickhouse vlm

hanars:


vlm/clickhouse_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    'password': os.environ.get('CLICKHOUSE_PASSWORD'),
+    'database': os.environ.get('CLICKHOUSE_DATABASE', 'seqr'),
+}
+
+
+def get_clickhouse_variant_counts(locus: hl.LocusExpression, ref: str, alt: str, genome_build: str) -> hl.Struct:
+    locus = hl.eval(locus)
+    client = clickhouse_connect.get_client(**CLICKHOUSE_CONNECTION_PARAMS)
+    results = client.query(
+         "SELECT plus(gt_stats.1, gt_stats.2), plus(gt_stats.3, gt_stats.4) FROM (SELECT dictGet(%(dict_name)s, ('ac_wes', 'ac_wgs', 'hom_wes', 'hom_wgs'), key) AS gt_stats FROM %(table_name)s WHERE variantId=%(variant_id)s)",
```
bpblanken, line 18: Did this query come from the ORM generation?  Not super important, but it would be more reasonable to use `+` and the column names.

bpblanken (approved):

vlm/clickhouse_utils.py
hanars: Discussed offline, but related to issues in different environments whether or not `dictGet` returns a named tuple

PR #4922, opened 2025-08-08
Revert rocksdb to join for annotations

hanars:


clickhouse_search/migrations/0003_annotationsdiskmito_annotationsmito_entriesmito_and_more.py
```diff
@@ -36,7 +36,7 @@ class Migration(migrations.Migration):
             ],
             options={
                 'db_table': 'GRCh38/MITO/annotations_disk',
-                'engine': clickhouse_search.backend.engines.EmbeddedRocksDB(0, f'{CLICKHOUSE_DATA_DIR}/GRCh38/MITO/annotations', flatten_nested=0, primary_key='key'),
+                'engine': clickhouse_search.backend.engines.EmbeddedRocksDB(0, f'{CLICKHOUSE_DATA_DIR}/v0/GRCh38/MITO/annotations', flatten_nested=0, primary_key='key'),
```
hanars, line 39: Now if you run all the migrations from scratch, you create rocksdb memory/disk tables, then delete them, and then recreate them. This causes an error to occur because there is a LOCK file set for the tables that were deleted and thus they can't be recreated in the same spot. Therefore, for the original table creations that will never be used (since they will be dropped and recreated) use a different directory.
Since we have already run these migrations, modifying them will have no impact on our actual prod/dev installations. And before we release this code to local installations we plan to run djangos collapse migrations command and this will be pruned as part of that so it will never be run anywehere outside of tests, so it feels like a reasoanble workaround

bpblanken: I like the idea of a version flag, the LOCK issue has been a consistent pain.

bpblanken (approved):

PR #4931, opened 2025-08-13
gnomad genomes dict

bpblanken:


clickhouse_search/models.py
```diff
@@ -883,6 +883,28 @@ class GtStatsSv(models.ClickhouseModel):
...
+            order_by='key'
+        )
+
+class GnomadGenomesSnvIndel(models.ClickhouseModel):
+    key = OneToOneField('AnnotationsSnvIndel', db_column='key', primary_key=True, on_delete=CASCADE)
+    filter_af = models.DecimalField(max_digits=9, decimal_places=5)
+
+    class Meta:
+        db_table = 'GRCh38/SNV_INDEL/gnomad_genomes'
+        engine = models.ReplacingMergeTree(
```
bpblanken, line 903: I went with this engine because we'll be writing directly from the pipeline for now, with limited ability to deduplicate.  Eventually, if this were pulled directly from source and keyed by `variantId`, the engine would be different.

PR #4936, opened 2025-08-15
Improve inheritance search performance

hanars:
Removes maps from inheritance search to greatly improve performance. This does remove the ability to safely have additional samples available in an entry without the corresponding sample loaded in the seqr postgres db


clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -338,6 +338,62 @@
         ["HG00731", 1, 58, 0.17241, 29]
       ]
     }
+}, {
+    "model": "clickhouse_search.entriessnvindel",
+    "pk": 52,
+    "fields": {
+      "key": 2,
+      "project_guid": "R0001_1kg",
+      "family_guid": "F000002_2_x",
```
hanars, line 347: previously were testing an edge case by removing a sample from postgres and testing behavior, but this no longer works as seqr no longer ignores calls with no postgres sample. Therefore we are adding a copy of the data with the desired samples under a modified family. changes to search tests are related to these fixture changes, not other behavior changes

clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -351,7 +407,6 @@
       "filters": [],
       "sign": 1,
       "calls": [
-        ["NA20872", 1, 99, 0.34285714285714286, 35],
```
hanars, line 354: this call has no corresponding sample in the postgres database and was included in the fixture data to check that it was properly ignored, but now needs removal as this behavior has changed

PR #4942, opened 2025-08-19
feat: implement new fields

bpblanken:


hanars (commented):
this looks good but lets hold off on merging until the db in dev has these columns

PR #4953, opened 2025-08-21
chore: add handling for clinvar edge case

bpblanken:


seqr/management/tests/reload_clinvar_all_variants_tests.py
```diff
@@ -87,6 +87,11 @@ class ReloadClinvarAllVariantsTest(DifferentDbTransactionSupportMixin, TestCase)
     databases = '__all__'
     fixtures = ['clinvar_all_variants']
 
+    def tearDown(self):
```
bpblanken: inserts from one of the other tests were leaking.  I didn't see this pattern elsewhere though...

hanars: clickhouse does not support transactions in the same way that postgres does, so it will not reset fixture data between tests. Since we are basically only ever testing clcikhouse as read only this is not an issue in any of our other tests (the only place we actually test modifying the data is for deleting a project, and that is the ony test in the class that actually looks at clickhouse data so the between test teardown is moot)

One thing you might want to look into here is removing the `DifferentDbTransactionSupportMixin`. By default if django sees that some databases support transactions and some do not it treats all databses as if they do not support transactions and does a full database reset for all databases between tests, which really slowed down all our tests that use postgres and also broke some of how our tests initialize postgres test data. But I wrote this mixin assuming that we do not want to reset clickhouse between test, which is not the case for this test. In this case, I actually think djangos default set up/ clean up for no-transaction databases will be what ou want here

bpblanken: removing `DifferentDbTransactionSupportMixin` seems to work, but I'm digging a little deeper.  The line we have in the overwritten `tearDownClass` that calls the "flush" command appears to work, and propagates TRUNCATE commands to clickhouse via the djanog-clickhouse's correct implementation of [sql_flush_table](https://github.com/jayvynl/django-clickhouse-backend/blob/59c1fe549290f516c3adcd78c6d8807f47e70168/clickhouse_backend/backend/operations.py#L281).  In theory, the code should work as is 🤔.

bpblanken: Oh, tearDownClass doesn't run between tests 🤦.

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -259,11 +259,15 @@ def handle(self, *args, **options):
...
                     # Drop any currently existing variants in the table that may exist due to a
-                    # previously failed partial run.
-                    clinvar_run_sql(
-                        Template(f"ALTER TABLE `$reference_genome/$dataset_type/clinvar_all_variants` DROP PARTITION '{new_version}';")
-                    )
+                    # previously failed partial run.  Note that we validate that the Postgresql existing version
+                    # is present in ClickHouse to account for the situation where Postgresql has an incorrect
+                    # version.
+                    if existing_version_obj and existing_version_obj.version in loaded_clinvar_versions:
+                        clinvar_run_sql(
```
hanars, line 267: instead of fetching an entire list from the database and checking if the value is in the exist, why not just query for it directly - `ClinvarAllVariantsSnvIndel.objects.filter(version=existing_version_obj.version).exists()`

hanars (approved):

PR #4963, opened 2025-08-27
Sv geneid prefilter

hanars:


clickhouse_search/backend/functions.py
```diff
@@ -131,7 +131,9 @@ def process_lhs(self, compiler, connection):
 
     def process_rhs(self, compiler, connection):
         rhs, rhs_params = super().process_rhs(compiler, connection)
-        return f'bitmapBuild{rhs.split("::")[0]}', rhs_params
+        cast_type = self.lhs.output_field.base_field.cast_db_type(connection)
+        rhs = rhs.split("::")[0].replace('%s', f'%s::{cast_type}')
+        return f'bitmapBuild{rhs}', rhs_params
```
hanars, line 136: forces the integer values for the filtered gene ids to have the same type as the actual ids in the database

clickhouse_search/models.py
```diff
@@ -645,6 +647,7 @@ class Meta(BaseEntries.Meta):
         )
 
 class EntriesSv(BaseEntries):
+    MAX_XPOS_FILTER_INTERVALS = 0
```
hanars, line 650: for SNV_INDELs for small gene lists its more efficient to just use an xpos range instead of the bitmap filter. For SVs, since we essentially can't use the xpos filter at all its always more effective to use the bitmap filter

bpblanken (approved):

PR #4980, opened 2025-09-02
Deprecate dag trigger

hanars:


clickhouse_search/search.py
```diff
@@ -633,3 +633,7 @@ def delete_clickhouse_project(project, dataset_type=None, **kwargs):
             cursor.execute(f'SYSTEM WAIT VIEW "{view_name}"')
             cursor.execute(f'SYSTEM RELOAD DICTIONARY "{table_base}/gt_stats_dict"')
     return f'Deleted all {dataset_type} search data for project {project.name}'
+
+
+def delete_clickhouse_family(project, family_guid, dataset_type=None, **kwargs):
```
hanars: There are 2 places in the code where the delete family DAG was triggered, I have replaced them with this placeholder function so in the future when we implement this functionality we do not miss a place where it should be utilized

PR #4985, opened 2025-09-04
Cache intermediate search results

hanars:
There are still some searches where for users with access to more than just our internal data (currently one user) we are getting time outs on all project search. While we have longer term plans to support searches on this many families, for now as a workaround this PR adds intermediate caching of the SNV_INDEL search results (the bottleneck for these searches) so that if the cumulative time for all the search components times out the bulk of the query result will be cached so on page refresh the search will succeed. While this is not ideal behavior, it is a better than being fully unable to run searches until the refactor is completed


hanars:
Actually came up with a better solution for this

PR #5005, opened 2025-09-10
Check samples job hail backend cleanup

hanars:


seqr/management/commands/reload_saved_variant_annotations.py
```diff
@@ -1,27 +0,0 @@
...
-class Command(BaseCommand):
-    help = 'Reload shared variant annotations for all saved variants'
-
-    def add_arguments(self, parser):
-        parser.add_argument('data_type', choices=sorted(DATA_TYPE_CHOICES))
-        parser.add_argument('genome_version', choices=sorted(GENOME_VERSION_LOOKUP.values()))
-        parser.add_argument('chromosomes', nargs='*', help='Chromosome(s) to reload. If not specified, defaults to all chromosomes.')
-
-    def handle(self, *args, **options):
-        backend_specific_call(lambda: True, lambda: True, _clickhouse_error)()
```
hanars, line 26: although this backend specific call implies that this command can be called successfully for the elasticsearch backend, in actuality the helper function calls a hail-specific function and this would error if ever called for ES, which has a totally different mechanism for updating savwed variants thats untouched by this PR. With the hail backend deprecated, there is no need for this helper and command anymore

seqr/management/tests/check_for_new_samples_from_pipeline_tests.py
```diff
@@ -741,20 +681,18 @@ def test_command(self, mock_email, mock_temp_dir, mock_max_reload_variants):
...
-            'GRCh38/SNV_INDEL': create_snv_indel_samples_logs + airtable_logs + update_sample_qc_logs + self.RELOAD_PROJECT_VARIANT_LOGS,
+        self._test_call(num_runs=2, run_loading_logs={
+            'GRCh38/SNV_INDEL': create_snv_indel_samples_logs + airtable_logs + update_sample_logs,
         })
 
-@mock.patch('seqr.utils.search.hail_search_utils.HAIL_BACKEND_SERVICE_HOSTNAME', MOCK_HAIL_HOST)
-class LocalCheckNewSamplesTest(AuthenticationTestCase, CheckNewSamplesTest):
-    fixtures = ['users', '1kg_project']
+
+class LocalCheckNewSamplesTest(DifferentDbTransactionSupportMixin, AuthenticationTestCase, CheckNewSamplesTest):
```
hanars, line 693: previously this test case was testing a local install-like configuration with the hail backend, now it is testing local install with the clickhouse backend. While we already were testing the clickhouse backend on a broad-like configuration its still worth it to have 2 different test cases testing the same backend because they cover difference in local vs gcs filesystems and also behavior with and without airflow

PR #5011, opened 2025-09-10
update docs for dataproc + clickhouse

bpblanken:


deploy/LOCAL_INSTALL_HELM.md
```diff
@@ -95,19 +95,20 @@ kubectl exec seqr-POD-ID -c seqr -it -- bash
 gcloud auth list
 ```
 This service account must be permissioned to create and modify dataproc clusters, and read and write to the GCS buckets that will host your data.  We also recommend verifying that the Cloud API access scopes are permissioned for both IAM and Compute Engine access.  
+- [Create GCS service account HMAC key id and secret key](https://clickhouse.com/docs/integrations/gcs#generate-an-access-key) for your service account.  These keys are used by the direct Clickhouse integration with Google Cloud Storage.
+- Set those values as secrets within your Kubernetes envirionment:
+```
+kubectl create secret generic clickhouse-hmac-secrets
```
bpblanken, line 101: We've ended up with these named confusingly.  GCS refers to these values as "key" and "secret" but Clickhouse has this mapping:

Screenshot 2025-09-10 at 2 24 02 PM

and we shortened "secret key" to "key" in our most recent rename.

hanars: oops. Do we want to change this before go live?

deploy/LOCAL_INSTALL_HELM.md
```diff
@@ -117,8 +118,11 @@ pipeline-runner:
     GCLOUD_ZONE: 'zone-of-dataproc-cluster'
 seqr:
   environment:
-    # Bucket used for caching and worker communcation by the pipeline.
-    # Must be identical to `LOADING_DATASETS_DIR` stored above.
+    # Stores authoritative copies of newly annotated variants
```
hanars, line 121: shouldn't this be the global environment?

hanars (commented):
Would you mind branching this off of my documentation branch instead of dev: https://github.com/broadinstitute/seqr/pull/4996
I would like to be able to release the documentation updates as soon as we have released the actual helm chart and announcement without having to do a code release along with it

deploy/LOCAL_INSTALL_HELM.md
bpblanken: yeah, I dropped it accidentally.

deploy/LOCAL_INSTALL_HELM.md
bpblanken: I think we might as well.  I will do it, and make the change to the secrets and tgg-helm pr.

bpblanken: To me, the clearest names to an external user would be `pipeline_data_dir_hmac_key` and `pipeline_data_dir_hmac_secret`.  We will manage the Clickhouse poor naming when we create the named collection.

PR #5017, opened 2025-09-11
seqr resiliency during clickhouse outage

hanars:
Updates seqr to be more resilient when clickhouse is unavailable. This was done by identifying places in the code where we were fetching more data from the backend than the UI needed and paring that down, and then adding explicit error handling to places where the UI will work mostly fine if certain clickhouse information is missing
Also update the status check to actually detect a clickhouse outage

### DIFF
--- a/seqr/views/status_tests.py
+++ b/seqr/views/status_tests.py
@@ -11,9 +11,19 @@ from seqr.utils.search.elasticsearch.es_utils_tests import urllib3_responses
 @mock.patch('clickhouse_search.search.CLICKHOUSE_SERVICE_HOSTNAME', '')
 @mock.patch('seqr.views.status.redis.StrictRedis')
 @mock.patch('seqr.views.status.logger')
-class StatusTest(object):
+class ElasticsearchStatusTest(TestCase):
+    databases = '__all__'
 
-    def teardown(self):
+    SEARCH_BACKEND_ERROR = 'No response from elasticsearch ping'
+    HAS_KIBANA = True
+    ES_HOSTNAME = 'testhost'
+
+    def setUp(self):
+        patcher = mock.patch('seqr.utils.search.elasticsearch.es_utils.ELASTICSEARCH_SERVICE_HOSTNAME', self.ES_HOSTNAME)
+        patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def _post_teardown(self):
         for conn in connections.all():
             conn.connection = None
             conn.ensure_connection()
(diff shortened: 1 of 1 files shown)


seqr/views/status.py
```diff
@@ -17,7 +17,8 @@ def status_view(request):
     # Test database connection
     for db_connection_key in DATABASES.keys():
         try:
-            connections[db_connection_key].cursor()
+            with connections[db_connection_key].cursor() as cursor:
+                cursor.execute('SELECT 1')
```
hanars, line 21: different backends implement the `cursor` method differently, and it turns out the postgres's will throw an error in the connection is not usable but clickhouses will not. Explicitly executing against the database gets around this issue

seqr/views/status_tests.py
```diff
@@ -8,42 +9,58 @@
...
         if self.HAS_KIBANA:
             calls.append(mock.call('Search Admin connection error: Kibana Error 400: Bad Request'))
         mock_logger.error.assert_has_calls(calls)
-        mock_logger.reset_mock()
 
-    @mock.patch('seqr.views.status.redis.StrictRedis')
-    @mock.patch('seqr.views.status.connections')
-    @mock.patch('seqr.views.status.logger')
     @urllib3_responses.activate
-    def test_status(self, mock_logger, mock_db_connections, mock_redis):
```
hanars, line 35: I updated this test to no longer mock the database connections and instead actually mess them up during the test, but this necessitated restructuring the tests a bit

ui/pages/Project/reducers.js
```diff
@@ -256,13 +256,9 @@ export const getMmeMatches = submissionGuid => (dispatch, getState) => {
   const state = getState()
   const submission = state.mmeSubmissionsByGuid[submissionGuid]
   if (submission && !submission.geneVariants) {
-    const { familyGuid } = state.individualsByGuid[submission.individualGuid]
     dispatch({ type: REQUEST_MME_MATCHES })
     new HttpRequestHelper(`/api/matchmaker/get_mme_matches/${submissionGuid}`,
       (responseJson) => {
-        dispatch({
```
hanars, line 263: this was updating the tracker we have in the state for which families have saved variants loaded to prevent fetching again unneccessarily when navigating to a new page. Since we no longer return the full saved variants in this response, we will always want to refetch on other pages so we do not want this update anymore

seqr/views/status.py
bpblanken: ah!

PR #5022, opened 2025-09-15
Sv lookup cleanup

hanars:
In the hail backend the variant lookup for SVs had to function differently than for SNV_INDEL due to the absence of a lookup table. This is no longer needed for clickhouse so we can remove much of the sv-specific logic completely and simplify the logic and helper functions. There is still one major difference for SVs, which is that we will search across both sample types for them, but this logic no longer needs to run as a full search and is instead inlined into the clickhouse lookup function, allowing us to have better syntax for the `padded_interval` search

### DIFF
--- a/seqr/utils/search/utils.py
+++ b/seqr/utils/search/utils.py
@@ -329,7 +329,7 @@ def _query_variants(search_model, user, previous_search_results, genome_version,
 
     _validate_search(parsed_search, samples, previous_search_results)
 
-    variant_results = _execute_search(
+    variant_results = backend_specific_call(get_es_variants, get_clickhouse_variants)(
         samples, parsed_search, user, previous_search_results, genome_version,
         sort=sort, num_results=num_results, **kwargs,
     )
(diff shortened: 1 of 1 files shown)


clickhouse_search/managers.py
```diff
@@ -1043,18 +1043,14 @@ def _family_passes_expression(pass_field):
             mapped_expression='x.1', output_field=models.ArrayField(models.StringField()),
         )
 
-    def filter_locus(self, exclude_intervals=False, intervals=None, gene_intervals=None, gene_ids=None, variant_ids=None, padded_interval=None, require_gene_filter=False, **kwargs):
+    def filter_locus(self, exclude_intervals=False, intervals=None, gene_intervals=None, gene_ids=None, variant_ids=None, require_gene_filter=False, **kwargs):
```
hanars, line 1046: padded_interval was never available from the actual search UI , it was only passed by the lookup code, so now we can make the functionality and syntax more explicit

clickhouse_search/search.py
```diff
@@ -565,12 +567,29 @@ def clickhouse_variant_lookup(user, variant_id, data_type, genome_version=None,
...
-                variant = _clickhouse_variant_lookup(lifted_id, lifted_genome_version, data_type, samples)
+                variant = _clickhouse_variant_lookup(lifted_id, lifted_genome_version, data_type)
 
     if not variant:
         raise ObjectDoesNotExist('Variant not present in seqr')
 
-    return variant
+    variants = [variant]
+
+    if is_sv and variant['svType'] in {'DEL', 'DUP'}:
```
hanars, line 577: moves the other sample type into the lookup function directly without running a full search

PR #5024, opened 2025-09-16
remove unnecessary parsed locus processing

hanars:


clickhouse_search/managers.py
```diff
@@ -54,6 +54,11 @@ def _seqr_pop_fields(self, seqr_populations):
                 seqr_pop_fields += [f"'{sub_fields['hom']}_{sample_type}'" for sample_type in sample_types]
         return seqr_pop_fields
 
+    def _format_gene_intervals(self, genes):
+        return [
+            {field: gene[f'{field}Grch{self.genome_version}'] for field in ['chrom', 'start', 'end']} for gene in genes.values()
```
bpblanken, line 59: is there a reason `Grch` isn't `GRCh` here (like, other than it was necessary because it's like this elsewhere in the codebase so it's consistent)?

hanars: yeah because this is how the field names are serialized when the genes are parsed to json, it just is simple camel-casing of snake cased field names

bpblanken (approved):

PR #5026, opened 2025-09-17
Airflow deprecation

hanars:
Removes seqr integration with airflow and istead triggers loading through the pipeline runner service. 
This should not be released until prod deployment has been updated with a running pipeline runner


seqr/utils/search/add_data_utils.py
```diff
@@ -132,9 +153,9 @@ def _upload_data_loading_files(individual_ids: list[int], vcf_sample_id_map: dic
             raise e
 
 
-def _write_gene_id_file(load_data_dir, user):
+def _write_gene_id_file(user):
     file_name = 'db_id_to_gene_id'
-    if does_file_exist(f'{load_data_dir}/{file_name}.csv.gz'):
+    if does_file_exist(f'{LOADING_DATASETS_DIR}/{file_name}.csv.gz'):
```
hanars, line 158: in airflow utils we had hardcoded a `SEQR_V3_PEDIGREE_GS_PATH` variable that was passed through here, but now it has the same value as `LOADING_DATASETS_DIR` so its cleaner not to have this as an arg

PR #5030, opened 2025-09-18
Dataset type specific discovery genes

hanars:
We were not correctly looking up transcripts for their intended dataset types


seqr/views/apis/project_api_tests.py
```diff
@@ -390,7 +390,7 @@ def test_project_families(self):
         gene_ids = self._assert_expected_project_families(url, response_keys)
         self.assertSetEqual(gene_ids, {'ENSG00000135953', 'ENSG00000240361'})
 
-    def _assert_expected_project_families(self, url, response_keys):
+    def _assert_expected_project_families(self, url, response_keys, no_discovery_tags=False):
```
hanars, line 393: the behavior for how we handle variants that are absent from clickhouse when clickhouse is down changed a bit in this PR. Since we have essentially no discovery variants that are missing keys and this is only an outage mitigation situation, this change is insignificant

PR #5041, opened 2025-09-25
Delete search functionality cleanup

hanars:
After [enabling the trigger delete data commands for local users](https://github.com/broadinstitute/seqr/pull/5037), the fact that  there is also a separate manage command or manual sample deactivation that needs to be done as well is not intuitive and would lead to user error. The solution is to just move sample deactivation directly into the trigger endpoint behavior

### DIFF
--- a/seqr/management/commands/deactivate_project_search.py
+++ b/seqr/management/commands/deactivate_project_search.py
@@ -1,8 +1,7 @@
 from django.core.management.base import BaseCommand, CommandError
 
-from clickhouse_search.search import delete_clickhouse_project
 from seqr.models import Project, Sample
-from seqr.utils.search.utils import backend_specific_call
+from seqr.utils.search.utils import es_only
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


seqr/management/commands/deactivate_project_search.py
```diff
@@ -12,6 +11,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('project')
 
+    @es_only
```
hanars, line 14: the behavior of this manage command for the clickhouse backend can now fully be achieved through the UI so there is no need to maintain it as a separate command

ui/pages/DataManagement/components/TriggerSearchDataUpdatePages.jsx
```diff
@@ -46,25 +46,25 @@ const FAMILY_FIELDS = [
   DATASET_TYPE_FIELD,
 ]
 
-const TriggerSearchDataUpdateForm = ({ path, fields }) => (
+const TriggerSearchDataUpdateForm = ({ entity, fields }) => (
   <SubmitFormPage
-    header={`Trigger ${snakecaseToTitlecase(path)}`}
-    url={`/api/data_management/trigger_${path}`}
+    header={`Trigger Delete ${snakecaseToTitlecase(entity)} Search`}
```
hanars, line 51: I thought "Trigger Delete Project Search" was a better message for what happens with this action than the current "Trigger Delete Project"

PR #5045, opened 2025-09-26
Store gene ids on postgres saved variants

hanars:
There are several places where we need to access the genes associated with a saved variant, and querying clickhouse every time can be unwieldy, especially for pre-filtering results (as happens in the linked bug). Since VEP updates incredibly rarely, storing this information directly on the postgres model gives us a performance boost without any loss of accuracy or need for regular syncing between dbs (in the event that we ever do update VEP, a on-time job to update this will be very straightforward to write)
When going through the code to identify places where this new field could be used instead, I also identified a couple of places where we were returning an unnecessary amount of transcript data and changed the response/UI accordingly


matchmaker/views/matchmaker_api.py
```diff
@@ -50,44 +48,30 @@ def get_individual_mme_matches(request, submission_guid):
 
     results = MatchmakerResult.objects.filter(submission=submission)
 
-    response_json = get_json_for_saved_variants_with_tags(
```
hanars, line 53: we were returning all the saved variants and their tags etc to show on the matchmaker summary page, but this is actually unnecessary as all we need for that page are the saved variants associated with the submission. All the saved variants are needed if the user is updating a submission, but opening the modal to do that editing alrady triggers a load of all the needed data, so returning the data here as well as redundant

seqr/migrations/0081_savedvariant_gene_ids.py
```diff
@@ -0,0 +1,73 @@
...
+from clickhouse_search.backend.functions import ArrayMap
+from clickhouse_search.search import get_annotations_queryset
+
+BATCH_SIZE = 10000
+
+class JsonKeys(models.Func):
+    function = 'jsonb_object_keys'
+
+
+def populate_variant_gene_ids(apps, schema_editor):
```
hanars, line 16: this runs in about a minute so should not be a problem to run as a normal migration

seqr/views/apis/saved_variant_api.py
```diff
@@ -137,26 +137,20 @@ def _create_variant_note(saved_variants, note_json, user, genome_version):
 
 
 def _variant_gene_id(variant, genome_version):
-    main_transcript_id = variant.selected_main_transcript_id or variant.saved_variant_json.get('mainTranscriptId')
-    if main_transcript_id:
+    if variant.selected_main_transcript_id and len(variant.gene_ids) > 1:
```
hanars, line 140: gene_ids are sorted so the gene associated with the `mainTranscriptId` is always first, so there is no longer any need to determine which gene is the "main" one

seqr/views/apis/summary_data_api.py
```diff
@@ -115,11 +112,7 @@ def saved_variants_page(request, tag):
     saved_variant_models = saved_variant_models.filter(family__project__guid__in=get_project_guids_user_can_view(request.user))
 
     if gene:
-        gene_filter = Q(saved_variant_json__transcripts__has_key=gene) | backend_specific_call(
-            lambda *args: Q(),
-            _saved_variant_with_clickhouse_gene_q,
-        )(saved_variant_models, gene, is_all_tags)
-        saved_variant_models = saved_variant_models.filter(gene_filter)
+        saved_variant_models = saved_variant_models.filter(gene_ids__overlap=[gene])
```
hanars, line 115: this is what fixes the linked bug

seqr/views/utils/project_context_utils.py
```diff
@@ -118,63 +118,69 @@ def add_child_ids(response):
 
 
 def families_discovery_tags(families, genome_version, project=None):
-    families_by_guid = {f['familyGuid']: dict(discoveryTags=[], **f) for f in families}
+    families_by_guid = {f['familyGuid']: dict(discoveryGeneIds=[], **f) for f in families}
```
hanars, line 121: we were previously returning a list of tags with transcripts and main and selected transcripts IDs, but then all the UI was doing was determining which of those is the single main gene to show. Moving that server side makes this much easier and removes the need to query clickhouse for most variants

PR #5054, opened 2025-09-30
Benb/change communicate

bpblanken:


seqr/utils/file_utils.py
```diff
@@ -117,8 +117,8 @@ def run_gsutil_with_wait(command, gs_path, user=None, **kwargs):
     return process
 
 
-def _run_gsutil_with_stdout(command, gs_path, user=None, allow_missing=False):
-    process = _run_gsutil_command(command, gs_path, user=user, pipe_errors=True)
+def _run_gsutil_with_stdout(command, gs_path, user=None, allow_missing=False, **kwargs):
```
hanars, line 120: I don't think this approach will work at all because `_google_bucket_file_iter` needs to work with files that are too large to read all the way into memory and I think `communicate` will put the whole output into memory, not an iterator like `process.stdout`

PR #5055, opened 2025-09-30
Trigger delete families

hanars:


ui/pages/DataManagement/components/TriggerSearchDataUpdatePages.jsx
```diff
@@ -43,7 +43,6 @@ const FAMILY_FIELDS = [
     placeholder: 'Search for a family',
     validate: validators.required,
   },
-  DATASET_TYPE_FIELD,
```
hanars, line 46: Deleting family data now runs for all dataset types at once, so remove the field from the UI

bpblanken (approved):

PR #5066, opened 2025-10-03
Gene variant look

hanars:


clickhouse_search/search_tests.py
```diff
@@ -1464,3 +1467,56 @@ def test_multi_data_type_comp_het_sort(self):
                 [{'selectedGeneId': 'ENSG00000277258'}, {'selectedGeneId': 'ENSG00000277258'}],
             ]
         )
+
+    def test_gene_variant_lookup(self):
+        url = reverse(gene_variant_lookup)
```
hanars, line 1484: The pattern we generallyhave for search tests is we have a test that tests the request and response but mocks the actual backend helper function, and then we have tests that call the helpers directly and test their behavior. This is not an ideal way to test things as its vulnerable to missed bugs if the function signature changes at all, and is also a bit hard to read/maintain. The reason we have this elsewhere is in order to test shared vs different behavior across different backends, but for this entirely new function thats only available in clickhouse it felt better to just actually test the request end-to-end without any mocking. Once elasticsearch is fully deprecated all the other tests in this file will eventually be converted to this pattern and combined with the other request tests

clickhouse_search/search_tests.py
```diff
@@ -21,15 +21,17 @@
 from seqr.utils.search.search_utils_tests import SearchTestHelper
 from seqr.utils.search.utils import query_variants, variant_lookup, get_variant_query_gene_counts, get_single_variant, InvalidSearchException
 from seqr.views.utils.json_utils import DjangoJSONEncoderWithSets
-from seqr.views.utils.test_utils import DifferentDbTransactionSupportMixin
+from seqr.views.utils.test_utils import AnvilAuthenticationTestCase
+from seqr.views.apis.variant_search_api import gene_variant_lookup
 
 
-class ClickhouseSearchTests(DifferentDbTransactionSupportMixin, SearchTestHelper, TestCase):
+class ClickhouseSearchTests(SearchTestHelper, AnvilAuthenticationTestCase):
```
bpblanken: I think this will conflict with the changes I made to the transaction management here on the affected status.  Will we need to split `DifferentDbTransactionSupportMixin` out from `AnvilAuthenticationTestCase`?

hanars: good point, I will refactor this so it can be integrated with your changes without conflict

hanars: Okay I have refactored the test case to instead be a mixin that is usable here

seqr/views/utils/test_utils.py
```diff
@@ -603,6 +612,19 @@ def assert_no_extra_anvil_calls(self):
         self.mock_get_group_members.assert_not_called()
 
 
+class AnvilAuthenticationTestCase(DifferentDbTransactionSupportMixin, AnvilAuthenticationTestMixin, TestCase):
```
bpblanken, line 615: 👍

PR #5078, opened 2025-10-16
Handle updates to affected status

hanars:
Restricts updating affected status except where intentional, and then explicitly triggers reload where necessary

### DIFF
--- a/seqr/utils/search/add_data_utils.py
+++ b/seqr/utils/search/add_data_utils.py
@@ -83,7 +83,7 @@ def trigger_delete_families_search(project, family_guids, user=None):
 
 def trigger_rebuild_gt_stats(project, user):
     logger.info(f'Triggering rebuild_gt_stats for {project.guid}', user)
-    _enqueue_pipeline_request('rebuild_gt_stats', {'project_guids': project.guid}, user, raise_error=False)
+    _enqueue_pipeline_request('rebuild_gt_stats', {'project_guids': [project.guid]}, user, raise_error=False)
 
 def trigger_data_loading(projects: list[Project], individual_ids: list[int], sample_type: str, dataset_type: str,
                          genome_version: str, data_path: str, user: User, raise_error: bool = False, skip_expect_tdr_metrics: bool = True,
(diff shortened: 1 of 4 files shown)


seqr/utils/search/add_data_utils.py
```diff
@@ -119,21 +123,23 @@ def trigger_data_loading(projects: list[Project], individual_ids: list[int], sam
     return success
 
 
-def _enqueue_pipeline_request(name: str, variables: dict, user: User, raise_error: bool = True):
-    response = requests.post(f'{PIPELINE_RUNNER_SERVER}/{name}_enqueue', json=variables, timeout=60)
+def _enqueue_pipeline_request(name: str, variables: dict, user: User, raise_error: bool = True, log_error: bool = True):
+    response = None
     error = None
     try:
+        response = requests.post(f'{PIPELINE_RUNNER_SERVER}/{name}_enqueue', json=variables, timeout=60)
```
bpblanken, line 130: I made a ticket [to eventually make this cleaner](https://github.com/broadinstitute/seqr-loading-pipelines/issues/1175), now that there are generic jobs being run.

seqr/utils/search/add_data_utils.py
```diff
@@ -119,21 +123,23 @@ def trigger_data_loading(projects: list[Project], individual_ids: list[int], sam
...
         response.raise_for_status()
         logger.info(f'Triggered {_to_title_case(name)}', user, detail=variables)
-    except requests.HTTPError as e:
+    except requests.RequestException as e:
         error = str(e)
-        if response.status_code == 409:
+        if response is not None and response.status_code == 409:
```
bpblanken, line 135: If I change the affected status, and the pipeline queue happens to be full, will the UI display the exception to the user?  Or will it fail in the background?

hanars: how we handle the exception is different for every type of enqueue:
- loading fails for our installation - shows success to the user and posts the error to slack
- loading fails for local installs - shows the user the error
- delete projects fails - shows the error to the user
- update affected status fails - shows success to the user and logs an error (which notifies us). The affected status itself is successfully changed in seqr

bpblanken (approved):

PR #5079, opened 2025-10-17
Benb/update partitioning

bpblanken:


clickhouse_search/migrations/0019_projectpartitions.py
```diff
@@ -0,0 +1,109 @@
...
+from django.db import migrations, connections
+import django.db.models.deletion
+import django.db.models.manager
+
+
+CLICKHOUSE_WRITER_PASSWORD = os.environ.get('CLICKHOUSE_WRITER_PASSWORD', 'clickhouse_test')
+CLICKHOUSE_WRITER_USER = os.environ.get('CLICKHOUSE_WRITER_USER', 'clickhouse')
+
+PROJECT_PARTITIONS_DICT = Template("""
+CREATE DICTIONARY `GRCh38/SNV_INDEL/project_partitions_dict`
```
bpblanken, line 16: I ended up running into a set of very nasty issues with defining the partitions table as a Join table.  When flushing/truncating the table within the tests, ClickHouse freezes up and starts blocking any edit to the table.  I wasn't able to fully diagnose or fix it, giving up and going with a dictionary approach.  

The issues I encountered made me queasy enough about the dynamic join partitioning approach that I've instead followed a safer path that materializes the partition index and partition number on the entries table as columns.  This adds complexity to the table definition and the load, but should be safer.

clickhouse_search/migrations/0019_projectpartitions.py
```diff
@@ -0,0 +1,109 @@
...
+            return
+        cursor.execute('SELECT COUNT(*) FROM `GRCh38/SNV_INDEL/entries`;')
+        if cursor.fetchone()[0] > 0:
+            return
+        create_table_query = create_table_query.replace(
+            'PARTITION BY project_guid',
+            'PARTITION BY (project_guid, partition_id)',
+        )
+        cursor.execute('DROP TABLE `GRCh38/SNV_INDEL/entries`')
+        cursor.execute(create_table_query)
```
bpblanken, line 54: The conditional drop and recreate approach:
- Leaves the existing squashed migration (0015) as is.
- Allows adding the two new columns.
- does NOT modify PARTITION BY if the table exists and has data in it.  Users that have already loaded data will not have the new structure.  

I also considered just modify-ing the existing migration, but in order to make that work we'd also need to move the new project_partitions table and the new columns to that migration and remove this one entirely.  That is a valid approach, but I thought less clear and maintainable.

clickhouse_search/backend/engines.py
```diff
@@ -5,7 +5,6 @@
 def _no_validate(value, name):
     return value
 
-
```
hanars, line 8: not very important but I think it is the convention to have 2 blank lines between top level functions

clickhouse_search/migrations/0017_projectpartitionssnvindel.py
```diff
@@ -0,0 +1,109 @@
...
+            'PARTITION BY (project_guid, partition_id)',
+        )
+        cursor.execute('DROP TABLE `GRCh38/SNV_INDEL/entries`')
+        cursor.execute(create_table_query)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('clickhouse_search', '0016_add_affected_status'),
```
hanars: I think this should be updated as we have since merged other migrations

clickhouse_search/migrations/0019_projectpartitions.py
```diff
@@ -0,0 +1,109 @@
...
+        if 'farmHash64(family_guid)' in create_table_query:
+            return
+        cursor.execute('SELECT COUNT(*) FROM `GRCh38/SNV_INDEL/entries`;')
+        if cursor.fetchone()[0] > 0:
+            return
+        create_table_query = create_table_query.replace(
+            'PARTITION BY project_guid',
+            'PARTITION BY (project_guid, partition_id)',
+        )
+        cursor.execute('DROP TABLE `GRCh38/SNV_INDEL/entries`')
```
hanars, line 53: are we not updating partitioning for GRCh37?

clickhouse_search/migrations/0019_projectpartitions.py
```diff
@@ -0,0 +1,109 @@
...
+            state_operations=[
+                migrations.AddField(
+                    model_name='entriessnvindel',
+                    name='n_partitions',
+                    field=clickhouse_search.backend.fields.MaterializedUInt8Field(expression="dictGetOrDefault('GRCh38/SNV_INDEL/project_partitions_dict', 'n_partitions', project_guid, 1)"),
+                )
+            ],
+        ),
+        migrations.RunSQL(
+            "ALTER TABLE `GRCh38/SNV_INDEL/entries` ADD COLUMN `partition_id` UInt8 MATERIALIZED farmHash64(family_guid) % n_partitions AFTER `n_partitions`;",
```
hanars, line 98: out of scope for this PR but it seems like the only reason we need custom SQL for all of our `ADD COLUMN` migrations is that we need the ability to specify an AFTER column, and it would only be like an hour or 2 of work for me to add support for that in the clickhouse schema editor. I think its worth making a ticket for this assuming that is the only reason

clickhouse_search/migrations/0019_projectpartitions.py
bpblanken: I felt like we shouldn't.  There isn't a project in GRCh37 that would come close to needing re-partitioning.  I think all of our GRCh37 data put together is ~15% of the Gregor multi site.

clickhouse_search/migrations/0019_projectpartitions.py
bpblanken: The other piece the default SQL generates is a `mutations_sync = 1` setting, which we very much do not want.

hanars: that should also be handleable in the the update so just include that in the ticket as well

PR #5083, opened 2025-10-22
bugfix: entries to project gt stats mv

bpblanken:


clickhouse_search/migrations/0017_fix_affected_status_column_order.py
```diff
@@ -0,0 +1,107 @@
...
+
+from settings import DATABASES
+
+ENTRIES_TO_PROJECT_GT_STATS = Template("""
+CREATE MATERIALIZED VIEW `$reference_genome/$dataset_type/entries_to_project_gt_stats_mv`
+TO `$reference_genome/$dataset_type/project_gt_stats`
+AS SELECT
+    project_guid,
+    key,
+    $columns
```
bpblanken, line 11: I'd lifted the dictGet to be shared amongst all tables, hoping that the view would correctly handle the column order swap between sample type and affected status.  it did not!

PR #5088, opened 2025-10-23
bugfix: clinvar download file, then parse

bpblanken:


bpblanken:
Forgot to handle tmp file deletion, adding it now!

PR #5091, opened 2025-10-23
feat: affected status migration

bpblanken:


clickhouse_search/migrations/0018_queue_rebuild_gt_stats_jobs.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    db_alias = schema_editor.connection.alias
+    for batch in batch_project_guids(
+        list(
+            set(
+                ProjectGtStatsGRCh37SnvIndel.objects.using(db_alias)
+                .values_list("project_guid", flat=True)
+                .distinct()
+            )
+            | set(
```
hanars, line 33: this works fine, but project guids are distinct and only associated with a single genome build so you could also just do `list(ProjectGtStatsGRCh37SnvIndel...) + list(ProjectGtStatsSnvIndel...)`

PR #5100, opened 2025-10-27
feat: move clinvar to new pattern

bpblanken:


seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -274,7 +274,7 @@ def handle(self, *args, **options):
                         # version.
                         if existing_version_obj and ClinvarAllVariantsSnvIndel.objects.filter(version=existing_version_obj.version).exists():
                             clinvar_run_sql(
-                                Template(f"ALTER TABLE `$reference_genome/$dataset_type/clinvar_all_variants` DROP PARTITION '{new_version}';")
+                                Template(f"ALTER TABLE `$reference_genome/$dataset_type/reference_data/clinvar/all_variants` DROP PARTITION '{new_version}';")
```
hanars: You can actually get the table name off of the model's meta field which would make this code more resilient - you just use `ClinvarAllVariantsSnvIndel._meta.db_table`. It would also get rid of the need to template out variables

seqr/management/commands/reload_clinvar_all_variants.py
```diff
@@ -66,7 +66,6 @@ def clinvar_run_sql(sql: str):
         ]:
             cursor.execute(sql.substitute(reference_genome=reference_genome, dataset_type=dataset_type))
```
hanars, line 67: not particularly related to this PR, but this file should really be in the `clickhouse_search/management/commands` folder not the seqr commands folder

bpblanken:
`clinvar/all_variants` and the `clinvar` Join table are just renames, so will have all of the data, leaving seqr search untouched.  The `seqr_variants` table will be empty though, until either the job runs or a manual intervention.  

I _think_ this is fine, but there is a potential edge case where the two materialized views we create here run concurrently upon creation; the second one (the `seqr_variants` -> `clinvar` Join) will have an empty source table until the first completes, resulting in an empty Join table.   

One hacky resolution would be to add a RunPython that runs `'SYSTEM REFRESH VIEW` in the correct order at the end of this migration.

clickhouse_search/migrations/0020_clinvar_new_reference_data.py
```diff
@@ -0,0 +1,594 @@
...
+CREATE MATERIALIZED VIEW `$reference_genome/$dataset_type/reference_data/clinvar/seqr_variants_to_search_mv`
+REFRESH EVERY 10 YEAR
+TO `$reference_genome/$dataset_type/reference_data/clinvar`
+AS 
+SELECT 
+DISTINCT ON (key) *
+FROM `$reference_genome/$dataset_type/reference_data/clinvar/seqr_variants`
+""")
+
+def build_materialized_view(reference_genome: str, dataset_type: str, materialized_view: str):
```
bpblanken, line 33: added a mechanism for refreshing and waiting for each view.

PR #5101, opened 2025-10-28
feat: reference data range dictionaries

bpblanken:


clickhouse_search/migrations/0020_pext_range_dicts.py
```diff
@@ -0,0 +1,239 @@
...
+GNOMAD_NON_CODING_CONSTRAINT_VIEW = """
+CREATE MATERIALIZED VIEW `GRCh38/SNV_INDEL/reference_data/gnomad_non_coding_constraint/all_variants_mv`
+REFRESH EVERY 10 YEAR
+TO `GRCh38/SNV_INDEL/reference_data/gnomad_non_coding_constraint/all_variants`
+AS SELECT
+    replaceOne(chrom, 'chr', '') as chrom,
+    toUInt32(assumeNotNull(start)) as start,
+    toUInt32(assumeNotNull(end)) as end,
+    z
+FROM url('https://storage.googleapis.com/gcp-public-data--gnomad/release/3.1/secondary_analyses/genomic_constraint/constraint_z_genome_1kb.qc.download.txt.gz')
```
hanars: will this work out of the box for local installs, or is there any set up we mauy need to do in the seqr image or helm chart to ensure that accessing GCP works from here?

clickhouse_search/migrations/0020_pext_range_dicts.py
```diff
@@ -0,0 +1,239 @@
...
+SCREEN_VIEW = """
+CREATE MATERIALIZED VIEW `GRCh38/SNV_INDEL/reference_data/screen/all_variants_mv`
+REFRESH EVERY 10 YEAR
+TO `GRCh38/SNV_INDEL/reference_data/screen/all_variants`
+AS SELECT
+    replaceOne(c1, 'chr', '') as chrom,
+    toUInt32(assumeNotNull(c2)) as start,
+    toUInt32(assumeNotNull(c3)) as end,
+    splitByChar(',', assumeNotNull(c6))[1] as regionType
+FROM url('https://downloads.wenglab.org/V3/GRCh38-cCREs.bed')
```
hanars: I do not have a great solution for this problem, but I think that in the future its going to be very hard to track down what the sources are for all our reference datasets. Making these into variables in any shared code (like models.py) and then importing them to the migration is not a good idea because this migration should stay static always even if we later change the source. The easiest solution I have come up with is to make them constants at the top of this migration file (to make them easier to find)  and then rename to migration something like `0020_refererence_sources` and name all future migrations that add sources like this `N_refererence_sources` so its easy to see which migration files to open to check.

The other alternative I can think of is that I actually implement a `MaterializedView` class and we add models for all the views in models.py. I guess the django for this would look something like
```
class ScreenMv(ClickhouseMaterializedView):
    chrom = Enum8Field(return_int=False, choices=BaseAnnotations.CHROMOSOME_CHOICES, primary_key=True)
    ...

    class Meta:
        db_table = 'GRCh38/SNV_INDEL/reference_data/screen/all_variants_mv'
        to_table = 'GRCh38/SNV_INDEL/reference_data/screen/all_variants'
        source_url = 'https://downloads.wenglab.org/V3/GRCh38-cCREs.bed'
        column_selects = {
            'chrom': 'replaceOne(c1, 'chr', '')',
            ...
        }
```
I think getting this to work and to generate the code I want is probably like 2 days of work. The advantage to that would be that the live source versions would all be defined in models.py, and if in the future it needed change we would be able to use a more normal migration pattern to change them. The disadvantage would be that this is a bunch of work that would delay this code (although I could also do this after the code goes live instead and backfill it)

 I am also open to other suggestions

bpblanken: I had similar thoughts while doing this, eventually just implementing in the simplest way.

I'm a little lukewarm on the Django Model approach, just because these views are not objects that will ever be queried by the application and are likely to exhibit strange behavior if treated as relational tables.  I think it's overkill for what are basically static templates, and when we update them we will always drop and recreate rather than alter.

What about a combination of a shared python string template for the SQL and an additional python dictionary config in a separate file?  

The config would have the source url and column selects, but the table definitions are actually static here.

clickhouse_search/migrations/0020_pext_range_dicts.py
bpblanken: Yep! These are public gcp urls, not hosted by us.

clickhouse_search/migrations/0020_pext_range_dicts.py
hanars: I think that a file that exists purely as a tracking file with no actual utilization in code is really annoying to keep up to date and unlikely to stay accurate. And actually importing the config to the migration would be breaking later onn so I do not think we should do that. At this point I would rather just use a naming convention to make the relevant migrations easy to find if we need to figure things out, which feels like the easiest path forward 

Also, there are lots of actual tables in our models.py that are never queried/ directly used by the application because they are just the table part of a view so (to me) its actually clearer to have something else django-y defined with them to make it clear why they are there in the first place, like having the table and view defined and clearly linked makes more sense than having them only defined in the migrations. But I do not have strong feelings on it, and definitely would want to keep it out of scope for this PR

clickhouse_search/migrations/0020_pext_range_dicts.py
hanars: Yeah I was more worried that like they are public but you need to pass some sort of credential or requester pays or something thats not standard here, but I now realize that that also would not have worked for you when you ran it if that was the case

PR #5109, opened 2025-10-30
Do not trigger gt stats update on individual create

hanars:


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -30,7 +30,7 @@
 
 LOAD_SAMPLE_DATA_ALL_PENDING = LOAD_SAMPLE_DATA + [
     ["2", "HG00731", "", "HG00732", "HG00733", "Female", "Affected", "HP:0001508", "", ""],
-    ["2", "HG00732", "", "", "", "Male", "Unaffected", "", "", ""],
+    ["2", "HG00732", "", "", "", "Male", "Affected", "HP:0001508", "", ""],
```
hanars, line 33: this change means that there is an individual who is changing affected status with this loading request, so we are testing that gt stats will not rebuild when that happens

bpblanken (approved):

PR #5110, opened 2025-10-31
bugfix: quote passwords

bpblanken:
Resolves an issue with passwords with special characters.

We _could_ do this with a new migration with REPLACE, but the migrations actually crash as is.


bpblanken:
I think only an issue if doing a clean new install with a special char password.  If you have a running install now, you chose a safe password (like us).

PR #5114, opened 2025-11-03
use backend specific filters for in silico fields

hanars:
Some of the in silico fields in seqr have configurations to support both the enum options used by elasticsearch and the numeric options used by clickhouse. Currently any field with an enum is shown as an enum select field in the search UI - this PR changes it so that behavior is maintained for elasticsearch but for clickhouse it will preferentially use the numeric field if available. I have confirmed that this change will correctly map the search UI to the actual data types used in the clickhouse backend. Note that mut_taster is still an enum in clickhouse which is why we do not want to fully get rid of that functionality

### DIFF
--- a/ui/pages/Search/components/VariantSearchFormPanelConfigs.jsx
+++ b/ui/pages/Search/components/VariantSearchFormPanelConfigs.jsx
@@ -69,7 +69,7 @@ const REQUIRE_SCORE_FIELD = {
   labelHelp: 'Only return variants where at least one filtered predictor is present. By default, variants are returned if a predictor meets the filtered value or is missing entirely',
 }
 
-const FORMATTED_IN_SILICO_FUIELDS = [...ORDERED_PREDICTOR_FIELDS.filter(({ displayOnly }) => !displayOnly).map(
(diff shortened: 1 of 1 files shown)


ui/pages/Search/components/VariantSearchFormPanelConfigs.jsx
```diff
@@ -69,42 +69,48 @@ const REQUIRE_SCORE_FIELD = {
...
+      step: max ? 1 : 0.05,
+      ...filterField,
+    } : null
+    return [enumField, decimalField]
+  },
+)]
+
+const IN_SILICO_FIELDS = [
+  REQUIRE_SCORE_FIELD,
+  ...FORMATTED_IN_SILICO_FUIELDS.map(([enumField, decimalField]) => decimalField || enumField),
```
bpblanken: FORMATTED_IN_SILICO_FUIELDS is misspelled!

hanars: 😮

bpblanken (approved):

PR #5141, opened 2025-11-19
Affected status dict

hanars:
For multi-family search use the affected status dict in queries instead of constructing sample ID maps


clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -1317,7 +1317,7 @@
       "sign": 1,
       "calls": [
         ["NA21234", 2, 4, 27, true, 38717327, 38719636, 3, ["ENSG00000275023"], true, false, false],
-        ["NA21987", 1, 3, 51, false, 38717327, 38719636, 3, ["ENSG00000275023"], false, false, true],
```
hanars, line 1320: we use a custom affected filter in all our SV searches due to how WGS data is configured in the fixtures, but now there is one edge case where we add GCNV data and do a mulit-family search and that custom affected filter is not applied (as its impossible to add custom affected in the UI for multi family search), so updating the fixture data to pass the inheritance filter with the actual affected stats and not rely on the custom override is this simplest fix

PR #5142, opened 2025-11-19
Show seqr affected ac

hanars:


clickhouse_search/test_utils.py
```diff
@@ -394,6 +399,7 @@
     },
     'populations': {
         'seqr': {'ac': 2, 'hom': 1, 'ac_wes': 2, 'ac_wgs': 0, 'hom_wes': 1, 'hom_wgs': 0},
+        'seqr_affected': {'ac': 0, 'hom': 0},
```
hanars: @bpblanken I am seeing some weird behavior for some of the `entries_to_project_gt_stats_mv` tables in the tests, including for the build 37 variant here. When I connect to the db during a test a see the following:
```
SELECT *
FROM test_seqr.`GRCh37/SNV_INDEL/entries_to_project_gt_stats_mv`

Query id: 4fd852ca-6609-4cd9-a578-a01f2f35afd9

   ┌─project_guid──────────────┬─key─┬─sample_type─┬─affected─┬─ref_samples─┬─het_samples─┬─hom_samples─┐
1. │ R0004_non_analyst_project │  21 │ WGS         │ A        │           0 │           1 │           0 │
2. │ R0001_1kg                 │  11 │ WES         │ N        │           0 │           0 │           1 │
   └───────────────────────────┴─────┴─────────────┴──────────┴─────────────┴─────────────┴─────────────┘
```
But when I re-create the select statement that backs the view I do see the expected affected row for key 11:
```
SELECT
    project_guid,
    key,
    sample_type,
    dictGetOrDefault(test_seqr.seqrdb_affected_status_dict, 'affected', (family_guid, calls.sampleId), 'U') AS affected,
    sumIf(sign, calls.gt = 'REF'),
    sumIf(sign, calls.gt = 'HET'),
    sumIf(sign, calls.gt = 'HOM')
FROM test_seqr.`GRCh37/SNV_INDEL/entries`
ARRAY JOIN calls
GROUP BY
    project_guid,
    key,
    sample_type,
    affected

Query id: 7d7f6936-ba6a-44fd-b17c-11ff17aeb916

   ┌─project_guid──────────────┬─key─┬─sample_type─┬─affected─┬─sumIf(sign, ⋯gt, 'REF'))─┬─sumIf(sign, ⋯gt, 'HET'))─┬─sumIf(sign, ⋯gt, 'HOM'))─┐
1. │ R0004_non_analyst_project │  21 │ WGS         │ A        │                        0 │                        1 │                        0 │
2. │ R0001_1kg                 │  11 │ WES         │ N        │                        0 │                        0 │                        0 │
3. │ R0001_1kg                 │  11 │ WES         │ A        │                        0 │                        0 │                        1 │
   └───────────────────────────┴─────┴─────────────┴──────────┴──────────────────────────┴──────────────────────────┴──────────────────────────┘
```

I tried refreshing the view but its apparently no refreshable? Any ideas?

bpblanken: I'd make sure you're querying `test_seqr`.`GRCh37/SNV_INDEL/project_gt_stats`.  The view itself is just a query definition; it look like you can query it but it might not be reliable.  

There also _might_ be eventual consistency issues with the insert into entries followed by an immediate query of the `project_gt_stats` table.  I've mostly seen weirdness when inserting multiple rows with the some ORDER BY values, but that doesn't look to be the case here...

hanars: to clarify, the values in `gt_stats_dict` for this variant are wrong - they are 0 when they should not be. I was showing here that the source of this error is somehow in the construction of the `entries_to_project_gt_stats_mv` view, as the rows in the view do not match the rows if you run the query thats supposed to back the view. I am not sure what to do to force this view to be accurate, but there is probably no way for `project_gt_stats` to be accurate if the backing view for it is incorrect

hanars: The unit test failure I think is also related to this issue - when we delete all the data for a project we run a whole [bunch of refreshes ](https://github.com/broadinstitute/seqr/blob/master/clickhouse_search/search.py#L665-L676) and then at the end what we expect is that there is an updated affected count but what we are getting in this test is 0 for all the affected counts

hanars: Seems like the issue is with `project_gt_stats` itself
```

SELECT *
FROM test_seqr.`GRCh37/SNV_INDEL/project_gt_stats`

Query id: 9771b915-0d94-4d6e-8a33-f6eb55df4d6a

   ┌─project_guid──────────────┬─key─┬─sample_type─┬─affected─┬─ref_samples─┬─het_samples─┬─hom_samples─┐
1. │ R0004_non_analyst_project │  21 │ WGS         │ A        │           0 │           1 │           0 │
2. │ R0001_1kg                 │  11 │ WES         │ N        │           0 │           0 │           1 │
   └───────────────────────────┴─────┴─────────────┴──────────┴─────────────┴─────────────┴─────────────┘
```

Also FWIW this is the entries table - `HG00731` is affected and `HG00733` is unaffected
```
SELECT *
FROM test_seqr.`GRCh37/SNV_INDEL/entries`

Query id: 03c8291f-da36-415b-b5e6-2a29cebc886f

   ┌─key─┬─project_guid──────────────┬─family_guid─┬─sample_type─┬───────xpos─┬─is_gnomad_gt_5_percent─┬─is_annotated_in_any_gene─┬─geneId_ids─┬─filters────────────────────────┬─calls────────────────────────────────────────────────────┬─sign─┐
1. │  21 │ R0004_non_analyst_project │ F000014_14  │ WGS         │ 1000010439 │ true                   │ false                    │ []         │ []                             │ [('NA21234','HET',87,0.531,27)]                          │    1 │
2. │  11 │ R0001_1kg                 │ F000002_2   │ WES         │ 7143270172 │ true                   │ true                     │ [73,74]    │ ['VQSRTrancheSNP99.90to99.95'] │ [('HG00733',NULL,0,NULL,NULL),('HG00731','HOM',48,1,16)] │    1 │
   └─────┴───────────────────────────┴─────────────┴─────────────┴────────────┴────────────────────────┴──────────────────────────┴────────────┴────────────────────────────────┴──────────────────────────────────────────────────────────┴──────┘
```

hanars: @bpblanken merging in the changes you made to the order by has fixed this issue for all the search test and I am seeing reasonable looking affected coutns for all variants

However, the test for the gt stats after deleting a project are still [failing](https://github.com/broadinstitute/seqr/actions/runs/19642342781/job/56248881272?pr=5142) here. Locally it is passing for me with values for the affected counts, but on the github test runner its failing because they are getting affected counts of 0 for all variants. 
The actual SQL we run when dropping a project is here: https://github.com/broadinstitute/seqr/blob/master/clickhouse_search/search.py#L665-L676
Is it possible theres an additional statement we should be running there to ensure the affected counts are correct? Or do we need to be doing something involving inserts instead of wholesale dropping partitions?

bpblanken: My initial thought is to try with the approach we made in the search fixtures [here](https://github.com/broadinstitute/seqr/blob/master/clickhouse_search/search_tests.py#L40-L55).  The dictionary needs to be reloaded (from postgres -> clickhouse) _before_ the 'clickhouse_search' runs, as the functionality that aggregates runs as rows are inserted into the entries table.

hanars: okay so adding the refresh before the test run changes it so that in my local tests I am also getting an affected ac of 0 for everything. It appears that after refreshing it everything is showing as unaffected

```
SELECT
    project_guid,
    key,
    sample_type,
    dictGetOrDefault(test_seqr.seqrdb_affected_status_dict, 'affected', (family_guid, calls.sampleId), 'U') AS affected,
    calls.sampleId
FROM test_seqr.`GRCh38/SNV_INDEL/entries`
ARRAY JOIN calls

Query id: babb71c1-8fbb-4589-bbf8-0402680226b3

   ┌─project_guid──────────────┬─key─┬─sample_type─┬─affected─┬─calls.sampleId─┐
1. │ R0003_test                │   1 │ WES         │ U        │ NA20885        │
2. │ R0003_test                │   5 │ WES         │ U        │ NA20885        │
3. │ R0003_test                │   2 │ WES         │ U        │ NA20885        │
4. │ R0003_test                │   1 │ WGS         │ U        │ NA20885        │
5. │ R0003_test                │   5 │ WGS         │ U        │ NA20885        │
6. │ R0003_test                │   2 │ WGS         │ U        │ NA20885        │
7. │ R0004_non_analyst_project │  22 │ WGS         │ U        │ NA21234        │
8. │ R0004_non_analyst_project │  22 │ WGS         │ U        │ NA21987        │
9. │ R0004_non_analyst_project │  22 │ WGS         │ U        │ NA21654        │
   └───────────────────────────┴─────┴─────────────┴──────────┴────────────────┘
```

At the same point in my code as that executed, I ran a check on the postgres data and it looked correct:
```
Individual.objects.filter(individual_id='NA20885').values('affected')
<QuerySet [{'affected': 'A'}]>
```

bpblanken: Hrm.  Did you also move the test case to a `TransactionTestCase`?  I would check that you can view the affected status values in postgres not just from that point in the code, but from ClickHouse and also from a postgres connection created outside of the tests (with psql or a python terminal).  The affected status in clickhouse being `U` (I think) points to an issue accessing actual COMMITTED rows.

hanars: oh yeah thats it. Actually the easiest solution to this by far is just to move the test into the clickhouse search file, especially since all this endpoint does is call a clickhouse helper function thats defined in the clickhouse search.py file so its a reasonable place to test it

bpblanken: 👍 Cool!  glad it worked.  iirc we actually decided to minimize the scope of the `TranscationTestCase`, leaving this issue until now, which effectively footgun-ed us here.  It's hard to maintain things that require a human having to remember edge cases, so maybe worth additional thought.

PR #5145, opened 2025-11-20
Sample agg search cleanup

hanars:
Now that we are using the affected status dict for search filtering, for multi-project search we should not query all the sample data out of postgres before running the search and instead just query the aggregate data we actually need


clickhouse_search/search_tests.py
```diff
@@ -429,24 +429,6 @@ def test_inheritance_filter(self):
             ],
         )
 
-        self.results_model.families.set(Family.objects.filter(guid__in=['F000002_2', 'F000014_14']))
```
hanars, line 432: This test case relies on running a multi-project search with a custom affected filter (for the SV data). This is impossible in seqr (custom affected is only available for single family search) and this test is now failing. Fixing it is not worth it as its testing an unreachable edge case and everything in this part of the test is already covered early in the test function, so I am just removing it

PR #5174, opened 2025-12-08
feat: first batch of reference datasets

bpblanken:


.github/workflows/unit-tests.yml
```diff
@@ -77,6 +77,7 @@ jobs:
           docker exec clickhouse clickhouse-client --query "CREATE SETTINGS PROFILE clickhouse_settings SETTINGS flatten_nested=0, join_use_nulls=1 TO clickhouse_test, clickhouse_read_only"
           docker exec clickhouse clickhouse-client --query "GRANT SELECT, SYSTEM VIEWS, dictGet ON *.* TO clickhouse_read_only"
           docker exec clickhouse clickhouse-client --query "CREATE NAMED COLLECTION seqr_postgres_named_collection AS user='postgres', password='pgtest', host='postgres', port='5432'"
+          docker exec clickhouse clickhouse-client --query "CREATE NAMED COLLECTION pipeline_data_access AS access_key_id='123', secret_access_key='456'"
```
hanars: In our open source helm template, this collection only exists for users using gs:
https://github.com/broadinstitute/seqr-helm/blob/0891f02e41b0b3f129bb69aaee08d999782586d0/charts/seqr/templates/clickhouseconfig.yaml#L35

Will this be a problem for local users not using gs?

clickhouse_search/models.py
```diff
@@ -581,6 +581,310 @@ class ClinvarMito(BaseClinvarJoin):
...
+            primary_key=('key'),
+            order_by=('key'),
+        )
+
+class TopmedSeqrVariantsSnvIndel(BaseTopmed):
+    key = OneToOneField('AnnotationsSnvIndel', db_column='key', primary_key=True, on_delete=CASCADE)
+
+    class Meta:
+        db_table = 'GRCh38/SNV_INDEL/reference_data/topmed/seqr_variants'
+        engine = models.MergeTree(
```
hanars, line 745: Wouldn't we want the seqr tables to be rocksdb or join tables? Doing  a dozen joins on regular clickhouse tables feels like it would be prohibitively slow

bpblanken: The "seqr_variants" tables are the on-disk representation backing "search" entities.... dictionaries for almost all reference datasets and JOIN tables for Clinvar and HGMD because they are smaller and have enums.

hanars: 👌

.github/workflows/unit-tests.yml
bpblanken: Good question. It's only used right now in the hgmd table, as we need the credentials to query the private reference data bucket.  The logic there is like:

```
if os.environ.get('HGMD_URL'):
  hgmd_source = gcs(
                            pipeline_data_access,
                            url=os.environ.get('HGMD_URL'),
                            format='TSV',
                            structure='{HGMD_INFO_STRUCTURE}'
                        )
else:
  hgmd_source = null(structure='{HGMD_INFO_STRUCTURE}')
```

which makes everything flow nicely.  Maybe we want to lift the `gcs` and `pipeline_data_access` components into the environment variable?  


fwiw we can actually remove this line from the unit tests.yml... it was from an earlier version that approached the mocking differently.

hanars: I think since hgmd does not need to work for local installs its okay to leave the configuration/ env vars as is unless you feel strongly. I would say we should take this out of unit test set up if its not needed though

clickhouse_search/migrations/0025_screen_reference_data.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Original file sourced from `https://downloads.wenglab.org/V3/GRCh38-cCREs.bed`
+SCREEN_VIEW = Template("""
+$mv_header
+AS SELECT
+    replaceOne(c1, 'chr', '') as chrom,
+    toUInt32(assumeNotNull(c2)) as start,
+    toUInt32(assumeNotNull(c3)) as end,
+    splitByChar(',', assumeNotNull(c6))[1] as regionType
+FROM url('https://storage.googleapis.com/seqr-reference-data/clickhouse/GRCh38/screen/GRCh38-cCREs.bed')
```
bpblanken, line 19: Note that BITS did create seqr-reference-data.broadinstitute.org for us, so we can go change some of these!

hanars: Is there a reason not to do that as part of this PR?

clickhouse_search/migrations/0026_gnomad_noncoding_reference_data.py
```diff
@@ -0,0 +1,74 @@
...
+
+from clickhouse_search.migration_templates import ALL_VARIANTS_MV_HEADER, conditionally_refresh_reference_dataset, render_search_dictionary
+
+GNOMAD_NON_CODING_CONSTRAINT_VIEW = Template("""
+$mv_header
+AS SELECT
+    replaceOne(chrom, 'chr', '') as chrom,
+    toUInt32(assumeNotNull(start)) as start,
+    toUInt32(assumeNotNull(end)) as end,
+    z
```
hanars: is the z here on purose?

clickhouse_search/migrations/0031_splice_ai.py
```diff
@@ -0,0 +1,158 @@
...
+
+from clickhouse_search.migration_templates import ALL_TO_SEQR_MV, ALL_VARIANTS_MV_HEADER, conditionally_refresh_reference_dataset, render_search_dictionary
+
+SPLICE_AI_ALL_VARIANTS_MV = Template("""
+$mv_header
+AS SELECT
+    variant_id as variantId,
+    delta_score as score,
+    splice_consequence_id as consequence
+FROM gcs('https://storage.googleapis.com/seqr-reference-data/v3.1/$reference_genome/splice_ai/1.1.parquet/*.parquet')
```
hanars, line 17: out of curiosity, where if anywhere does the code that generated this parquet live?

clickhouse_search/models.py
```diff
@@ -581,6 +581,651 @@ class ClinvarMito(BaseClinvarJoin):
     class Meta(BaseClinvarJoin.Meta):
         db_table = 'GRCh38/MITO/reference_data/clinvar'
```
hanars, line 583: I think we should transition from a single large models.py file into a multi-file models package -
its very straightforward to do (https://docs.djangoproject.com/en/4.2/topics/db/models/#organizing-models-in-a-package) and I think it will make it a lot easier to maintain if we separate out the models into 2 or 3 files (I was thinking one for reference data and one for search models but am open to suggestions). I'm okay if we put that in its own PR or just do it now

clickhouse_search/migrations/0031_splice_ai.py
bpblanken: It's [this code](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_parquet.py).

clickhouse_search/migrations/0025_screen_reference_data.py
bpblanken: I'm working on it.  there's maybe an issue with the SSL-cert... but we should be able to do it on this PR.

clickhouse_search/migrations/0026_gnomad_noncoding_reference_data.py
bpblanken: Yes, the model for this one is:

```
class GnomadNonCodingConstraintAllVariantsSnvIndel(models.ClickhouseModel):
    chrom = Enum8Field(return_int=False, choices=BaseAnnotations.CHROMOSOME_CHOICES, primary_key=True)
    start = models.UInt32Field()
    end = models.UInt32Field()
    z = models.DecimalField(max_digits=9, decimal_places=5)
```

happy to change to `score` if we think it's more maintainable..

bpblanken: Yeah, it's `z` in the raw file; we rename to `z_score` in the pipeline, then just `gnomad_noncoding` in the export.

```

SELECT *
FROM url('https://storage.googleapis.com/gcp-public-data--gnomad/release/3.1/secondary_analyses/genomic_constraint/constraint_z_genome_1kb.qc.download.txt.gz')
LIMIT 1

Query id: c252d035-16ef-4949-8377-542c7e0ad866

   ┌─chrom─┬──start─┬────end─┬─element_id─────────┬─possible─┬───────────expected─┬─observed─┬─oe─────────────────┬─z──────────────────┐
1. │ chr1  │ 783000 │ 784000 │ chr1-783000-784000 │     2443 │ 159.09212499779832 │      142 │ 0.8925646068400002 │ 1.3551011656250318 │
   └───────┴────────┴────────┴────────────────────┴──────────┴────────────────────┴──────────┴────────────────────┴────────────────────┘

1 row in set. Elapsed: 5.378 sec.
```

bpblanken: I will rename (this and any others) to `score`.  It's more consistent.

clickhouse_search/migrations/0025_screen_reference_data.py
bpblanken: It's working now!  All the links are updated.

hanars: per offline discussion, it is not actually working and we are not changing it

hanars (approved):

clickhouse_search/migrations/0027_hgmd_reference_data.py
```diff
@@ -0,0 +1,220 @@
...
+            hints={"clickhouse": True},
+        ),
+        migrations.RunPython(
+            build_hgmd_all_variants_mv(
+                reference_genome="GRCh37",
+                hgmd_url=HGMD_GRCH37_URL,
+            ),
+        ),
+        migrations.RunPython(
+            build_hgmd_all_variants_mv(
```
hanars, line 209: don't we want to build the MVs after running `conditionally_refresh_reference_dataset`? I think this is an issue inseveral files

bpblanken: No, the `build` only creates the table/schema.  Refresh actually runs the "REFRESH VIEW" that materializes.

bpblanken: conditionally_refresh_reference_dataset should always be last, I will make sure that it is.

PR #5182, opened 2025-12-09
fix logic for grouping comp hets

hanars:
Fixes a bug Lynn identified while looking into the prioritized variants. The current logic for getting paired variants is to go variant by variant, check all the note and tag guids and find the correct sets within them. Rather than try to debug that logic I simplified it and changed it to go through the notes and tags and identify all the unique pairs

### DIFF
--- a/ui/shared/components/panel/variants/selectors.test.js
+++ b/ui/shared/components/panel/variants/selectors.test.js
@@ -37,19 +37,16 @@ test('getPairedSelectedSavedVariants', () => {
   const savedFamilyVariants = getPairedSelectedSavedVariants(
     STATE_WITH_2_FAMILIES, { match: { params:  { familyGuid: 'F011652_1' } } }
   )
-  expect(savedFamilyVariants.length).toEqual(3)
+  expect(savedFamilyVariants.length).toEqual(2)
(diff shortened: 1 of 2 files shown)


ui/shared/components/panel/variants/selectors.test.js
```diff
@@ -37,19 +37,16 @@ test('getPairedSelectedSavedVariants', () => {
   const savedFamilyVariants = getPairedSelectedSavedVariants(
     STATE_WITH_2_FAMILIES, { match: { params:  { familyGuid: 'F011652_1' } } }
   )
-  expect(savedFamilyVariants.length).toEqual(3)
+  expect(savedFamilyVariants.length).toEqual(2)
   expect(savedFamilyVariants[0].variantGuid).toEqual('SV0000004_116042722_r0390_1000')
   expect(savedFamilyVariants[1].variantGuid).toEqual('SV0000002_1248367227_r0390_100')
-  expect(savedFamilyVariants[2].variantGuid).toEqual('SV0000002_SV48367227_r0390_100')
```
hanars, line 43: in the test fixture data this variant is not associated with any tags or notes, and the new logic does not return it. Since There is no way to return saved variants from the server without tags or notes this difference is moot

PR #5195, opened 2025-12-16
Materialized view django model

hanars:
Creates a django model class for materialized views and moves all of our existing materialized views from templated `RunSql` definitions into these models. This makes it clearer what "tables" exist in the database and their relationship to one another just by looking at the models files and not needing to go back through old migrations to find definitions

### DIFF
--- a/clickhouse_search/backend/table_models.py
+++ b/clickhouse_search/backend/table_models.py
@@ -5,14 +5,25 @@ from django.db import connections
 MATERIALIZED_VIEW_META_FIELDS = ['to_table', 'source_table', 'source_sql', 'column_selects', 'refreshable']
 
 
-class MaterializedView(models.ClickhouseModel):
-
-    @classmethod
-    def refresh(cls):
-        if getattr(cls._meta, 'refreshable', False):
-            with connections['clickhouse_write'].cursor() as cursor:
-                cursor.execute(f'SYSTEM REFRESH VIEW "{cls._meta.db_table}"')
-                cursor.execute(f'SYSTEM WAIT VIEW "{cls._meta.db_table}"')
+class IncrementalMaterializedView(models.ClickhouseModel):
 
     class Meta:
         abstract = True
+
+
+class RefreshableMaterializedView(IncrementalMaterializedView):
+
+    @classmethod
+    def refresh(cls):
+        with connections['clickhouse_write'].cursor() as cursor:
+            cursor.execute(f'SYSTEM REFRESH VIEW "{cls._meta.db_table}"')
+            cursor.execute(f'SYSTEM WAIT VIEW "{cls._meta.db_table}"')
+
+    class Meta:
+        abstract = True
+
+
+class RefreshableMaterializedViewMeta:
+    refreshable = True
+    column_selects = {}
+    source_sql = ''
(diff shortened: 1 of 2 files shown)


clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -222,10 +222,22 @@
       "sorted_regulatory_feature_consequences": []
     }
 }, {
-    "model": "clickhouse_search.clinvarsnvindel",
-    "pk": 1,
+    "model": "clickhouse_search.keylookupsnvindel",
+    "pk": "1-10439-AC-A",
     "fields": {
-      "key": 1,
```
hanars, line 228: rather than directly insert fixture rows into the clinvar join table, insert into the all variants tables and rely on the materialized view to work correctly. Changing the data like this ensures that out unit tests will catch if those views stop working as intended

clickhouse_search/backend/base.py
```diff
@@ -7,7 +7,13 @@
 
 class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
 
+    def _is_materialzed_view(self, model):
```
bpblanken: `_is_materialzed_view` -> `_is_materialized_view`

clickhouse_search/backend/table_models.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import connections
+
+
+MATERIALIZED_VIEW_META_FIELDS = ['to_table', 'source_table', 'source_sql', 'column_selects', 'refreshable']
+
+
+class MaterializedView(models.ClickhouseModel):
+
+    @classmethod
+    def refresh(cls):
```
bpblanken, line 17: It might be cleaner and more explicit here to just define disjoint `IncrementalMaterializedView` and a `RefreshableMaterializedView`.  Clickhouse itself is not particularly clear about the distinctness but they work very differently.

clickhouse_search/backend/base.py
```diff
@@ -17,14 +23,40 @@ def table_sql(self, model):
...
         prev_quote_value = self.quote_value   # pylint: disable=access-member-before-definition
         if isinstance(engine, Join):
             self.quote_value = self.no_quote_value
         expression = super()._get_engine_expression(model, engine)
         self.quote_value = prev_quote_value
         return expression
 
+    def _get_materialized_view_engine_expression(self, model):
+        sql = f'TO {self._table_name(model._meta, model._meta.to_table)}'
+        if getattr(model._meta, 'refreshable', False):
```
bpblanken, line 56: One feature that was added in the new reference data views that wasn't in the existing ones was the `EMPTY` keyword, which prevents an immediate refresh when the view is created.  

It's less important for the gt_stats views (when they are created everything should be empty), but we should have support for it, and default to using it, going forwards.

bpblanken: [here](https://github.com/broadinstitute/seqr/blob/master/clickhouse_search/migration_templates.py#L24) is an example

hanars: I am going to keep this out of scope for now and when I move the reference data views to be defined as models I will extend the behavior

bpblanken (approved):

PR #5196, opened 2025-12-17
Dictionary django models

hanars:


clickhouse_search/models.py
```diff
@@ -1161,6 +1162,62 @@ class Meta(ProjectsToGtStatsMeta):
...
+    key = models.UInt32Field(primary_key=True)
+    ac_wgs = models.UInt32Field()
+    ac_affected = models.UInt32Field()
+    hom_wgs = models.UInt32Field()
+    hom_affected = models.UInt32Field()
+
+    class Meta(GtStatsDictMeta):
+        db_table = 'GRCh38/SV/gt_stats_dict'
+        source_table = 'GtStatsSv'
+        layout = f'FLAT(MAX_ARRAY_SIZE {int(5e6)})'
```
bpblanken: I think these can just be typed out.

clickhouse_search/backend/base.py
```diff
@@ -57,6 +64,26 @@ def _get_materialized_view_engine_expression(self, model):
             sql = 'REFRESH EVERY 10 YEAR ' + sql
         return sql
 
+    def _dictionary_sql(self, model):
+        original_sql_create_table = self.sql_create_table  # pylint: disable=access-member-before-definition
+        meta = model._meta
+        source_table = self._table_name(meta, meta.source_table)
+        self.sql_create_table = f"""
+        CREATE DICTIONARY %(table)s (%(definition)s) %(extra)s
+        SOURCE(CLICKHOUSE(USER {CLICKHOUSE_WRITER_USER} PASSWORD {CLICKHOUSE_WRITER_PASSWORD} TABLE {source_table}))
```
bpblanken: password needs to be [single quoted](https://github.com/broadinstitute/seqr/pull/5110/changes)!

clickhouse_search/search_tests.py
```diff
@@ -1715,15 +1715,41 @@ def test_trigger_delete_project(self):
...
 
-        updated_seqr_pops_by_key = dict(AnnotationsSnvIndel.objects.all().join_seqr_pop().values_list('key', 'seqrPop'))
+        seqr_pops = []
+        seqr_pop_fields = []
+        annotations_qs = AnnotationsSnvIndel.objects.all().join_seqr_pop()
+        annotations_qs._get_seqr_pop_expressions(seqr_pops, seqr_pop_fields)
+        annotations_qs = annotations_qs.annotate(seqr_pop=Tuple(*seqr_pops, output_field=NamedTupleField(seqr_pop_fields)))
+        updated_seqr_pops_by_key = dict(annotations_qs.values_list('key', 'seqr_pop'))
         self.assertDictEqual(updated_seqr_pops_by_key, {
-            1: (2, 2, 1, 1, 4, 2),
```
hanars, line 1720: these values did not actually change but the order of the tuple did. Since its not the tuple order thats important but rather whether the values map to the correct field names I changed this to test that the seqr populations as dictionaries are correct. This will also mean that future changes will be easier to assess for accuracy

PR #5217, opened 2026-01-07
misc pext/dbnsfp bugs.

bpblanken:
These work!

### DIFF
--- a/clickhouse_search/migrations/0033_dbnsfp.py
+++ b/clickhouse_search/migrations/0033_dbnsfp.py
@@ -35,7 +35,7 @@ SELECT
     CAST(`fathmm-XF_coding_score` as Nullable(Decimal(9, 5))) as fathmm,
     CAST(predictor_parse(`MPC_score`) as Nullable(Decimal(9, 5))) as mpc,
     CAST(predictor_parse(`MutPred2_score`) as Nullable(Decimal(9, 5))) as mut_pred,
-    if(predictor_parse(`MutationTaster_pred`) = '', NULL, `MutationTaster_pred`) as mut_taster,
+    if(predictor_parse(`MutationTaster_pred`) = '', NULL, predictor_parse(`MutationTaster_pred`)) as mut_taster,
     CAST(predictor_parse(`Polyphen2_HVAR_score`) as Nullable(Decimal(9, 5))) as polyphen,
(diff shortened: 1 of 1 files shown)


clickhouse_search/migrations/0024_pext_reference_data.py
```diff
@@ -16,9 +16,9 @@
         '-',
         splitByChar(':', assumeNotNull(locus))[2],
         '-',
-        JSONExtractArrayRaw(assumeNotNull(alleles))[1],
+        JSONExtract(assumeNotNull(alleles), 'Array(String)')[1],
```
bpblanken, line 19: alleles were coming out with additional double quotes.

clickhouse_search/migrations/0024_pext_reference_data.py
```diff
@@ -36,7 +36,7 @@ class Migration(migrations.Migration):
             name='PextAllVariantsSnvIndel',
             fields=[
                 ('variant_id', clickhouse_backend.models.StringField(db_column='variantId', primary_key=True, serialize=False)),
-                ('score', clickhouse_backend.models.DecimalField(blank=True, decimal_places=5, max_digits=9, null=True)),
+                ('score', clickhouse_backend.models.DecimalField(decimal_places=5, max_digits=9)),
```
bpblanken, line 39: `score` should not be nullable.

clickhouse_search/migrations/0033_dbnsfp.py
```diff
@@ -66,7 +66,7 @@
     CAST(`fathmm-XF_coding_score` as Nullable(Decimal(9, 5))) as fathmm,
     CAST(predictor_parse(`MPC_score`) as Nullable(Decimal(9, 5))) as mpc,
     CAST(predictor_parse(`MutPred2_score`) as Nullable(Decimal(9, 5))) as mut_pred,
-    predictor_parse(`MutationTaster_pred`) as mut_taster,
+    if(predictor_parse(`MutationTaster_pred`) = '', NULL, predictor_parse(`MutationTaster_pred`)) as mut_taster,
```
bpblanken, line 69: interestingly, the empty string '' is the default NULL value for the `arrayFirst(Array(String))` function, but does not actually coalesce to NULL downstream.  So we need an additional handler.

hanars (approved):

PR #5228, opened 2026-01-12
geneIds bugfix

bpblanken:


clickhouse_search/migrations/0036_geneIds_bugfix.py
```diff
@@ -0,0 +1,120 @@
...
+                if row is None:
+                    n_partitions = 1
+                else:
+                    n_partitions = row[0]
+                for n in range(0, n_partitions):
+                    cursor.execute(f"OPTIMIZE TABLE `GRCh38/SNV_INDEL/entries` PARTITION ('{project_guid}', {n}) FINAL ASYNC")
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('clickhouse_search', '0035_promoterai'),
```
hanars: this should be updated so it is against the latest migration that was merged today

hanars (approved):

PR #5231, opened 2026-01-14
use hgmd reference table

hanars:


clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -42,7 +42,7 @@
       "caid": null,
       "lifted_over_chrom": "1",
       "lifted_over_pos": 11854476,
-      "hgmd": ["CM981315", 3],
+      "hgmd": ["CM981315", 1],
```
hanars, line 45: changing the value on the annotations table ensures that unit tests will fail if we end up using the wrong source

clickhouse_search/models/reference_data_models.py
```diff
@@ -479,14 +479,14 @@ class Meta:
         )
 
 class HgmdGRCh37SnvIndel(BaseHgmd):
-    key = ForeignKey('EntriesGRCh37SnvIndel', db_column='key', related_name='hgmd_join', primary_key=True, on_delete=PROTECT)
+    key = ForeignKey('AnnotationsGRCh37SnvIndel', db_column='key', related_name='hgmd_join', primary_key=True, on_delete=PROTECT)
```
hanars, line 482: Since hgmd is used as an or with the transcripts filters so it can not be used to reduce the number of entries it makes more sense to join it directly to the annotations table

PR #5234, opened 2026-01-16
Dict get expression output

hanars:
Moves the logic for determining the fields for a dict get into the dict class itself. When starting to work on the referece data dicts i realized their was going to be alot of copy paste and code duplication around this

### DIFF
--- a/clickhouse_search/managers.py
+++ b/clickhouse_search/managers.py
@@ -188,7 +188,8 @@ class AnnotationsQuerySet(SearchQuerySet):
         if len(pop_expressions) == 1:
             pop_expressions = [(subfield, pop_expressions[0][1])]
         elif len(pop_expressions) > 1:
-            pop_expressions = pop_expressions + [(subfield, Plus(*[expr for _, expr in pop_expressions]))]
+            combined = Plus(*[expr for _, expr in pop_expressions])
+            pop_expressions.append((subfield, combined))
         if subfield_name and subfield_name != subfield:
             pop_expressions = [(name.replace(subfield, subfield_name), expr) for name, expr in pop_expressions]
         return pop_expressions


clickhouse_search/managers.py
```diff
@@ -104,34 +104,10 @@ def gt_stats_dict(self):
...
-        if not seqr_pop_fields:
+        if self.gt_stats_dict is None:
             return results
-        return results.annotate(seqrPop=self.gt_stats_dict.dict_get_expression(
-            'key',
-            seqr_pop_fields,
-            output_field=models.TupleField([models.UInt32Field() for _ in seqr_pop_fields])
-        ))
-
-    def _seqr_subfield_pop_expressions(self, subfield, subfield_name=None, affected=False):
```
hanars, line 123: this was moved without much changed to the annotation queryset as its not longer needed for entries

clickhouse_search/managers.py
```diff
@@ -199,6 +175,20 @@ def _get_seqr_pop_expressions(self, seqr_pops, population_fields):
 
         return seqr_pops
 
+    def _seqr_subfield_pop_expressions(self, subfield, subfield_name=None, affected=False):
+        seqr_pop_fields = [field_name for field_name, _ in self.gt_stats_dict.base_fields()]
```
hanars, line 183: the only line that changed in this function is this, using the new helper function from the dict model

clickhouse_search/search_tests.py
```diff
@@ -1717,41 +1715,16 @@ def test_trigger_delete_project(self):
...
-            },
-            6: {
-                'seqr': {'ac': 0, 'ac_wes': 0, 'ac_wgs': 0, 'hom': 0, 'hom_wes': 0, 'hom_wgs': 0},
-                'seqr_affected': {'ac': 0, 'hom': 0},
-            },
-            22: {
-                'seqr': {'ac': 3, 'ac_wes': 0, 'ac_wgs': 3, 'hom': 1, 'hom_wes': 0, 'hom_wgs': 1},
-                'seqr_affected': {'ac': 3, 'hom': 1},
-            },
+            1: {'ac_wes': 2, 'ac_wgs': 2, 'hom_wes': 1, 'hom_wgs': 1, 'ac_affected': 4, 'hom_affected': 2},
```
hanars, line 1721: this data did not change but with the new utility function the `seqrPop` expression has a named tuple for the output field so django returns a dictionary instead of the plain tuple it was returning before, so given that the output is now clear on its own theirs no need for an additional transform

clickhouse_search/managers.py
```diff
@@ -203,6 +179,20 @@ def _get_seqr_pop_expressions(self, seqr_pops, population_fields):
...
+        pop_expressions = [
+            (field, F(f'seqrPop__{seqr_pop_fields.index(field)}')) for field in seqr_pop_fields
+            if field.startswith(subfield) and (field.endswith('affected') == affected)
+        ]
+        if len(pop_expressions) == 1:
+            pop_expressions = [(subfield, pop_expressions[0][1])]
+        elif len(pop_expressions) > 1:
+            pop_expressions = pop_expressions + [(subfield, Plus(*[expr for _, expr in pop_expressions]))]
```
bpblanken: might be cleaner as

```
combined = Plus(*(expr for _, expr in pop_expressions))
pop_expressions.append((subfield, combined))
```

PR #5236, opened 2026-01-20
Ref data predicitions from dicts

hanars:


clickhouse_search/fixtures/clickhouse_search.json
```diff
@@ -16,7 +16,7 @@
       "lifted_over_pos": 10439,
       "hgmd": [null, null],
       "screen_region_type": null,
-      "predictions": [4.668000221252441, null, null, null, null, null, null, null, null, null, null, null, null, null],
+      "predictions": [1, null, null, null, null, null, null, null, null, null, null, null, null, null],
```
hanars, line 19: I changed the values of the annotations table predictions so we can't accidentally introduce a regression where we use these instead of the dict values. IRL these would be the same as whats in the MVs and dicts

clickhouse_search/models/reference_data_models.py
```diff
@@ -1919,6 +1922,7 @@ class Meta(ReferenceDataMvMeta):
         db_table = 'GRCh37/SNV_INDEL/reference_data/eigen/all_variants_to_seqr_variants_mv'
         to_table = 'EigenSeqrVariantsGRCh37SnvIndel'
         source_table = 'EigenAllVariantsGRCh37SnvIndel'
+        source_sql = _all_variants_to_seqr_source_sql('GRCh37', 'SNV_INDEL')
```
hanars, line 1925: this was a bug I introduced when I did the move to using classes for MVs, the eigen seqr variants table for GRCh37 was filtered based on the GRCh38 key table. I have added a migration to drop and rebuild this view as well as to reload the downstream dictionary.

clickhouse_search/managers.py
```diff
@@ -513,19 +533,20 @@ def _clinvar_star_q(min_stars):
...
 
-    def _parse_in_silico_qs(self, in_silico, require_score):
-        in_silico_qs, in_silico_missing_qs = super()._parse_in_silico_qs(in_silico, require_score)
+    def _parse_in_silico_qs(self, results, in_silico, require_score, in_silico_qs, in_silico_missing_qs):
+        in_silico_q = None
+        missing_q = None
 
-        if (in_silico or {}).get('alphamissense') and 'alphamissensePathogenicity' in self.sorted_transcript_consequence_fields:
-            in_silico_qs.append(Q(sorted_transcript_consequences__array_exists={
+        if self.has_annotation('pass_in_silico'):
```
hanars, line 539: in general in silico filtering is now done at the entries table stage not the annotations table stage, the exception is `alphamissense` which is a transcript-specific score

clickhouse_search/managers.py
```diff
@@ -755,6 +781,15 @@ def _interval_query(self, chrom, start, end, *args, **kwargs):
             contains_q  &= Q(end_chrom__isnull=True)
         return start_range_q | end_range_q | contains_q
 
+    def _parse_in_silico_qs(self, results, in_silico, require_score, in_silico_qs, in_silico_missing_qs):
+        for score, _ in self.model.PREDICTION_FIELDS:
```
hanars, line 784: keep the old method fo prediction filtering for SVs

seqr/management/commands/tag_seqr_prioritized_variants.py
```diff
@@ -580,7 +580,7 @@ def _run_search(cls, search_name, config_search, family_variant_data, family_gui
     @classmethod
     def _run_comp_het_search(cls, search_name, config_search, family_variant_data, family_guid_map, dataset_type, sample_data, samples, **kwargs):
         queryset = get_data_type_comp_het_results_queryset(
-            GENOME_VERSION_GRCh38, dataset_type, sample_data, **kwargs,
```
hanars, line 583: the changes to this file are not related to the prediction change, but due to I guess an unrelated change in the order variants come out of the database it did reveal a bug that already existed with this code:
VARIANT3 and VARIANT4 both match on both the primary and secondary annotations for the recessive search, however once you apply the MANE transcript restrictions VARIANT4 only matches on secondary, not primary. Because of how the db query is written, the variant "pair" that comes out of the database is sorted so the variant that matched the primary criteria is first and the secondary criteria is second. This means what actually come out is both the pair [VARIANT3, VARIANT4] and [VARIANT4, VARIANT3], but we arbitrarily deduplicate so we only get back one of those. Then in the MANE filtering step this order becomes crucial because one ordering is valid but one is not. To remedy this I changed it so we no longer arbitrarily deduplicate and instead return all valid orderings, filter for MANE, and then deduplicate

clickhouse_search/migrations/0040_gnomadnoncodingconstraintdict.py
```diff
@@ -0,0 +1,81 @@
...
+            managers=[
+                ('objects', django.db.models.manager.Manager()),
+                ('_overwrite_base_manager', django.db.models.manager.Manager()),
+            ],
+        ),
+        migrations.RunSQL(
+            'SYSTEM RELOAD DICTIONARY "GRCh38/SNV_INDEL/reference_data/gnomad_non_coding_constraint"',
+            hints={'clickhouse': True},
+        ),
+        migrations.RunSQL(
```
hanars, line 65: for the eigen view the all variants table is fine tis just the MV to join with the seqr variants that was an issue, so we can just refresh this view and the dictionary once it has been rebuilt, not need to trigger a full dataset reload

PR #5238, opened 2026-01-20
eigen

bpblanken:


clickhouse_search/migrations/0034_absplice2.py
```diff
@@ -33,7 +33,7 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='Absplice2SeqrVariants',
             fields=[
-                ('key', models.OneToOneField(db_column='key', on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='clickhouse_search.annotationsmito')),
+                ('key', models.OneToOneField(db_column='key', on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='clickhouse_search.annotationssnvindel')),
```
bpblanken, line 36: noticed these were wrong.

clickhouse_search/migrations/0038_eigen.py
```diff
@@ -0,0 +1,191 @@
...
+    dependencies = [
+        ('clickhouse_search', '0037_geneIds_bugfix'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='EigenAllVariantsGRCh37SnvIndel',
+            fields=[
+                ('score', clickhouse_backend.models.DecimalField(decimal_places=5, max_digits=9)),
+                ('variant_id', clickhouse_backend.models.StringField(db_column='variantId', primary_key=True, serialize=False)),
```
hanars, line 20: does the order for these columns matter? The order is the same for `EigenAllVariantsSnvIndel` as well, and for the seqr variants tables the key column is also second

bpblanken: yes, it does.  I will fix.  

(this is actually driving me crazy.. I fix the order and then re-run `makemigrations` for whatever reason and then forget)

PR #5241, opened 2026-01-20
feat: add variants tables

bpblanken:


clickhouse_search/migrations/0039_variantsdiskgcnv_variantsdiskgrch37snvindel_and_more.py
```diff
@@ -0,0 +1,279 @@
...
+    dependencies = [
+        ('clickhouse_search', '0038_eigen'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='VariantsDiskGRCh37SnvIndel',
+            fields=[
+                ('key', clickhouse_search.backend.fields.UInt32FieldDeltaCodecField(primary_key=True, serialize=False)),
+                ('variant_id', clickhouse_backend.models.StringField(db_column='variantId')),
```
bpblanken: I dropped `chrom`, `pos`, `xpos`, `ref`, and `alt`.  This is ambitious, but in theory those fields are recoverable from variantId?

clickhouse_search/migrations/0039_variants_tables.py
```diff
@@ -0,0 +1,279 @@
...
+            managers=[
+                ('objects', django.db.models.manager.Manager()),
+                ('_overwrite_base_manager', django.db.models.manager.Manager()),
+            ],
+        ),
+        migrations.CreateModel(
+            name='VariantDetailsGRCh37SnvIndel',
+            fields=[
+                ('key', models.OneToOneField(db_column='key', on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='clickhouse_search.variantsgrch37snvindel')),
+                ('rsid', clickhouse_backend.models.StringField(blank=True, null=True)),
```
bpblanken: I moved both `rsid` and `CAID` onto variant details.

clickhouse_search/migrations/0039_variantsdiskgcnv_variantsdiskgrch37snvindel_and_more.py
hanars: arguably we could also drop variant ID here and get it from the key lookup?

clickhouse_search/models/search_models.py
```diff
@@ -41,6 +41,42 @@ class BaseAnnotations(FixtureLoadableClickhouseModel):
     class Meta:
         abstract = True
 
+class BaseVariants(FixtureLoadableClickhouseModel):
+    ANNOTATION_CONSTANTS = {
+        'genomeVersion': GENOME_VERSION_GRCh38,
+        'liftedOverGenomeVersion': GENOME_VERSION_GRCh37,
+    }
+    SV_TYPE_FILTER_PREFIX = ''
+    GENOTYPE_OVERRIDE_FIELDS = {}
```
hanars: I would prefer not to port over any constants from the annotations models, I am going to have to refactor the managers pretty heavily when I change over to use the new tables and I would rather move constants as needed

clickhouse_search/models/search_models.py
```diff
@@ -41,6 +41,42 @@ class BaseAnnotations(FixtureLoadableClickhouseModel):
...
+            ('lofFilters', models.ArrayField(models.StringField(null=True, blank=True))),
+        ], null_empty_arrays=True)),
+        ('majorConsequence', models.Enum8Field(null=True, blank=True, return_int=False, choices=CONSEQUENCE_TERMS)),
+        ('transcriptId', models.StringField()),
+        ('transcriptRank', models.UInt8Field()),
+    ]
+
+    key = UInt32FieldDeltaCodecField(primary_key=True)
+    variant_id = models.StringField(db_column='variantId')
+    lifted_over_pos = models.UInt32Field(db_column='liftedOverPos', null=True, blank=True)
```
hanars: this is not needed at search time so I think it would actually belong on the variant detail table instead

clickhouse_search/models/search_models.py
```diff
@@ -41,6 +41,42 @@ class BaseAnnotations(FixtureLoadableClickhouseModel):
...
+
+    objects = AnnotationsQuerySet.as_manager()
```
hanars: I would rather add a new manager to these tables when I actually start using them, as right now this is just going to merge conflict with changes I am already making to the manager classes

clickhouse_search/models/search_models.py
```diff
@@ -94,6 +130,32 @@ class BaseAnnotationsSvGcnv(BaseAnnotations):
     class Meta:
         abstract = True
 
+class BaseVariantsSvGcnv(BaseVariants):
```
hanars: This looks very similar to `BaseAnnotationsSvGcnv` - maybe have this inherit from that to make it clear whet, if anything changed?

clickhouse_search/models/search_models.py
```diff
@@ -170,6 +232,23 @@ class BaseAnnotationsGRCh37SnvIndel(BaseAnnotationsMitoSnvIndel):
     class Meta:
         abstract = True
 
+class BaseVariantsGRCh37SnvIndel(BaseVariants):
+    ANNOTATION_CONSTANTS = {
```
hanars: per above, I would prefer not to add manager-specific code to the new tables and would instead rather do that as needed when I switch over that tables

clickhouse_search/models/search_models.py
```diff
@@ -170,6 +232,23 @@ class BaseAnnotationsGRCh37SnvIndel(BaseAnnotationsMitoSnvIndel):
...
+        'genomeVersion': GENOME_VERSION_GRCh37,
+        'liftedOverGenomeVersion': GENOME_VERSION_GRCh38,
+    }
+    SORTED_TRANSCRIPT_CONSQUENCES_FIELDS = [
+        ('canonical', models.UInt8Field(null=True, blank=True)),
+        ('consequenceTerms', models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=BaseVariants.CONSEQUENCE_TERMS))),
+        ('geneId', models.StringField(null=True, blank=True))
+    ]
+
+    lifted_over_chrom = Enum8Field(db_column='liftedOverChrom', return_int=False, null=True, blank=True, choices=CHROMOSOME_CHOICES)
```
hanars: this is not needed for search so should go on the details table

clickhouse_search/models/search_models.py
```diff
@@ -211,6 +302,29 @@ class BaseAnnotationsSnvIndel(BaseAnnotationsGRCh37SnvIndel):
     class Meta:
         abstract = True
 
+class BaseVariantsSnvIndel(BaseVariantsGRCh37SnvIndel):
+    ANNOTATION_CONSTANTS = BaseVariants.ANNOTATION_CONSTANTS
+    SORTED_TRANSCRIPT_CONSQUENCES_FIELDS = sorted([
+        ('alphamissensePathogenicity', models.DecimalField(null=True, blank=True, max_digits=9, decimal_places=5)),
```
hanars, line 271: would it be very hard to add mane here? We do not need the full string, just a boolean indicating whether or not `maneSelect` is null

clickhouse_search/models/search_models.py
```diff
@@ -211,6 +302,29 @@ class BaseAnnotationsSnvIndel(BaseAnnotationsGRCh37SnvIndel):
...
+        ('extendedIntronicSpliceRegionVariant', models.BoolField(null=True, blank=True)),
+        ('fiveutrConsequence', models.Enum8Field(null=True, blank=True, return_int=False, choices=[(1, '5_prime_UTR_premature_start_codon_gain_variant'), (2, '5_prime_UTR_premature_start_codon_loss_variant'), (3, '5_prime_UTR_stop_codon_gain_variant'), (4, '5_prime_UTR_stop_codon_loss_variant'), (5, '5_prime_UTR_uORF_frameshift_variant')])),
+        *BaseVariantsGRCh37SnvIndel.SORTED_TRANSCRIPT_CONSQUENCES_FIELDS,
+    ])
+
+    sorted_transcript_consequences = NestedField(SORTED_TRANSCRIPT_CONSQUENCES_FIELDS, db_column='sortedTranscriptConsequences')
+    sorted_motif_feature_consequences = NestedField([
+        ('consequenceTerms', models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=[(0, 'TFBS_ablation'), (1, 'TFBS_amplification'), (2, 'TF_binding_site_variant'), (3, 'TFBS_fusion'), (4, 'TFBS_translocation')]))),
+        ('motifFeatureId', models.StringField(null=True, blank=True)),
```
hanars: not sure whether this would have a substantial enough performance/ space boost to be worth it, but for `sorted_motif_feature_consequences` and `sorted_regulatory_feature_consequences` we could put the full nested field on the details table and here just have the `consequenceTerms` enum as a top level field on this table as thats all thats needed at query time. So something like
```
sorted_motif_consequence_terms = models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=[...])))

```

clickhouse_search/migrations/0039_variantsdiskgcnv_variantsdiskgrch37snvindel_and_more.py
bpblanken: key lookup is inverted, keyed by `variantId`.  So we actually do need it here!

clickhouse_search/models/search_models.py
bpblanken: Ok, I can make these only columns!

clickhouse_search/models/search_models.py
bpblanken: It's identical.  I actually copied (rather than inherit) on purpose rather than inherited because I thought that we'd just whole-sale delete the models in the future.  I can inherit in this case as it probably makes more sense.

hanars: Yeah I was thinking that if we had 
```
class BaseAnnotationsSvGcnv:
    <stuff>

class BaseVariantsSvGcnv(BaseAnnotationsSvGcnv):
    pass
```

then in the future it would be very easy to change that to 
```
class BaseVariantsSvGcnv:
    <stuff>
```

hanars: oh or if we did the inheritance the other way around and had
```
class BaseAnnotationsSvGcnv(BaseVariantsSvGcnv):
    pass
```
then cleanup would be even easier

bpblanken:
```
class BaseVariantsSvGcnv(BaseAnnotationsSvGcnv):
    pass
```

is exactly what I did!

bpblanken: will push in a minute.

clickhouse_search/models/search_models.py
bpblanken: added `hasManeSelect` here.

clickhouse_search/models/search_models.py
bpblanken: I dropped everything but `consequenceTerms` (leaving it nested) from `Variants` and added the full set of fields onto the `VariantDetails` table.

clickhouse_search/migrations/0039_variantsdiskgcnv_variantsdiskgrch37snvindel_and_more.py
hanars: So we need variant ID for both display/returning to the client and also for filtering. However, since the list of variant IDs we are filtering to is generally quite small, I feel like it would be much faster to do a dict lookup for the allowed variantIds to get the keys and then filter the entries table to that set of keys rather than do the full entries table, join with the variant table and then filter by the variant ID strings. So given that we can move variant ID to the details table and not this variant table. Does that make sense?

clickhouse_search/models/search_models.py
hanars: nitpick, but it would be more accurate to say `isManeSelect`

clickhouse_search/models/search_models.py
```diff
@@ -358,6 +512,30 @@ class BaseAnnotationsGcnv(BaseAnnotationsSvGcnv):
...
+        ('sv_callset', NamedTupleField([
+            ('ac', models.UInt32Field()),
+            ('af', models.DecimalField(max_digits=9, decimal_places=8)),
+            ('an', models.UInt32Field()),
+            ('het', models.UInt32Field()),
+            ('hom', models.UInt32Field()),
+        ])),
+    ]
+    SV_TYPE_FILTER_PREFIX = 'gCNV_'
+    GENOTYPE_OVERRIDE_FIELDS = {
```
hanars: same as previous review, I would prefer not to put constantas on these new classes for the manager to use until its time to actually start using the new tables

clickhouse_search/models/search_models.py
```diff
@@ -211,6 +271,28 @@ class BaseAnnotationsSnvIndel(BaseAnnotationsGRCh37SnvIndel):
...
+    ])
+    SORTED_MOTIF_FEATURE_CONSEQUENCES_FIELDS = sorted([
+        ('consequenceTerms', models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=[(0, 'TFBS_ablation'), (1, 'TFBS_amplification'), (2, 'TF_binding_site_variant'), (3, 'TFBS_fusion'), (4, 'TFBS_translocation')]))),
+    ])
+    SORTED_REGULATORY_FEATURE_CONSEQUENCES_FIELDS = sorted([
+        ('consequenceTerms', models.ArrayField(models.Enum8Field(null=True, blank=True, return_int=False, choices=[(0, 'regulatory_region_ablation'), (1, 'regulatory_region_amplification'), (2, 'regulatory_region_variant'), (3, 'regulatory_region_fusion')]))),
+    ])
+
+    sorted_transcript_consequences = NestedField(SORTED_TRANSCRIPT_CONSQUENCES_FIELDS, db_column='sortedTranscriptConsequences')
+    sorted_motif_feature_consequences = NestedField(SORTED_MOTIF_FEATURE_CONSEQUENCES_FIELDS, db_column='sortedMotifFeatureConsequences', null_when_empty=True)
```
hanars, line 285: Is there a reason this needs to be a nested field with only one column instead of just a single top level field? Its clearer if its just a single array field, unless there is a compelling reason to have it like this? For `sorted_transcript_consequences` theres multiple columns we need which is why it makes sense for it to be nested

bpblanken: I left it like this because I thought it was _slightly_ easier in the loading code to keep it nested.  If you feel strongly though I can change it.

hanars: nah slightly easier for loading code is a good enough reason

PR #5248, opened 2026-01-23
draft: set eux

bpblanken:


deploy/docker/seqr/entrypoint.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -x
+set -eux
```
bpblanken, line 3: `u` unset variables are errors
`e` script exits on error

PR #5260, opened 2026-01-30
Variant clickhouse tables

hanars:
Refactors seqr to use the new variant table models. Should not be merged until ready to do a major breaking version release. Major changes include
- Adding a queryset manager for the new variant detail tables and refactor the managers to share behavior
- Small tweaks to search behavior itself, mostly around the absence of a variantId at search time and needing to use the xpos from the entries table
- refactors the `tag_seqr_prioritized_variants` command to take advantage of the availabilty of mane select at query time
-  updates to the helpers for bulk creating saved variants to correctly pull and format the fields needed from clcikhouse given that the variantId can now be on a different table

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## dev
 
+## 2/11/26
+* Adds eigen reference data to clickhouse (REQUIRES DB MIGRATION)
+  * Note that for users with Clickhouse already enabled, this migration will enqueue tasks in the loading pipeline. 
+  You may want to monitor the pipeline for any failures and re-trigger them in the event of any failures.
+* Updates VLM to the finalized v1 schema
+
 ## 1/8/26
 * Moves PanelApp updates from their own manage command into the main `update_all_reference_data` command.
 May require changes to custom cron jobs (REQUIRES DB MIGRATION)
--- a/clickhouse_search/fixtures/clickhouse_search.json
+++ b/clickhouse_search/fixtures/clickhouse_search.json
@@ -317,6 +317,24 @@
       "consequence": 4
     }
 }, {
+    "model": "clickhouse_search.pextallvariantssnvindel",
+    "pk": "1-38724419-T-G",
+    "fields": {
+      "score": 0.03
+    }
+}, {
+    "model": "clickhouse_search.promoteraiallvariants",
+    "pk": "1-38724419-T-G",
+    "fields": {
+      "score": 0.64
+    }
+}, {
+    "model": "clickhouse_search.absplice2allvariants",
+    "pk": "1-38724419-T-G",
+    "fields": {
+      "score": 0.11
+    }
+}, {
     "model": "clickhouse_search.gnomadnoncodingconstraintallvariantssnvindel",
     "pk": 1,
     "fields": {
@@ -1080,6 +1098,12 @@
     "pk": "M-11597-A-G",
     "fields": {
       "score": 0.75
+    }
+}, {
+    "model": "clickhouse_search.pextallvariantsmito",
+    "pk": "M-11597-A-G",
+    "fields": {
+      "score": 0.13
     }
 }, {
     "model": "clickhouse_search.localconstraintmitoallvariantsmito",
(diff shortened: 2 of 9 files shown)


clickhouse_search/managers.py
```diff
@@ -619,13 +701,37 @@ def _allowed_consequences_filters(self, allowed_consequences):
     def _consequence_term_filter(consequences, **kwargs):
         return {'consequenceTerms': (consequences, 'hasAny({value}, {field})'), **kwargs}
 
+    def _annotate_filtered_transcripts(self, results, consequence_field, transcript_filters, *args, require_mane_canonical=False, **kwargs):
+        if require_mane_canonical:
+            filtered_expr = ArrayFilter(consequence_field, conditions=[{'canonical': (0, '{field} > {value}')}])
+            if 'isManeSelect' in self.sorted_transcript_consequence_fields:
```
hanars, line 696: this is the only new behavior added in this PR - adds the `require_mane_canonical` flag that uses the new `isManeSelect` field so restricting to primary consequence is doen at query time instead of in post-processing

clickhouse_search/search.py
```diff
@@ -762,43 +759,96 @@ def get_clickhouse_variant_by_id(variant_id, parsed_variant_id, samples, genome_
...
     lookup = {}
-    fields = ('variant_id', 'key') if not reverse else ('key', 'variant_id')
+    fields = ('variant_id', 'key')
     for i in range(0, len(variants_ids), BATCH_SIZE):
         batch = variants_ids[i:i + BATCH_SIZE]
         lookup.update(dict(key_lookup_class.objects.filter(variant_id__in=batch).values_list(*fields)))
     return lookup
 
 
+def get_variant_main_transcripts_by_key(genome_version, dataset_type, selected_transcripts_by_key, include_clinvar=False, additional_values=None):
```
hanars, line 804: slightly different versions of this logic were implemented in 3 different places in the code, this moves the logic into a single shared helper

clickhouse_search/migrations/0039_variantsdiskgcnv_variantsdiskgrch37snvindel_and_more.py
```diff
@@ -283,4 +283,70 @@ class Migration(migrations.Migration):
                 ('_overwrite_base_manager', django.db.models.manager.Manager()),
             ],
         ),
+        migrations.RunSQL(
```
hanars, line 286: this is a no-op as the foreign keys are not actually enforced in clickhouse at the db level, but this just tells django that they have been updated

seqr/management/commands/tag_seqr_prioritized_variants.py
```diff
@@ -5,12 +5,8 @@
 from django.db.models import Q, F
 from django.db.models.functions import JSONObject
```
hanars, line 6: The logic for the command changed quite a bit, as previously we had to get the full results back from search and then post-process them to filter based on whether or not they pass their transcript filters with a valid mane transcript. Since mane select in now available on the main variants table this filtering is done at search time and no post-processing is needed. Therefore, rather than getting back the formatted results from search we get back just the minimal genotypes and keys. We then use the shared utility function for bulk saving variants to handle getting and formatting the fields we need for actually saving a variant model, as this allows us to encapsulate that functionality in one place instead of having multiple different implementations. A restriction of this is that we need to call the bulk update method on a per-dataset type bases instead of once for everything, but that should not have a real performance impact

seqr/views/utils/test_utils.py
```diff
@@ -624,6 +624,19 @@ def setUpTestData(cls):
         super().setUpTestData()
         cls.set_up_users()
 
+    @classmethod
+    def setUpClass(cls):
+        cls._clean_up_clickhouse_db()
+        super().setUpClass()
+
+    @classmethod
+    def _clean_up_clickhouse_db(cls):
```
hanars: To be clear, this exact TRUNCATE sql is run as part of every test class tear down, it juts doe snot always propogate, and this was the only way I could reliably get tests to pass

PR #5263, opened 2026-02-03
bugfix: use a node14 build container

bpblanken:


deploy/docker/seqr/Dockerfile
```diff
@@ -20,37 +40,19 @@ ENV VIRTUAL_ENV=/opt/venv
...
 
-COPY ./ui/ .
-
-RUN npm run build
-
-FROM python:3.11-slim-bullseye as runtime
+############################
+# Runtime stage
+############################
+FROM python:3.11-slim-bullseye AS runtime
```
bpblanken, line 54: hrm, turns out we actually also need ```

ENV VIRTUAL_ENV=/opt/venv
RUN python3 -m venv $VIRTUAL_ENV
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
``` here due to python3.9/python3.11 version conflicts, which is not great.

File: deploy/docker/seqr/Dockerfile [hunk 2]
```diff
@@ -1,16 +1,34 @@
-FROM python:3.11-slim-bullseye as build
+############################
+# UI build (Node.js 14 only)
+############################
+FROM node:14-bullseye AS ui-build
 
-LABEL maintainer="Broad TGG"
```
hanars, line 3: did we mean to remove this?

deploy/docker/seqr/Dockerfile
```diff
@@ -20,41 +38,27 @@ ENV VIRTUAL_ENV=/opt/venv
...
 
-FROM python:3.11-slim-bullseye as runtime
+# Copy Python virtualenv from build stage
+COPY --from=build /opt/venv /opt/venv
+ENV VIRTUAL_ENV=/opt/venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    cron \
```
hanars, line 57: 🎉

deploy/docker/seqr/Dockerfile
bpblanken: Yes, the removal was pseudo-intentional; the LABEL has to go under a `FROM` and I couldn't decide which one.   I put it back under the `node:14` one.

PR #5270, opened 2026-02-05
bug: parse VUS-mid to Uncertain Significance

bpblanken:


clickhouse_search/management/commands/reload_clinvar_all_variants.py
```diff
@@ -116,6 +116,9 @@ def parse_pathogenicity_and_assertions(classified_record_node: xml.etree.Element
     ).replace(
         'Likely pathogenic/Likely pathogenic',
         'Likely pathogenic'
+    ).replace(
```
bpblanken, line 119: Snippet in the XML:

```
    <Classifications>
      <GermlineClassification DateLastEvaluated="2026-01-22" NumberOfSubmissions="1" NumberOfSubmitters="1" DateCreated="2026-02-01" MostRecentSubmission="2026-02-01">
        <ReviewStatus>no assertion criteria provided</ReviewStatus>
        <Description>VUS-mid</Description>
        <Citation Type="general">
          <ID Source="PubMed">38484052</ID>
        </Citation>
```

This strategy _does not_ handle the case where there are conflicting pathogenicities and `VUS-mid` is listed as one of  several different classifications.  The handling there is actually more restrictive than we have when parsing the top level pathogenicity.

hanars: that seems fine, as we aren;t really expecting to keep seeing this as a consequence

PR #5272, opened 2026-02-06
squash migrations

hanars:
Since all exisitng installations will have to check out the latest seqr version before upgrading to the next major version, this is a good opportunity to squash the existing migrations. This allows us to remove a lot of the table alterations and instead cleanly create tables the way we want them

### DIFF
--- a/clickhouse_search/migrations/0042_delete_annotationsdiskgcnv_and_more.py
+++ b/clickhouse_search/migrations/0042_delete_annotationsdiskgcnv_and_more.py
@@ -16,7 +16,7 @@ def validate_data_migrated(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('clickhouse_search', '0041_delete_annotationsdiskgcnv_and_more'),
+        ('clickhouse_search', '0041_variant_details_path'),
     ]
 
     operations = [


clickhouse_search/migrations/0040_gnomadnoncodingconstraintdict.py
```diff
@@ -58,24 +4473,173 @@ class Migration(migrations.Migration):
...
+            },
+            managers=[
+                ('objects', django.db.models.manager.Manager()),
+                ('_overwrite_base_manager', django.db.models.manager.Manager()),
+            ],
         ),
-        migrations.RunSQL(
-            'SYSTEM RELOAD DICTIONARY "GRCh37/SNV_INDEL/reference_data/eigen"',
-            hints={'clickhouse': True},
+        migrations.RunPython(
```
hanars, line 4528: I was pretty sure we always want these to run on startup for a clean db so I left them in

clickhouse_search/migrations/0040_gnomadnoncodingconstraintdict.py
```diff
@@ -1,42 +1,4457 @@
...
+                'db_table': 'GRCh38/SNV_INDEL/entries',
+                'engine': clickhouse_search.backend.engines.CollapsingMergeTree('sign', deduplicate_merge_projection_mode='rebuild', index_granularity=8192, order_by=('project_guid', 'family_guid', 'sample_type', 'is_gnomad_gt_5_percent', 'is_annotated_in_any_gene', 'key'), partition_by='project_guid'),
+                'projection': clickhouse_search.models.search_models.Projection('xpos_projection', order_by='is_gnomad_gt_5_percent, is_annotated_in_any_gene, xpos'),
+            },
+            managers=[
+                ('objects', django.db.models.manager.Manager()),
+                ('_overwrite_base_manager', django.db.models.manager.Manager()),
+            ],
+        ),
+        migrations.RunSQL(
```
hanars: this is the only executed SQL I left in as it does not look like we actually encoded the default on the model itself and that feels out of scope for this change. Note that I fixed the bug associated with this default

PR #5276, opened 2026-02-08
feat: manage.py comand for updating caids

bpblanken:


clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,334 @@
...
+
+def build_url(
+    genome_version: Literal[GENOME_VERSION_GRCh37, GENOME_VERSION_GRCh38],
+) -> str:
+    login = CLINGEN_ALLELE_REGISTRY_LOGIN
+    password = CLINGEN_ALLELE_REGISTRY_PASSWORD
+    base_url = ALLELE_REGISTRY_URL.format(ALLELE_REGISTRY_GNOMAD_IDS[genome_version])
+
+    # Adapted from https://reg.clinicalgenome.org/doc/scripts/request_with_payload.py
+    identity = hashlib.sha1(f"{login}{password}".encode("utf-8")).hexdigest()  # noqa: S324
```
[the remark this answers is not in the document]
bpblanken: hash lib sha1 is required on the server end.

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+        for genome_version, variant_details_model, entries_model in [
+            (GENOME_VERSION_GRCh37, VariantDetailsGRCh37SnvIndel, EntriesGRCh37SnvIndel),
+            (GENOME_VERSION_GRCh38, VariantDetailsSnvIndel, EntriesSnvIndel),
+        ]:
+            version_obj = DataVersions.objects.filter(
+                data_model_name=f"{genome_version}/ClingenAlleleRegistry"
+            ).first()
+
+            if not version_obj:
+                max_key_id = entries_model.objects.aggregate(max_key=Max("key_id"))["max_key"] or 0
```
bpblanken, line 295: I set this up so that we'd not have to support a migration in any way.  The first run will set the current data version to the max of the entries table.

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+                qs = variant_details_model.objects.join_series(
+                    curr_key + 1,
+                    curr_key + 1 + batch_size,
+                )
+                variants = list(qs)
+                if not variants:
+                    break
+
+                try:
+                    max_key = register_caids(genome_version, variants)
```
bpblanken, line 317: The SQL generated here is very nice and exactly what we want:

```
# ALTER TABLE "GRCh38/SNV_INDEL/variants/details" UPDATE "CAID" = CASE WHEN ("key" = 4) THEN 'CA997563840' WHEN ("key" = 5) THEN 'CA997563845' WHEN ("key" = 6) THEN NULL ELSE NULL END::Nullable(String) WHERE "key" IN (4, 5, 6)
```

I have not, however, tested with ~10k variants.

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+                    variant_details_model.objects.using('clickhouse_write').bulk_update(variants, ["caid"])
+                except Exception:
+                    logger.exception(
+                        f"Failed in {genome_version}/ClingenAlleleRegistry curr_key: {curr_key}"
+                    )
+                    break
+
+                # Save current key on every iteration
```
bpblanken: Each successful batch will result in an increment to the data versions table.  If a batch fails, we will stop at that key and pick up on the next run.

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+        "#CHROM POS ID  REF ALT QUAL    FILTER  INFO",
+    ],
+}
+ALLELE_REGISTRY_URL = "https://reg.genome.network/alleles?file=vcf&fields=none+@id+genomicAlleles+externalRecords.{}.id"
+HTTP_REQUEST_TIMEOUT_S = 420
+VCF_DEFAULT_VALUE = "."
+
+
+@dataclasses.dataclass
+class AlleleRegistryError:
```
bpblanken, line 101: This is ported as is, we could maybe do it differently in the future but it works so I left it.

clickhouse_search/managers.py
```diff
@@ -887,6 +887,9 @@ def _filter_frequency(self, results, freqs=None, pathogenicity=None, **kwargs):
 
 
 class VariantDetailsQuerySet(VariantsQuerySet):
+    def join_series(self, min_: int, max_: int):
```
bpblanken: Just to clarify:

- SELECT FROM x WHERE key > 0 and key < 10000 does a full table scan.
- I struggled with getting the `generate_series` into Django because it's a generating function, and they don't appear to be fully supported.
- I couldn't get:
```
SELECT * FROM X
where key in (
   SELECT key from generate_series(0, 10000)
)
``` to perform either.

File: clickhouse_search/management/commands/register_caids.py [hunk 7]
```diff
@@ -0,0 +1,323 @@
...
+    EntriesGRCh37SnvIndel,
+    EntriesSnvIndel,
+    VariantDetailsGRCh37SnvIndel,
+    VariantDetailsSnvIndel,
+)
+from reference_data.models import DataVersions
+from reference_data.models import GENOME_VERSION_GRCh37, GENOME_VERSION_GRCh38
+from seqr.utils.communication_utils import safe_post_to_slack
+from settings import (
+    CLINGEN_ALLELE_REGISTRY_LOGIN,
```
hanars, line 24: if this involves moving env vars around in the helm charts we should make sure to document that change with the new release

clickhouse_search/managers.py
hanars: I would prefer we put this in the register caid itself instead of as a utility helper on the queryset - if in the future we need to share this behavior elsewhere we can always move it in but I do not think this is something we want to like use in most cases

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+                curr_key = max_key
+                version_obj.version = curr_key
+                version_obj.save()
+
+            if min_key != max_key:
+                slack_message = (
+                    f"Successfully called {genome_version}/ClingenAlleleRegistry "
+                    f"for variants {min_key} -> {max_key}."
+                )
+                safe_post_to_slack(
```
hanars, line 334: I do not think we want to send a slack per batch, I think we want to send one slack at the end with all the batches

clickhouse_search/management/commands/register_caids.py
hanars: wouldn't we want that saving inside the try block though? Otherwise we are sort of just doing an unnecessary save

clickhouse_search/management/commands/register_caids.py
```diff
@@ -0,0 +1,323 @@
...
+                    break
+
+                try:
+                    max_key = register_caids(genome_version, variants)
+                    variant_details_model.objects.using('clickhouse_write').bulk_update(variants, ["caid"])
+                except Exception:
+                    logger.exception(
+                        f"Failed in {genome_version}/ClingenAlleleRegistry curr_key: {curr_key}"
+                    )
+                    break
```
hanars: I think we want to raise a command error here so that the job will be detected as failed

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,218 @@
...
+            match=[
+                responses.matchers.query_param_matcher({
+                    "file": "vcf",
+                    "fields": "none @id genomicAlleles externalRecords.gnomAD_4.id",
+                }, strict_match=False),
+            ],
+            status=200,
+            json={'errorType': 'InternalServerError'}
+        )
+        call_command("register_caids", batch_size=5)
```
hanars: if the command completely fails we want it to raise an exception, not log an error but still be marked as succeeded

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,218 @@
...
+            "First unmappable variant:\n{'@id': 'http://reg.genome.network/allele/CA16716503', 'genomicAlleles': [{'chromosome': '1', 'coordinates': [{'allele': 'C', 'end': 10131, 'referenceAllele': '', 'start': 10131}], 'referenceGenome': 'GRCh38'}]}"
+        )
+        mock_logger.warning.assert_called_with(
+            '1 failed. First error: \n'
+            'TYPE: InternalServerError\n'
+            'DESCRIPTION: Given allele cannot be mapped in consistent way to reference genome.\n'
+            'MESSAGE: 1\t10469\trs370233998\tC\tG\t.\t.\t.\n'
+            'INPUT_LINE: Cannot align NC_000001.10 [10468,10469).',
+        )
+        vd = VariantDetailsSnvIndel.objects.filter(variant_id='1-91511686-T-G').first()
```
hanars: instead of `.filter().first()` you can use `.get()`

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,218 @@
...
+            },
+        ],
+    },
+]
+
+@mock.patch('clickhouse_search.management.commands.register_caids.logger')
+@mock.patch("clickhouse_search.management.commands.register_caids.safe_post_to_slack")
+class RegisterCaidsTest(TestCase):
+    databases = "__all__"
+    fixtures = ["variant_details_for_update"]
```
hanars, line 118: we should also have a test class in this file with empty fixtures so we can test that the initial version setting and loading flow works as expected on an empty database

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+        mock_logger.warning.assert_called_with(
+            '1 failed. First error: \n'
+            'TYPE: InternalServerError\n'
+            'DESCRIPTION: Given allele cannot be mapped in consistent way to reference genome.\n'
+            'MESSAGE: 1\t10469\trs370233998\tC\tG\t.\t.\t.\n'
+            'INPUT_LINE: Cannot align NC_000001.10 [10468,10469).',
+        )
+        vd = VariantDetailsSnvIndel.objects.filter(variant_id='1-91511686-T-G').first()
+        self.assertEqual(vd.caid, 'CA997563840')
```
hanars, line 238: I would add another `call_command` at the end of this test so we can test the behavior when the cron job is run but theres nothing new to register

clickhouse_search/management/commands/register_caids.py
bpblanken: 👍 I got [these](https://github.com/broadinstitute/seqr-private/pull/1733/changes#diff-8ea7e761f4431f97008201a35a24b3f9cf0d041f0689fcb2e61063f8d43d9517R52-R69)

clickhouse_search/management/commands/register_caids.py
bpblanken: 👍

clickhouse_search/management/commands/register_caids.py
bpblanken: The `break` would prevent the save but I think it's clearer to move to the try block for sure.

clickhouse_search/management/commands/register_caids.py
bpblanken: Yep, this block is outside of the while loop and should send at most one per run.  `min_key` will be the initial key and `max_key` would be the last seen one.

clickhouse_search/managers.py
bpblanken: 👍 makes sense!

bpblanken:
Ready for re-review!

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,257 @@
...
+                }, strict_match=False),
+            ],
+            status=200,
+            json=MOCK_RESPONSE
+        )
+        call_command("register_caids", batch_size=3)
+        mock_logger.info.assert_called_with(
+            'Attempting to register caids from key: 0'
+        )
+        mock_logger.warning.assert_not_called()
```
hanars, line 107: can we also test that the `DataVersions` object is created as expected and also that `responses` captured the expected API calls ( I would assume that would be no calls, but if I am wrong about that we would want to be explicit)

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,257 @@
...
+
+@mock.patch('clickhouse_search.management.commands.register_caids.logger')
+@mock.patch("clickhouse_search.management.commands.register_caids.safe_post_to_slack")
+class RegisterCaidsEmptyDatabaseTest(TestCase):
+    databases = "__all__"
+    fixtures = []
+
+    @responses.activate
+    def test_register_caids(self, mock_safe_post_to_slack, mock_logger):
+        responses.add(
```
hanars: if the database is empty why would we need to be calling api at all? If we do not expect to call it its better not to add any matchers

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,257 @@
...
+        mock_logger.reset_mock()
+
+        responses.add(
+            responses.PUT,
+            "https://reg.genome.network/alleles",
+            match=[
+                responses.matchers.query_param_matcher({
+                    "file": "vcf",
+                    "fields": "none @id genomicAlleles externalRecords.gnomAD_4.id",
+                }, strict_match=False),
```
hanars, line 149: why not use `strict_match`?

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,257 @@
...
+                    [
+                        *ALLELE_REGISTRY_HEADERS['38'],
+                        '7\t143270172\t.\tA\tG\t.\t.\t.',
+                        '7\t9310123\t.\tT\tC\t.\t.\t.',
+                    ]
+                ) + "\n",
+            ],
+        )
+        mock_safe_post_to_slack.assert_has_calls([
+            mock.call(SEQR_SLACK_DATA_ALERTS_NOTIFICATION_CHANNEL, "Successfully called 38/ClingenAlleleRegistry for variants 3 -> 10."),
```
hanars: This channel is used by the PMs and analysts to track data updates in seqr, and I think this message would be a bit confusing for them. Maybe have this message logged so engineers reading job logs can get this level of granularity, but for the slack message lets go with something like "Registered Clingen Allele IDs for 7 recently loaded GRCh38 variants"

clickhouse_search/management/tests/register_caids_tests.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # Ensure re-calling is a no-op
+        mock_safe_post_to_slack.reset_mock()
+        mock_logger.reset_mock()
+        call_command("register_caids", batch_size=3)
+        mock_logger.info.assert_called_with(
+            'Attempting to register caids from key: 10',
+        )
+        mock_logger.warning.assert_not_called()
+        mock_safe_post_to_slack.assert_not_called()
```
hanars, line 249: We should also test that the DataVersion is unchanged and no new API calls were made

clickhouse_search/management/tests/register_caids_tests.py
bpblanken: I updated the message which I think should be sufficient enough for debugging.   We'll have logs in the Clickhouse query log too.

clickhouse_search/management/tests/register_caids_tests.py
bpblanken: We're appending query params to the request:

```
    # Adapted from https://reg.clinicalgenome.org/doc/scripts/request_with_payload.py
    identity = hashlib.sha1(f"{login}{password}".encode("utf-8")).hexdigest()  # nosec
    gb_time = str(int(time.time()))
    token = hashlib.sha1(f"{base_url}{identity}{gb_time}".encode("utf-8")).hexdigest()  # nosec
    return f"{base_url}&gbLogin={login}&gbTime={gb_time}&gbToken={token}"
```

PR #5287, opened 2026-02-11
start clinvar view

hanars:


clickhouse_search/migrations/0020_clinvar_new_reference_data.py
```diff
@@ -9,6 +9,9 @@
 def build_materialized_view(reference_genome: str, dataset_type: str, materialized_view: str):
     def inner(apps, schema_editor):
         with connections['clickhouse_write'].cursor() as cursor:
+            cursor.execute(
```
bpblanken, line 12: Is it out of scope to use `RefreshableMaterializedView.refresh()` here?

bpblanken (approved):

clickhouse_search/migrations/0020_clinvar_new_reference_data.py
hanars: it doesn't work in the migration, the class returned by `apps.get_model` is not the actual model class with the helper implemented its a weird migration-specific class

PR #5291, opened 2026-02-12
rename variants details table

bpblanken:


clickhouse_search/migrations/0041_variant_details_path.py
```diff
@@ -0,0 +1,69 @@
...
+import clickhouse_backend.models
+import clickhouse_search.backend.fields
+from django.db import migrations, models
+import django.db.models.deletion
+import django.db.models.manager
+
+from settings import CLICKHOUSE_DATA_DIR
+
+
+class Migration(migrations.Migration):
```
bpblanken, line 12: Intentionally left the existing migration as is so that we don't run into path on conflicts.

clickhouse_search/migrations/0041_variant_details_path.py
```diff
@@ -0,0 +1,69 @@
...
+
+    dependencies = [
+        ('clickhouse_search', '0040_gnomadnoncodingconstraintdict'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
hanars: is there a reason not to use `migrations.DeleteModel`? 

Generally speaking when I want to recreate a model I comment it out, run `makemigrations`, uncomment it, make any changes I want, run `makemigrations` again, and then combine the output of the 2 migration files into 1

bpblanken: I'd just started the migration trying to do an `ALTER TABLE ... RENAME``` so had the RunSQL setup.  I don't think there's a reason not to use `migrations.DeleteModel` if that's preferred though.

hanars: there are certian django actions like `squashmigrations` that we may want to run in the future and that works better if we use the django utilities

bpblanken: 👍

Issue #5295, opened 2026-02-17
Secondary project owner


lynnpais:
**Is your feature request related to a problem? Please describe.**
Need to have a backup owner for projects.

**Describe the solution you'd like**
On the Project Page, have the option to list the name and email of a primary and secondary project owner. They would have the same project-level permissions. 

We would need to email all project owners of this once live.

**Describe alternatives you've considered**
NA

**Additional context**
Discussed in the seqr planning meeting.

PR #5298, opened 2026-02-20
Full phenotype share opt out

hanars:


hanars:
Will be easier to do as a single update with https://github.com/broadinstitute/seqr/issues/5295

PR #5324, opened 2026-03-03
Single backend decorator deprecation

hanars:
- removes the clickhouse_only decorator leaving it's decorated code unchanged
- removed the es_only decorator and removes the functions it decorates
- refactors some of the code in dataset_utils that was shared between elasticsearch and the check samples command, and moves the code directly in that command


seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -257,10 +252,10 @@ def _load_new_samples(cls, metadata_path, genome_version, dataset_type, run_vers
             except Exception as e:
                 logger.error(f'Error updating individuals sample qc {run_version}: {e}')
 
-        # Reload saved variant JSON
-        update_projects_saved_variant_json([
```
hanars, line 261: this helper function was only used in one other place in es-specific code, so with that gone it makes more sense to move this helper directly into this file

seqr/management/commands/check_for_new_samples_from_pipeline.py
```diff
@@ -394,28 +389,124 @@ def _update_individuals_sample_qc(cls, sample_type, family_guids, sample_qc_map)
...
                 variant.genotypes = genotypes
                 variants.append(variant)
-            logger.info(f'Reloading genotypes for {len(variants)} {clickhouse_dataset_type} variants in family {family_guid}')
+            logger.info(f'Reloading genotypes for {len(variants)} {dataset_type} variants in family {family_guid}')
             SavedVariant.bulk_update_models(None, variants, ['genotypes'])
             updates.update({v.id: v for v in variants})
         return updates
 
+    @classmethod
+    def _match_and_update_search_samples(cls, sample_project_tuples, sample_type, dataset_type, **sample_data):
```
hanars, line 420: adapted from code removed from dataset_utils. The only other place those helpers were used were es-only, so this reimplements them here without any of the conditional logic to support es. Its a lot of copy paste, but some refactoring as well

seqr/urls.py
```diff
@@ -241,7 +238,6 @@
     'project/(?P<project_guid>[^/]+)/upload_individuals_table': receive_individuals_table_handler,
     'project/(?P<project_guid>[^/]+)/save_individuals_table/(?P<upload_file_id>[^/]+)': save_individuals_table_handler,
     'project/(?P<project_guid>[^/]+)/upload_igv_dataset': receive_igv_table_handler,
-    'project/(?P<project_guid>[^/]+)/add_dataset/variants': add_variants_dataset_handler,
```
hanars, line 244: the ui code that calls all the URLs removed in this PR was already removed in https://github.com/broadinstitute/seqr/pull/5313

PR #5338, opened 2026-03-06
Search util test cleanup

hanars:
We have a test file that mocks all search calls and tests that the helper functions call the search function correctly. Now that theres only one search backend it does not make sense to mock all the search calls so this PR gets rid of that test file and moves any tests that are not already covered by the clickhouse tests into that file

### DIFF
--- a/clickhouse_search/search_tests.py
+++ b/clickhouse_search/search_tests.py
@@ -145,7 +145,7 @@ class ClickhouseSearchTests(ClickhouseSearchTestCase):
             gene_counts_json = get_variant_query_gene_counts(results_model, self.user)
             self.assertDictEqual(gene_counts_json, gene_counts)
 
-    def _assert_expected_variants(self, variants, expected_results, results_page=None, cache_key=None, format_cached_variants=None, sort='xpos', **kwargs):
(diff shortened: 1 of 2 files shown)


seqr/utils/search/utils.py
```diff
@@ -162,9 +161,6 @@ def _get_previous_search_results(search_model, sort, page, num_results, load_all
     if total_results is not None:
         end_index = min(end_index, total_results)
 
-    if end_index > MAX_VARIANTS:
```
hanars, line 165: this constant is also used in the clickhouse file itself and results are already validated there, so this code is now unreachable

seqr/utils/search/utils.py
```diff
@@ -173,7 +169,7 @@ def _get_previous_search_results(search_model, sort, page, num_results, load_all
     return previous_search_results, cached_page, num_results
 
 
-def query_variants(search_model, sort=XPOS_SORT_KEY, skip_genotype_filter=False, load_all=False, user=None, page=1, num_results=100):
```
hanars, line 176: this argument was used by elasticsearch but is just passed into clickhouse as an ignored kwarg

bpblanken (approved):

PR #5343, opened 2026-03-10
Search test cleanup

hanars:
To test search results we currently have the main clickhouse search tests that directly call the search helper functions and test the actual search behavior. We also have variant_search_api_test that call the actual api endpoints but with those helper function mocked and those test request parsing/validation and how the search results are post-processed and formatted as responses. Now that there is only one backend this split in tests adds a large maintenance burden as it makes it very hard to change around function signatures and also mock helper functions do not always have representative return values. 
This PR does 2 primary changes:
- In the clickhouse tests, changes all helper function calls to API calls
- removes all the search-specific tests from variant_search_api_test and moves them into the main clickhouse tests

### DIFF
--- a/clickhouse_search/search_tests.py
+++ b/clickhouse_search/search_tests.py
@@ -664,11 +664,13 @@ class ClickhouseSearchTests(ClickhouseSearchTestCase):
             VARIANT1, VARIANT2, [VARIANT3, VARIANT2], [GCNV_VARIANT4, GCNV_VARIANT3],
         ]})
 
-        exclude = {'previousSearch': True, 'previousSearchHash': 'abc1234'}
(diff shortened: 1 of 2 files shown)


clickhouse_search/search.py
```diff
@@ -730,7 +730,7 @@ def clickhouse_variant_lookup(user, variant_id, parsed_variant_id, dataset_type,
 
 def _add_liftover_genotypes(variant, data_type, variant_id, affected_only, hom_only):
     lifted_entry_cls = ENTRY_CLASS_MAP.get(variant.get('liftedOverGenomeVersion'), {}).get(data_type)
-    if not lifted_entry_cls:
```
hanars, line 733: this is actually a bug we now have test coverage for - variant lookup had an unhandled error when looking up variants where liftover failed in the loading pipeline (which routinely happens for a small subset of poorly mapped variamts)

PR #5356, opened 2026-03-19
Add saved variant fields

hanars:


seqr/views/apis/summary_data_api_tests.py
```diff
@@ -116,8 +116,6 @@
     'condition_id': 'OMIM:616126',
     'condition_inheritance': 'Autosomal recessive',
     'known_condition_name': 'Immunodeficiency 38',
-    'ClinGen_allele_ID-1': 'CA1501729',
-    'ClinGen_allele_ID-2': None,
```
hanars, line 120: This is a small functionality change, as we were previously showing the CAID as part of the individual metadata summary table. However, this field is not especially relevant to that report and was mostly included because we already had it and it was easier than writing custom logic to drop it. with the refactors in this PR it would now be messier to include this so I chose to remove it and updated the UI accordingly

clickhouse_search/management/commands/set_saved_variant_key.py
```diff
@@ -14,156 +14,10 @@
...
-    'CMG.phase1_CMG_DEL_chr2_5843', 'CMG.phase1_CMG_DEL_chr3_168', 'CMG.phase1_CMG_DEL_chr3_174', 'CMG.phase1_CMG_DEL_chr3_175',
-    'CMG.phase1_CMG_DEL_chr3_744', 'CMG.phase1_CMG_DEL_chr5_1661', 'CMG.phase1_CMG_DEL_chr5_1936',
-    'CMG.phase1_CMG_DEL_chr7_3298', 'CMG.phase1_CMG_DEL_chrX_1942', 'CMG.phase1_CMG_DEL_chrX_521', 'CMG.phase1_CMG_DEL_chrX_524',
-    'CMG.phase1_CMG_DEL_chrX_544', 'phase2_DEL_chr16_4247',
-}
-
 
 class Command(BaseCommand):
 
     def handle(self, *args, **options):
```
hanars, line 20: the migration in this PR sets the genotypes and dataset types where missing so we no longer need to set them as part of this command

clickhouse_search/management/commands/set_saved_variant_key.py
```diff
@@ -14,156 +14,10 @@
 
 BATCH_SIZE = 10000
 
-SV_DROPPED_IDS = {
```
hanars, line 17: This PR includes a migration which resets the variant IDs for our remaining unmapped SVs so if we ever did run this manage command again (which we are not planning on) this dropped IDs list would be inaccurate anyways, so theres no need to maintain it

bpblanken (approved):

PR #5394, opened 2026-04-16
Fix delete project

hanars:
The partitioning for the 38 SNV_INDEL entries table is different in the model definition/ real seqr db than in the migrations and test data, which meant that our code for deleting project data has been broken but we missed it. This PR fixes the partition dropping step of the project deletion to work with the actual partition naming, and also updates the migrations to create consistent db schemas

### DIFF
--- a/clickhouse_search/search.py
+++ b/clickhouse_search/search.py
@@ -5,6 +5,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.db import connections
 from django.db.models import Count, F, Min, Q
 from django.db.models.functions import JSONObject
+from django.db.utils import OperationalError
 import json
 
 from clickhouse_search.backend.fields import NamedTupleField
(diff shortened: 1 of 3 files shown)


clickhouse_search/search.py
```diff
@@ -1030,8 +1030,16 @@ def delete_clickhouse_project(project, dataset_type, sample_type=None):
     if dataset_type == Sample.DATASET_TYPE_SV_CALLS and sample_type == Sample.SAMPLE_TYPE_WES:
         dataset_type = 'GCNV'
     table_base = f'{GENOME_VERSION_LOOKUP[project.genome_version]}/{dataset_type}'
+    partition_id = f"'{project.guid}'"
+    partition_ids = [partition_id]
+    if project.genome_version == GENOME_VERSION_GRCh38 and dataset_type == Sample.DATASET_TYPE_VARIANT_CALLS:
```
bpblanken, line 1035: fwiw we're using `(project_guid,)` one item tuples for the non-GRCh38/SNV_INDEL data set types.

Also, IIRC we have released open source versions that are not repartitioned and never provided a mandatory update process, as it was painful.  We only provided [this](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/loading_pipeline/ops/repartition_clickhouse_grch38_snv_indel.py).  I've been trying to write portable code that handles both cases (we have tests for both), even though it is extremely annoying!

hanars: I do not think supporting both is viable in the seqr code base from a maitenance perspective. The approach I have taken here is that all new installations will conform to our installation and the code path for deleting projects will work for these, and if there are legacy installations that run into an issue deleting projects they will reach out to us and we can send them update instructions at that time. Since this is a code path that is very rarely accessed, that feels acceptable to me

hanars: you can also delete all families in the project instead of the whole project as a workaround in a pinch

bpblanken: Can you add a nicely interpretable error message then?  So that it's clear to the user why the command failed rather than relying on us to remember the state of seqr releases?

bpblanken (approved):

PR #5430, opened 2026-05-08
reload affected dict when status updated

hanars:
This adds a necessary refresh of the affected status dictionary whenever affected status is updated. Adding this functionality had an unintended consequences of messing up the affected dict setup in the delete data test. Although I was not able to debug exactly where this issue was occurring after a day of investigation, I did discover that reordering the test suite fixes the issue. While this is not ideal and I would rather figure out how to update the set up/ tear down to work fully,  given the relatively urgent nature of this bug fix I think it makjes sense to release it with the reordered tests


seqr/views/utils/individual_utils.py
```diff
@@ -70,10 +70,12 @@ def add_or_update_individuals_and_families(project, individual_records, user, ge
 
     updated_family_models = Family.objects.filter(id__in=updated_family_ids)
     _remove_pedigree_images(updated_family_models, user)
-    if updated_affected and not skip_gt_stats_rebuild:
-        trigger_rebuild_gt_stats(project, user)
+    if updated_affected:
```
bpblanken, line 73: This will also handle the reload for a newly created project right?

hanars: yep! this function is called there as well

bpblanken (approved):

PR #5448, opened 2026-05-20
Vlm shared test db

hanars:
Since the initial VLM implementation used a very small subset of the clickhouse database we used a lightweight standalone setup for just those tables in our unit tests. With the coming addition of an endpoint that utilizes more of the DB, having this separate DB setup opens us up to potentially serious bugs if and when the database schema changes. Therefore, we want our VLM unit tests to set up and populate data in the clickhouse DB using the same django mechanism as the main seqr app and its tests

## Pull request overview

This PR moves VLM tests from a bespoke ClickHouse setup script to the shared seqr ClickHouse test database fixture and updates VLM unit test CI to provision the supporting Postgres/ClickHouse services.

**Changes:**
- Adds VLM pytest database setup that loads `clickhouse_search` fixture data and refreshes SNV/INDEL stats views/dictionaries.
- Updates VLM test expected counts to match the shared ClickHouse fixture.
- Replaces the old standalone ClickHouse seed script in CI with Django-backed test DB setup and removes an unused JSX test script.


larrybabb (approved):
This looks like a very useful and well defined PR. I didn't catch anything useful but Claude did identify one thing that might be useful to address. I will defer to you, this is not a show stopper so feel free to push it.

...
1. pytest-django is not in `vlm/requirements-test.txt`. `vlm/conftest.py` uses pytest-django fixtures (`db, django_db_blocker, django_db_keepdb, _django_db_marker`), but pytest-django is only installed ad-hoc inside the throwaway venv in the "Set up Clickhouse Test Database" step. During the actual `coverage run -m pytest vlm/ step, pytest-django` isn't installed, so the conftest's fixtures never wire up — the DB and grants are preloaded by the prior step, and conftest.py is effectively a no-op there. This works but is subtle
**Two consequences**:
    - A developer running `pytest vlm/` locally now needs the full seqr Python env plus pytest-django, plus clickhouse + postgres services. The "VLM is standalone" boundary called out
  in CLAUDE.md is meaningfully weakened.
    - Anyone reading `conftest.py` will reasonably assume it runs during tests. It doesn't.

hanars:
Regarding your point, this is very intentional design decision. The VLM application runs a fully isolated docker image that does not have django installed. As such, it is crucial that our unit tests run without django installed, as otherwise we could easily introduce a bug in which our local environments and test environments succeed because they rely on django but our production release would catastrophically fail. 
While this adds a (minimal) set up burden for local developers who want to run the test suite locally and not rely on the cloud runner (which is how I ave been testing), this minor inconvenience for one-time test set up is greatly outweighed by the need to have our tests run in an environment that  closely mirrors production

PR #5450, opened 2026-05-20
Remove hail from VLM

hanars:
## Pull request overview

This PR removes the Hail dependency from the VLM service by switching variant liftover/validation to `pyliftover`, and updates packaging/CI to run on a standard Python 3.11 base image.

**Changes:**
- Replace Hail-based locus/liftover logic with `pyliftover` and direct `(chrom, pos)` handling for ClickHouse queries.
- Add `vlm/requirements.in` + compiled `vlm/requirements.txt`, and update test dependency compilation/installation.
- Move Dockerfile and GitHub Actions unit test container from `hailgenetics/hail:*` to `python:3.11-slim-bullseye`; adjust tests for updated variantId formatting (no `chr` prefix).


vlm/test_vlm.py
```diff
@@ -52,7 +52,7 @@ async def test_match(self, mocked_responses):
                         'id': 'TestVLM',
                         'label': 'TestVLM browser',
                     },
-                    'url': 'https://test-seqr.org/variant_lookup?genomeVersion=38&variantId=chr1-38724419-T-G',
+                    'url': 'https://test-seqr.org/variant_lookup?genomeVersion=38&variantId=1-38724419-T-G',
```
hanars, line 55: this was a bug, the seqr lookup page does not necessarily handle the chr prefix

bpblanken:
Is it worth doing any validation that Hail liftover and pyliftover give similar results?

hanars:
> Is it worth doing any validation that Hail liftover and pyliftover give similar results?

Not really, there are definitely some differences in how these liftovers work and if we do that validation we will absolutely find some differences, but theres nothing we can do about them. Running an increasingly insecure and unmaintainable system in order to preserve the particular liftover version we used at go live, which is not regulated or in any way required to remain stable according to the network rules, is not worth the tradeoff. The fact that we are using the same chain file and only doing single point conversions (instead of like a range conversion for an indel) should minimize differences to the greatest extent we can

bpblanken (approved):

PR #5460, opened 2026-05-28
Fix issue with viewing RNA junction tracks

mike8115:
Fixed issue raised in #5425.

Haven't implemented any changes that would check whether `process.returncode == 0` since I'm not sure how to elegantly handle any exceptions that could get raised here.

### DIFF
--- a/seqr/utils/file_utils.py
+++ b/seqr/utils/file_utils.py
@@ -69,6 +69,9 @@ def file_iter(file_path, byte_range=None, raw_content=False, user=None, **kwargs
         if file_path.endswith("gz") and not raw_content:  # Avoids unzipping BED files
             command += " | gunzip -c - "
         process = run_command(command, user=user)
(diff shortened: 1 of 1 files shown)


hanars:
> Haven't implemented any changes that would check whether `process.returncode == 0` since I'm not sure how to elegantly handle any exceptions that could get raised here.

So seqr's middleware handles all exceptions for API requests so you would not have to handle them, just raise them. You would realistically want to add a few lines of code that behave the same as the error handling we have implemented in this file in the `run_gsutil_with_wait` function:
https://github.com/broadinstitute/seqr/blob/master/seqr/utils/file_utils.py#L116-L118

mike8115:
Noted and done! Thanks @hanars!

mike8115:
Tests appear to be passing now. 

@hanars: Let me know if there's any other changes that we need to do.

PR #5468, opened 2026-05-29
disable rna reference tracks for grach37

hanars:
## Pull request overview

Disables GRCh38-specific RNA reference tracks (GTEx normalized/aggregate and mappability) for GRCh37 projects, since the underlying track URLs only exist for GRCh38.

**Changes:**
- In `updateReads`, only auto-populate `rnaReferences` from the tissue lookup when the project's genome version is GRCh38.
- Gate the Reference Tracks divider, GTEx selectors, and Mappability checkbox group on an `isGrch38` flag derived from the open family's project.


bpblanken (approved):

bpblanken:
this is into `master`, that was on purpose right?

hanars:
> this is into `master`, that was on purpose right?

nope good catch!

PR #5472, opened 2026-06-03
sync tag updates to postgres dicts

hanars:
This pull request removes a bug where some in-memory dictionaries (used for fast lookups, such as gene IDs and individual metadata) are not being kept in sync with database changes. The main updates ensure that these dictionaries are reloaded automatically whenever relevant reference data or model changes occur, improving data consistency across the application. The changes also add comprehensive logging and expand test coverage to verify that dictionary reloads and audit logs are correctly triggered.

**Dictionary Reload Triggers and Logging**

* Added calls to `IndividualMetadataDict.reload()` and related dictionary reloads (e.g., `GeneIdDict.reload()`, `AffectedDict.reload()`, `SexDict.reload()`) after key updates or deletions in families, individuals, and reference data, ensuring in-memory dictionaries stay in sync with the database. These are now triggered in handlers for family/individual updates, deletions, and project creation.

* Ensured `GeneIdDict.reload()` is called after updating `GeneInfo` reference data, and corresponding log messages are produced.

**Test Coverage and Logging Verification**

* Updated and expanded test cases in `family_api_tests.py` and `individual_api_tests.py` to assert that dictionary reloads are logged as expected and that audit logs reflect the correct update operations. This includes new log assertions after updates, deletions, and bulk operations. 

* Added `databases = '__all__'` to relevant test classes to ensure tests are still able to run  now that application behavior requires access t the clickhouse database

These changes collectively improve the robustness and observability of how reference and metadata dictionaries are managed, ensuring that the application always serves up-to-date information after any relevant data changes.

### DIFF
--- a/seqr/views/apis/family_api.py
+++ b/seqr/views/apis/family_api.py
@@ -5,7 +5,7 @@ from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import Count, Max, Q, F
 from django.db.models.fields.files import ImageFieldFile
 
-from clickhouse_search.models.postgres_dicts import IndividualMetadataDict
+from clickhouse_search.models.postgres_dicts import AffectedDict, SexDict, IndividualMetadataDict
 from matchmaker.models import MatchmakerSubmission
 from reference_data.models import Omim
 from seqr.utils.gene_utils import get_genes_for_variant_display
@@ -211,6 +211,8 @@ def delete_families_handler(request, project_guid):
     # delete families
     Family.bulk_delete(request.user, project=project, guid__in=family_guids_to_delete)
 
+    AffectedDict.reload(request.user)
+    SexDict.reload(request.user)
     IndividualMetadataDict.reload(request.user)
 
     # send response
--- a/seqr/views/apis/family_api.py
+++ b/seqr/views/apis/family_api.py
@@ -211,7 +211,7 @@ def delete_families_handler(request, project_guid):
     # delete families
     Family.bulk_delete(request.user, project=project, guid__in=family_guids_to_delete)
 
-    IndividualMetadataDict.reload()
+    IndividualMetadataDict.reload(request.user)
 
     # send response
     return create_json_response({
@@ -238,7 +238,7 @@ def update_family_fields_handler(request, family_guid):
         'display_name',
     ] + immutable_keys, updated_fields=updated_fields)
     if updated_fields.intersection({'post_discovery_omim_numbers', 'post_discovery_mondo_id', 'analysis_status'}):
-        IndividualMetadataDict.reload()
+        IndividualMetadataDict.reload(request.user)
 
     return create_json_response({
         family.guid: _get_json_for_model(family, user=request.user, process_result=_set_display_name)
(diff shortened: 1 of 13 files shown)


seqr/views/apis/family_api.py
```diff
@@ -230,9 +233,12 @@ def update_family_fields_handler(request, family_guid):
 
     request_json = json.loads(request.body)
     immutable_keys = [] if external_anvil_project_can_edit(family.project, request.user) else ['family_id']
+    updated_fields = set()
     update_family_from_json(family, request_json, user=request.user, allow_unknown_keys=True, immutable_keys=[
         'display_name',
-    ] + immutable_keys)
+    ] + immutable_keys, updated_fields=updated_fields)
+    if updated_fields.intersection({'post_discovery_omim_numbers', 'post_discovery_mondo_id', 'analysis_status'}):
+        IndividualMetadataDict.reload(request.user)
```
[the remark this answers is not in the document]
hanars: `edit_families_handler` can not be called with updates to any of the impacted fields, theres a small white list of fields that can be bulk updated in this way and non of them are included in the metadata

seqr/views/apis/project_api.py
```diff
@@ -80,7 +82,7 @@ def update_project_handler(request, project_guid):
     check_project_permissions(project, request.user, can_edit=True)
 
     request_json = json.loads(request.body)
-    updated_fields = None
+    updated_fields = set()
     consent_code = request_json.get('consentCode')
```
larrybabb, line 86: **`update_project_handler` re-assigns instead of `.add.`**  After initializes `updated_fields = set()`, then if consent code changes, does `updated_fields =
  {'consent_code'}` — an overwrite, not `.add().` It happens to work because nothing else mutates the set before this point, but the pattern is fragile (anyone adding a new check above
  would have it silently clobbered). Trivial fix: `updated_fields.add('consent_code')`.

seqr/views/apis/saved_variant_api_tests.py
larrybabb, line 1: **`LocalSavedVariantAPITest `(the non-AnVIL test variant) was deleted** (`saved_variant_api_tests.py` — see removed lines 732-734 in the old file). The remaining class inherits `ClickhouseSearchTestCase` which derives from `AnvilAuthenticationTestCase`. That means saved-variant write paths are no longer tested under the local (`AuthenticationTestCase`) auth mode. If non-AnVIL deployments are a supported configuration, this is a regression in coverage breadth that deserves a sentence in the PR description. 

Is this intentional consolidation, or was the local class collateral damage from the merge with `ClickhouseSearchTestCase`?

larrybabb (approved):
I see that co-pilot's review caught a similar issue below that you addressed related to the possible correctness gaps with individual_Id and family changes that aren't triggering any reloads. Your response makes sense so I will defer to you again on the alignment of those bulk update fields.

I provided 2 other comments for your consideration, but again, I'm going ahead with the approval as I don't think these comments rise to the level of hindering the release.

seqr/views/apis/saved_variant_api_tests.py
hanars: So the Anvil-specific test case does test the difference between anvil and non-anvil Auth, but it is also historically how we tested different backends as well, with the anvil test case using the latest supported backend and the local test case testing the oldest supported backend. The reason we had the 2 test modes for this file was only ever to cover behavior for different backends, theres never been any auth-mode specific behavior in this file that needs to be tested. Since support for multiple backends was deprecated, this test file has essentially been running an identical set of tests twice. As such, while the number of tests being run is changing with this refactor, theres actually no functional changes to test coverage.

PR #5474, opened 2026-06-08
clean up db config for clickhouse

hanars:
Our DB configuration still has conditional code left over from when we had multiple backend support - this cleans up the configuration to remove that configurability and simplify the set up


larrybabb (approved):
Looks straightforward.

Issue #5470, opened 2026-06-01
Panelapp Australia upgrading Gencode/Ensembl version


drtconway:
Hi Seqr-people,

We're working on a major upgrade for Panelapp Australia to migrate the underlying gene annotation from Gencode 27 (Ensembl 90) to Gencode 49 (Ensembl 115). This entails quite a few genes changing symbol (name), and Ensembl ID, and in some cases both.

This will lead to some problems - some fairly direct, and some more subtle.

I'm still investigating, and will add more details as I discover them, but let me start with what we have determined so far.

Where the blob of JSON from the Panelapp API is parsed, the hard coded "90" here, will break:

https://github.com/broadinstitute/seqr/blob/7e611fcb5549339863e5d89216709d33681a4262/panelapp/panelapp_utils.py#L23

But there's a more subtle and pervasive problem, which is that there will be gene symbols and Ensembl IDs that will be inconsistent with GeneInfo. GeneInfo can be updated with the relevant gencode version, I'm still trying to elucidate what that "fixes" and what remains problematic.

One area of concern is in gene_utils.py

https://github.com/broadinstitute/seqr/blob/7e611fcb5549339863e5d89216709d33681a4262/seqr/utils/gene_utils.py#L198

where gene symbols are resolved into gene IDs for query resolution. The code currently finds the most "recent" match only, which means existing projects/samples stored in the backend will not match the query without being reloaded.

We thought we should give a heads-up, since this will affect any Seqr instance using Panelapp Australia for gene lists. I understand that Genomics England are also working toward an update of their underlying genome annotation, but I can't speak to any detail about that.

We would be keen to work with you to make the transition as smooth as possible for everyone!

Tom, for the Bioinformatics Methods Group at MCRI.

ssadedin:
@hanars - just paging you to alert to this.

We are proposing to implement the above change on the PanelApp API this coming Friday, which will stop panels updating from PanelApp Australia until at least a basic patch is implemented along the lines above.

We will propose a patch later today to address the problem, along with a couple of other changes that improve the safety when gene symbols have multiple definitions across different versions of Ensembl / Gencode.

Apologies for the short timeline for this change!

hanars:
Hi all, really appreciate the proactive conversation and support you all are providing for this change! To address some of the specific issues raised above:

> The code currently finds the most "recent" match only, which means existing projects/samples stored in the backend will not match the query without being reloaded.

[...]

PR #5476, opened 2026-06-09
get the latest Ensembl ID, not hard coded 90

drtconway:
As discussed in https://github.com/broadinstitute/seqr/issues/5470, when retrieving gene lists from Panelapp Australia, the code currently looks for a hard coded Ensembl release of '90'. Panelapp Australia has upgraded its underlying annotation to Ensembl release 115, so the code will fail.

This change scans for the most recent Ensembl ID.

### DIFF
--- a/panelapp/panelapp_utils.py
+++ b/panelapp/panelapp_utils.py
@@ -14,13 +14,22 @@ class TooManyRequestsError(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


hanars:
Can you update the unit tests to confirm that this works with the newer version/ prevent a regression in the future?

PR #5479, opened 2026-06-15
handle updated hpo download values

hanars:
Our update for HPO broke over the weekend because the format for one of the fields in the download has changed to allow new values that do not work with our current parsing logic. This updates the parsing and unit tests to support the real data


larrybabb (approved):
+1 very straightforward

PR #5488, opened 2026-06-18
Add validation options to data loading UI

hanars:
## Pull request overview

Adds support for selectively skipping specific VCF validations during data loading, wiring the selected skip options from the Data Management UI through the `load_data` API into the pipeline trigger payload. Also extends intenral Broad load controls to include skipping TDR metrics and updates API tests accordingly.

**Changes:**
- UI: adds a “Skip Validations” multi-select (checkbox group) and adds a “Skip TDR Metrics” toggle for Broad internal load flow.
- API: forwards `validationsToSkip` to `trigger_data_loading` as `validations_to_skip`, and refactors kwargs passed to the loader trigger.
- Tests/utilities: updates test request bodies and expected pipeline trigger variables; updates loader utility to conditionally include `validations_to_skip`.

### DIFF
--- a/seqr/utils/add_data_utils.py
+++ b/seqr/utils/add_data_utils.py
@@ -102,7 +102,7 @@ def trigger_data_loading(projects: list[Project], individual_ids: list[int], sam
     conditional_variables = {
         'skip_check_sex_and_relatedness': skip_check_sex_and_relatedness,
         'skip_expect_tdr_metrics': skip_expect_tdr_metrics,
-        'validations_to_skip': validations_to_skip,
+        'validations_to_skip': _valid_validation_to_skip(validations_to_skip, user),
     }
     variables.update({k: v for k, v in conditional_variables.items() if v})
     file_path = _get_pedigree_path(genome_version, sample_type, dataset_type)
@@ -125,6 +125,16 @@ def trigger_data_loading(projects: list[Project], individual_ids: list[int], sam
         ]))
 
     return success
+
+
+def _valid_validation_to_skip(validations_to_skip: list[str], user: User) -> list[str]:
+    invalid_validations = [validation for validation in (validations_to_skip or []) if validation not in {
+        'validate_expected_contig_frequency', 'validate_sample_type', 'validate_no_duplicate_variants',
+    }]
+    if invalid_validations:
+        validations_to_skip = [validation for validation in validations_to_skip if validation not in invalid_validations]
+        logger.error(f'Omitted invalid skip validations from loading request: {", ".join(invalid_validations)}', user)
+    return validations_to_skip
 
 
 def _enqueue_pipeline_request(name: str, variables: dict, user: User, raise_error: bool = True, log_error: bool = True):
@@ -186,7 +196,7 @@ def _upload_data_loading_files(individual_ids: list[int], vcf_sample_id_map: dic
 
 def _write_gene_id_file(user):
     file_name = 'db_id_to_gene_id'
-    if does_file_exist(f'{LOADING_DATASETS_DIR}/{file_name}.csv.gz'):
+    if does_file_exist(f'{LOADING_DATASETS_DIR}/{file_name}.csv.gz', user):
         return
 
     gene_data_loaded = (GeneInfo.objects.filter(gencode_release=int(GeneInfo.CURRENT_VERSION)).exists() and
--- a/seqr/utils/add_data_utils.py
+++ b/seqr/utils/add_data_utils.py
@@ -90,7 +90,7 @@ def trigger_rebuild_gt_stats(project, user):
 
 def trigger_data_loading(projects: list[Project], individual_ids: list[int], sample_type: str, dataset_type: str,
                          genome_version: str, data_path: str, user: User, raise_error: bool = False, skip_expect_tdr_metrics: bool = True,
-                         skip_check_sex_and_relatedness: bool = True, vcf_sample_id_map=None,
+                         skip_check_sex_and_relatedness: bool = True, validations_to_skip: list[str] = None, vcf_sample_id_map=None,
                          success_message: str = None,  error_message: str = None, success_slack_channel: str = SEQR_SLACK_LOADING_NOTIFICATION_CHANNEL):
     variables = {
         'projects_to_run': sorted([p.guid for p in projects]) if projects else None,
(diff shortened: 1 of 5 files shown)


ui/pages/DataManagement/components/LoadData.jsx
```diff
@@ -55,17 +55,49 @@ const FILE_PATH_FIELD = {
...
+    ...GENOME_VERSION_FIELD,
     component: ButtonRadioGroup,
-    options: [SAMPLE_TYPE_EXOME, SAMPLE_TYPE_GENOME].map(value => ({ value, text: value })),
     validate: validators.required,
   },
+  {
+    name: 'validationsToSkip',
+    label: 'Skip Validations',
+    component: CheckboxGroup,
+    options: [
```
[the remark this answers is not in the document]
hanars: This suggestion will not actually work, but I updated the `CheckboxGroup` component itself so that all usages of `value` handle cases where it is undefined

seqr/views/apis/data_manager_api.py
```diff
@@ -304,19 +304,18 @@ def load_data(request):
     if errors:
         raise ErrorsWarningsException(errors)
 
-    is_local = True
-    success_message = None
-    error_message = None
-    if AirtableSession.is_airtable_enabled():
-        is_local = False
-        success_message = f'*{request.user.email}* triggered loading internal {sample_type} {dataset_type} data for {len(individual_ids)} samples in {len(projects)} projects ({"; ".join(sorted(project_counts))})'
-        error_message = f'ERROR triggering internal {sample_type} {dataset_type} loading'
```
hanars, line 313: Most of the options here can never be set for local installs, but the way we were passing the kwargs made it very ambiguous which were local vs. Broad internal. For example, always passing `skip_check_sex_and_relatedness` from the request json made it seem like this was always a possible option, but in fact this toggle is completely unavailable in the local UI and therefore could never be passed through. Similarly, `vcf_sample_id_map` was always passed through but is always empty for local installs because its populated from Airtable. By restructuring the kwargs here it makes it very clear which options are available in different environments, making the behavior more maintainable

seqr/views/apis/data_manager_api_tests.py
```diff
@@ -1214,16 +1221,18 @@ def _assert_expected_load_data_requests(self, dataset_type='SNV_INDEL', sample_t
             'dataset_type': dataset_type,
             'reference_genome': 'GRCh38',
         }
-        if self.SKIP_TDR:
+        if validations_to_skip:
+            body['validations_to_skip'] = validations_to_skip
+        if skip_tdr or self.SKIP_TDR:
             body['skip_expect_tdr_metrics'] = True
-        if skip_check_sex_and_relatedness:
```
hanars, line 1219: this change reflects the only real "behavior change" from this refactor - previously for local installations `skip_check_sex_and_relatedness` was always passed as False, and now it is always passed as True. In reality, the `skip_check_sex_and_relatedness` and `skip_expect_tdr_metrics` parameters are no-ops for local installs as they are globally always True in our local install pipeline configuration. This means that previously we were always sending `skip_check_sex_and_relatedness` as False which was then ignored by the pipeline and set to True, and now we are just passing True by default to align the seqr app and pipeline behaviors. See point 2 in Ben's comment on the ticket: https://github.com/broadinstitute/seqr-private/issues/1748#issuecomment-4500416449

seqr/utils/add_data_utils.py
```diff
@@ -99,11 +99,12 @@ def trigger_data_loading(projects: list[Project], individual_ids: list[int], sam
...
         'sample_type': sample_type,
     }
-    bool_variables = {
+    conditional_variables = {
         'skip_check_sex_and_relatedness': skip_check_sex_and_relatedness,
         'skip_expect_tdr_metrics': skip_expect_tdr_metrics,
+        'validations_to_skip': validations_to_skip,
     }
-    variables.update({k: v for k, v in bool_variables.items() if v})
+    variables.update({k: v for k, v in conditional_variables.items() if v})
```
larrybabb, line 107: minor...
the `if v` filter that skips falsy values now applies to heterogeneous types not just `boolean`s.

so if someone adds an integer `0` or an intentionally empty list as a meaningful value in the future, the filter will silently drop it. A comment explaining the "skip if falsy" contract would help remind a future dev in the event that use case occurs. but I suppose a python dev would inherently know this. your call of course.

seqr/views/apis/data_manager_api.py
```diff
@@ -304,19 +304,18 @@ def load_data(request):
...
+        'success_message': f'*{request.user.email}* triggered loading internal {sample_type} {dataset_type} data for {len(individual_ids)} samples in {len(projects)} projects ({"; ".join(sorted(project_counts))})',
+        'error_message': f'ERROR triggering internal {sample_type} {dataset_type} loading',
+    } if AirtableSession.is_airtable_enabled() else {'raise_error': True}
 
     success = trigger_data_loading(
         projects_by_guid.values(), individual_ids, sample_type, dataset_type, request_json['genomeVersion'],
-        _callset_path(request_json), user=request.user,
-        skip_check_sex_and_relatedness=request_json.get('skipSRChecks', False), vcf_sample_id_map=vcf_sample_id_map,
-        raise_error=is_local, skip_expect_tdr_metrics=is_local, success_message=success_message, error_message=error_message,
+        _callset_path(request_json), user=request.user, validations_to_skip=request_json.get('validationsToSkip'),
```
larrybabb, line 317: `validations_to_skip=request_json.get('validationsToSkip'),`

The UI seems to be the only barrier to enforce the allowed values (`validate_expected_contig_frequency`, `validate_sample_type`, `validate_no_duplicate_variants`). If the pipeline has a way to handle unknown validation names, I would make sure that potential invalid or misspelled names are handled properly. 
Adding validation against the known set would be consistent with how seqr validates other enum-like inputs.

ui/pages/DataManagement/components/LoadData.jsx
```diff
@@ -55,17 +55,49 @@ const FILE_PATH_FIELD = {
...
+      {
```
larrybabb, line 74: I would have not caught this. And it may not be legit, but I think you should assess since it is a cosmetic concern regarding the formatting of the tooltip output.

----

The multi-line template literals for description fields end with trailing spaces (from indentation before the closing backtick continuation). Example:


```
description: `By default, VCFs will be checked to ensure they have a reasonable number of variants in each 
  chromosome, and loading will fail...`
```

The line break + leading whitespace will render as a double-space in the tooltip/description. This is cosmetic but noticeable in the UI.

larrybabb (approved):
3 comments for your review. I'm pre-approving since I don't think any one of the 3 are worthing of holding things up. I defer to you to fix and close if you feel it is warranted. These are all really minor.

The refactor of the loading_kwargs is cleaner and easier to follow than before.

ui/pages/DataManagement/components/LoadData.jsx
hanars: This is just not true though - I looked at the tooltips specifically to check the whitespace (I am always making whitespace errors in JSX so I check it a lot)  and the whitespace looked right

seqr/views/apis/data_manager_api.py
hanars: agreed that this validation is a good idea, since I think the pipeline just silently ignores invalid validation names

PR #5493, opened 2026-06-24
Metrics dashboard

hanars:
## Pull request overview

Adds new metrics and export functionality to the Report “Seqr Stats” view, expanding the dashboard to include user login/activity counts and enabling download of sample loading history.

**Changes:**
- Updated Seqr stats API aggregation to return user activity counts and to support grouped sample-load exports.
- Added a new `sample_stats_download` API endpoint (wired into URLs) that exports daily sample load counts (TSV/XLS).
- Updated the Seqr Stats UI to show a Users table and provide an export button for sample loading history.

### DIFF
--- a/seqr/views/apis/report_api.py
+++ b/seqr/views/apis/report_api.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 
 from datetime import datetime, timedelta
 from django.db.models import Count, Q, F, Value
+from django.db.models.expressions import Case, When
 from django.contrib.auth.models import User
 from django.contrib.postgres.aggregates import ArrayAgg
 import json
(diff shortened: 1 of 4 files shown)


seqr/views/apis/report_api.py
```diff
@@ -39,55 +42,114 @@
...
+    individuals_count = {}
+    for agg in project_aggs:
+        key = _agg_key(agg)
+        project_counts[key] = agg['count']
+        families_count[key] = agg['family_count']
+        individuals_count[key] = agg['individual_count']
+
+    user_counts = User.objects.filter(is_active=True).aggregate(
+        total=Count('id'),
+        multipleLogins=Count('id', filter=Q(last_login__isnull=False) & ~Q(last_login=F('date_joined'))),
```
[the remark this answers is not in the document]
hanars: This is wrong and does not understand how django user timestamps work with the social auth integration. All users created after 2015 have a date_joined exactly equal to last_login when the user logs in the first time and therefore users who have logged in exactly once have the same value. If they have different values, it means they logged in again later

seqr/views/apis/report_api.py
```diff
@@ -39,55 +42,114 @@
...
+    return export_table('seqr_sample_loading', header, rows, file_format)
+
+
+def _format_export_row(loaded, count, sample_type, dataset_type, demo, no_anvil=0, is_internal=0):
+    row = [loaded.strftime('%Y-%m-%d'), count, sample_type, DATASET_TYPE_LOOKUP.get(dataset_type, 'Unknown'), demo]
+    if no_anvil:
+        row.append('No AnVIL')
+    elif is_internal is not 0:
+        row.append('Internal' if is_internal else 'External')
+    return row
```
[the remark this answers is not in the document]
hanars: correct identification of an issue , but this fix breaks local install behavior

seqr/views/apis/report_api.py
```diff
@@ -39,55 +42,115 @@
...
+
+    project_aggs = project_qs.annotate(
+        count=Count('id', distinct=True),
+        family_count=Count('family', distinct=True),
+        individual_count=Count('family__individual', distinct=True),
+    )
+    project_counts = {}
+    families_count = {}
+    individuals_count = {}
+    for agg in project_aggs:
```
larrybabb, line 60: The original code — `Project.objects.filter(is_demo=True).count()` unconditionally counted all demo projects. The new aggregation implicitly assumes each `_agg_key` is produced by exactly one (demo, no_anvil, is_internal) combination. 

Is that assumption true in practice (demo projects almost certainly don't have AnVIL workspaces)?

But the code  by construction will silently produce incorrect counts if the data ever violates it. Maybe this should either be documented as an assumption or fixed with `+= agg['count']`?

larrybabb: The above is based on the following concern identified by Claude

>In AnVIL mode, project_qs is grouped by (demo, no_anvil, is_internal). Multiple distinct combinations can map to the same key — e.g., a demo project without an AnVIL workspace (demo=True, no_anvil=True) and a demo project with an AnVIL workspace (demo=True, no_anvil=False) both map to 'demo'. If both groups exist, only the last one's count survives. The same overwrite risk applies to grouped_sample_counts on line 50.

larrybabb (approved):
@hanars I'm pre-approving since I don't think my identified comments are worthy or a re-review as I don't think they have a significant impact.

seqr/views/apis/report_api.py
hanars: good catch, this is actually worth fixing as its possible to end up with both anvil and non-anvil demo projects

PR #5495, opened 2026-06-30
Tiered access to projects

hanars:
## Pull request overview

Implements “tiered access” by allowing users to access subsets of a project via Analysis Groups backed by AnVIL workspaces, and propagates that concept through UI routing, API permissions/filters, and ClickHouse search behavior.

**Changes:**
- Add AnalysisGroup-level workspace fields and permissions logic to support partial project access via AnVIL workspaces.
- Update multiple APIs (projects, variants, dashboard, awesomebar, IGV, summary exports) to filter data by either full-project access or analysis-group access.
- Update UI navigation/components and expand test fixtures + test coverage for partial-access scenarios.


ui/pages/Project/components/AnalysisGroupButtons.jsx
```diff
@@ -174,7 +201,7 @@ export const UpdateAnalysisGroupButton = connect(mapUpdateStateToProps, mapDispa
 
 const navigateProjectPage = (history, projectGuid) => () => history.push(`/project/${projectGuid}/project_page`)
 
-export const DeleteAnalysisGroup = React.memo(({ project, analysisGroup, onSubmit, size, iconOnly, history }) => (
+const DeleteAnalysisGroup = React.memo(({ project, analysisGroup, onSubmit, size, iconOnly, history }) => (
   canUpdateGroup(project, analysisGroup) ? (
     <DeleteButton
```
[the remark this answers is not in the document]
hanars: So if you call `canUpdateGroup` without a user it will return the proper permissions if its a group without an anvil workspace, and will always return false if there is an anvil workspace, which is actually the behavior we want here and does match the server side behavior. While I agree with copilot that omitting the user seems like a possible error and obscures intent, I disagree that the solution is to re-implement part of `canUpdateGroup` inline here like they suggest. Rather, I think a better option is to explicitly pass a null value for user here, to make it clear that the omission of the user permission override is intentional

ui/pages/Project/reducers.js
```diff
@@ -63,7 +63,26 @@ export const loadProjectExportData = () => (dispatch, getState) => Promise.all([
...
+    new HttpRequestHelper(`/api/project/${currentProjectGuid}/analysis_groups/${analysisGroupGuid}/get_collaborators`,
+      (responseJson) => {
+        dispatch({ type: RECEIVE_DATA, updatesById: responseJson })
+        dispatch({ type: RECEIVE_PROJECT_COLLABORATORS })
+      },
+      (e) => {
+        dispatch({ type: RECEIVE_PROJECT_COLLABORATORS, error: e.message })
+      }).get()
+  }
+}
```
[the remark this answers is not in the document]
hanars: Some reducers in this file return promises but some do not. Notably, this function is replacing a call to `loadProjectCollaborators` which in turn was calling `loadProjectChildEntities` which does NOT return anything. This proposed change would therefore be a much larger behavior change than the current implementation, with the potential to break behavior with no real upsides

clickhouse_search/search.py
```diff
@@ -64,32 +64,35 @@ def get_clickhouse_variants(families, user, genome_version=None, sort=None, samp
         except InvalidDatasetTypeException:
             continue
 
+        sample_data = sample_data_by_dataset_type[dataset_type]
+        exclude_dt_keys = (exclude_keys or {}).get(dataset_type)
+        dataset_results = []
+
+        if sample_data:
```
hanars, line 71: this is just a control flow order change, this code block was essentially copied from below

clickhouse_search/search.py
```diff
@@ -64,32 +64,35 @@ def get_clickhouse_variants(families, user, genome_version=None, sort=None, samp
...
+            if run_x_linked_male_search:
+                dataset_results += _get_x_linked_male_search_results(
+                    entry_qs, variants_qs, dataset_type, user, sample_data, exclude_keys=exclude_dt_keys,
+                    **search, **parsed_filters,
+                )
+
         if dataset_type == Dataset.DATASET_TYPE_VARIANT_CALLS and no_access_project_genome_version:
             results += _get_no_access_search_results(
                 entry_qs, variants_qs, has_comp_het, user, **search, **parsed_filters,
-                exclude_projects=sample_data_by_dataset_type[dataset_type].get('project_guids'), inheritance_mode=inheritance_mode,
```
hanars, line 70: When querying for variants in projects that users do not have access to, we exclude any variants that are returned from projects they do have access to. When project access was binary we did this by excluding the projects they have access to from this search. However, now that users potentially have access to a part of a project, this no longer works as it means that no search is ever run on the subset of families they do not have access to from the partial project. Rather than a complicated series of ors with project family subsets, its much more straightforward to utilize the existing `exclude_keys` filter and explicitly skip any variants returned from the accessible families search

seqr/views/apis/anvil_workspace_api.py
```diff
@@ -24,21 +24,10 @@
...
-from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL, POLICY_REQUIRED_URL, API_POLICY_REQUIRED_URL,\
-    SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL
+from seqr.views.utils.permissions_utils import check_workspace_perm, anvil_auth_and_policies_required
+from settings import BASE_URL, GOOGLE_LOGIN_REQUIRED_URL, POLICY_REQUIRED_URL, SEQR_SLACK_ANVIL_DATA_LOADING_CHANNEL
 logger = SeqrLogger(__name__)
 
-anvil_auth_required = user_passes_test(is_anvil_authenticated, login_url=GOOGLE_LOGIN_REQUIRED_URL)
-
-
-def anvil_auth_and_policies_required(wrapped_func=None, policy_url=API_POLICY_REQUIRED_URL):
```
hanars, line 35: this was moved unchanged into `permissions_utils`

seqr/views/apis/family_api.py
```diff
@@ -288,7 +288,7 @@ def update_family_analysed_by(request, family_guid):
     create_model_from_json(FamilyAnalysedBy, {'family': family, 'data_type': request_json['dataType']}, request.user)
 
     return create_json_response({
-        family.guid: {'analysedBy': list(get_json_for_queryset(family.familyanalysedby_set.all()))}
```
hanars, line 291: not super related to this PR but lack of deterministic order was causing a test flap

seqr/views/apis/igv_api.py
```diff
@@ -183,9 +184,9 @@ def update_individual_igv_sample(request, individual_guid):
 
 
 @login_and_policies_required
-def fetch_igv_track(request, project_guid, igv_track_path):
```
hanars, line 186: IGV tracks are associated with one specific sample, so it is appropriate to check access at the family instead of project level

seqr/views/apis/project_api.py
```diff
@@ -69,12 +69,6 @@ def create_project_handler(request):
     })
 
 
-def _is_valid_anvil_workspace(request_json, user):
```
hanars, line 72: this was moved unchanged into `permissions_utils`

seqr/views/apis/users_api.py
```diff
@@ -49,10 +49,10 @@ def get_all_user_group_options(request):
 
 @login_and_policies_required
 def get_project_collaborator_options(request, project_guid):
-    project = get_project_and_check_view_permission(project_guid, request.user)
+    project, analysis_groups = get_project_analysis_groups_and_check_view_permission(project_guid, request.user)
     user_fields = {'display_name', 'username', 'email'}
-    users = get_project_collaborators_by_username(
-        request.user, project, fields=user_fields, expand_user_groups=True,
+    users = None if analysis_groups else get_project_collaborators_by_username(
```
hanars, line 54: This endpoint is used to get the list of collaborators you can assign to a case. If you have full access to the project you can assign anyone with project access. If you only have partial access you can only assign yourself

seqr/views/apis/variant_search_api.py
```diff
@@ -415,11 +414,22 @@ def search_context_handler(request):
...
-    response.update(get_projects_child_entities(projects, project_guid, request.user))
+    family_project_guids = families.values_list('project__guid', flat=True)
+    project_guid = family_project_guids[0] if len(family_project_guids) == 1 else None
+
+    project_guids, analysis_group_guids = get_project_analysis_group_guids_user_can_view(request.user)
+    full_access_projects = set(family_project_guids).intersection(project_guids)
+    partial_access_projects = set(family_project_guids) - set(project_guids)
+    family_q = group_project_q = Q(project__guid__in=full_access_projects)
+    group_q = Q()
+    if partial_access_projects:
```
hanars, line 425: the context endpoint needs to return all families and groups that a user has access to even if they are not included in the current context, because these are used to populate the UI so users can add more cases to their search

seqr/views/apis/variant_search_api.py
```diff
@@ -445,6 +455,37 @@ def search_context_handler(request):
     return create_json_response(response)
 
 
+def _add_projects_response_context(response, project_guids, project_guid, user):
```
hanars, line 458: this was adapted from `get_projects_child_entities` to be more explicit about handling partial project entities but its about 80% copy-paste

seqr/views/apis/variant_search_api_tests.py
```diff
@@ -426,3 +429,11 @@ def test_search_results_redirect(self):
...
         self.assertEqual(response.status_code, 301)
         self.assertEqual(response.url, '/variant_search/results/6ebb895dfca0f63c34be1ca59d950205?page=2&sort=cadd')
+
+
+class LocalVariantSearchAPITest(AuthenticationTestCase, VariantSearchAPITest):
+    fixtures = ['users', 'social_auth', '1kg_project', 'reference_data', 'variant_searches', 'clickhouse_saved_variants']
+
+
+class AnvilVariantSearchAPITest(AnvilAuthenticationTestCase, VariantSearchAPITest):
+    fixtures = ['users', 'social_auth', '1kg_project', 'reference_data', 'variant_searches', 'clickhouse_saved_variants']
```
hanars, line 439: this change means we now runs the test suite twice, once with and once without anvil enabled, so we can properly test search context behavior with partial access

seqr/views/utils/orm_to_json_utils.py
```diff
@@ -543,25 +547,36 @@ def get_project_collaborators_by_username(user, project, fields, include_permiss
 
         for collaborator, perms in collaborator_perms.items():
             collaborators[collaborator.username] = _get_collaborator_json(
-                collaborator, fields, include_permissions, can_edit=CAN_EDIT in perms)
+                collaborator, ['username', 'email', 'display_name'], include_permissions, can_edit=CAN_EDIT in perms)
 
     elif project_has_anvil(project):
-        permission_levels = get_workspace_collaborator_perms(user, project.workspace_namespace, project.workspace_name)
```
hanars, line 549: this was all abstracted into `get_workspace_collaborators_by_username` unchanged, the diff just looks bad because of the indent

seqr/views/utils/permissions_utils.py
```diff
@@ -91,6 +92,16 @@ def decorator(view_func):
     return decorator
 
 
+anvil_auth_required = user_passes_test(is_anvil_authenticated, login_url=GOOGLE_LOGIN_REQUIRED_URL)
+
+def anvil_auth_and_policies_required(wrapped_func=None, policy_url=API_POLICY_REQUIRED_URL):
```
hanars, line 97: copied from `anvil_workspace_api`

seqr/views/utils/permissions_utils.py
```diff
@@ -167,8 +194,14 @@ def has_workspace_perm(user, permission_level, namespace, name, can_share=False,
     return workspace_permission if meta_fields else True
 
 
+def is_valid_anvil_workspace(request_json, user):
```
hanars, line 197: copied from `project_api`

seqr/views/utils/project_context_utils.py
```diff
@@ -13,42 +13,11 @@
 logger = SeqrLogger(__name__)
 
 
-def get_projects_child_entities(projects, project_guid, user):
```
hanars, line 16: this was moved to `variant_search_api` which is the only place it was used

ui/pages/Project/components/AnalysisGroupButtons.jsx
```diff
@@ -125,17 +138,29 @@ const DECORATORS = [
...
-const canUpdateGroup = (project, analysisGroup) => (
-  project.canEdit && (!analysisGroup?.analysisGroupGuid || analysisGroup.projectGuid)
+const canUpdateGroup = (project, analysisGroup, user) => (
+  project.canEdit && (!analysisGroup?.analysisGroupGuid || analysisGroup.projectGuid) && (
+    !analysisGroup?.workspaceNamespace || user?.isPm
+  )
 )
 
-export const UpdateAnalysisGroup = React.memo(({ project, analysisGroup, onSubmit, iconOnly, createDynamic }) => {
-  if (!canUpdateGroup(project, analysisGroup)) {
```
hanars, line 133: the base components were originally exported from this file for unit testing, but this is an old anti pattern and we prefer to test the connected component instead, which we are now doing

ui/pages/Project/components/ProjectPageUI.jsx
```diff
@@ -100,17 +100,19 @@ const ProjectPageUI = React.memo(({ analysisGroupGuid, load, loading, familiesLo
             />
           </ProjectSection>
           <VerticalSpacer height={10} />
-          <ProjectSection label="Variant Tags" linkPath="saved_variants" linkText="View All" loading={loading}>
+          <ProjectSection label="Variant Tags" linkPath={analysisGroupGuid ? `saved_variants/analysis_group/${analysisGroupGuid}` : 'saved_variants'} linkText="View All" loading={loading}>
             <VariantTags analysisGroupGuid={analysisGroupGuid} />
           </ProjectSection>
         </Grid.Column>
         <Grid.Column width={4}>
-          <ProjectSection label="Notifications">
```
hanars, line 108: it actually never made sense to show notifications on the analysis group page because they reflect aggregated data and can not be filtered to a family subset

larrybabb (approved):
Nice work. This is an impressive piece of work. This is the first time I went back to observe how you go through the dev process with your commits. Very nice job. You've been busy.
No concerns at all.

PR #5500, opened 2026-07-07
fix gene breakdown results in external projects

hanars:
## Pull request overview

Fixes a crash in the variant gene-breakdown API when variant results (notably from external/no-access projects) omit `familyGuids`, and updates ClickHouse search tests to cover the corrected behavior.

**Changes:**
- Make `get_variant_gene_breakdown` tolerant of missing `familyGuids` by iterating over an empty list when absent.
- Extend `test_gene_variant_lookup` to assert gene-breakdown responses in scenarios where `familyGuids` is intentionally removed from returned variants.


larrybabb (approved):
this looks pretty straightforward. no concerns.

PR #5509, opened 2026-07-15
update requirements

hanars:
In general its usually enough for us to use automated vulnerability detection to bump pinned dependencies to get security fixes as needed, but over time we have been accruing issues with more complex dependency resolution issues and have been getting a build up of known issue with available fixes. Therefore, I fully re-ran pip compile to get us to the latest viable versions for all the packages we use in order to resolve our outstanding issues with known vulnerabilities and prune our pinned downstream dependencies


seqr/views/apis/feature_updates_api.py
```diff
@@ -26,7 +26,7 @@ def get_feature_updates(request):
     entries = []
     for entry in fetch_feature_updates():
         # Atom feeds can have multiple content elements per feed entry
-        markdown = "".join(md(content.value) for content in entry.content)
```
hanars, line 29: the latest version of markdownify does a better job removing unneccessary whitespace, which in turn means we can't rely on the presence of the buggy extra whitespace anymore

larrybabb (approved):
LGTM

PR #5511, opened 2026-07-16
Vlm lift bug

hanars:
## Pull request overview

Fixes a VLM liftover issue by adjusting chromosome naming passed into `pyliftover` so GRCh37/GRCh38 coordinate conversion works correctly, and adds a regression test to cover GRCh37 requests that rely on liftover.

**Changes:**
- Update `_liftover_variant` to use build-specific chromosome prefixing when calling `LiftOver.convert_coordinate`.
- Add a test request exercising `assemblyId=GRCh37` that should return results via liftover (lifotver was previously tested but only for lifting 38->37)


larrybabb (approved):
+1

PR #5527, opened 2026-07-28
5384: (elanfisher) Fix header nav items being cut off on smaller screens

### LINKED ISSUE #5384: UI issue: Gene Lookup addition causing features to be cut Off
**Is your feature request related to a problem? Please describe.**
After the gene lookup feature was added to the top tab, the features on the right are being partially cut off.

**Describe the solution you'd like**
Can we adjust the spacing, such as shifting the layout slightly to the left? There’s unused space before “seqr” on the left. Another option is to reduce the size of the search box. 

**Describe alternatives you've considered**
NA

**Additional context**

Image

elanfisher:
Fixes #5384

At 1512px, the width in the ticket's screenshot, the header overflows by 178px and "Log out" ends up 164px past the right edge where you can't click it. Zooming does the same since it shrinks the viewport.

The header reserves 100px gutters either side plus a fixed 350px search box. This lets the search shrink when space runs out, trims the gutters below 1800px, and only lets the bar wrap below 1400px. That last part matters because flexbox wraps before it shrinks, so wrapping unconditionally gives two rows at widths where shrinking alone was fine. I also moved "Log out" into the user dropdown, which is both convention and 77px of headroom. 
**Let me know if you want me to move it back out as it did declutter the header but was slightly outside of the scope of the issue.**

You mentioned manager fixes shouldn't come at the cost of the core user base, so I checked both roles. See screenshots below, I tried to be complete.

Managers go from 178px clipped to a single row, and at 1280px the bar wraps rather than clipping anything. Non-managers keep the same 350px search box and 100px gutters they have today, with "Log out" now in the dropdown.

Below 1400px the nav doesn't fit on one line. An overflow menu might be better long term if thats more in the style of the rest of the `seqr` repo. Id be happy to take that up if needed.

Lint is clean and all 86 suites and 165 tests pass when I ran them. I also added including two new ones in Header.test.js just to be safe.

Non-Manager (100% zoom):
image

Non-Manager (150% zoom):
image

Non-Manager Migrated Login Drop down:
image

image

Manager View (100% zoom):
image

Manager View (150% zoom):
image

### DIFF
--- a/ui/shared/components/page/Header.jsx
+++ b/ui/shared/components/page/Header.jsx
@@ -17,41 +17,22 @@ const HeaderMenu = styled(Menu)`
   padding-left: 100px;
   padding-right: 100px;
 
-  /* 
-    @elanfisher:
-    The nav items need ~1780px at the default gutters, so on smaller screens
-    or when the page is zoomed the right hand items (eg. "Log out") were
-    clipped and thus occluded. Shifted the gutters and search box and then allowed 
-    the bar to wrap rather than overflow.
-  */
   @media (max-width: 1800px) {
     padding-left: 24px;
     padding-right: 24px;
   }
 
-  /*
-    @elanfisher: 
-    Flexbox wraps before it shrinks, so only allow wrapping once the search box
-    has run out of room to give. Above this the bar stays on one line. 
-  */
   @media (max-width: 1400px) {
     flex-wrap: wrap;
   }
+`
 
-  /* 
-    @elanfisher:
-    Keep the search box at its full 350px whenever it fits, and let it be the
-    first thing to give way when it does not. The basis and max are 382px,
-    350px of search plus the menu item's own 2 x 16px padding, so users who
-    were never short of space see exactly the previous layout.
-  */
-  .item.awesomebar {
-    flex: 0 1 382px;
-    min-width: 216px;
-    max-width: 382px;
-  }
+const AwesomeBarMenuItem = styled(Menu.Item)`
+  flex: 0 1 382px !important;
+  min-width: 216px;
+  max-width: 382px;
 
-  .item.awesomebar .ui.search {
+  .ui.search {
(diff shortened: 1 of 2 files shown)


ui/shared/components/page/Header.jsx
```diff
@@ -31,7 +69,8 @@ const PageHeader = React.memo(({ user, oauthLoginProvider, onSubmit, lastFeature
         <Menu.Item key="summary_data" as={Link} to="/summary_data" content="Summary Data" />,
         (user.isAnalyst || user.isPm) ? <Menu.Item key="report" as={Link} to="/report" content="Reports" /> : null,
         (user.isDataManager || user.isPm) ? <Menu.Item key="data_management" as={Link} to="/data_management" content="Data Management" /> : null,
-        <Menu.Item key="awesomebar" fitted="vertically"><AwesomeBar newWindow inputwidth="350px" /></Menu.Item>,
+        // @elanfisher: Made the searchbar flex properly rather than be fixed at 350px.
+        <Menu.Item key="awesomebar" className="awesomebar" fitted="vertically"><AwesomeBar newWindow inputwidth="350px" /></Menu.Item>,
```
hanars: It is an anti pattern to use `className` in this repo. Instead, you should use a styled component. You could do something like 
```
const AwesomeBarMenuItem = styled(Menu.Item)`
  flex: 0 1 382px;
  min-width: 216px;
  max-width: 382px;

  .ui.search {
    width: 100%;
  }
`
```

hanars (commented):
Thanks so much for submitting this, overall it looks great! I did leave one comment about the use of css classes that I would prefer to have addressed before merging, but overall the approach is solid. Additionally, the comments are very helpful for PR review but I think a bit bulky/ unnecessary for long term support, so  I would suggest removing them

hanars (approved):

PR #5532, opened 2026-07-31
allow loading data to proejcts with other sample type data

hanars:
## Pull request overview

This PR updates the “loaded projects” selection logic for variant-call datasets so projects are not excluded solely because they already contain datasets of the other sample type (WES vs WGS), enabling loading variant calls into projects that previously would have been filtered out. It also updates API tests to reflect the new project-list behavior and ensure `dataTypeLastLoaded` is reported per requested sample type.

**Changes:**
- Relaxed `get_loaded_projects` filtering for `Dataset.DATASET_TYPE_VARIANT_CALLS` to allow projects that may already have the opposite sample type loaded.
- Kept the “must already have data for this sample type” restriction for non-variant dataset types.
- Updated test expectations to validate returned project lists and `dataTypeLastLoaded` behavior for WGS vs WES requests.


larrybabb (approved):
This is a small clean change. no concerns

PR #5540, opened 2026-08-10
Load anvil data for no update individuals

hanars:
## Pull request overview

This PR updates seqr’s AnVIL loading flow to ensure sample loading is triggered even when the uploaded pedigree doesn’t cause any individual metadata changes, by returning DB IDs for all processed individuals (not just updated ones).

**Changes:**
- Updated `add_or_update_individuals_and_families` to optionally return DB IDs for all individuals encountered in the input records.
- Updated AnVIL workspace loading to use the new “all individual IDs” behavior when triggering pipeline loading.
- Adjusted AnVIL workspace API tests (mocking gsutil-backed reads) and updated the 1KG fixture to add test coverage for unchanged metadata


seqr/views/apis/anvil_workspace_api_tests.py
```diff
@@ -565,13 +565,10 @@ def setUp(self):
         patcher = mock.patch('seqr.utils.add_data_utils.logger')
         self.mock_add_data_utils_logger = patcher.start()
         self.addCleanup(patcher.stop)
-        patcher = mock.patch('seqr.views.apis.anvil_workspace_api.load_uploaded_file')
```
hanars, line 568: removes higher-level mocking to consistently be able to check all interactions with gsutil. Not strictly related to this PR but it was related to the debugging I was doing and we do want the lower level behavior tested to prevent regression

larrybabb (approved):
LGTM. seems clean.

PR #5543, opened 2026-08-11
Pipeline runner to seqr repo

hanars:
This PR moves the current main branch of the seqr loading pipeline into the seqr repo. Once released, we will deprecate the separate pipeline repo and all subsequent changes to the pipeline will be managed in this one repository. 

This PR is almost entirely copy paste. The few exceptions are:
- the .gitignore and .dockerignore files, where I merged the pipeline's versions of these files into the existing seqr repo files
- moving the cloudbuild and github workflow files into the existing directories for those files in the seqr repo. 
- small updates to the pipeline unit test runner/ build workflows and docker file to handle small changes to the file structure

The initial commit is a direct copy-paste of most of the files from the pipeline. For a slightly more readable diff that includes the (very small number of) files I actually modified, see here: https://github.com/broadinstitute/seqr/pull/5543/changes/5de4c680a313f63b02c35a936ead9db0e6d48161..59a95488dfa4524baf9bdfee37ea0c2a00e45aec


.dockerignore
```diff
@@ -6,6 +6,7 @@ static/*
 deploy/*
 !deploy/docker/seqr/*
 !deploy/docker/vlm/*
+*test*
```
larrybabb: .dockerignore adds *test* globally — This pattern excludes any path containing "test" from Docker builds across the entire repo, not just loading_pipeline/. If seqr's Docker image needs any file with "test" in its name (e.g., test_utils.py), it will be silently missing. Should be scoped to loading_pipeline/*test*.

.gitignore
```diff
@@ -18,3 +18,174 @@ pedigree_images/*
 seqr_settings
 
 django_key
+
```
larrybabb, line 21: .gitignore adds ~170 generic Python ignore patterns globally — Patterns like *.vcf.gz, temp, *.log, build/, dist/ are broad enough to cause surprises in a genomics project. These should be scoped under loading_pipeline/ rather than applied repo-wide.

larrybabb (dismissed):
This is a big PR and I get it's basically a move from the loading pipeline repo. I think I'd like to chat with you about it to make sure I know how much effort I should put into this.

In the meantime, I added 2 concerns that Claude identified that you should review and consider.

After you review and we chat I'll approve.

.gitignore
hanars: I disagree - these are good best practices that prevent accidentally committing private data to public repositories. I did an audit and no files currently in github are now ignored with these changes, so there is no harm to the existing codebase. By including these for the entire seqr repo we improve our security overall - if in the future we really need to add a specific file that matches these new patterns, we can always explicitly include it

larrybabb (approved):
this is a satisfactory lift over of the code base.

PR #5547, opened 2026-08-13
Pipeline unit test django

hanars:
Update pipeline tests to always run against the latest clickhouse database schema

## Pull request overview

This PR updates the loading pipeline’s ClickHouse-related unit tests to run via a minimal Django + `django-clickhouse-backend` setup (fixtures + migrations) instead of a hand-maintained SQL schema file. In order to support this minimal set up, it refactors some shared domain constants into a single module. It also adjusts CI workflows and pipeline dev dependencies to support the new test approach.

**Changes:**
- Centralize common constants (sex/affected status/sample/dataset/genome version) into `seqr/utils/constants.py` and update call sites.
- Replace pipeline ClickHouse schema test bootstrapping from a raw SQL file with Django migrations + pytest-django settings/fixtures.
- Update pipeline dependencies/locks and GitHub Actions workflows to keep CI aligned with the new test harness.


loading_pipeline/lib/misc/clickhouse_test.py
```diff
@@ -1243,96 +1172,41 @@ def test_rebuild_gt_stats(self):
...
             """,
         )
-        self.assertCountEqual(gt_stats, [(1, 0)])
-
-    @patch.object(
-        ClickhouseReferenceDataset,
-        'for_reference_genome_dataset_type',
-        return_value=[ClickhouseReferenceDataset.CLINVAR],
-    )
-    def test_repartitioned_entries_table(self, mock_for_reference_genome_dataset_type):
```
hanars, line 1286: the previous test suite was testing against an older partition schema and this test was added to have a test against the new partition schema. We now only test against the latest supported schema in all tests, so this test is now redundant

loading_pipeline/lib/misc/clickhouse_test.py
```diff
@@ -38,313 +40,176 @@
...
-                )
+
+class ClickhouseTest(MockedDatarootTestCase, ClickhouseSchemaTestCase):
+    fixtures: ClassVar = ['clickhouse_test']
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        #  Postgres dicts are not managed by django in this test suite and therefore cannot be loaded from fixtures
+        with connections['clickhouse_write'].cursor() as cursor:
```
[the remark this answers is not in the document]
hanars: no test reloads the dict, this is a fake problem

clickhouse_search/models/search_models.py
```diff
@@ -461,7 +453,7 @@ class EntriesGcnv(BaseEntries):
     class Meta(BaseEntries.Meta):
         db_table = 'GRCh38/GCNV/entries'
 
-class VariantDetailsGRCh37SnvIndel(models.ClickhouseModel):
+class VariantDetailsGRCh37SnvIndel(FixtureLoadableClickhouseModel):
```
larrybabb, line 456: Please confirm that it is safe that the `.save()` on these two classes will now do INSERT-only behavior even on updates.

(from Claude for verificiation)
VariantDetailsGRCh37SnvIndel and VariantDetailsSnvIndel changed base class from models.ClickhouseModel to FixtureLoadableClickhouseModel, which forces force_insert=True on every .save(). These models use EmbeddedRocksDB engines (no ALTER support), so INSERT-only semantics are likely correct — but if any production code path calls .save() on existing instances expecting UPDATE behavior, it would silently change. Worth confirming.

larrybabb (changes requested):
one item to verify see `search_models.py` line 456 and 473.

clickhouse_search/models/search_models.py
hanars: It would be a total violation of all supported usages of seqr to ever create rows in any table defined in search_models.py using django- as such no production code path is ever calling `save` either explicitly or implicitly. All data written to these tables must always be written using the loading pipeline, which executes targeted SQL commands. The only time we use django to write to these tables is during django's test set up.

larrybabb (approved):
After chatting with Hana and re-reviewing I am happy with this PR as-is

PR #5551, opened 2026-08-18
Deprecate pipeline migrations

hanars:
## Pull request overview

This PR removes legacy Luigi tasks and scripts that supported historical Helm upgrade/migration workflows (notably ClickHouse/variants-table migrations), aligning the loading pipeline code with the current 3.x.x deployment reality described in seqr-helm docs. That documentation explicitly instructs users to run these migrations off of the `2.19.3-annotations-final` helm chart, so removing them here is safe.

**Changes:**
- Removed deprecated ClickHouse/variants migration task modules, their corresponding tests, and CLI entry scripts.
- Simplified pipeline logic by removing migration-specific branching in variant table/parquet tasks and Dataproc cluster naming.
- Dropped the unused `migrations` global and migration run-id constants that were only needed for those legacy flows.


larrybabb (approved):
straightforward.

PR #5559, opened 2026-08-22
Prevent duplicate emails from blocking login

### LINKED ISSUE #1439: Seqr will allow for the creation of two users with the same email, and then will allow neither to sign in
**Description**
Using the documented `docker-compose exec seqr python manage.py createsuperuser` command, two users can be created that share the same email address. Two users cannot be created that share the same username because the script will throw an error if you try to do that. Once two users are created with the same email address, neither user can sign in.

**Steps to replicate**
Create two users with `createsuperuser`.
User 1:
 - Username: example@example.com
 - Email: example@example.com
 - Password: example1

User 2:
 - Username: example
 - Email: example@example.com
 - Password: example2

Now neither user can sign in. Attempting to sign in results in a 'invalid credentials (401)' error on the login page.

**Steps to resolve**
Sign in to the postgres database and `\connect seqrdb`, then `select * from auth_user;` and get the row ID of the second user, then run `delete from auth_user where id = X`, with X being the row ID. Now user 1 can sign in.

**Link to page(s) where bug is occurring**
seqr login page, local setup

mannyakosah:
## Summary
- Fixes #1439: `createsuperuser` could create two accounts with the same email, and login then rejected both with `Invalid credentials`.
- Reject saving a user whose email already exists (case-insensitive, same matching as login).
- When duplicates already exist, login returns `Multiple users found with this email` instead of a generic 401.

## Test plan
- [x] `./manage.py test seqr.views.apis.auth_api_tests seqr.management.tests.createsuperuser_tests`
- [x] `createsuperuser` with a new email succeeds; a second user with the same email (any case) fails with `That email is already taken.`
- [x] Login with a unique email still works; wrong email/password still returns `Invalid credentials`
- [x] If two users already share an email, login shows `Multiple users found with this email`

### DIFF
--- a/seqr/apps.py
+++ b/seqr/apps.py
@@ -3,3 +3,6 @@ from django.apps import AppConfig
 class SeqrConfig(AppConfig):
     name = 'seqr'
 
+    def ready(self):
+        import seqr.signals  # noqa: F401
+
--- a/seqr/apps.py
+++ b/seqr/apps.py
@@ -4,5 +4,10 @@ class SeqrConfig(AppConfig):
     name = 'seqr'
 
     def ready(self):
-        import seqr.signals  # noqa: F401
+        from django.contrib.auth.models import User
+        from django.db.models.signals import pre_save
+        from seqr.signals import validate_unique_email
+        pre_save.connect(
+            validate_unique_email, sender=User, dispatch_uid='seqr.validate_unique_email',
+        )
 
--- a/seqr/management/tests/createsuperuser_tests.py
+++ b/seqr/management/tests/createsuperuser_tests.py
@@ -1,0 +1,25 @@
+from django.contrib.auth.models import User
+from django.core.management import call_command
+from django.core.management.base import CommandError
+from django.test import TestCase
+
+
+class CreateSuperuserTest(TestCase):
+    fixtures = ['users']
+
+    def test_command(self):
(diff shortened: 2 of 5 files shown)


mannyakosah:
before 

https://github.com/user-attachments/assets/cb73721a-57de-435f-bba2-817cf2b676b2


after

https://github.com/user-attachments/assets/00d8e8a2-fbae-42ac-aa46-716de7b40ce0

hanars:
Thank you for your interest in contributing to seqr! At this time, we will unfortunately not be able to upstream this PR into the main seqr codebase. While the bug you address is real, it can only be reached by ignoring seqr's setup documentation and best practices, and there is a manual resolution workflow to resolve the issue in the case that a user does reach this edge case through such a violation. As such, our team does not feel that taking on the long term technical debt a change like this accumulates is worth the fix. I have updated the ticket to reflect that our team has decided not to fix the issue.
In the future, if you are pulling an old ticket out of our icebox, I might recommend commenting on the ticket first to get any insights on it from our team. While we do our best to keep our tickets as well curated as possible, our small dev team does not have the bandwidth to always keep everything in a ready state for external contributors, and as such a quick check in can ensure that a ticket is up-to-date before work is begun. 
Than you again for contributing to seqr!

PR #5560, opened 2026-08-22
Name selected families and individuals in the delete confirm

### LINKED ISSUE #2940: Edit Families & Individuals filter is not applied when selecting all
**Describe the bug**

When "Editing individuals" from the "Edit Families & Individuals" page on any project, you can use the checkbox to select all individuals, **when filtering this applies the select to all individuals outside the selection**.

The confirmation dialogue doesn't indicate how many participants it's about to delete.

**Link to page(s) where bug is occurring**
Project page -> Edit Families & Individuals

**Scope of the bug**
Any uses of the EditRecordForm (EditIndividuals + EditFamilies).

**Screenshots**

image

image

Happy to put together PR to address this if there is a preferred approach.

mannyakosah:
## Summary
- Fixes #2940: the Edit Families / Edit Individuals delete confirm said `the selected families/individuals` and did not list which rows were checked.
- The confirm now names the selected IDs, e.g. `Are you sure you want to delete the following individuals: NA19678, NA19675?`
- Disable **Delete Selected** until at least one row is checked, and fix the button label typo (`Deleted` → `Delete`).

## Test plan
- [x] `npm test -- --testPathPattern='EditFamiliesForm.test|EditIndividualsForm.test' --watchAll=false`
- [ ] Project page → Edit Families & Individuals → Edit Families: with nothing checked, Delete Selected is disabled
- [ ] Check one or more families; confirm lists those family IDs; cancel
- [ ] Edit Individuals tab: same for individual IDs (including after filter + select-all)

### DIFF
--- a/ui/shared/components/form/EditRecordsForm.jsx
+++ b/ui/shared/components/form/EditRecordsForm.jsx
@@ -108,6 +108,9 @@ class EditRecordsForm extends React.PureComponent {
     const rowsToDelete = Object.entries(recordData).reduce((acc, [recordId, { toDelete }]) => (
       { ...acc, [recordId]: toDelete }
     ), {})
+    const selectedIds = Object.values(recordData).filter(({ toDelete }) => toDelete).map(
+      record => record[filterColumn] || record[idField],
+    )
 
     const getRowFilterVal = filterColumn ? row => row[filterColumn] : null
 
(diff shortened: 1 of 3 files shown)


mannyakosah:
**before**

https://github.com/user-attachments/assets/50e4eac6-ad00-4f0a-9094-e26a5af01f76




**after**

https://github.com/user-attachments/assets/06abcb3e-16ea-4ba6-9575-22b9f8dbb666

ui/shared/components/form/EditRecordsForm.jsx
```diff
@@ -139,8 +142,9 @@ class EditRecordsForm extends React.PureComponent {
                   initialValues={this.getFilteredRecords(recordData, record => record.toDelete)}
                   onSubmit={onSubmit}
                   onSuccess={closeParentModal}
-                  confirmDialog={`Are you sure you want to delete the selected ${entityKey}?`}
-                  buttonText="Deleted Selected"
+                  confirmDialog={`Are you sure you want to delete the following ${entityKey}: ${selectedIds.join(', ')}?`}
```
hanars, line 145: So based on my reading of the code, this should be causing the whole footer and possibly the entire DataTable to fully remount every time a value in the form is changed. Can you validate that that is not happening? While you would not see much of a performance hit on a small project like the one you include in your video, for large projects a remount like this would create an unusably laggy experience. 
If I am wrong about the remounting, then thats great. However, if I am correct, then the real solution to this would be to update the underlying component to support a dynamic `getConfirmDialog` property that could compute a confirm dialog based on form input at render time

hanars (changes requested):

PR #5561, opened 2026-08-22
Treat non-array HPO term fields as empty on the family page

mannyakosah:
## Summary
- The family page calls `.reduce` on `nonstandardFeatures`. If that JSON field is a string (the 1kg fixture stored `"[]"`), the panel throws and the page goes blank.
- Treat a non-array as empty in `getHpoTermsForCategory` / `FeatureSection`, and store a real list in the 1kg fixture.

## Test plan
- [x] `npm test -- --testPathPattern='HpoPanel.test' --watchAll=false`
- [ ] After `loaddata` of `1kg_project`, open `/project/R0001_1kg/family_page/F000001_1` — Family 1 renders (HPO terms for NA19675_1)
- [ ] Family 2 still renders as before
- [ ] Before/after recordings attached as a comment


mannyakosah:
**before**

https://github.com/user-attachments/assets/37a6d146-6b89-4464-b26b-23d57f3f7c48


 

**after**

https://github.com/user-attachments/assets/25f68e04-4fe2-459b-bb36-a32eaca3237e

hanars:
The django fixtures are designed to be used for tests, not to be loaded into a running application and explored. In order for us to consider updating this behavior, please demonstrate a workflow using the UI to update the `nonstandardFeatures` that can replicate this bug

PR #5574, opened 2026-09-10
update debian to supported version

hanars:
Debian 11 (bullseye) reached end of long term support on 8/31, so updating to Debian 13 (trixie)


larrybabb (approved):
+1

PR #5580, opened 2026-09-11
fix pipeline zipfile build

hanars:
The change to directory structure broke the zipfile used for dataproc runs. I have run this manually to fix the current pipeline production zipfiles, but these changes are needed going forward


.github/workflows/pipeline-unit-tests.yaml
```diff
@@ -9,7 +9,7 @@ on:
       - master
     paths:
       - 'loading_pipeline/**'
-      - '.github/workflows/pipeline-*-release.yaml'
+      - '.github/workflows/*-pipeline-release.yaml'
```
larrybabb, line 12: the file rename still says `.yaml` while the actual files are `.yml`. this won't match and retrigger unit tests.

larrybabb (changes requested):
Just one minor change which may be needed if I'm reading that correctly. The rest looks great.

.github/workflows/pipeline-unit-tests.yaml
hanars: good catch! I'll update the files so that they are .yaml and consitent with all other workflow files in this repo

larrybabb: +1

larrybabb (approved):
+1

PR #5582, opened 2026-09-11
Join new entry variants in clickhouse

hanars:
This pull request refactors how variant level annotations are added to the entries table - currently existing and new variants are joined with the entries in hail and exported to the entries parquet that is then loaded. With this change, the entries parquet includes no variant level data, and it is all added at INSERT time by joining the relevent tables before writing to clickhouse.

**1. ClickHouse Insert and Export Logic Updates**
- Refactored `insert_new_entries` in `clickhouse.py` to standardize column selection, ensure inclusion of `key` and `xpos`, and update the logic for populating `geneId_ids` from the correct gene consequence field depending on dataset type.
- Updated the export logic for existing variants to always use the `KEY_LOOKUP` table when appropriate and to select the correct fields for different dataset types, improving compatibility and downstream processing.

**2. Test Data and Test Logic Improvements**
- Enhanced test data for both SNV/INDEL and GCNV pipelines in `clickhouse_test.py`, including more realistic gene IDs, updated keys, and comprehensive GCNV entry records to better reflect production scenarios.
- Adjusted test assertions to align with the new data and logic, ensuring tests accurately verify the revised pipeline behavior.

**3. GCNV Annotation Logic Simplification**
- Simplified the GCNV annotation functions in `gcnv.py` by removing conditional logic and always returning the relevant fields directly, which reduces complexity and potential for inconsistent annotations. Previously, values were being mapped to missing and repopulated later in the pipeline flow, now they are computed properly up front

**4. Export Fields and Annotation Refactoring**
- Refactored export field logic in `fields.py` to remove dataset-specific field selection functions and instead directly reference the correct annotation fields, further simplifying the export code and reducing redundancy. 

**5. Documentation Update**
- Updated the pipeline diagram in `Diagrams.md` to reflect the new data flow and task dependencies, ensuring documentation matches the revised pipeline structure.

### DIFF
--- a/loading_pipeline/lib/misc/clickhouse.py
+++ b/loading_pipeline/lib/misc/clickhouse.py
@@ -974,7 +974,7 @@ def export_existing_variants_to_parquet(
         dataset_type,
         run_id,
     )
-    variants_table = variants_table = table_name_builder.dst_table(
+    variants_table = table_name_builder.dst_table(
         ClickHouseTable.KEY_LOOKUP
         if dataset_type.should_write_new_variant_details
         else ClickHouseTable.VARIANTS_MEMORY,
--- a/loading_pipeline/lib/misc/clickhouse.py
+++ b/loading_pipeline/lib/misc/clickhouse.py
@@ -661,7 +661,11 @@ def insert_new_entries(
 
     if 'geneId_ids' in dst_cols:
         common.append('geneId_ids')
-        gene_list_field = 'sortedGeneConsequences' if table_name_builder.dataset_type == DatasetType.SV else 'sortedTranscriptConsequences'
+        gene_list_field = (
+            'sortedGeneConsequences'
+            if table_name_builder.dataset_type == DatasetType.SV
+            else 'sortedTranscriptConsequences'
+        )
         overrides['geneId_ids'] = f"""
             arrayFilter(
                 x -> x IS NOT NULL,
@@ -684,9 +688,6 @@ def insert_new_entries(
         overrides['is_gnomad_gt_5_percent'] = f"""
             dictGetOrDefault({ClickhouseReferenceDataset.GNOMAD_GENOMES.search_path(table_name_builder)}, 'filter_af', key, 0) > 0.05
         """
-
-    if table_name_builder.dataset_type == DatasetType.GCNV:
-        overrides
 
     dst_list = ', '.join(common)
     src_list = ', '.join([overrides.get(c, f'e.{c}') for c in common])
--- a/loading_pipeline/lib/misc/clickhouse.py
+++ b/loading_pipeline/lib/misc/clickhouse.py
@@ -974,7 +974,11 @@ def export_existing_variants_to_parquet(
         dataset_type,
         run_id,
     )
-    variants_table = table_name_builder.dst_table(ClickHouseTable.VARIANTS_MEMORY)
+    variants_table = variants_table = table_name_builder.dst_table(
+        ClickHouseTable.KEY_LOOKUP
+        if dataset_type.should_write_new_variant_details
+        else ClickHouseTable.VARIANTS_MEMORY,
+    )
     export_table = table_name_builder.src_table(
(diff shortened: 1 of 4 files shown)


loading_pipeline/lib/misc/clickhouse.py
```diff
@@ -682,12 +691,21 @@ def insert_new_entries(
...
+        FROM (
+            SELECT
+                dst.key,
+                COLUMNS('.*') EXCEPT(variantId, key)
+            FROM {table_name_builder.src_table(ClickHouseTable.ENTRIES)} src
+            INNER JOIN {table_name_builder.dst_table(ClickHouseTable.KEY_LOOKUP)} dst
+            ON {ClickHouseTable.KEY_LOOKUP.join_condition}
+        ) e
+        INNER JOIN {table_name_builder.dst_table(ClickHouseTable.VARIANTS_MEMORY)} v
+        ON assumeNotNull(e.key) = v.key
```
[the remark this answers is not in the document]
hanars: fixed!

loading_pipeline/lib/tasks/write_existing_variants_parquet_test.py
```diff
@@ -46,20 +46,18 @@ def _run_task(
...
             [
-                {
-                    'key_': 1,
-                    'variant_id': '1-878314-G-C',
-                    'geneIds': ['ENSG00000177000'],
-                },
-                {'key_': 7, 'variant_id': '7-1234567-AGT-A', 'geneIds': []},
-                {'key_': 10, 'variant_id': '10-987654-G-A', 'geneIds': []},
+                {'key_': 1, 'variant_id': '1-878314-G-C'},
+                {'key_': 3, 'variant_id': '3-133456789-A-G'},
```
hanars, line 57: reflects updated fixture data

loading_pipeline/lib/tasks/exports/write_new_entries_parquet_test.py
```diff
@@ -433,7 +390,7 @@ def test_gcnv_write_new_entries_parquet(self):
                             'defragged': False,
                             'start': 100017585,
                             'end': 100023213,
-                            'numExon': 1,
+                            'numExon': 3,
```
hanars, line 393: the numExon for a couple of calls was being overridden by the annotations.ht added in test setup, which is not reflective of real world behavior and is now removed for this test. The actual behavior for this value is unchanged

larrybabb (commented):
@hanars I did my best to understand the logic. It seems reasonable based on my limited knowledge of the underlying designs.
I did get one interesting report from an anthropic review which I think you should be aware of. It may be that EntriesGcnv will never have a gene_ids column then this may be moot. I defer to your assessment.
Otherwise this looks pretty complete to me (and Claude).

> **One finding, confirmed as a real (if currently dormant) bug:** in loading_pipeline/lib/misc/clickhouse.py, the gene_list_field ternary picks sortedTranscriptConsequences for any dataset type that isn't SV — but GCNV's variants_memory model only defines sortedGeneConsequences (it shares a base class with SV, not with SNV_INDEL/MITO). It doesn't fire today only because EntriesGcnv has no geneId_ids column yet, so the if 'geneId_ids' in dst_cols: guard short-circuits. The moment someone adds that column to the GCNV entries schema, this will break with an unknown-column error in ClickHouse. Worth fixing the condition to dataset_type in (SV, GCNV) now while the context is fresh, rather than leaving it as a landmine.

larrybabb (approved):
I'm pre-approving with the assumption that you will review the previous comment and respond appropriately.

hanars:
> The moment someone adds that column to the GCNV entries schema, this will break with an unknown-column error in ClickHouse.

I do not approve of building magically future proof logic for features that will never exist. For one thing, SV and GCNV are being fully replaced by the end of the year, so theres really no reason to future proof them. But even if they weren't, making the code worse so that a theoretical future change would not break things is bad engineering practice. There are unit tests that would immediately catch such an error. If in the future if the tables change we can change the logic that writes to them

PR #5590, opened 2026-09-16
fix dev  pipeline release workflow

hanars:


.github/workflows/dev-pipeline-release.yaml
```diff
@@ -49,9 +49,9 @@ jobs:
         run: |-
           gcloud storage rm -r gs://seqr-pipeline-runner-builds/dev/latest/ || echo 'No latest release'
           gcloud storage cp loading_pipeline/bin/* gs://seqr-pipeline-runner-builds/dev/latest/bin/
-          gcloud storage cp cd loading_pipeline/dist/*.whl gs://seqr-pipeline-runner-builds/dev/latest/pyscripts.zip
+          gcloud storage cp cd pipeline_build/dist/*.whl gs://seqr-pipeline-runner-builds/dev/latest/pyscripts.zip
```
larrybabb, line 52: Stray `cd` token — `gcloud storage cp` now receives `cd` as an extra source argument. This mirrors the same bug in the line below and won't match the working `prod-pipeline-release.yaml` pattern.

```suggestion
          gcloud storage cp pipeline_build/dist/*.whl gs://seqr-pipeline-runner-builds/dev/latest/pyscripts.zip
```

.github/workflows/dev-pipeline-release.yaml
```diff
@@ -49,9 +49,9 @@ jobs:
...
           gcloud storage cp loading_pipeline/bin/* gs://seqr-pipeline-runner-builds/dev/${{ github.event.workflow_run.head_sha }}/bin/
-          gcloud storage cp cd loading_pipeline/dist/*.whl gs://seqr-pipeline-runner-builds/dev/${{ github.event.workflow_run.head_sha }}/pyscripts.zip
+          gcloud storage cp cd pipeline_build/dist/*.whl gs://seqr-pipeline-runner-builds/dev/${{ github.event.workflow_run.head_sha }}/pyscripts.zip
```
larrybabb, line 54: Same stray `cd` token here.

```suggestion
          gcloud storage cp pipeline_build/dist/*.whl gs://seqr-pipeline-runner-builds/dev/${{ github.event.workflow_run.head_sha }}/pyscripts.zip
```

larrybabb (commented):
Pragmatic review: the directory rename looks right, but both `pyscripts.zip` copy lines still carry a stray `cd` argument left over from before this PR, so `gcloud storage cp` gets 3 positional args instead of 2 and this step will still fail. Compare with `prod-pipeline-release.yaml`, which has the correct `gcloud storage cp pipeline_build/dist/*.whl gs://...` form without `cd`. Suggested fixes inline.

PR #5594, opened 2026-09-18
delete dataproc cluster that has entered error state

hanars:
The retry for creating a dataproc cluster is happening faster than automated pruning of error clusters, so we are not adequately attempting to recreate cluster when initial creation fails


larrybabb (changes requested):
@hanars I think you should review this discovery. It seems like it found an existing pattern for dealing with this type of issue that may be worth applying. I'm curious to see whether this is a worthwhile change. 

> **One real gap I'd flag:** self.client.delete_cluster(...) is a GCP long-running operation — like create_cluster() elsewhere in this same file, it returns immediately without waiting for completion (the codebase's established pattern is fire-then-poll via safely_get_cluster(), but there's no corresponding poll added here). Since it's fired right before a raise, nothing waits for the deletion to actually finish. On the next retry attempt (30s later, per @retry(tries=5, delay=30)'s default 30s delay), two bad outcomes are both plausible:
> 
> 1. If deletion is still in progress, safely_get_cluster() finds the cluster in DELETING state — not RUNNING, not in FAILURE_STATUSES — so create_cluster() is skipped and the code drops into the polling while loop. Once GCP finishes the delete, safely_get_cluster() starts returning None, and the very next loop iteration does cluster.status.state == ...RUNNING on cluster=None → uncaught AttributeError.
> 2. Alternatively, if create_cluster() does get reached while the old cluster is still mid-deletion, GCP will raise AlreadyExists.
>
> Both outcomes are silently absorbed by @retry's broad except Exception, but each one burns through the limited 5-try budget on a bug in this code path rather than a genuine transient GCP failure — for a real multi-node cluster where deletion legitimately takes longer than 30s, this could exhaust retries before the cluster is ever successfully recreated.
> 
> **Suggested fix:** wait for the delete operation to complete (e.g., self.client.delete_cluster(...).result()) before raising, or at minimum guard the polling loop against cluster is None.

hanars:
> @hanars I think you should review this discovery. It seems like it found an existing pattern for dealing with this type of issue that may be worth applying. I'm curious to see whether this is a worthwhile change. 

AI has been atrociously wrong about pretty much every assessment they have done about how dataproc management works. I can tell you that in order to get loading to succeed I have been manually deleting clusters during runs and the retry logic immediately starts spinning up a new cluster, even while the old cluster is still deleting

